PR #49, opened 2013-05-28
Unbreak doxygen. Possibly it also needs illumos headers fixes 

pyhalov:
(https://www.illumos.org/issues/3787) .

I successfully built and installed doxygen in my build zone, but it has patches applied from mentioned above illumos ussue. Maybe it will not compile on non-patched system.

### DIFF
--- a/components/components.ignore
+++ b/components/components.ignore
@@ -31,7 +31,6 @@
 /^cups$/d
 /^curl$/d
 /^dejagnu$/d
-/^doxygen$/d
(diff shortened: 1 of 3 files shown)


0xffea:
Alexander,
it would be best you post a fix for 3787 to the developer list (with the improvement recommend by Rich). This affects
other packages too and should be better fixed in the gate.  If you need help with the webrev i can assist you.

PR #50, opened 2013-06-01
Unbreak tcl/expect.

pyhalov:
It seems to be a bug in aucoconf or packet configuration.
This workaround is a bit ugly, but it works.


0xffea:
Thanks, did build fine.

PR #52, opened 2013-06-25
Enable ncurses and bump to v5.9

erolms:
Enabled ncurses library and bumped to version 5.9. Both 32 and 64 bit versions built.


0xffea:
nice, thanks.

PR #53, opened 2013-06-25
Added cxa_finalize binary which is necessary for clang++ and can be

pyhalov:
used by several versions simultaneously (tested on clang/llvm 3.3/3.2).


pyhalov:
I've decided to separate this component, because it can be used by several versions of clang++, it is built in a rather special way and in future we may have this functionality natively.

PR #81, opened 2013-07-15
Unbreak readline

pyhalov:


pyhalov:
Also need to change components.ignore...

PR #182, opened 2013-08-17
Add sqlite-3.7.17

DrLou:


pyhalov:
Please, don't merge this now.  To do propper sqlite update we firstly have to import all dependent components to oi-userland. So I'll close it without merge for now.

PR #189, opened 2013-08-19
Rename component to sqlite3 - fix Makefile, manifest. component.ignore for the moment.

DrLou:


pyhalov:
Lou is going to renew this request so that it can be merged automatically.

pyhalov:
Lou is going to renew this request so that it can be merged automatically.

PR #265, opened 2013-09-22
Update and enable bcc.

pyhalov:


pyhalov:
Don't know if someone needs it...

PR #326, opened 2013-11-07
These are patches we make to openvpn 2.2.2 (ported onto 2.3.2)

gwr:
Configure our socket for IPsec bypass, when appropriate.
(You need this when you're on an "edge" machine that's using IPsec.)

Fix up 1m vs. 8 in the man page.

I just tested this (updating my machine from 2.2.2 -> 2.3.2 and it worked fine.

I see that easy-rsa went away between those versions, which we use.
Perhaps I should find a way to package that from somewhere...

### DIFF
--- a/components/openvpn/Makefile
+++ b/components/openvpn/Makefile
@@ -26,6 +26,17 @@ include ../../make-rules/prep.mk
 include ../../make-rules/configure.mk
 include ../../make-rules/ips.mk
 
+PATCH_LEVEL = 0
+
+CPPFLAGS +=	-I$(WS_TOP)/components/tuntap/build/prototype/$(MACH)/usr/include
+CFLAGS +=	$(CPP_LARGEFILES)
+LDFLAGS +=	$(LD_Z_DEFS) $(LD_Z_TEXT) -lpthread
+
+CONFIGURE_OPTIONS += CPPFLAGS="$(CPPFLAGS)"
+CONFIGURE_OPTIONS += CFLAGS="$(CFLAGS)"
+CONFIGURE_OPTIONS += LDFLAGS="$(LDFLAGS)"
+CONFIGURE_OPTIONS += GREP=/usr/gnu/bin/grep
+CONFIGURE_OPTIONS += --disable-lzo
 CONFIGURE_OPTIONS += --disable-static
 CONFIGURE_OPTIONS += --enable-shared
 
--- a/components/openvpn/openvpn.p5m
+++ b/components/openvpn/openvpn.p5m
@@ -1,6 +1,6 @@
 #
 # This file and its contents are supplied under the terms of the
-# Common Development and Distribution License ("CDDL)". You may
+# Common Development and Distribution License ("CDDL"). You may
 # only use this file in accordance with the terms of the CDDL.
 #
 # A full copy of the text of the CDDL should have accompanied this
@@ -52,9 +52,10 @@ file path=usr/share/doc/openvpn/README.auth-pam
 file path=usr/share/doc/openvpn/README.down-root
 file path=usr/share/doc/openvpn/README.polarssl
 file path=usr/share/doc/openvpn/management-notes.txt
+
 dir  path=usr/share/man
-dir  path=usr/share/man/man8
-file path=usr/share/man/man8/openvpn.8
+dir  path=usr/share/man/man1m
+file usr/share/man/man8/openvpn.8 path=usr/share/man/man1m/openvpn.1m
 
 depend type=require fmri=driver/network/tun variant.opensolaris.zone=global
 depend type=require fmri=driver/network/tap variant.opensolaris.zone=global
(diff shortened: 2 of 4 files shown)


xen0l:
Hi,

first of all, thanks for further enhancing openvpn in OI and fixing some issues in the first packaging.
easy-rsa can be obtained from https://github.com/downloads/OpenVPN/easy-rsa/easy-rsa-2.2.0_master.tar.gz, which I will be packaging now. I will make openvpn depend on it, so it's pulled when installing openvpn.

Any objections to this?

xen0l:
Also why disable LZO support, which is worth using and we have liblzo in userland?

gwr:
Hi Adam,

I did not realize liblzo was there now.  I pulled this out of a much older
userland I had.

BTW, I'm somewhat new to your new integration process for hipster.
I assumed that when I issue a pull request, people would review it.
Is there another venue for requesting review first?

Anyway, as for the items you mention, I agree they should be as you suggest.

Thanks,
Gordon

On Thu, Nov 7, 2013 at 3:55 PM, Adam Števko notifications@github.comwrote:

> Also why disable LZO support, which is worth using and we have liblzo in
> userland?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/OpenIndiana/oi-userland/pull/326#issuecomment-28005446
> .

## 

Gordon Ross gwr@nexenta.com
Nexenta Systems, Inc.  www.nexenta.com
Enterprise class storage for everyone

gwr:
I think you mean make openvpn depend on easy-rsa so the latter gets pulled
in?
Anyway, easy-rsa is small (just some files in /usr/share, IIRC) so there
should be little harm in having openvpn just pull that in.

I have patches we needed to get parts of that to compile.  Not sure if
they're still needed, but I'll attach them below.  Just some fixes to the
makefile install rules to run the right "install".

Thanks again,
Gordon

On Thu, Nov 7, 2013 at 3:54 PM, Adam Števko notifications@github.comwrote:

> Hi,
> 
> first of all, thanks for further enhancing openvpn in OI and fixing some
> issues in the first packaging.
> easy-rsa can be obtained from
> https://github.com/downloads/OpenVPN/easy-rsa/easy-rsa-2.2.0_master.tar.gz,
> which I will be packaging now. I will make openvpn depend on it, so it's
> pulled when installing openvpn.
> 
> Any objections to this?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/OpenIndiana/oi-userland/pull/326#issuecomment-28005413
> .

## 

Gordon Ross gwr@nexenta.com
Nexenta Systems, Inc.  www.nexenta.com
Enterprise class storage for everyone

xen0l:
Hi, 

I packaged it already. I created a new component and made openvpn dependent on easy-rsa.
Btw, any chance that openvpn IPSec bypass patch could be upstreamed?

Adam

gwr:
On Thu, Nov 7, 2013 at 5:51 PM, Adam Števko notifications@github.comwrote:

> Hi,
> 
> I packaged it already. I created a new component and made openvpn
> dependent on easy-rsa.
> 
> Cool - thanks!
> 
> Btw, any chance that openvpn IPSec bypass patch could be upstreamed?
> 
> Sure, I just haven't taken the time to try, as yet.
> I appreciate the help with that.

Gordon

xen0l:
Hi, 

no problem at all, I have fixed those. Also thanks for pointing out the typo in CDDL header.

The proposed changes:
https://github.com/OpenIndiana/oi-userland/pull/327

Speaking of reviews, it’s a bit complicated. Hipster was created so the OI development moves forward and is done w/o any unwanted bureaucracy.
It’s a rolling release branch and stuff is expected to break sometimes. However, people dislike that and we are more careful.

We just do reviews either on IRC or mailing list, but it’s not strictly said how many LGTMs you need.
The problem here is that people ask for stable hipster (it has some newer packages than /dev), but aren’t that many reviewers and it would stop the whole process dramatically.
So, we simply put the stuff together, ask for review and wait some time. If we don’t get it in reasonable time (depends on the changes, could vary from few hours to few days), the changes are simply merges and the
problems are fixed as they appear. This is not ideal, but it doesn’t slow the pace of development and allows us to do some progress. But as I mentioned, things break sometimes.

I should find some time and definitely write these things down on wiki.

Btw, if you are missing some packages for build env, just install build-essential. It provides you with everything you need for userland and illumos-gate development.

Adam

gwr:
On Thu, Nov 7, 2013 at 6:05 PM, Adam Števko notifications@github.comwrote:

> Hi,
> 
> no problem at all, I have fixed those. Also thanks for pointing out the
> typo in CDDL header.
> 
> Thanks.
> 
> The proposed changes:
> https://github.com/OpenIndiana/oi-userland/pull/327
> 
> Speaking of reviews, it’s a bit complicated. Hipster was created so the OI
> development moves forward and is done w/o any unwanted bureaucracy.
> [...] the changes are simply merged and the problems are fixed as they
> appear. This is not ideal, but it doesn’t slow the pace of development and
> allows us to do some progress. But as I mentioned, things break sometimes.

That model seems to be mostly working, aside from the temporary loss of the
"lzo" feature due to my merging old changes :)
If the people pressing the "merge pull request" are reasonably careful, I
think this works.
I'm very impressed with how it's working for OI.

Thanks again,
Gordon

PR #339, opened 2013-11-12
Fix mysql build after dtrace changes.

pyhalov:
ndb strorage engine is included. However, there's no any confidence that it's usable.


pyhalov:
The main change is enabling ndb. So, some defines which enable dtrace logic is ndb storage engine are set. This just fixes mysql build, I don't know in ndb is usable in any way.

PR #354, opened 2013-11-21
Initial commit of bash-completion 2.1

alarcher:
This is an initial addition of the bash-completion macros maintained by Debian as I need this to handle the environment-modules in a clean way.
- Some macros are not useful on OpenIndiana and other might need modifications but at least this commit would give the possibility for people to get started with.
- A variable BASH_COMPLETIONS_PATH has been added to shared-macros.mk to allow installation of macros by other packages in the same directory.

As it is my first addition to oi-userland I would appreciate your comments and advice on the best practices.
Best regards,

Aurelien

### DIFF
--- a/components/bash-completion/bash-completion.p5m
+++ b/components/bash-completion/bash-completion.p5m
@@ -1,0 +1,608 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2013 Aurelien Larcher. All rights reserved.
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value=org.opensolaris.category.2008:System/Shells
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=.*\.py$ -> default pkg.tmp.autopyc false>
+
+depend type=require fmri=pkg:/shell/bash
+
+dir  path=etc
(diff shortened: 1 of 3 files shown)


components/bash-completion/bash-completion.p5m
```diff
@@ -0,0 +1,606 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=.*\.py$ -> default pkg.tmp.autopyc false>
+
+dir  path=etc
+dir  path=etc/profile.d
+file path=etc/profile.d/bash_completion.sh
```
pyhalov: Here should be preserve attribute and 0644 file mode

alarcher: Thank you, is there a page listing attributes that should be usually enforced in the manifest ?
On another topic, should I force a dependency to bash or is it generated automatically ?

pyhalov: I don't know about such page. But you could look at other manifests. I think that dependency on bnash will not be generated automaticaly - such  dependency is usually  based on requiring particular shared libraries. And this package doesn't provide anything for pkgdepend to find out that bash is required. So, yes, manual depend action seems to be required.

xen0l:
Hi,

and what about existing /etc/bash/bash_completion.example from bash? I guess it would be best to move this as well to the package if possible.

alarcher:
Hi,
I added the attributes for bash_completion.sh and an explicit dependency to shell/bash.
I am not sure what to do with /etc/bash/bash_completion.example because it might be redundant with the completions present in /ust/share/bash-completion/completions.
Best regards,

Aurelien

pyhalov:
Pehaps, we can just remove this file from bash manifest?

xen0l:
Let's do it then. But please do so in a separate commit.

alarcher:
I cleaned up the commit history. I will remove the bash_completion.example file in another pull request.

components/bash-completion/bash-completion.p5m
```diff
@@ -0,0 +1,608 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=.*\.py$ -> default pkg.tmp.autopyc false>
+
+depend type=require fmri=pkg:/shell/bash
+
+dir  path=etc
+dir  path=etc/profile.d
```
pyhalov, line 29: the same note - please, firstly add group=sys rule for profile.d and profile.d/\* to transforms/defaults

pyhalov:
please, modify pull request so that it can be automatically merged

alarcher:
Should be OK now.

PR #355, opened 2013-11-21
Add fortran to configure.mk

alarcher:
Adding FORTRAN compilers and flags to the configure options and environment in configure.mk using the newly introduced definitions in shared-macros.mk


pyhalov:
It generally looks good. However, I don't like to have "Merge" commits in the gate. Could you create clean pull request, which doesn't contain "merge" commits?

PR #360, opened 2013-11-24
Requested review for initial addition of mpich

alarcher:
Hi,
this is an initial proposal for the integration of mpich to oi-userland, mainly for reviewing purpose as I am not familiar with Github and IPS.

As pointed in an email sent to the mailing-list, we should consider an installation for a triple (MPI implementation, compiler, 32 or 64 bit), which requires an installation prefix which is dependent on the MPI implementation and the compiler used (gcc or studio).

The proposed scheme is to install the MPI bundle and compiled libraries in /usr/lib/$ARCH/$MPI/$COMPILER and name the IPS packages pkg:/library/$MPI/$COMPILER

The variants mpich/gcc and mpich/studio can a priori be built using the proposed files as the compiler is not hardcoded.

I provided two modulefiles which allow loading the environment (using environment-modules) for mpich/gcc/32 or mpich/gcc/64 if gcc is used for example:

alarcher@phainos> module avail

------------------------ /usr/share/Modules/modulefiles ------------------------
mpich/gcc/32/3.0.4  mpich/gcc/64/3.0.4
mpich/gcc/32/latest mpich/gcc/64/latest

alarcher@phainos> module list
Currently Loaded Modulefiles:
  1) mpich/gcc/64/latest

alarcher@phainos> which mpicc
/usr/lib/amd64/mpich/gcc/bin/mpicc

alarcher@phainos> module switch mpich/gcc/64 mpich/gcc/32

alarcher@phainos> module list
Currently Loaded Modulefiles:
  1) mpich/gcc/32/latest

alarcher@phainos> which mpicc
/usr/lib/mpich/gcc/bin/mpicc

The mpich and openmpi modules are exclusive i.e only one triple (MPI, compiler, 32/64) can be loaded in the environment.

I would like to make sure we agree on a reasonable way to provide libraries for different MPI bundle, so feel free to give me your feeling about it.
The pull request for bash-completion, environment-modules and mpich rely on the same addition of several variables in shared-macros.mk. I am not sure how to make it clean to avoid duplicate commits.
But if we merge them in the right order I can clean them one by one.

As soon as mpich is merged, I can work efficiently on adding the numerical libraries.
If time allows I might be able to add Paraview and Scilab  in the coming weeks.
Best regards,

Aurelien

### DIFF
--- a/components/mpich/Makefile
+++ b/components/mpich/Makefile
@@ -1,0 +1,89 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2013 Aurelien Larcher.  All rights reserved.
+#
+
+include ../../make-rules/shared-macros.mk
+include $(WS_MAKE_RULES)/mpi-macros.mk
+
+COMPONENT_NAME =	mpich/$(COMPILER)
+COMPONENT_VERSION =	3.0.4
+COMPONENT_FMRI =	library/$(COMPONENT_NAME)
+COMPONENT_PROJECT_URL =	http://www.mpich.org/
+COMPONENT_SRC_NAME = mpich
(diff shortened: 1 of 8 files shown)


pyhalov:
1) Please, remove *.la files.
2) Is delivering static libraries necessary or useful in any way? We usually avoid shipping them if it's not necessary (e.g. this library is delivered only as static and it is used by other packages).

alarcher:
I will remove the *.la files. For numerical libraries at the laboratory we usually chose static linking over dynamic but I am not sure if it is critical.

PR #368, opened 2013-12-02
Add /var and /kernel/drv macros to shared-macros.mk

alarcher:
I propose the addition of two paths to shared-macros, which will be further needed (by at least openafs):
- /var (for delivery of cache and logs directories)
- /kernel/drv (for the afs driver)

Thank you.


xen0l:
Could you add /usr/kernel/drv macro, so it's complete? Thanks

alarcher:
OK. I will add the 64 versions as well.
The USRLIBDIR64 macro is placed at the end of the list and thus not in the same order as the PROTO macros.
May I put it right after USRLIBDIR ?

xen0l:
Sure, do so. I checked how it got there and it seems that it was a chaos commit, which brought it multiple things.
Thanks for adding and fixing this.

alarcher:
To summuarize, I added:
- /var
- /kernel/drv
- /kernel/drv/$(MACH64)
- /usr/kernel/drv
- /usr/kernel/drv/$(MACH64)

I fixed the ordering that we discussed and added PROTOSBINDIR.

I wonder why commit cbf24f4 appears in the pull request, any idea ?

xen0l:
Changes looks good. 

Did you create branch from the latest changeset? I would bet  that commit wasn't merged into your branch, just commited, so Github thinks that it is a new commit. Could you try to merge that commit first?

alarcher:
Strange thing is ... yes, I created the branch from the latest changeset and it appears in the log of the branch.
I can create a new branch and a new pull request or do you have any other idea ?

alarcher:
Should be good.

xen0l:
Thanks.

PR #370, opened 2013-12-02
Initial add of openmpi version 1.6.5 and modulefiles

alarcher:
Hello,
Here is OpenMPI using the same recipe as for MPICH.

One question: should I set group=sys for /usr/lib/openmpi/gcc/etc/\* and /usr/lib/amd64/openmpi/gcc/etc/\* or should I leave group=bin ?

Best regards,

Aurelien

### DIFF
--- a/components/openmpi/Makefile
+++ b/components/openmpi/Makefile
@@ -1,0 +1,90 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2013 Aurelien Larcher.  All rights reserved.
+#
+
+include ../../make-rules/shared-macros.mk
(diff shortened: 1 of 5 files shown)


pyhalov:
Do I understand correctly that this package can be installed simultaneously with mpich?

alarcher:
Yes the package can be installed together with mpich.
Contrary to the solution adopted by several Linux distributions (including Debian if I am not mistaken), it is left to the user to decide which MPI implementation he/she wants to use, and can switch between them.
Libraries linked against one of the MPI toolchain should be delivered in the corresponding prefix.

The environment-modules that I pushed earlier allows to switch between them.
The current possibilities are: mpich/gcc/32, mpich/gcc/64, openmpi/gcc/32 and openmpi/gcc/64.
We could have studio as well, or people could publish themselves on their machine.
They are mutually exclusive thanks to a conflict rule in the modulefiles.

I started writing something at http://www.distrans.org/doku.php?id=openindiana:scientific to explain.

This method is commonly used to deliver different versions of numerical libraries and different toolchains on clusters so that user can load the environment they need.

alarcher:
If people know that want to select the same environment all the time they can add 'module add mpich/gcc/64' in ~/.modules

For applications like Paraview or Scilab we need to choose which default implementation we want to use and set the dependencies accordingly.

I think it is quite robust and not too overly fancy.

PR #371, opened 2013-12-02
Fixed missing "and_64" in build target for mpich

alarcher:
Due to a mix up between commits.
By chance, it is required by install_32_and_64 so it has built so far.

### DIFF
--- a/components/mpich/Makefile
+++ b/components/mpich/Makefile
@@ -62,7 +62,7 @@ CONFIGURE_OPTIONS+=	--enable-mpe
 
 COMPONENT_TEST_TARGETS = test
 
-build:		$(BUILD_32)
+build:		$(BUILD_32_and_64)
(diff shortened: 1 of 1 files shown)


pyhalov:
And what about changes in *.p5m manifests? I see only 32-bit libraries there...

pyhalov:
And what about changes in *.p5m manifests? I see only 32-bit libraries there...

alarcher:
Sorry I do not understand, the p5m file contains 64 librairies in "dir  path=usr/lib/$(MACH64)/mpich/gcc/lib" from line 81 to 112: https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/mpich/mpich.p5m What is missing ?

pyhalov:
sorry, haven't noticed

PR #375, opened 2013-12-05
Initial commit of libspectre 0.2.6

alarcher:
Needed by upcoming Evas for the Enlightenment project.
The libraries are currently shipped with evince from JDS.

I keep the pull request open and modify it, it should be taken into consideration after the following steps are in order:
<pyhalov>
1) rollback libspectre to 0.2.6 
2) add poppler to be compiled with g++ and delivered to /usr/g++
3) bring evince to oi-userland, make it use new poppler version
4) bring desktop tracker and gimp to oi-userland.
</pyhalov>

### DIFF
--- a/components/libspectre/Makefile
+++ b/components/libspectre/Makefile
@@ -1,0 +1,55 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.illumos.org/license/CDDL.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+# Copyright (c) 2013, Aurelien Larcher. All rights reserved.
+#
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		libspectre
+COMPONENT_VERSION=	0.2.6
+COMPONENT_FMRI=     	library/libspectre
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	http://www.freedesktop.org/wiki/Software/libspectre/
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:2f637c62322c8040514284c00f63a5c310a28801e7dcfbe2ba2791be4fac0dd3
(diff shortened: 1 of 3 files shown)


pyhalov:
There are two issues with this pacakge. 1) (easy one) why there is no 64-bit library version ? 2) To deliver this package you have to recompile evince (and possibly dependent packages), because libspectre 0.2.6 is delivered with evince.

alarcher:
Oh I did not see it. I pkg searched spectre instead of libspectre. Since it is delivered with evince I guess it is not useful to add it.

pyhalov:
I think the following sequence will allow to deliver packages without breaking existing systems. 
a) deliver libspectre 0.2.6 (the same version which is used in evince)
b) deliver poppler 0.4.15 (the same version which is used in evince)
c) deliver evince component. On package install poppler and libspectre will be also installed. If we are lucky, they will be compatible with current evince libraries and everything will work. If not, evince and dependent packages should be recompiled. The question is the following: does libpoppler and libspectre contain c++ code?

alarcher:
I was actually looking at poppler right now. Yes it poppler contains cpp wrappers so on my laptop I had recompiled it with g++ and installed in /usr/g++. libspectre is only C. What do you recommend ?

pyhalov:
There are packages which depend on evince: 
pkg:/image/editor/gimp
pkg:/library/desktop/search/tracker
pkg:/library/python-2/python-gnome-desktop-24
pkg:/library/python-2/python-gnome-desktop-26

python-gnome-desktop-24 can be marked obsolete. Other should be recompiled.

alarcher:
About the lack of 64-bit version, I thought that 64-bit should be delivered if necessary but not systematically.
Since libspectre serves for desktop applications isn't it sufficient to deliver the 32-bit version ?

pyhalov:
I'd prefer to see 64-bit version for each library if it is not very hard.

alarcher:
OK. So, are these steps correct ?
1) rollback libspectre to 0.2.6 and enable 64-bit (or should I try to build and test everything with 0.2.7 ?)
2) add poppler 0.14.5 component to be compiled with studio, because of the c++ wrappers
3) bring evince 2.30.3 to oi-userland
4) recompile desktop tracker and gimp or should gimp 2.6.12 be brought from jds to oi-userland as well ?

Should we maintain a page on the wiki to see with Milan Jurik which package should be moved from jds to userland ?

pyhalov:
I think that all packages from JDS should be moved to userland. The only question is a manpower. 
1) rollback libspectre to 0.2.6 andtry to build 64-bit version
2) add poppler (don't matter what version) to be compiled with g++ and delivered to /usr/g++
3) bring evince to oi-userland, make it use new poppler version
4) bring desktop tracker and gimp to oi-userland.

I'll take a look at python-gnome-desktop-26 (and if have spare time - on  other packages)

pyhalov:
BTW, thanks to Andrzej, https://github.com/aszeszo/userland/tree/oi/components-jds has some necessary components, they can be consulted and borrowed. But I'd prefer to split packages according their natural boundaries.

alarcher:
OK !  As libspectre was a dependency for Enlightenment/evas I will prioritize your short list and then go back to it.
Good, I will get started from Andrzej files then !
I can maintain a list of packages moved from jds to userland and the ones present in Andrzej repository if I get access to the wiki.

Thank you !

alarcher:
Just an update.
I remember now why I did not deliver a 64-bit version for libspectre, it is because ghostscript only ships 32-bit versions.
Cheers,

Aurelien

PR #383, opened 2013-12-11
Initial commit of scientific-py 2.8 (Requires Pyro to be added first !)

alarcher:
General scientific computing module, not to be mistaken with scipy.
Needed by upcoming FIAT.

### DIFF
--- a/components/scientific/python/scientific-py/scientific-py-26.p5m
+++ b/components/scientific/python/scientific-py/scientific-py-26.p5m
@@ -1,0 +1,199 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2013 Aurelien Larcher. All rights reserved.
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)-26@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification \
+    value=org.opensolaris.category.2008:Development/Python
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# force a dependency on the Python runtime
+depend type=require fmri=__TBD pkg.debug.depend.file=python2.6 \
+    pkg.debug.depend.path=usr/bin
+
+# force a dependency on the unversioned package
+depend type=require \
+    fmri=$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+
+link path=usr/bin/bsp_virtual \
+    target=bsp_virtual2.6 \
+    mediator=python mediator-version=2.6
+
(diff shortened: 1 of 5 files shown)


components/scientific/python/scientific-py/scientific-py-26.p5m
```diff
@@ -0,0 +1,199 @@
...
+    target=bsp_virtual2.6 \
+    mediator=python mediator-version=2.6
+
+link path=usr/bin/task_manager \
+    target=task_manager2.6 \
+    mediator=python mediator-version=2.6
+
+file path=usr/bin/bsp_virtual2.6 mode=0555 pkg.depend.bypass-generate=.*
+file path=usr/bin/task_manager2.6 mode=0555 pkg.depend.bypass-generate=.*
+file path=usr/include/python2.6/Scientific/PyBSP_API.h
```
pyhalov, line 43: Are you sure that headers should go here, but not to /usr/lib/python2.6/vendor-packages/Scientific/  subdirectory? 
As I see, Debian doesn't ship them at all.  /\* It's not a recommendation, I'm just asking */

components/scientific/python/scientific-py/scientific-py-26.p5m
```diff
@@ -0,0 +1,199 @@
...
+
+# force a dependency on the unversioned package
+depend type=require \
+    fmri=$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+
+link path=usr/bin/bsp_virtual \
+    target=bsp_virtual2.6 \
+    mediator=python mediator-version=2.6
+
+link path=usr/bin/task_manager \
```
pyhalov, line 37: In Debian  they seem to rename this binary to task_manager_pyscientific (likely, to avoid conflict with sympa). Does it have sense?

components/scientific/python/scientific-py/scientific-py-26.p5m
alarcher, line 43: I will look into it. I am just using the Python module along with FEniCS so I need to understand whether other software might need them.

components/scientific/python/scientific-py/scientific-py-26.p5m
alarcher, line 37: Yes I saw it as well. Another suggestion I saw was to move the task_manager from Sympa to /usr/sbin.
There are positive and negative aspects to both: the developer of ScientificPython seemed concern that renaming would break existing scripts but on the other end since Debian is widely used it and have an active Debian Science team it might be also good to follow their scheme for some packages.
What would your advice be ?

pyhalov, line 37: I just don't know how it is used. I checked, because this name is very generic. If you feel that renaming is safe, I'd prefer to rename it. If not - we'll return to this when someone will port Sympa.

alarcher:
I need to think a bit more about that... I am on short vacation right now. My impression is that Debian delivering /usr/bin/task_manager is quite specific as I could not find it in RedHat, Mageia and OpenSuse.
The 'task_manager' in ScientificPython is supposed to be used by regular users to submit jobs on the command line while the task_manager.pl in Sympa is to be used punctually by administrators. It would make sense to have Sympa task_manager.pl either in /usr/sbin or in a prefix dedicated to sympa as other distributions than Debian do. Thus in the end I feel that Debian's decision seems misguided.

pyhalov:
OK. Let it be so. And what about headers?

alarcher:
It seems that OpenSuse delivers the headers as a -devel package. We could do the same. I am sending a message to the developer of ScientificPython to know if it is necessary.

alarcher:
Hi Alexander,
I contacted Konrad Hinsen, the main developer of ScientificPython who told me that the headers should be delivered as they are used by some solvers and that he is actually discussing with the Debian Science team to add them.

components/scientific/python/scientific-py/scientific-py-26.p5m
alarcher, line 43: I checked that RPM-centric distributions ship the headers in this prefix.

PR #403, opened 2014-01-11
Add bacula

pyhalov:


pyhalov:
Bacula director requires additional configuration before startup - at least you should change passwords, perhaps change listen address to something reasonable and create default database using scripts in /etc/bacula/scripts/. 
You can select the database type which you are actually using with baculadb mediator (postgresql,sqlite3 or mysql) and install corresponding package bacula-common/DBNAME.

PR #424, opened 2014-01-18
Add legacy autoconf-2.13

netbsduser:
This adds the legacy autoconf version 2.13 which is still required to build some packages; one of these is SeaMonkey which I hope to add to oi-userland later.


components/autoconf/autoconf-2.13/Makefile
```diff
@@ -0,0 +1,64 @@
...
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION = ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+# CONFIGURE_OPTIONS and CONFIGURE_ENV are reset because
+# autoconf 2.13 doesn't like the defaults (in any case,
+# autoconf is but a set of scripts and needn't be compiled)
+CONFIGURE_OPTIONS   =		--prefix=/usr 
+CONFIGURE_OPTIONS  +=		--srcdir=.
```
xen0l, line 49: Why do you use . instead of $(SOURCE_DIR)?

netbsduser:
@xen0l - I use . as that is also used in the autoconf 2.68 Makefile.

PR #490, opened 2014-02-09
Add nano component

ghost:
Newer version than in SFW (2.2.6 vs 2.0.9), same config flags.


xen0l:
And resolve.deps file is not needed.

PR #491, opened 2014-02-10
Add indri

ghost:
Import from upstream/oracle/userland-gate

### DIFF
--- a/components/dnsmasq/Makefile
+++ b/components/dnsmasq/Makefile
@@ -1,0 +1,64 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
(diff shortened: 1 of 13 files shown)


components/dnsmasq/Makefile
```diff
@@ -0,0 +1,64 @@
...
+COMPONENT_BUILD_ARGS += CFLAGS="$(CFLAGS)"
+COMPONENT_BUILD_ARGS += PREFIX="/usr"
+COMPONENT_BUILD_ARGS += LOCALEDIR="/usr/share/locale"
+COMPONENT_BUILD_ARGS += LDFLAGS="$(LDFLAGS) -m64"
+COMPONENT_BUILD_TARGETS = "all-i18n"
+COMPONENT_INSTALL_ARGS += PREFIX="/usr"
+COMPONENT_INSTALL_TARGETS = "install-i18n"
+
+# common targets
+build:		$(BUILD_64)
```
pyhalov, line 56: Component should be 32bit  or (32bit and 64bit) if there are no strong reasons for it to be not. You can use isaexec to make startup scripts simple.

alarcher:
Should we just add 32bit builds ?

pyhalov:
why not? We still claim that we support 32-bit sytems. And I hope dnsmasq doesn't require 64bit address space

alarcher:
I meant that if it is OK I can update the branch to build 32bit only and be able to merge. ;)

pyhalov:
It's OK

alarcher:
This PR can be closed as changes are duplicated in #1257 and #1258.

PR #591, opened 2014-04-19
Add SPARC support to gcc47 component

netbsduser:
A simple set of changes - it simply sets the Makefile to use Sun Assembler when building with SPARC, and alters the manifests to allow packaging of a SPARC build.


pyhalov:
Please, update COMPONENT_REVISION. Set it to 1 if it's not already set.

netbsduser:
Have now updated COMPONENT_REVISION.

PR #628, opened 2014-05-13
Bug #4859 Add refresh and stop methods to rsyslog manifest

ggendel:
Please review changes to rsyslog.xml.  Somehow I lost the CDDL license header in the original commit.  The subsequent commits put it back.


xen0l:
Changes seems good to me. However, could you please re-submit, so we do not alter CDDL header?

ggendel:
It matches perfectly now.

xen0l:
Modified CDDL is part of your commit history, which I do not want to enter the repository. Could you please create new pull request with a single commit?

PR #844, opened 2014-10-20
Add check 0.9.14

alarcher:
More unit testing for C ...

### DIFF
--- a/components/check/Makefile
+++ b/components/check/Makefile
@@ -1,0 +1,79 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2014 Aurelien Larcher.  All rights reserved.
+#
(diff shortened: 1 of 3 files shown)


0xffea:
LGTM.

It has tests in the dir 'tests' but how to run them all?

alarcher:
gmake check runs the tests but it seems the profiling tests take a while, I am not sure if it is a good idea to activate them.

pyhalov:
If checks are passed, why not enable it? They run only on explicit request.

alarcher:
Checks added.
It seems some tests that are not supposed to be executed, according to the configuration, are and fail: the system still returns 7 tests PASSED and exits normally.

pyhalov:
Could you add some comment to makefile: like 
failed tests: a,b,c (or 14 failed tests...). So that someone who touches this component later will not be surprised about these failures?

alarcher:
Should we instead deliver a file for components with failed tests instead of commenting in the Makefile ? Maybe providing the log of failed tests would help pinpoint issues in the long run and identify easily which components have issues by find . -type f -name "*.test.failed" for instance.
Otherwise I do as you say and I will open a ticket containing more information.

pyhalov:
Honestly, I don't know. Perhaps, it's a topic for oi-dev ML. For now I think comment will be enough.

alarcher:
OK. My build zone is not functional now (mount error after an update). I will update asap.

alarcher:
Updated description of tests result.

PR #845, opened 2014-10-20
Add OpenJPEG 2.1.0

alarcher:
The patch is quite ugly, if you have a better idea.
The issue is that index_t is defined in sys/types.h and there is no way to hide it, since POSIX discourages type names with _t suffix, it should be fixed upstream.

### DIFF
--- a/components/openjpeg/Makefile
+++ b/components/openjpeg/Makefile
@@ -1,0 +1,68 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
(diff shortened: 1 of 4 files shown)


pyhalov:
I see that openjpip.h is not shipped. What if we just do something like ifdef-d s:index_t:index_param_t in the code? On other hand, if this header is not shipped, perhaps it's not so bad to do this #define.

alarcher:
I will have a look at it later this week, long overdue.

alarcher:
Updated.

PR #1016, opened 2015-02-10
Disable openblas for now.

pyhalov:
Build server can't build it:
getarch_2nd.c:12:35: error: ‘SGEMM_DEFAULT_UNROLL_M’ undeclared (first use in this function)


alarcher:
It seems related to a CPU detection failure according to similar issues reported on the project's bugtracker.

PR #1021, opened 2015-02-11
Force PRESCOTT architecture in OpenBLAS

alarcher:
Force PRESCOTT architecture as least common denominator as GENERIC is broken and up number of threads to 64 as Debian does.

The developer suggests using a least common denominator architecture until the issue is fixed:
https://github.com/xianyi/OpenBLAS/issues/502

### DIFF
--- a/components/components.ignore
+++ b/components/components.ignore
@@ -55,9 +55,6 @@
 /^visual-panels\/zconsole$/d
 /^gcc44/d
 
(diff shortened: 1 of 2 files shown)


pyhalov:
Please, also remove corresponding entry from components.ignore

alarcher:
Updated.

PR #1024, opened 2015-02-11
Initial add of harfbuzz 0.9.38

alarcher:
GLib support is not activated as requirement seems >= 2.36 while we have 2.28.


alarcher:
Requires graphite2

PR #1109, opened 2015-03-18
Fix nimbus download url.

schwer-q:
The /download at the end may confuse some CLI download tools.


pyhalov:
Which download tools? It is intended for userland-fetch.

schwer-q:
I don't remember well, it was a long time ago... But I already have downloaded the web page pointed by the /download url which is reserved for web browser. Maybe it was a SF issue and they may have fixed it. Conventionally, download tools don't the /download url.

pyhalov:
It's not oi-userland issue, so I close it.

PR #1251, opened 2015-06-21
Boost 1.58.0: MPI + 64bit support

alarcher:
Update to latest release, added 64 bit build and MPI support compiled with MPICH.
Fix to broken html documention.


pyhalov:
Please, comment on  components/boost/patches/boost-gpp-02-usr-include.patch.  Why have you removed changes to config_all.hpp? Also, please preserve comment to the patch.

alarcher:
OK will add the comment. The patch cannot be applied as the file changed in Boost so I had to make the diff again. Changes to config_all.hpp were preserved.

pyhalov:
I see 
```C
#      if !defined(BOOST_TR1_DISABLE_INCLUDE_NEXT) && !defined(__ICC) \
            && (defined(__FreeBSD__) || defined(linux) || defined(__linux) || defined(__linux__) || defined(__GNU__) || defined(__GLIBC__))
         // Disable use of #include_next on Linux as typically we are installed in a directory that is searched
         // *after* the std lib include path:
#        define BOOST_TR1_DISABLE_INCLUDE_NEXT
#      endif
```
in config_all.hpp. We should define BOOST_TR1_DISABLE_INCLUDE_NEXT also on OI.

alarcher:
Right, let me rebuild with the updated patch to make sure and I will then update the pull request.

alarcher:
Updated. Make sure mpich/gcc and mpich/gcc/modulefiles are installed on the buildserver.

PR #1257, opened 2015-06-26
Add indri 5.4

alarcher:
Split previous pull request #491, no change in Solaris userland since then.

### DIFF
--- a/components/indri/Makefile
+++ b/components/indri/Makefile
@@ -1,0 +1,68 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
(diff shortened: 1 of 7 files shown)


components/indri/Makefile
```diff
@@ -0,0 +1,72 @@
...
+COMPONENT_ARCHIVE_URL	= http://sourceforge.net/projects/lemur/files/lemur/$(COMPONENT_SRC)/$(COMPONENT_SRC).tar.gz/download
+COMPONENT_BUGDB		= library/indri
+
+include ../../make-rules/prep.mk
+include ../../make-rules/configure.mk
+include ../../make-rules/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION	 = $(CLONEY) $(SOURCE_DIR) $(@D)
+COMPILER			 = gcc
+CONFIGURE_OPTIONS		+= "CFLAGS=$(CFLAGS)"
```
pyhalov: It seems these 3 lines can be removed, we have them in configure.mk. COMPILER=gcc is also not necessary.

alarcher:
Updated.

PR #1258, opened 2015-06-26
Add dnsmasq

alarcher:
Second part of pull request #491, updated patches to recent Solaris userland.
Only 32-bit according to previous comments.

### DIFF
--- a/components/dnsmasq/Makefile
+++ b/components/dnsmasq/Makefile
@@ -1,0 +1,68 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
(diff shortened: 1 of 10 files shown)


components/dnsmasq/Makefile
```diff
@@ -0,0 +1,69 @@
...
+TPNO=			16307
+
+include ../../make-rules/prep.mk
+include ../../make-rules/justmake.mk
+include ../../make-rules/ips.mk
+
+# Enable ASLR for this component
+ASLR_MODE = $(ASLR_ENABLE)
+
+COMPONENT_BUILD_ENV += CC="$(CC)"
```
pyhalov: It seems this section also can be simplified.

alarcher: Since it is just a Makefile, build args have to be passed.
I just removed the redundant COMPONENT_BUILD_ENV line.

PR #1275, opened 2015-07-14
Add bullet 2.83

alarcher:


components/bullet/Makefile
```diff
@@ -0,0 +1,64 @@
...
+#
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		bullet
+COMPONENT_VERSION=	2.83.5
+COMPONENT_FMRI=     	library/bullet
+COMPONENT_CLASSIFICATION=System/Libraries
+COMPONENT_SRC=		bullet3-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	http://bulletphysics.org/
+COMPONENT_ARCHIVE=	$(COMPONENT_VERSION).tar.gz
```
pyhalov: Please, don't call archive 2.83.5.tar.gz, this is ugly (especially when it's stored in USERLAND_ARCHIVES directory). Use something like bullet-2.83.5.tar.gz for COMPONENT_ARCHIVE and fix COMPONENT_ARCHIVE_URL instead to use $(COMPONENT_VERSION).tar.gz or similar.

alarcher: OK, I did not realize that the tarball would be renamed to $(COMPONENT_ARCHIVE), I should add some note to make it clear on the wiki.

alarcher:
Updated.

PR #1315, opened 2015-07-25
Fix imagemagick to 6.8.3-10

alarcher:
Some comments:
- pkg-config files for the current component are incorrect for Magick++
- archive url cannot be reached, neither can the newer in solaris-userland so fallback to official repository tarball which is a different version
- docdir path has changed to ImageMagick-%{major_version} instead of ImageMagick-%{version}
- executables are now symlinked from /usr/lib/%{arch}/ImageMagick-$(COMPONENT_VERSION)/bin-Q16 in /ur/bin/%{arch}
- symlinks in /usr/sfw are kept but should they be dropped eventually ?


alarcher:
Missing mode=0555 for some files. Fixing now.

pyhalov:
Is it ready now? Where pkg-config is fixed or was it fixed in imagemagick itself ?

alarcher:
It is fixed in ImageMagick itself, I used this version to build inkscape.

PR #1316, opened 2015-07-25
Add libsamplerate 0.1.8

alarcher:
All tests passed.


alarcher:
Note: FFTW-3 should be ported from sfw-gate at some point.

PR #1336, opened 2015-07-27
Add librevenge 0.0.2

alarcher:
LibreOffice prerequirement.

### DIFF
--- a/components/librevenge/Makefile
+++ b/components/librevenge/Makefile
@@ -1,0 +1,56 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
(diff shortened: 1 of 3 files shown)


components/librevenge/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+CONFIGURE_OPTIONS+=	--enable-shared
+CONFIGURE_OPTIONS+=	--disable-static
+CONFIGURE_OPTIONS+=	--disable-werror
+
+build: $(BUILD_32)
```
pyhalov, line 47: Why only 32-bit?

alarcher, line 47: I assumed that we will only provide LibreOffice as 32bit and it is the only consumer of it.

pyhalov, line 47: I prefer  to have both 32-bit and 64-bit libraries available, but don't have strong opinion. I any case someone who needs 64-bit library can deal with it later.

PR #1448, opened 2015-09-29
Initial commit of recipe for gnome-connection-manager-1.1.0

jimklimov:
The gnome-connection-manager is a tabbed/windowed session manager for X11, similar to SecureCRT or MTPutty on Windows. It can remember session passwords, post-login scriptlets, etc. to simplify work with multiple remote hosts.

The committed component has passed `gmake clobber publish` and `pkg install`

### DIFF
--- a/components/gnome-connection-manager/Makefile
+++ b/components/gnome-connection-manager/Makefile
@@ -1,0 +1,59 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2015 Jim Klimov
+#
+
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		gnome-connection-manager
+COMPONENT_NAME_SHORT=	gcm
+COMPONENT_VERSION=	1.1.0
+COMPONENT_PROJECT_URL=	http://kuthulu.com/gcm
+COMPONENT_SRC=		$(COMPONENT_NAME_SHORT)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)_$(COMPONENT_VERSION)_all.tgz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:71a128676d16153c735499cec9f0a1af1e2dc1107ed5c57cea1994eadc120c4c
+COMPONENT_ARCHIVE_URL=	$(COMPONENT_PROJECT_URL)/$(COMPONENT_ARCHIVE)
+COMPONENT_SUMMARY=	GCM - an Advanced SSH connections manager for X/Windows
+COMPONENT_FMRI=		terminal/gnome-connection-manager
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_LICENSE=	GPLv3
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
+include ../../make-rules/prep.mk
+include ../../make-rules/ips.mk
+
+CLEAN_PATHS+= $(BUILD_DIR)
+
+# rules are very simple, because the project is a script - just deliver and run
+# unpacked directory is among search paths for the p5m installation anyway
+
+$(BUILD_DIR)/.built:	$(SOURCE_DIR)/.prep
+	$(RM) -r $(@D) ; $(MKDIR) $(@D)
+	$(CLONEY) $(SOURCE_DIR) $(@D)
+	find $(@D) -name '*.pyc' -exec rm -f '{}' \;
+	$(PYTHON) -m compileall $(@D)
+	$(TOUCH) $@
+
+$(BUILD_DIR)/.installed:	$(BUILD_DIR)/.built
+	$(TOUCH) $@
+
+build:		$(BUILD_DIR)/.built files/gcm Makefile
+
+install:	$(BUILD_DIR)/.installed Makefile
+
+test:		$(NO_TESTS)
+
+BUILD_PKG_DEPENDENCIES =	$(BUILD_TOOLS)
+
+include ../../make-rules/depend.mk
--- a/components/gnome-connection-manager/files/gcm
+++ b/components/gnome-connection-manager/files/gcm
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# Simple start wrapper (C) 2015 by Jim Klimov for gnome-connection-manager
+# Install this script as /usr/bin/gcm
+# Requires the GCM itself: http://kuthulu.com/gcm/?module=download
+# unpacked as /usr/share/gcm (may be a symlink to actual version)
+# and its dependencies (python v2.6+, glade, X11, etc.)
+
+if [ -z "$DISPLAY" ]; then
+	echo "FATAL: The DISPLAY variable is not set and is required!" >&2
+	exit 1
+fi
+
+# Empirically it seems that for GCM interface to be properly localized,
+# the LANGUAGE should be just the two-character code
+if [ -z "${LANGUAGE-}" ]; then
+	[ -n "${LANG-}" ] && \
+		LANGUAGE="$LANG" || \
+		LANGUAGE=en
+fi
+[ "$LANGUAGE" = en ] || \
+	LANGUAGE="`echo "$LANGUAGE" | sed 's,^\(..\).*$,\1,'`"
+
+export LANGUAGE LANG LC_ALL DISPLAY
+echo "Starting gnome-connection-manager with LANGUAGE=$LANGUAGE ..."
+locale 2>/dev/null
+exec /usr/share/gnome-connection-manager/gnome_connection_manager.py "$@"
+
--- a/components/gnome-connection-manager/gnome-connection-manager.p5m
+++ b/components/gnome-connection-manager/gnome-connection-manager.p5m
@@ -1,0 +1,63 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2015 Jim Klimov
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=usr/share/applications/.* ->  default restart_fmri svc:/application/desktop-cache/desktop-mime-cache:default>
+<transform file path=usr/share/applications/.* ->  default restart_fmri svc:/application/desktop-cache/mime-types-cache:default>
+
+# Add-on files for this package
+file files/gcm path=usr/bin/gcm mode=555
+file files/gcm.desktop path=usr/share/applications/gcm.desktop
+file icon.png path=usr/share/pixmaps/gcm.png
+# Deliver the icon into the path expected by script as well
+# so it goes well into the window title bar (TODO: sym/hard-link?)
+file icon.png path=usr/share/gnome-connection-manager/icon.png
+# This is not from tarball but is compiled by our Makefile
+file gnome_connection_manager.pyc path=usr/share/gnome-connection-manager/gnome_connection_manager.pyc
+
+# Files delivered by source tarball, as of version 1.1.0, listed via:
+#   find . -type f | sed 's,^./\(.*\)$,file \1 path=usr/share/gnome-connection-manager/\1,'
+file donate.gif path=usr/share/gnome-connection-manager/donate.gif
+file gnome_connection_manager.py path=usr/share/gnome-connection-manager/gnome_connection_manager.py mode=0555
+file gnome-connection-manager.glade path=usr/share/gnome-connection-manager/gnome-connection-manager.glade
+file lang/de_DE.po path=usr/share/gnome-connection-manager/lang/de_DE.po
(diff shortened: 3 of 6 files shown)


components/gnome-connection-manager/Makefile
```diff
@@ -0,0 +1,83 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
```
pyhalov: Please, use illumos header , look for example at x11/xf86-video-ast/Makefile header.

components/gnome-connection-manager/Makefile
```diff
@@ -0,0 +1,83 @@
...
+COMPONENT_NAME=		gnome-connection-manager
+COMPONENT_NAME_SHORT=	gcm
+COMPONENT_VERSION=	1.1.0
+COMPONENT_PROJECT_URL=	http://kuthulu.com/gcm
+COMPONENT_SRC=		$(COMPONENT_NAME_SHORT)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)_$(COMPONENT_VERSION)_all.tgz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:71a128676d16153c735499cec9f0a1af1e2dc1107ed5c57cea1994eadc120c4c
+COMPONENT_ARCHIVE_URL=	$(COMPONENT_PROJECT_URL)/$(COMPONENT_ARCHIVE)
+COMPONENT_BUGDB=	utility/$(COMPONENT_NAME)
```
pyhalov: BUGDB is Oracle thing, we don't need it

components/gnome-connection-manager/Makefile
```diff
@@ -0,0 +1,83 @@
...
+COMPONENT_SUMMARY=	Advanced SSH connections manager for X11
+COMPONENT_FMRI=		terminal/gnome-connection-manager
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_LICENSE=	GPLv3
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
+include ../../make-rules/prep.mk
+include ../../make-rules/justmake.mk
```
pyhalov: Why do you need justmake.mk if you redefine $(BUILD_32) and $(INSTALL_32) targets? 
Perhaps, it's better to remove justmake.mk and simplify standard targets - e.g., remove PARFAIT rules, 
$(INSTALL_32) also seems overcomplicated. BTW, why use tar in one cases and $(CP) in other?
Just an idea: sometimes to avoid incorporating such rules in component Makefile you can deliver Makefile in files/Makefile and copy it to sources after extraction.

components/gnome-connection-manager/gnome-connection-manager.p5m
```diff
@@ -0,0 +1,54 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# Add-on for this package
+file path=usr/bin/gcm
+# Files delivered by source tarball, as of version 1.1.0, listed via:
+#   find . -type f | sed 's,^./,file path=usr/share/gnome-connection-manager/,'
+file path=usr/share/gnome-connection-manager/donate.gif
+file path=usr/share/gnome-connection-manager/gnome_connection_manager.py
```
pyhalov: You don't set execute bit on the file. This would allow pkgdepend to find dependencies.

components/gnome-connection-manager/resolve.deps
```diff
@@ -0,0 +1,8 @@
+system/library
```
pyhalov: oi-userland ignores resolve.deps files, you can remove it.

components/gnome-connection-manager/files/gcm
```diff
@@ -0,0 +1,28 @@
...
+		LANGUAGE="$LANG" || \
+		LANGUAGE=en
+fi
+[ "$LANGUAGE" = en ] || \
+	LANGUAGE="`echo "$LANGUAGE" | sed 's,^\(..\).*$,\1,'`"
+
+export LANGUAGE LANG LC_ALL DISPLAY
+echo "Starting gnome-connection-manager with LANGUAGE=$LANGUAGE ..."
+locale 2>/dev/null
+exec /usr/bin/python /usr/share/gnome-connection-manager/gnome_connection_manager.py "$@"
```
pyhalov: Do we have all required dependencies for python 2.7? If we don't, I think this should be explicit call to python2.6

pyhalov:
And one more note - please, merge all your changes in one commit.

components/gnome-connection-manager/gnome-connection-manager.p5m
```diff
@@ -0,0 +1,54 @@
...
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2014 Alexander Pyhalov. All rights reserved.
```
pyhalov: Please, remove my copyright, and also remove ". All rights reserved." from your (it's preferred for illumos code, we follow the same rule).

components/gnome-connection-manager/gnome-connection-manager.p5m
```diff
@@ -0,0 +1,54 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# Add-on for this package
+file path=usr/bin/gcm
```
pyhalov: I suppose we also need .desktop file and some icon

jimklimov:
Hello Alexander, 

First of all, thanks for the thorough review - as one can see, this is my first foray into this part of OI project and its traditions ;)

Regarding the first few formal-acceptance comments: I used some existing components (gnome-terminal, cvs, maybe a few others) as examples that I copied and tailored for GCM. As a result, their license headers, copyrights, variables, file sets etc. were used as a starting point. Of course, these can be tailored further to fit the modern requirements of the project.

Given the reasonable desire of one commit rather than a lot, I'll try to merge this and subsequent work on this PR into one patchset, or will spawn another branch to PR the one changeset when its contents will become acceptable ;)

I'll look at other comments and practical responses to them some time later in the week, hopefully.

Jim

components/gnome-connection-manager/Makefile
jimklimov: I'll see about that. Are there any other trivial components which are essentially deliveries of tarballed contents, no compilation/configuration/etc. to use as a starting point and inspiration? :)

In particular, is the expected "Makefile API" that it provides the custom "make build install test" recipes, including trivial no-ops, and is not required to follow/use the common make-rules?

Practically, I saw that `justmake` is nearly the most trivial of those includable files, though it still needed a bit too much for this tarball - e.g. the Makefile to be present in the project sources, so it could proceed to "make install" and "make build (or equivalent)". So I overrode the two recipes which failed to build and install GCM in their original form. and probably kept the rest of common-structure rules in place.

components/gnome-connection-manager/gnome-connection-manager.p5m
jimklimov: I tried mode=755 and the make process failed because pkgdepend's python inspector could not determine which version of Python to use - no explicit filename in shebang, and no hints in the code. I'm not a Python coder, so have little input on that, except possibly providing a patch to change the shebang to an explicit version. Currently I went the other way, keeping the script non-executable and calling `/usr/bin/python /path/to/script` from the `/usr/bin/gcm` wrapper.

The GCM project states (via README.txt in the tarball) that it needs "python-2.6 or newer". I'm not sure it is nice to explicitly pull in a dependency on the older version which we might want to get rid of soon, when 2.7 is also available in Hipster repos, and 3.x are in upstream. Currently I'm not at the PC where the work was done, but I think it has both python-26 and python-27 runtimes installed, and whatever is default via `/usr/bin/python` - suffices to run the script there.

components/gnome-connection-manager/files/gcm
jimklimov: Where/how should I list the rependencies, especially if `resolve.deps` is not used? 
Some explicit list in the *.p5m manifest? Or rely on `pkgdepend` and similar magic alone?
This is a matter of also requiring (some version of) ssh, glade and gtk2 and their python bindings, and X11 in general ;)

components/gnome-connection-manager/gnome-connection-manager.p5m
jimklimov: Can do. As stated earlier, this was a port of header from an existing component ;)

components/gnome-connection-manager/gnome-connection-manager.p5m
jimklimov: Any helpful pointers (how to define that in the manifest) that you could quickly think of? :)
Is it okay of the common icons (I'm thinking about one of the multi-windowed ones).

components/gnome-connection-manager/Makefile
pyhalov: You can look at components/docbook/docbook-dsssl or pgloader, former seems me a bit more cleaner.

components/gnome-connection-manager/gnome-connection-manager.p5m
pyhalov: Yes, usually you have to set explicit python version. I expect, GCM requires some GUI libraries, and we provide some of them only for python2.6. It's OK to patch it to use /usr/bin/python2.6 for now, as it's the default version now (we can't switch default because illumos-gate still depends on python 2.6). When default is switched. we'll correct this component.
BTW, we don't usually use mode=0755, mode=0555 is preferrable (and is checked by pkglint).

components/gnome-connection-manager/files/gcm
pyhalov: You should mostly rely on pkgdepend magic, and when it fails add 
depend actions (relying on pkgdepend, like in quilt/quilt.p5m or explicit like in gnome-applets/gnome-utility-applets.p5m).

components/gnome-connection-manager/gnome-connection-manager.p5m
pyhalov: Look at gkrellm/gkrellm.p5m, it delivers external desktop and icon files. Also note smf actuator for desktop-mime-cache. It's necessary for application to appear in gnome panel menu immediately, without user logout/login. Also you probably have to trigger svc:/application/desktop-cache/icon-cache:default.

jimklimov:
Hopefully, all comments have been addressed (except one changeset per PR).

components/gnome-connection-manager/gnome-connection-manager.p5m
```diff
@@ -0,0 +1,61 @@
...
+file lang/ko_KO.po path=usr/share/gnome-connection-manager/lang/ko_KO.po
+file lang/ko/LC_MESSAGES/gcm-lang.mo path=usr/share/gnome-connection-manager/lang/ko/LC_MESSAGES/gcm-lang.mo
+file lang/pl_PL.po path=usr/share/gnome-connection-manager/lang/pl_PL.po
+file lang/pl/LC_MESSAGES/gcm-lang.mo path=usr/share/gnome-connection-manager/lang/pl/LC_MESSAGES/gcm-lang.mo
+file lang/pt_BR.po path=usr/share/gnome-connection-manager/lang/pt_BR.po
+file lang/pt/LC_MESSAGES/gcm-lang.mo path=usr/share/gnome-connection-manager/lang/pt/LC_MESSAGES/gcm-lang.mo
+file lang/ru_RU.po path=usr/share/gnome-connection-manager/lang/ru_RU.po
+file lang/ru/LC_MESSAGES/gcm-lang.mo path=usr/share/gnome-connection-manager/lang/ru/LC_MESSAGES/gcm-lang.mo
+file pyAES.py path=usr/share/gnome-connection-manager/pyAES.py
+file pyAES.pyc path=usr/share/gnome-connection-manager/pyAES.pyc
```
pyhalov, line 59: I'm worried that we don't compile *.py files during installation, but deliver compiled files. pyc files format is not compatible between python versions.

components/gnome-connection-manager/Makefile
```diff
@@ -0,0 +1,60 @@
...
+##include ../../make-rules/justmake.mk
+include ../../make-rules/ips.mk
+
+# rules are very simple, because the project is a script - just deliver and run
+# unpacked directory is among search paths for the p5m installation anyway
+
+$(BUILD_DIR)/.built:	$(SOURCE_DIR)/.prep
+	$(RM) -r $(@D) ; $(MKDIR) $(@D)
+	$(CLONEY) $(SOURCE_DIR) $(@D)
+	$(COMPONENT_PRE_BUILD_ACTION)
```
pyhalov: COMPONENT_PRE/POST_BUILD_ACTION are empty and so not needed.

components/gnome-connection-manager/Makefile
```diff
@@ -0,0 +1,60 @@
...
+	$(COMPONENT_POST_BUILD_ACTION)
+	$(TOUCH) $@
+
+$(BUILD_DIR)/.installed:	$(BUILD_DIR)/.built
+	$(COMPONENT_PRE_INSTALL_ACTION)
```
pyhalov: The same is true for PRE/POST_INSTALL ACTION

components/gnome-connection-manager/Makefile
```diff
@@ -0,0 +1,60 @@
...
+    sha256:71a128676d16153c735499cec9f0a1af1e2dc1107ed5c57cea1994eadc120c4c
+COMPONENT_ARCHIVE_URL=	$(COMPONENT_PROJECT_URL)/$(COMPONENT_ARCHIVE)
+COMPONENT_SUMMARY=	GCM - an Advanced SSH connections manager for X/Windows
+COMPONENT_FMRI=		terminal/gnome-connection-manager
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_LICENSE=	GPLv3
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
+include ../../make-rules/prep.mk
+##include ../../make-rules/justmake.mk
```
pyhalov: please, remove unnecessary commented code

jimklimov:
Oops, my OI VM date seems stale... anyhow, the latest 3 commits should fulfill the suggestions from the last comments.

pyhalov:
And the last issue - gmake clean doesn't work. You have to add something like 
CLEAN_PATHS += $(BUILD_DIR) to Makefile

jimklimov:
Also done. Should I squash the multiple commits into one (e.g. make another PR, or do `git rebase`)?

pyhalov:
Yes, sqash commits and do force push.

jimklimov:
Eh, taking a timeout till local evening: apparently, when I squashed and
force-pushed another branch, it also overwrote this one with whatever was
available on the workstation of the moment. Gotta get back to home laptop
to rectify the real latest state, and squash into single commit, the
changesets relevant to ccache.

jimklimov:
Squashed for the merge

PR #1471, opened 2015-10-08
ccache 3.2.3: initial commit

jimklimov:
Compiles without warnings, package installs OK.

Verified to work: original compilation (of ccache itself) took 25 sec several times; then with Makefile edited to redefine `CC = /usr/bin/ccache /usr/gcc/4.8/bin/gcc` initial recompilation took 33 sec (part of the added overhead could be ccache's acquaintance with the newly encountered compiler - checksum, flag support and other metadata saved into the cache), several subsequent runs took 2-2.5 sec.

Package creates a /usr/lib/ccache directory with symlinks to several GCC filenames in order to simplify usage of the cache (just prepend it into the PATH).

Recommended for build farms, CI systems and similar projects that build same sources more than once. Hopefully will speed up cycles for illumos-gate and oi-userland among others ;)

Can be used with clang/llvm, crosscompilers, etc.

### DIFF
--- a/components/ccache/Makefile
+++ b/components/ccache/Makefile
@@ -1,0 +1,51 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2015 Jim Klimov
+#
+
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		ccache
+COMPONENT_VERSION=	3.2.3
+COMPONENT_PROJECT_URL=	http://ccache.samba.org/
+COMPONENT_DOWNLOAD_URL=	http://samba.org/ftp/$(COMPONENT_NAME)/
+COMPONENT_DOCUMENTATION_URL =	http://ccache.samba.org/manual.html
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.bz2
+COMPONENT_ARCHIVE_HASH=	\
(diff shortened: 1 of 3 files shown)


components/ccache/Makefile
```diff
@@ -0,0 +1,51 @@
...
+COMPONENT_FMRI=		developer/ccache
+COMPONENT_CLASSIFICATION=	Development/C
+COMPONENT_LICENSE=	"GPLv3,murmurhash,papowell/jhweiss,PostgreSQL,Python,zlib"
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
+include ../../make-rules/prep.mk
+include ../../make-rules/configure.mk
+include ../../make-rules/ips.mk
+
+CONFIGURE_OPTIONS  +=           --prefix=/usr
```
pyhalov: Is it really required? CONFIGURE_OPTIONS should include this by default

jimklimov: As in other cases - copypaste from existing components. On another hand, I
kind of hardcode the expected paths in p5m (symlinks), so for certainty
might as well force it here.

## 

09.10.2015 14:18 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov: Just remove this line, it shouldn't be necessary as it is defined in configure.mk

jimklimov:
Also done. Should I squash the multiple commits into one (e.g. make another PR, or do `git rebase`)?

pyhalov:
You can squash them and do force push to this branch

components/ccache/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_CLASSIFICATION=	Development/C
+COMPONENT_LICENSE=	"GPLv3,murmurhash,papowell/jhweiss,PostgreSQL,Python,zlib"
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
+include ../../make-rules/prep.mk
+include ../../make-rules/configure.mk
+include ../../make-rules/ips.mk
+
+CONFIGURE_OPTIONS  +=           --sysconfdir=/etc
+CONFIGURE_OPTIONS  +=           CFLAGS="$(CFLAGS) -D_GNU_SOURCE -D__EXTENSIONS__"
```
pyhalov: perhaps, change to CFLAGS += - "-D_GNU_SOURCE -D__EXTENSIONS__" several lines earlier ?

jimklimov:
Squashed for the merge

PR #1473, opened 2015-10-13
incorporations 0.151.1.8 -> 2013.0.0.0 for 151a9 -> hipster update

hrosenfeld:
All of this is about updating OI 151a9 to latest hipster.

I have a script using pkgmogrify that bumps all old 0.151.1.8 versions to 2013.0.0.0 in a hipster pkg repo. I have tested image-update from a 3 months old hipster as well as from 151a9, although I won't claim the 151a9 update is completely bug-free. It works fine on my laptop, though, except for some completely unrelated issues with some of the newer versions in hipster that have nothing to do with the update procedure per se.

This change is to adjust X-incorporation, admin-incorporation, nspg-incorporation, and solaris_re-incorporation to incorporate other packages at version 2013.0.0.0 instead of the old version 0.151.1.8. This will be necessary when they will be rebuilt against the new 2013.0.0.0 versions.

For completeness, here's the script that I've been using:

``` shell.ksh
#!/usr/bin/ksh93

cd /export/pkg

cat >/tmp/bump-version.mog <<EOF
<transform set name=pkg.fmri -> edit value '-0\.151\.1\.8.*:' '-2013.0.0.0:'>
<transform depend type=incorporate -> edit fmri '-0\.151\.1\.8.*' '-2013.0.0.0'>
EOF

pkgrecv -s /export/pkg/hipster-2015 -d incoming --raw $(pkgrecv -s /export/pkg/hipster-2015 --newest | grep -v userland-incorporation)

for pdir in incoming/*/* ; do
        pkgmogrify $pdir/manifest /tmp/bump-version.mog > $pdir/manifest.newpub
        pkgsend -s /export/pkg/testrepo publish -d $pdir $pdir/manifest.newpub
done

/export/ws/oi-userland/tools/userland-incorporator --repository /export/pkg/testrepo --version=0.5.11,5.11-2015.0.2.0 -p openindiana.org -c userland --destdir incoming
pkgsend -s /export/pkg/testrepo publish -d incoming incoming/userland-incorporation.p5m
```

It should be obvious that the paths will need adjusting.


jimklimov:
I am a bit worried about "usability" of this change: since the software versions and binaries remain the same, I guess, shouldn't we refer to OI-dev versioning as well so users can easily distinguish what fits where? E.g. use "2013.151.1.8" rather than "2013.0.0.0" so one sees that this overrides oi_151a8 level? (it was "8" and "8", right? or the number-shift remained, with "8" vs. "9" in the packaging vs. release naming?)

hrosenfeld:
The version 2013.0.0.0 was suggested by Alexander Pyhalov based on the fact that the hipster project started in January 2013. Technically it can be anything thats greater than 0.151.*.

Since the goal is to allow updating from 0.151.1.9 to hipster encoding 151.1.8 in the version would IMHO just create confusion.

pyhalov:
I see the following issue with this: you haven't fixed build number in meta-packages/gui-install/slim_install.p5m and meta-packages/history/*.p5m, they should be bumped from 0.151.1.8 to 2015.0.2.0, so that meta-packages which we publish overwrite binary packages in repository. One thing to check here is if any package from this set is restricted by some incorporate dependency.

hrosenfeld:
I see four packages that set their version explicitly and which are not "history":

```
components/pgbouncer/pgbouncer83.p5m:set name=pkg.fmri value=pkg:/database/postgres-83/pgbouncer@1.1.2,5.11-0.151.1.8.1
components/postgresql-jdbc/postgresql83-jdbc.p5m:set name=pkg.fmri value=pkg:/database/postgres-83/connector/jdbc@8.3.603,5.11-0.151.1.8.1
components/meta-packages/amp/web_amp.p5m:set name=pkg.fmri value=pkg:/web/amp@0.5.11,5.11-0.151.1.8
components/meta-packages/gui-install/slim_install.p5m:set name=pkg.fmri value=pkg:/slim_install@0.1,5.11-0.151.1.8
```

I guess I need to change all of them to use $(IPS_COMPONENT_VERSION),$(BUILD_VERSION) as version?

Is changing all the history packages also necessary, given that they are obsoleted and empty?

pyhalov:
I think mentioned manifests should be moved to components/meta-packages/history. I'd prefer to set build version explicitly to 2015.0.2.0 (current build number), avoid using $(IPS_COMPONENT_VERSION),$(BUILD_VERSION), as we don't want user to think that they are updated on the next build number change. Also using IPS_COMPONENT_VERSION is incorrect as components/meta-packages/history/Makefile doesn't define COMPONENT_VERSION. This should also be done for all manifests in components/meta-packages/history (even for marked obsolete or renamed). Else one can't recreate our repository without black magic.

hrosenfeld:
The additional commit changes all manifests in components/meta-packages/history and also the four others mentioned before.

Moving pgbouncer83.p5m, postgresql83-jdbc.p5m, and web_amp.p5m to components/meta-packages/history should be done separately. What shall I do about the old manifests for entire? Do they need to be changed, too?

pyhalov:
I don't know. If you don't touch them, noone can create ISO image like old one. But I doubt this is strictly necessary, as old entires belong to history. What do you think will be the best?

hrosenfeld:
I'd say that anyone wanting to build an old iso can checkout an old version of oi-userland without this change.

components/meta-packages/history/SUNWPython-extra.p5m
```diff
@@ -13,6 +13,6 @@
 # Copyright 2014 Andrzej Szeszo.  All rights reserved.
 #
 
-set name=pkg.fmri value=pkg:/SUNWPython-extra@0.5.11,5.11-0.151.1.8.1
+set name=pkg.fmri value=pkg:/SUNWPython-extra@0.5.11,5.11--2015.0.2.0
```
pyhalov: Something is wrong here (and in other files - two dashes instead of one)

PR #1489, opened 2015-11-01
Make perl headers Sun Studio friendly

mtelka:


pyhalov:
Please, also bump COMPONENT_REVISION in components/perl/perl522/Makefile

mtelka:
Done

pyhalov:
Shouldn't patch also handle __SUNPRO_CC?

mtelka:
No idea.  I did this to be able to build illumos-gate on recent hipster using Sun Studio.  I didn't notice that I need to do the same for __SUNPRO_CC so I didn't do it.  Maybe it is needed maybe not, I really do not know.  If it is needed we could do it later easily.

PR #1497, opened 2015-11-05
Add multitail package, with optional xclip support added

ivanrichwalski:
I enabled xclip support, but multitail does not depend on xclip being installed since it just calls the binary if it's available and enabled in /etc/multitail.conf


components/multitail/multitail.p5m
```diff
@@ -0,0 +1,49 @@
...
+
+dir  path=etc
+file etc/multitail.conf.new path=etc/multitail.conf preserve=renamenew
+dir  path=etc/multitail
+file etc/multitail/colors-example.pl path=etc/multitail/colors-example.pl
+file etc/multitail/colors-example.sh path=etc/multitail/colors-example.sh
+file etc/multitail/convert-geoip.pl path=etc/multitail/convert-geoip.pl
+file etc/multitail/convert-simple.pl path=etc/multitail/convert-simple.pl
+
+dir  path=usr
```
pyhalov, line 33: non-empty dir entries are not necessary (and not needed, when there's no directory permissions changes)

components/multitail/multitail.p5m
```diff
@@ -0,0 +1,49 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=usr/share/doc/$(COMPONENT_NAME)/.* -> default facet.doc true>
+
+dir  path=etc
+file etc/multitail.conf.new path=etc/multitail.conf preserve=renamenew
+dir  path=etc/multitail
+file etc/multitail/colors-example.pl path=etc/multitail/colors-example.pl
```
pyhalov, line 28: Do we really need to ship example files to /etc? If we need, perhaps they should be executable? Or is it better to do like Debian, shipping them in /usr/share/doc/multitail ?

components/multitail/multitail.p5m
```diff
@@ -0,0 +1,49 @@
...
+file etc/multitail/colors-example.sh path=etc/multitail/colors-example.sh
```
pyhalov, line 29: "file bla path=bla" is redundant, "file path=bla" should be enough.

components/multitail/Makefile
```diff
@@ -0,0 +1,44 @@
...
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tgz
+COMPONENT_ARCHIVE_HASH=	sha256:af1d5458a78ad3b747c5eeb135b19bdca281ce414cefdc6ea0cff6d913caa1fd
+COMPONENT_ARCHIVE_URL=	https://www.vanheusden.com/$(COMPONENT_NAME)/$(COMPONENT_ARCHIVE)
+
+include ../../make-rules/prep.mk
+include ../../make-rules/justmake.mk
+include ../../make-rules/ips.mk
+
+COMPONENT_BUILD_ENV += CFLAGS="-I/usr/include/ncurses"
```
alarcher, line 34: Should this flag not be appended to existing CFLAGS ?

pyhalov, line 34: Yes, it would be better just to set CFLAGS+="-I/usr/include/ncurses"

components/multitail/multitail.p5m
minikola, line 28: I think that example files are better within /etc because there is wher epeople expect them to find. 
I was looking in documentation multiple times and they are always there for many packages. it would be very strange not to find them there. 
Maybe they should be executable, If as examples they are called with other name that is not called by an app who's example it is, but when it is renamed to a name used by an app.

xen0l:
@ivanrichwalski any updates?

PR #1579, opened 2015-12-10
open-fabrics: Fix download URLs

mtelka:


pyhalov:
OpenFabrics are not enabled (see components.ignore) as noone built and tested it. If you know/want to how to build/test it, go ahead and resync it with userland-gate. If you don't want to do the full job, I'd leave it as is untill someone properly fix and test it.

mtelka:
I didn't tried to build it, but I still think it is better to have the proper download URLs even we do not know how to build it, than to leave it to rot more and more.  I consider the proper download links as a first step in a successful build.

PR #1584, opened 2015-12-12
Add libjpeg6 and provide libjpeg as meta-package

alarcher:
First shot, I am still not sure what is the best approach as Debian, Ubuntu, Fedora seem to have dropped IJG's libraries:

https://lists.debian.org/debian-devel-announce/2014/08/msg00000.html

I decided to keep them in subdirectories to minimize the probability of mixing and provided symlinks to provide compatbility with the former package from JDS.

So please consider this PR as a feedback request first.

### DIFF
--- a/components/libjpeg6-ijg/Makefile
+++ b/components/libjpeg6-ijg/Makefile
@@ -1,0 +1,53 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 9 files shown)


alarcher:
I actually created now a component building all APIs for libjpeg-turbo so we should definitely discuss first.

pyhalov:
Could you also add fix for CVE-2013-6629 (and perhaps, investigate a bit, what other CVEs appeared for libjpeg)?

alarcher:
I actually checked CVEs but missed this one, other seem to apply to libjpeg-turbo (up to 1.3.0) only but this is made difficult because both libjpeg and libjpeg-turbo are refered to as libjpeg in bug reports... :S

alarcher:
Added patch for CVE-2013-6629.

PR #1585, opened 2015-12-12
Add libjpeg-turbo 1.4.2 with support for API 6,7,8

alarcher:
Each API is installed in a different prefix.


alarcher:
Fixed missing license file and added dependency 'require' from libjpeg-turbo to current default API i.e. libjpeg6-turbo.

PR #1586, opened 2015-12-12
Add VLC 2.2.1 (encumbered)

alarcher:
I prepared a VLC 2.2.1 component which requires libjpeg6-turbo.
Various audio and video format can be played (h264,wmv,mov,xvid,divx), tested in VirtualBox from Videolan's and Wavetlan.com's samples.

There are known issues:
- lua_plugin.so does not seem to load.
- submenus can become overriden by the background.

but so far so good... if one would like to test it before PR.


components/encumbered/vlc/Makefile
```diff
@@ -0,0 +1,186 @@
...
+CONFIGURE_OPTIONS += --enable-jpeg
+CONFIGURE_OPTIONS += --disable-x262
+CONFIGURE_OPTIONS += --disable-x265
+CONFIGURE_OPTIONS += --enable-x264
+CONFIGURE_OPTIONS += --disable-mfx
+CONFIGURE_OPTIONS += --disable-fluidsynth
+CONFIGURE_OPTIONS += --disable-zvbi
+CONFIGURE_OPTIONS += --enable-telx
+CONFIGURE_OPTIONS += --enable-libass
+CONFIGURE_OPTIONS += --disable-kate
```
alarcher, line 128: My mistake, I did not see that we have opus so I can enable it.

pyhalov:
Merged modified version as https://github.com/OpenIndiana/oi-userland/pull/1616

PR #1589, opened 2015-12-13
Fix download for lynx

AndWac:
I am not sure whether the new COMPONENT_VERSION is correct.
I only downloaded and built lynx successfully with it.

BTW: Oracle userland gate offers the latest lynx (2.8.8) with two new patches.
OIs patch is not there anymore. As I don't use lynx I wasn't keen to update to
the new version, though.


AndWac:
Hmm, I don't know what went wrong with the 2nd commit. I just merged the latest jpeg work into my branch...

pyhalov:
Perhaps you should just go and update it to the latest minor version?

components/lynx/Makefile
```diff
@@ -23,13 +23,13 @@
 include ../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		lynx
-COMPONENT_VERSION=	2.8.7
+COMPONENT_VERSION=	2.8.7rel.2
```
alarcher: Should it not be numeric only ?

pyhalov: If no, IPS_COMPONENT_VERSION should be also set

AndWac:
Ok, I will try to update to 2.8.8 when time permits. Probably next week.

PR #1596, opened 2015-12-14
Bump hdf5 to 1.8.16

alarcher:


components/scientific/hdf5/hdf5.p5m
```diff
@@ -163,54 +165,52 @@ file path=usr/include/hdf5/h5s.mod
...
-    target=libhdf5hl_fortran.so.8.0.2
-link path=usr/lib/$(MACH64)/libhdf5hl_fortran.so.8 \
-    target=libhdf5hl_fortran.so.8.0.2
-file path=usr/lib/$(MACH64)/libhdf5hl_fortran.so.8.0.2
+    target=libhdf5hl_fortran.so.10.0.2
+link path=usr/lib/$(MACH64)/libhdf5hl_fortran.so.10 \
+    target=libhdf5hl_fortran.so.10.0.2
+file path=usr/lib/$(MACH64)/libhdf5hl_fortran.so.10.0.2
 file path=usr/lib/libhdf5.settings
-link path=usr/lib/libhdf5.so target=libhdf5.so.8.0.2
```
pyhalov, line 196: So, this update is ABI-incompatible. Do we have any packages dependent on hdf5?

alarcher, line 196: Not to my knowledge as I only use it for non-packaged software but it is worth checking.
I do not have OI running at home for now though, so I'll do it tomorrow.

alarcher, line 196: pkg search -r -H -o pkg.name 'depend:require:library/c++/hdf5' returns nothing

PR #1613, opened 2015-12-23
hexchat: Split out the hexchat-devel package

mtelka:


pyhalov:
Perhaps, it is not necessary, as files in devel package have facet.devel=all set (it is set automatically in oi-userland/transforms/devel):

<pre>
$ pkg contents -rm hexchat |grep facet.devel
file a34e2058797cebfb00a9c1563cc4af20de743e55 chash=c5218ccc2973564e3cbff4060bf47b23b84e5467 facet.devel=all group=bin mode=0444 owner=root path=usr/include/hexchat-plugin.h pkg.csize=2414 pkg.depend.bypass-generate=usr/lib(.*)/libpq.so.* pkg.size=13818
file 711c38ccde5474ac75f9b90ef993584ab150b8ce chash=c67a6c2193f7826093933abad28441f2113406b0 facet.devel=all group=bin mode=0444 owner=root path=usr/lib/pkgconfig/hexchat-plugin.pc pkg.csize=162 pkg.depend.bypass-generate=usr/lib(.*)/libpq.so.* pkg.size=214
</pre>

mtelka:
You are right.

PR #1643, opened 2016-01-17
Initial commit of recipe for LSOF

jimklimov:
As it says on the tin :)
Passes "gmake publish" and "gmake clean clobber", installed package works well - also regarding software and kernel bitness (via isaexec), and non-root user privileges (via setgid sys).

### DIFF
--- a/components/lsof/Makefile
+++ b/components/lsof/Makefile
@@ -27,12 +27,32 @@ COMPONENT_ARCHIVE_HASH=	\
     sha256:81ac2fc5fdc944793baf41a14002b6deb5a29096b387744e28f8c30a360a3718
 
 include ../../make-rules/prep.mk
-include ../../make-rules/configure.mk
+include ../../make-rules/justmake.mk
 include ../../make-rules/ips.mk
 
+# Snatched from espeak component
+BINDIR.32=/usr/bin
+BINDIR.64=/usr/bin/$(MACH64)
+
+LIBDIR.32=/usr/lib
+LIBDIR.64=/usr/lib/$(MACH64)
+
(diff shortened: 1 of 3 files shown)


components/lsof/Makefile
```diff
@@ -0,0 +1,48 @@
...
+# but it is not available at the time of this writing
+COMPONENT_ARCHIVE_URL=	ftp://sunsite.ualberta.ca/pub/Mirror/$(COMPONENT_NAME)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:81ac2fc5fdc944793baf41a14002b6deb5a29096b387744e28f8c30a360a3718
+
+include ../../make-rules/prep.mk
+include ../../make-rules/configure.mk
+include ../../make-rules/ips.mk
+
+COMPONENT_PREP_ACTION = \
```
pyhalov, line 54: Looks a bit ugly. Perhaps, this can be done with CONFIGURE_SCRIPT=Configure or using justbuild.mk ?

jimklimov, line 54: Per se CONFIGURE_SCRIPT does not cut it, unless I uglily override some
other variable to effect `yes ''` to accept the defaults. At least, this
commit point does build 32/64 bit bins so users can install them, and IMHO
improvements like finding real ways to pass build-settings explicitly can
be subject of another PR... at least I won't have time in nearby days to
pursue this.

I will look if there is a quick conversion to justmake in the time I have
today though. But if not - i'd suggest to leave it for another PR too.

## 

18.01.2016 10:05 пользователь "Alexander Pyhalov" notifications@github.com
написал:

PR #1656, opened 2016-01-23
build bind chroot when service starts

hrosenfeld:
This change will greatly improve support for running bind in a chroot. Minimal support for chroot was there already, but that required manually maintaining (and updating) the chroot, which I forgot every time I updated my system and every now and then it wouldn't come up cleanly because of this.

The dns-server method will now populate the chroot with everything needed. It will require additional privileges, but it will drop them when starting named. Additionally it will support running named as a non-root user.


components/bind/Solaris/dns-server.sh
```diff
@@ -99,12 +100,54 @@ case "$method" in
...
+      msg="$I : Configuration file ${configuration_file} does not exist!"
+      echo ${msg} >&2
+      /usr/bin/logger -p daemon.error ${msg}
+      # dns-server should be placed in maintenance state.
+      result=${SMF_EXIT_ERR_CONFIG}
+    fi
+
     # If chroot option is set, note zones(5) are preferred, then
     # configuration file lives under chroot directory.
     if [ "${chroot_dir}" != "" ]; then
```
pyhalov, line 153: I like this, but please, add also check for $chroot_dir != / , to disallow user to shoot himself in the foot.

pyhalov:
Also, please bump COMPONENT_REVISION in Makefile

hrosenfeld:
Added the requested check and set COMPONENT_REVISION in Makefile.

I also changed the way the configuration files are propagated to the chroot. I now create symlinks to the chrooted files in the original location so that a sysadmin can edit the files in their regular location and rndc still works.

components/bind/Solaris/dns-server.sh
```diff
@@ -99,12 +100,67 @@ case "$method" in
...
+      for file in ${configuration_files}; do
+        if [ ! -L ${configuration_dir}/${file} ]; then
+          pax -rw -H -pe ${configuration_dir}/${file} ${chroot_dir}
+          ln -sf ${chroot_dir}/${configuration_dir}/${file} ${configuration_dir}/${file}
+	fi
+      done
+
+      if [ ! -L ${configuration_file} ]; then
+        pax -rw -H -pe ${configuration_file} ${chroot_dir}
+        ln -sf ${chroot_dir}/${configuration_file} ${configuration_file}
```
pyhalov: Isn't it safer to mount configuration files to chroot dir, without replacing them with links?

hrosenfeld:
All config files and the pkcs11 libs will be lofs-mounted into the chroot. Unless you have further change requests I'd consider this ready for merge.

components/bind/Solaris/dns-server.sh
```diff
@@ -20,34 +20,39 @@
...
+             exit ${SMF_EXIT_ERR_CONFIG}
+        fi
 
-# Set defaults; SMF_FMRI should have been set, but just in case.
-if [ -z "$SMF_FMRI" ]; then
-    SMF_FMRI="svc:/network/dns/server:${instance}"
-fi
-server="/usr/sbin/named"
-I=`/usr/bin/basename $0`
+        umount ${c}/${f}
```
pyhalov: Shouldn't we check that file is mounted before unmounting it or at least suppress error messages?

components/bind/Solaris/dns-server.sh
```diff
@@ -99,40 +104,97 @@ case "$method" in
...
+            msg="$I : chroot_dir must not be /"
+            echo ${msg} >&2
+            /usr/bin/logger -p daemon.error ${msg}
+            # dns-server should be placed in maintenance state.
+            exit ${SMF_EXIT_ERR_CONFIG}
+        fi
+
+        server="env LD_NOLAZYLOAD=1 ${server}"
+
+        mkdir -p ${chroot_dir}
```
pyhalov, line 164: What if chroot_dir already exists?

hrosenfeld, line 164: It's not an error for mkdir -p if a directory already exists, it will leave it alone.

components/bind/Solaris/dns-server.sh
hrosenfeld: This is just to make sure it actually is not mounted already before mounting it. Under normal operation that is noop, so I chose to supress error messages.

PR #1685, opened 2016-02-06
Update of of liblzo2 (2.09) delivery

jimklimov:
Previously "library/lzo" was delivered by SFE (and an older version, and 32-bit only), and "gmake publish" of a project with this library as a dependency complained that it comes from a "forbidden repository: sfe". I wonder how "lzop" got built that has this as a dependency? ;)
Anyway, now I remade and updated it within oi/hipster.
Installs all that it builds except *.la files (as requested in another ticket).


pyhalov:
We have components/liblzo (library/lzo). You should update it, if you want.

jimklimov:
Oops, must have been very late in the night... updated it instead (rebuilding then pushing)....

pyhalov:
Please, merge commits

jimklimov:
Ok, squashed and re-commented, ready to go ;)
Thanks for the thorough and informative (educative) reviews. I hope to generate less hiccups next times ;)

PR #1687, opened 2016-02-07
fix dns/server in zones

hrosenfeld:
Apparently my previous change to dns/server (components/bind in oi-userland) broke zone support. The privilege sys_devices is not available in zones and so SMF can't even run the start method.

The only solution to that is to not create device nodes in the chroot. This is less than optimal but it seems unavoidable in order to continue to support zones. Instead, a warning is now printed if any of the required devices are missing. The sysadmin will have to create the device nodes in the chroot himself with mknod.


pyhalov:
Can we run start method with full root permissions, create devices and then drop permissions, without restricting permissions at smf level? Will it work in zones? Will changing pax with mknod help?

hrosenfeld:
Thats what the script originally did, but it won't work in zones. Zones never have the privilege sys_devices. So no matter what, device nodes cannot be created inside a zone. I also tried lofs mounting of the device nodes, but that failed for different reasons.

PR #1688, opened 2016-02-07
Initial commit of SDL2 (2.0.4)

jimklimov:
Added a recipe for libsdl2 and friends, since sdl1.2.x is more or less considered obsolete by the project, and I'm looking at some programs that would need SDL2 to compile.
Package builds and publishes cleanly. Usability for the job - remains to be seen ;)

### DIFF
--- a/components/library/SDL2/Makefile
+++ b/components/library/SDL2/Makefile
@@ -1,0 +1,51 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2016 Jim Klimov
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		SDL2
+COMPONENT_VERSION=	2.0.4
+COMPONENT_PROJECT_URL=	http://www.libsdl.org/
+COMPONENT_SUMMARY=	libsdl2 - Simple DirectMedia Layer 2.0
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:da55e540bf6331824153805d58b590a29c39d2d506c6d02fa409aedeab21174b
+COMPONENT_ARCHIVE_URL=	http://www.libsdl.org/release/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI =	library/sdl2
+# Note: license change from LGPL was significant in 2.x vs 1.x series, and now
+# static linking is permitted. See e.g. https://wiki.libsdl.org/Installation
+COMPONENT_LICENSE=	ZLIB
+COMPONENT_LICENSE_FILE=	COPYING.txt
+COMPONENT_CLASSIFICATION =	System/Libraries
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+PATH=/usr/gnu/bin:/usr/bin
+
+COMPONENT_PREP_ACTION = \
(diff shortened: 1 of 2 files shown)


components/SDL2/Makefile
```diff
@@ -0,0 +1,56 @@
...
+include $(WS_TOP)/make-rules/ips.mk
+
+PATH=/usr/gnu/bin:/usr/bin
+
+COMPONENT_PREP_ACTION = \
+			( cd $(@D) && $(CONFIG_SHELL) autogen.sh )
+
+CONFIGURE_OPTIONS +=	--sysconfdir=/etc
+CONFIGURE_OPTIONS +=	--libexecdir=$(CONFIGURE_LIBDIR.$(BITS))
+CONFIGURE_OPTIONS +=	--libdir=$(CONFIGURE_LIBDIR.$(BITS))
```
pyhalov: This should be automatic

components/SDL2/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+#
+# Copyright (c) 2016 Jim Klimov
+#
+
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		SDL2
+COMPONENT_VERSION=	2.0.4
+COMPONENT_REVISION=	1
```
pyhalov: COMPONENT_REVISION is not necessary for the first revision

components/SDL2/sdl2.p5m
```diff
@@ -0,0 +1,127 @@
...
+file path=usr/lib/$(MACH64)/libSDL2_test.a
+file path=usr/lib/$(MACH64)/libSDL2main.a
+file path=usr/lib/libSDL2-2.0.so.0.4.0
+link path=usr/lib/libSDL2-2.0.so.0 target=libSDL2-2.0.so.0.4.0
+link path=usr/lib/libSDL2-2.0.so target=libSDL2-2.0.so.0
+link path=usr/lib/libSDL2.so target=libSDL2-2.0.so
+file path=usr/lib/libSDL2.a
+file path=usr/lib/libSDL2_test.a
+file path=usr/lib/libSDL2main.a
+dir  path=usr/lib/cmake
```
pyhalov: We usually avoid delivering dir actions until it's  strictly necessary

jimklimov: and do we commonly have /usr/lib/cmake in a typical installation? which might not have cmake itself?

components/SDL2/Makefile
jimklimov: Copypasted from SDL component...

components/SDL2/sdl2.p5m
pyhalov: It doesn't matter, file action for usr/lib/cmake/SDL2/sdl2-config.cmake is enough

components/SDL2/Makefile
```diff
@@ -0,0 +1,55 @@
...
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+PATH=/usr/gnu/bin:/usr/bin
+
+COMPONENT_PREP_ACTION = \
+			( cd $(@D) && $(CONFIG_SHELL) autogen.sh )
+
+#CONFIGURE_OPTIONS +=	--sysconfdir=/etc
+#CONFIGURE_OPTIONS +=	--libexecdir=$(CONFIGURE_LIBDIR.$(BITS))
```
pyhalov: --libexecdir and --sysconfdir are not passed by default, but --libdir is added by default rules (see configure.mk)

components/SDL2/sdl2.p5m
jimklimov: fixed

jimklimov:
Hopefully addressed all the comments above (rebuilding and committing soon).

On a side note, do we have a good mechanism to use different HW-optimized features or builds as determined at runtime (like loopback mounts of libc.so back in the day)? (Anyhow a task for some other non-initial development and PR)

I ask because the SDL2 has a number of compile-time options for different CPU instructions that might make a difference on newer CPUs but might not work on older ones, but if I understand it correctly - it does not offer building a single binary that would pick correct implementation based on CPU capabilities, sadly...

pyhalov:
No, we don't have such mechanism. 
Please, move files to components/library/SDL2

jimklimov:
Is this good to go? IIRC it was a dependency for OpenTTD to be built... ;)

pyhalov:
Yes, please, merge commits.

PR #1689, opened 2016-02-08
Initial commit of libGLEW (1.13.0) as part of oi/hipster userland

jimklimov:
SFE provided an older copy (as I found in the end of my endeavour), but due to age / other-repo it was not useful for compiling projects within oi-userland. So here is a bump and a new feature, made from scratch.

There is an issue about building this in a zone where VirtualBox was installed, which can be written off as a VirtualBox bug (don't go to replace system libs) and should not happen in dedicated build zones, or one can change the symlinks for the duration of the build, as detailed in the Makefile.

### DIFF
--- a/components/library/libglew/Makefile
+++ b/components/library/libglew/Makefile
@@ -1,0 +1,100 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov <jimklimov@gmail.com>
+#
+
+# NOTE: This project links with /usr/openwin/lib/libGL.so.1, which unrolls
+# to /usr/lib/GL/libGL.so.1 which is a symlink to run-time (hardware-related?)
+# mediator in the ../../../var/run/opengl/lib/libGL.so.1. So far so good...
+# In my case (GZ on a VirtualBox host, even though it is dual-booted and runs
+# OI/Hipster natively at the moment) this library symlink points back to
+# /usr/lib/mesa/libGL.so.1 and this one object points to /usr/lib/VBoxOGL.so
+# rather than the nearby mesa objects. This breaks the build because pkgmogrify
+# insists on objects coming from known repo's, and because ultimate objects do
+# refer to virtualbox rather than libs expected available on all deployments!
+# Perhaps we should *require* do the builds in a dedicated local zone?..
+# TODO: Find out the root-cause and attach a patch-file with a fix to this.
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		glew
+COMPONENT_VERSION=	1.13.0
+COMPONENT_SUMMARY=	The OpenGL Extension Wrangler Library
+COMPONENT_CLASSIFICATION=System/Libraries
+COMPONENT_FMRI=		x11/library/lib$(COMPONENT_NAME)
+COMPONENT_LICENSE=	GLEW,Mesa3D,Khronos
+COMPONENT_LICENSE_FILE=	LICENSE.txt
+COMPONENT_PROJECT_URL=	http://$(COMPONENT_NAME).sourceforge.net/
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tgz
+COMPONENT_ARCHIVE_HASH= \
+	sha256:aa25dc48ed84b0b64b8d41cdd42c8f40f149c37fa2ffa39cd97f42c78d128bc7
+COMPONENT_ARCHIVE_URL= \
+	http://sourceforge.net/projects/$(COMPONENT_NAME)/files/$(COMPONENT_NAME)/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+
+PATH=/usr/gnu/bin:/usr/bin
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CUSTOM.LDFLAGS.GL = -lGL
+COMPONENT_PREP_ACTION = \
+    ( cd "$(SOURCE_DIR)/config" && cp -pf "$(COMPONENT_DIR)/files/Makefile.illumos-gcc" ./ )
+
+CFLAGS +=	-I$(SOURCE_DIR)/include
+CFLAGS +=	-D_FILE_OFFSET_BITS=64
+
+# These flags define the nuances of "oi-userland" in "glew"'s Makefile dialect.
+# They are common for build and install actions, so we define them once in an
+# extra variable, and re-use it below. Note we use "CUSTOM.*" flags defined in
+# the custom Makefile.illumos-gcc to append common oi-userland flag values
+# and not override the project's CFLAGS etc. in whole.
+
+LIBDIR.32=$(USRLIBDIR)
+LIBDIR.64=$(USRLIBDIR64)
+
+COMPONENT_COMMON_ARGS  =	SYSTEM=illumos-gcc
+COMPONENT_COMMON_ARGS +=	BINDIR="$(USRBINDIR$(BITS))"
+COMPONENT_COMMON_ARGS +=	LIBDIR="$(LIBDIR.$(BITS))"
+COMPONENT_COMMON_ARGS +=	INCDIR="$(USRINCDIR)"
+COMPONENT_COMMON_ARGS +=	DESTDIR="$(PROTO_DIR)"
+COMPONENT_COMMON_ARGS +=	GLEW_PREFIX="$(USRDIR)"
+COMPONENT_COMMON_ARGS +=	GLEW_DEST="$(USRDIR)"
+COMPONENT_COMMON_ARGS +=	CC="$(CC)"
+COMPONENT_COMMON_ARGS +=	CUSTOM.CFLAGS.EXTRA="$(CFLAGS)"
+COMPONENT_COMMON_ARGS +=	LD="$(CC)"
+COMPONENT_COMMON_ARGS +=	CUSTOM.LDFLAGS.EXTRA="$(LDFLAGS)"
+COMPONENT_COMMON_ARGS +=	CUSTOM.LDFLAGS.SO="$(LDFLAGS)"
+COMPONENT_COMMON_ARGS +=	CUSTOM.LDFLAGS.GL="$(CUSTOM.LDFLAGS.GL)"
+COMPONENT_COMMON_ARGS +=	AR="$(AR)"
+COMPONENT_COMMON_ARGS +=	CUSTOM.ARFLAGS.EXTRA="$(ARFLAGS)"
+
+### In case of updates, review what the glew Makefile defines (and expects)
+### these commands to be. In particular, their LN="ln -sf" now, and it matters.
+COMPONENT_COMMON_ARGS +=	RM="$(RM)"
+COMPONENT_COMMON_ARGS +=	LN="$(SYMLINK) -f"
+
+COMPONENT_BUILD_ARGS  +=	$(COMPONENT_COMMON_ARGS)
+COMPONENT_BUILD_TARGETS =	all
+COMPONENT_INSTALL_ARGS +=	$(COMPONENT_COMMON_ARGS)
+COMPONENT_INSTALL_TARGETS =	install.all
+
+build: $(BUILD_32_and_64)
+
+install: $(INSTALL_32_and_64)
+
+test: $(TEST_32_and_64)
+
+REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += x11/library/libx11
--- a/components/library/libglew/files/Makefile.illumos-gcc
+++ b/components/library/libglew/files/Makefile.illumos-gcc
@@ -1,0 +1,12 @@
+# This makefile snippet allows to compile GLEW for "oi-userland" repository
+# Generally this is like "solaris-gcc" (hence the include to be future-proof)
+
+include config/Makefile.solaris-gcc
+LDFLAGS.GL = -lm -lX11
+
+# Allow additions from the builder
+LDFLAGS.EXTRA+=	$(CUSTOM.LDFLAGS.EXTRA)
+ARFLAGS.EXTRA+=	$(CUSTOM.ARFLAGS.EXTRA)
+CFLAGS.EXTRA+=	$(CUSTOM.CFLAGS.EXTRA)
+LDFLAGS.SO+=	$(CUSTOM.LDFLAGS.SO)
+LDFLAGS.GL+=	$(CUSTOM.LDFLAGS.GL)
--- a/components/library/libglew/libglew.p5m
+++ b/components/library/libglew/libglew.p5m
@@ -1,0 +1,50 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov. All rights reserved.
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.upstream-url value="$(COMPONENT_PROJECT_URL)"
+set name=info.source-url value="$(COMPONENT_ARCHIVE_URL)"
+set name=info.classification value="org.opensolaris.category.2008:$(COMPONENT_CLASSIFICATION)"
+set name=org.opensolaris.consolidation value="$(CONSOLIDATION)"
+
+license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)"
+
+depend fmri=service/opengl/ogl-select type=require
+file path=usr/bin/$(MACH64)/glewinfo pkg.depend.bypass-generate=libGL\.so\.1
+file path=usr/bin/$(MACH64)/visualinfo pkg.depend.bypass-generate=libGL\.so\.1
+file path=usr/bin/$(MACH32)/glewinfo pkg.depend.bypass-generate=libGL\.so\.1
+file path=usr/bin/$(MACH32)/visualinfo pkg.depend.bypass-generate=libGL\.so\.1
+hardlink path=usr/bin/glewinfo target=../lib/isaexec pkg.linted=true
+hardlink path=usr/bin/visualinfo target=../lib/isaexec pkg.linted=true
+file path=usr/include/glew.h
+file path=usr/include/glxew.h
+file path=usr/include/wglew.h
+file path=usr/lib/$(MACH64)/libGLEW.so.1.13.0 pkg.depend.bypass-generate=libGL\.so\.1
+link path=usr/lib/$(MACH64)/libGLEW.so.1.13 target=libGLEW.so.1.13.0
+link path=usr/lib/$(MACH64)/libGLEW.so target=libGLEW.so.1.13
+file path=usr/lib/$(MACH64)/libGLEWmx.so.1.13.0 pkg.depend.bypass-generate=libGL\.so\.1
+link path=usr/lib/$(MACH64)/libGLEWmx.so.1.13 target=libGLEWmx.so.1.13.0
+link path=usr/lib/$(MACH64)/libGLEWmx.so target=libGLEWmx.so.1.13
+file path=usr/lib/$(MACH64)/pkgconfig/glew.pc
+file path=usr/lib/$(MACH64)/pkgconfig/glewmx.pc
+file path=usr/lib/libGLEW.so.1.13.0 pkg.depend.bypass-generate=libGL\.so\.1
+link path=usr/lib/libGLEW.so.1.13 target=libGLEW.so.1.13.0
+link path=usr/lib/libGLEW.so target=libGLEW.so.1.13
+file path=usr/lib/libGLEWmx.so.1.13.0 pkg.depend.bypass-generate=libGL\.so\.1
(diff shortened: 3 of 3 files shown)


jimklimov:
Note, do not merge yet - there are issues in 64-bit build (something about PIC is to be done, looking at it now...)

components/glew/Makefile
```diff
@@ -0,0 +1,100 @@
...
+# Perhaps we should *require* do the builds in a dedicated local zone?..
+# TODO: Find out the root-cause and attach a patch-file with a fix to this.
+# So far we have a check to abort the build if this condition is detected.
+
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		glew
+COMPONENT_VERSION=	1.13.0
+COMPONENT_SUMMARY=	The OpenGL Extension Wrangler Library
+COMPONENT_CLASSIFICATION=org.opensolaris.category.2008:System/Libraries
```
pyhalov: should be System/Libraries

components/glew/glew.p5m
```diff
@@ -0,0 +1,62 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.upstream-url value="$(COMPONENT_PROJECT_URL)"
+set name=info.source-url value="$(COMPONENT_ARCHIVE_URL)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=org.opensolaris.consolidation value="$(CONSOLIDATION)"
+
+license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)"
+
+dir  path=usr
```
pyhalov: avoid shipping dir actions

components/glew/Makefile
```diff
@@ -0,0 +1,100 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov <jimklimov@gmail.com>
+#
+
+# NOTE: This project links with /usr/openwin/lib/libGL.so.1, which unrolls
```
pyhalov: Why pkg.depend.bypass-generate=libGL.so.1 doesn't works for this?

components/glew/glew.p5m
```diff
@@ -0,0 +1,62 @@
...
+file path=usr/include/glew.h
+file path=usr/include/glxew.h
+file path=usr/include/wglew.h
+dir  path=usr/lib
+dir  path=usr/lib/$(MACH64)
+file path=usr/lib/$(MACH64)/libGLEW.a
+file path=usr/lib/$(MACH64)/libGLEW.so.1.13.0 pkg.depend.bypass-generate=libGL\.so\.1
+link path=usr/lib/$(MACH64)/libGLEW.so.1.13 target=libGLEW.so.1.13.0
+link path=usr/lib/$(MACH64)/libGLEW.so target=libGLEW.so.1.13
+file path=usr/lib/$(MACH64)/libGLEWmx.a
```
pyhalov: We avoid shipping static libraries, if project delivers corresponding dynamic library.

components/glew/Makefile
jimklimov: This is two-fold:
1) Linking produces a bad binary (references paths that may not exist on cutormer's deployment) if VBoxOGL is the target of libGL.so.1 symlinks
2) Packaging complains about library not in this repo... or something of the sort. It complains without the pkg.depend... flag, it builds with it ;)

components/glew/glew.p5m
jimklimov: Makes sense for common dirs; does this apply to apparently project-private dirs too? will pkg make them automagically to the best of its judgement?

components/glew/Makefile
jimklimov: We have too many examples in the code with too different patterns ;)
Fixed

alarcher: Just for reference, Phorcys had already packaged this in the Kodi branch:
https://github.com/phorcys/oi-userland/commit/4490407e0b37b85beda2f91771b6e55cf796274b

jimklimov: They didn't add a PR so we don't know we have it ;)
Also their Makefile seems more elaborate (they might have better idea what they were doing than myself... maybe), but they only deliver 32-bit libs and bins - after a quick "cool, it almost works right away!" I spent a quite while polishing the 64-bit part to actually compile into usable state. But thanks for the pointer ;)

alarcher: Phorcys' branch is actually a contribution that we are reviewing to get Kodi in oi-userland.
I thought this was a good opportunity to also merge his changes in the process.

jimklimov: Well, at the very least it makes sense to claim some one package fmri for
both versions; i can adapt in my PR;)

Also, why is theirs encumbered?

And no MACH64 for libs is a bad thing. We might pick what bins to build,
but libs should be available in both arch's imho. Otherwise it blocks
future evolution into thre absent arch. Been there ;)

вторник, 9 февраля 2016 г. пользователь Aurélien Larcher написал:

components/glew/Makefile
pyhalov: If we do this, we should do this globally for components, linking to libGL.so. But it seems we have not been beaten by the issue yet. I'd preserved things as they are for now - added 
depend fmri=service/opengl/ogl-select type=require and  <transform file -> add pkg.depend.bypass-generate libGL.so.1>

components/glew/glew.p5m
pyhalov: If something is delivered in this dirs (links/files/hardlinks), IPS will handle this. If it's something like log directory, it should be present in manifest.

jimklimov: Fixed earlier

components/glew/Makefile
jimklimov: I am not sure I follow what should be done... so in commit 31936d98cd8d24465516163feae76dbde310e4f4 I added
  depend fmri=service/opengl/ogl-select type=require 
into the p5m manifest, and left the files with explicit dependencies as I had them, e.g.
  file path=usr/bin/$(MACH64)/glewinfo pkg.depend.bypass-generate=libGL.so.1
(a dozen like these)... right?

components/glew/glew.p5m
jimklimov: It is a differentiator against earlier releases (that had different licensing) although indeed still not recommended by docs for technical reasons. Fixed earlier.

pyhalov:
During build ld is invoked directly with GCC flags:

/usr/bin/ld -shared -Wl,-soname=libGLEWmx.so.1.13 -o lib/libGLEWmx.so.1.13.0 tmp/illumos-gcc/mx/shared/glew.o -L/usr/openwin/lib -m32 -lGL -lX11 -lm 
                LINK EDITOR MEMORY MAP

output          input           virtual
section         section         address         size

.dynamic                              b4             138
                .dynamic              b4             138 (null)
.SUNW_sy                             1ec            2730
                .SUNW_sy             1ec              00 (null)
.hash                               291c            4e9c
                .hash               291c            4e9c (null)
.SUNW_ld                            77b8             280
                .SUNW_ld            77b8             280 (null)
.dynsym                             7a38            9cc0
                .dynsym             7a38            9cc0 (null)
.dynstr                            116f8           10626
                .dynstr            116f8           10426 (null)
.SUNW_dy                           21d20            27a4
                .SUNW_dy           21d20            27a4 (null)
.SUNW_re                           244c4            4e00
                .SUNW_re           244c4            4e00 (null)
.rel.plt                           292c4              30
                .rel.plt           292c4              30 (null)
.plt                               292f4              70
                .plt               292f4              70 (null)
.text                              29370           3d736
                .text              29370           3d736 tmp/illumos-gcc/mx/shared/glew.o
                .text.un              00              00 tmp/illumos-gcc/mx/shared/glew.o
.rodata                            66aa8           121c4
                .rodata.              00           10cea tmp/illumos-gcc/mx/shared/glew.o
                .rodata.           66aa8            2e89 tmp/illumos-gcc/mx/shared/glew.o
                .rodata.           69931            f33b (null)
.got                               89000            2700
                .got               89000            2700 (null)
.data                              8b700              00
                .data              8b700              00 tmp/illumos-gcc/mx/shared/glew.o
.data.re                           8b700              28
                .data.re           8b700              28 tmp/illumos-gcc/mx/shared/glew.o
.bss                               8b728            26e4
                .bss               8b728            26e4 tmp/illumos-gcc/mx/shared/glew.o

jimklimov:
Ok, I see that LDFLAGS.SO variable had to be updated for their Makefile, thanks.

Does this log look more correct?

/usr/gcc/4.9/bin/gcc -DGLEW_NO_GLU -DGLEW_MX -DGLEW_BUILD -O2  -Iinclude -I/usr/openwin/include -fPIC -m32 -O3  -I/home/jim/shared/oi-userland/components/library/glew/glew-1.13.0/include -D_FILE_OFFSET_BITS=64  -o tmp/illumos-gcc/mx/shared/glew.o -c src/glew.c
/usr/bin/ld -shared -Wl,-soname=libGLEWmx.so.1.13 -o lib/libGLEWmx.so.1.13.0 tmp/illumos-gcc/mx/shared/glew.o -L/usr/openwin/lib -m32 -lGL -lX11 -lm
                LINK EDITOR MEMORY MAP

output          input           virtual
section         section         address         size

.dynamic                              b4             138
                .dynamic              b4             138 (null)
.SUNW_sy                             1ec            2730
                .SUNW_sy             1ec              00 (null)
.hash                               291c            4e9c
                .hash               291c            4e9c (null)
.SUNW_ld                            77b8             280
                .SUNW_ld            77b8             280 (null)
.dynsym                             7a38            9cc0
                .dynsym             7a38            9cc0 (null)
.dynstr                            116f8           10626
                .dynstr            116f8           10426 (null)
.SUNW_dy                           21d20            27a4
                .SUNW_dy           21d20            27a4 (null)
.SUNW_re                           244c4            4e00
                .SUNW_re           244c4            4e00 (null)
.rel.plt                           292c4              30
                .rel.plt           292c4              30 (null)
.plt                               292f4              70
                .plt               292f4              70 (null)
.text                              29370           3d736
                .text              29370           3d736 tmp/illumos-gcc/mx/shared/glew.o
                .text.un              00              00 tmp/illumos-gcc/mx/shared/glew.o
.rodata                            66aa8           121c4
                .rodata.              00           10cea tmp/illumos-gcc/mx/shared/glew.o
                .rodata.           66aa8            2e89 tmp/illumos-gcc/mx/shared/glew.o
                .rodata.           69931            f33b (null)
.got                               89000            2700
                .got               89000            2700 (null)
.data                              8b700              00
                .data              8b700              00 tmp/illumos-gcc/mx/shared/glew.o
.data.re                           8b700              28
                .data.re           8b700              28 tmp/illumos-gcc/mx/shared/glew.o
.bss                               8b728            26e4
                .bss               8b728            26e4 tmp/illumos-gcc/mx/shared/glew.o
/bin/ln -s -f libGLEWmx.so.1.13.0 lib/libGLEWmx.so.1.13
/bin/ln -s -f libGLEWmx.so.1.13.0 lib/libGLEWmx.so
strip -x lib/libGLEWmx.so.1.13.0

pyhalov:
No.  ld doesn't have -m32 option.  Doesn't know how it deals with -Wl options

jimklimov:
Ok... I get that imperfect is not good, but how much should I dig into this? ;)
At least, the tool does not complain and the shared libs are built for correct platforms :

root@jimoi:/home/jim/shared/oi-userland/components/library/glew# file ./build/prototype/i386/usr/lib{,/_}/_.so
./build/prototype/i386/usr/lib/libGLEW.so:     ELF 32-bit LSB dynamic lib 80386 Version 1, dynamically linked, stripped
./build/prototype/i386/usr/lib/libGLEWmx.so:   ELF 32-bit LSB dynamic lib 80386 Version 1, dynamically linked, stripped
./build/prototype/i386/usr/lib/amd64/libGLEW.so:       ELF 64-bit LSB dynamic lib AMD64 Version 1, dynamically linked, stripped
./build/prototype/i386/usr/lib/amd64/libGLEWmx.so:     ELF 64-bit LSB dynamic lib AMD64 Version 1, dynamically linked, stripped

Is that not sufficient? Out of curiosity, is there a solid reason why? (So I can perhaps review or enforce the policy in other projects, etc.) Thanks, Jim

pyhalov:
I just see that something is built in a wrong way. I'd try to avoid direct references to ld

jimklimov:
Hi, I revised how other components do it - and most define LD="$(CC)" and some relevant CC flags. So I went this route in a recent commit - looks better now?

/usr/gcc/4.9/bin/gcc -DGLEW_NO_GLU -DGLEW_MX -DGLEW_BUILD -O2  -Iinclude -I/usr/openwin/include -fPIC -m32 -O3  -I/home/jim/shared/oi-userland/components/library/glew/glew-1.13.0/include -D_FILE_OFFSET_BITS=64  -o tmp/illumos-gcc/mx/shared/glew.o -c src/glew.c
/usr/gcc/4.9/bin/gcc -shared -Wl,-soname=libGLEWmx.so.1.13 -o lib/libGLEWmx.so.1.13.0 tmp/illumos-gcc/mx/shared/glew.o -L/usr/openwin/lib -m32 -m32 -lGL -lX11 -lm 
/

/usr/gcc/4.9/bin/gcc -DGLEW_NO_GLU -DGLEW_MX -DGLEW_BUILD -O2  -Iinclude -I/usr/openwin/include -fPIC -m64 -O3  -I/home/jim/shared/oi-userland/components/library/glew/glew-1.13.0/include -D_FILE_OFFSET_BITS=64  -o tmp/illumos-gcc/mx/shared/glew.o -c src/glew.c
/usr/gcc/4.9/bin/gcc -shared -Wl,-soname=libGLEWmx.so.1.13 -o lib/libGLEWmx.so.1.13.0 tmp/illumos-gcc/mx/shared/glew.o -L/usr/openwin/lib -m64 -m64 -lGL -lX11 -lm 

oi-userland/components/library/glew$ file build/prototype/i386/usr/lib/{,_/}_.so
build/prototype/i386/usr/lib/libGLEW.so:    ELF 32-bit LSB dynamic lib 80386 Version 1, dynamically linked, not stripped, no debugging information available
build/prototype/i386/usr/lib/libGLEWmx.so:  ELF 32-bit LSB dynamic lib 80386 Version 1, dynamically linked, not stripped, no debugging information available
build/prototype/i386/usr/lib/amd64/libGLEW.so:  ELF 64-bit LSB dynamic lib AMD64 Version 1, dynamically linked, not stripped, no debugging information available
build/prototype/i386/usr/lib/amd64/libGLEWmx.so:    ELF 64-bit LSB dynamic lib AMD64 Version 1, dynamically linked, not stripped, no debugging information available

components/library/glew/Makefile
```diff
@@ -0,0 +1,115 @@
...
+COMPONENT_ARCHIVE_URL= \
+	http://sourceforge.net/projects/$(COMPONENT_NAME)/files/$(COMPONENT_NAME)/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+
+PATH=/usr/gnu/bin:/usr/bin
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/justmake.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+COMPONENT_PREP_ACTION = \
```
pyhalov: Please, remove Virtualbox-GL hacks until some common rule to handle this is adopted.

jimklimov: I'd rather solve problems as they come. Compiling in an environment where
VBoxGL is essentially hacked into the system by its installer - fails to
produce "universally usable" binaries (will refer to VBoxGL and not run
where it is not available). The hack here may be ugly, suggestions on
improvement are welcome (perhaps pre-populate builddir with copies of mesa
libs and symlinks?) But removal of this check today will lead to sorrow.
Jim

## 

06.03.2016 14:05 пользователь "Alexander Pyhalov" notifications@github.com
написал:

alarcher: You should point to mesa libs and set rpath accordingly.
The mistake is to rely on the default path which points to runtime libs.
I fixed the header path in the mesa update, do not remember if I pushed it...

jimklimov: A couple of questions here:
1) how do you do that? I'm not a linker alien yet =D
Any examples in our components worth flattering and paying tribute to?
2) given presence of ogl-select in overall architecture, is mesa the only
valid variant, or one bound to be present even if e.g. hardware-specific
ones are given, or otherwise a lowest-common denominator we can safely link
against (and work anywhere, and preferably not only with mesa but
accelerated by the optimal driver/lib chosen via ogl-select at the end-user
machine)?

Thanks, Jim

## 

07.03.2016 18:34 пользователь "Aurélien Larcher" notifications@github.com
написал:

pyhalov: You set rpath with -R/path/to/dir flag. Evidently, you need both -L/path/to/dir -R/path/to/dir in LDFLAGS. We link against mesa and consider other libGL providers to be compatible. Usually it works.

pyhalov: But as I understand, if you set RPATH explicitly to Mesa, you'll always use mesa, and ogl-select mechanism will not work.

jimklimov: Long time, no see. I looked back at where this PR got stuck: the problem is that VirtualBox overrides the `/usr/lib/mesa/(amd64)/libGL.so.1` symlink in-place (other links and files are okay). There is no separate system directory that I can link against that would have all FS objects correctly from mesa.

So we have the following trail (just 32-bit `libGL.*` for simplicity):

```
/usr/lib/libGL.so -> libGL.so.1
/usr/lib/libGL.so.1 -> GL/libGL.so.1

/usr/lib/GL/libGL.so -> libGL.so.1
/usr/lib/GL/libGL.so.1 -> ../../../var/run/opengl/lib/libGL.so.1

/var/run/opengl/lib/libGL.so.1 -> ../../../../usr/lib/mesa/libGL.so.1

/usr/lib/mesa/libGL.so -> libGL.so.1.2.0
/usr/lib/mesa/libGL.so.1 -> /usr/lib/VBoxOGL.so
/usr/lib/mesa/libGL.so.1.2.0 (file)
```

Running as an unprivileged build-time user, the Makefile recipe can not do much more than detect the error and break, as for any other configure-time dependency (prep-time to be quicker, in this implementation). If the error is fixed, e.g. by building in an OI-only zone not hacked by VirtualBox packaging, the build would succeed.

I've separated the script from the Makefile, so it could be reused by other components that can suffer from this situation (e.g. cairo, as I recently discovered).

jimklimov: Given the options, I'd prefer to link this against explicitly `/usr/lib/mesa/libGL.so.1.2.0` (or just `libGL.so`), because my GLEW utilities linked against `libGL.so.1` provided by MESA do a SIGSEGV fault on a system with `libGL.so.1` overridden by VirtualBox; according to truss, it died after an ioctl into /dev/vboxguest.

Do we have any other expected OGL providers? Are they compatible on dynamic-linking level?

I see that `/lib/opengl/ogl_select` also has an nvidia setup script, but no more.

jimklimov: I looked at some options, but none worked for me so far to enforce linking with specific library (e.g. fixed mesa filename). GL programs not working with VirtualBox Guest Additions is probably a VBox bug, but quite an impedimentr nonetheless :(

I also tried `elfedit` - half-successfully: I can ADD references with `/usr/bin/elfedit -e 'dyn:value  -s NEEDED libGL.so' $filename` but there seems to be no way to remove old references.

pyhalov:
For example, glxgears, linked against mesa's libGL.so, work with NVidia's libGL.so

jimklimov:
Regarding direct reference to desired SONAME, this line from clamav configuration gives me inspiration:

```
checking how to link with libbz2... /usr/lib/amd64/libbz2.so -R/usr/lib/amd64
```

Sort of forgot that `-lNAME` is not the only way to plug in a library to the linker command line ;)

alarcher:
As this is a prerequisite to KODI, any chance to merge it soon-ish ?

jimklimov:
AFAIK as far as C/C++ and PKG goes - this PR is okay (or was at some point, maybe before experiments vs. virtualbox).

Issues are with Virtualbox Guest Additions' hackery, that in general sense should be solved elsewhere (and do, can or should impact other GL-using programs).

One-off or systemic workarounds to this could be done in a number of ways and locations, and I guess specifically libGLEW is not required to be the solution to this problem. Also, the issue does not impact users who do not have OpenGL programs running builds and/or built programs in zones that have the VBoxOGL.so hacking its way in. In particular, builds done by OI Jenkins farm should be "clean" in this regard.

jimklimov:
I've tried linking with the `/usr/lib/mesa/libGL.so`(`.1.2.0`) directly to bypass the magic around `-lGL` that leads to linking with `libGL.so.1` (which might point to `VBoxOGL.so`) - but still get binaries that rely on VBoxOGL in `ldd` outputs and segfault in practice :\

alarcher:
Can we push this ?

jimklimov:
I think yes. There still are unsolved issues about VBoxOGL caused by the
latter and whose proper solution is not part of GLEW.

Also, if anyone better with linking comes up with some solution to force
linking against mesa libGL.so(.x.y.z) rather than libGL.so.1 resolving at
runtime to the bad lib, they are very welcome to do so as a subsequent PR
against common oi-userland upstream. So far nobody volunteered fixes to try
in my private-fork recipe ;)

Likewise, not-using VBox Guest Additions is a valid organizational solution
that applies to all HW and NGZ users automatically for example.

Maybe rolling our own build or packaging of vbox, or integrating it with
ogl-select scripts (if any of that helps) are valid technical solutions -
but all out of scooe of libglew itself.

## 

17.06.2016 3:49 пользователь "Aurélien Larcher" notifications@github.com
написал:

Can we push this ?

—
You are receiving this because you authored the thread.
https://github.com/OpenIndiana/oi-userland/pull/1689#issuecomment-226661197,
or mute the thread
https://github.com/notifications/unsubscribe/ABj5lP6aJ5fQnP8Suz52BrdoVj-jOW1wks5qMf0_gaJpZM4HVpt7
.

jimklimov:
PS: maybe another solution that is out of GLEW scope too is to move forward
with newer XOrg and Mesa, so it becomes binary-compatible with what
VBoxOGL.so provides (maybe).

Aurelien - do you have a setup to test this theory? ;)

## 

17.06.2016 4:44 пользователь "Jim Klimov" jimklimov@gmail.com написал:

pyhalov:
I'd avoid all dances around libGL for now and removed debug bits. KISS...

jimklimov:
Rearranged the code in this branch to be KISS... even if likely to break under VirtualBox Guest Additions - should be OK in other envs. Does the result seem good to go?

pyhalov:
I'd also removed comment about libGL, but if you want to leave it, I'm OK with it. Please, merge commits and let's be done with it.

jimklimov:
I think the comment should stay. If this situation bites someone, this would reduce the "why?" guesswork and maybe prod someone to a better solution ;)

jimklimov:
Re-verified by `gmake clean publish` and pushed the squash.

PR #1732, opened 2016-03-08
Update GSL to v2.1

DrLou:
bump GS to 2.1. Build --disable-static.

### DIFF
--- a/components/gsl/Makefile
+++ b/components/gsl/Makefile
@@ -9,36 +9,39 @@
 #
 
 #
-# Copyright (c) 2013 Louis M. Picciano. All rights reserved.
+# Copyright (c) 2013-2016 Louis M. Picciano. All rights reserved.
 #
 
-include ../../make-rules/shared-macros.mk
+include $(WS_TOP)/make-rules/shared-macros.mk
 
 COMPONENT_NAME			= gsl
-COMPONENT_VERSION		= 1.16
-COMPONENT_SUMMARY	= GSL - GNU Scientific Library
+COMPONENT_VERSION		= 2.1
+COMPONENT_SUMMARY		= GSL - GNU Scientific Library
 COMPONENT_SRC			= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE		= $(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_HASH	= \
-	sha256:73bc2f51b90d2a780e6d266d43e487b3dbd78945dd0b04b14ca5980fe28d2f53
-COMPONENT_ARCHIVE_URL=	http://gnu.mirrorcatalogs.com/gsl/$(COMPONENT_ARCHIVE)
+	sha256:59ad06837397617f698975c494fe7b2b698739a59e2fcf830b776428938a0c66
+COMPONENT_ARCHIVE_URL	= ftp://ftp.gnu.org/gnu/gsl/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI			= library/math/gsl
+COMPONENT_CLASSIFICATION= System/Libraries
+COMPONENT_LICENSE		= GPL v3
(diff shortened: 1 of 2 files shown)


components/gsl/Makefile
```diff
@@ -9,36 +9,39 @@
 #
 
 #
-# Copyright (c) 2013 Louis M. Picciano. All rights reserved.
+# Copyright (c) 2013-2016 Louis M. Picciano. All rights reserved.
 #
 
-include ../../make-rules/shared-macros.mk
+include $(WS_TOP)/make-rules/shared-macros.mk
```
pyhalov, line 15: this seems suspicious. Where does WS_TOP come from?

pyhalov:
This is ABI-incompatible update. Have you checked inkscape build with updated gsl? 
Also, please move it to library/

components/gsl/Makefile
DrLou, line 15: Our ancient (I guess you're saying _no longer-used_) approach to the build system?

jimklimov, line 15: Actually, it is quite used (and @alp - you could see a lot of it in my recent commits). This facilitates especially the restructuring of directory layout you do now - the amount of parent ".." components has to only be fixed for the one include of "shared-macros.mk"; the rest of included makefiles are in a fixed location under now-known WS_TOP. So I'd rather convert other Makefiles to use it, as part of the effort to change directory locations or otherwise a best practice ;)

alarcher, line 15: Except that WS_TOP comes from shared-macros.mk so it can only be used after including this file.
For subsequent includes WS_TOP can be used.

jimklimov, line 15: Ah... my bad - saw the question in email but did not see the context ;-\
Anyhow, the point stands - other locations welcome to use explicit double-dots just once per Makefile ;)

DrLou, line 15: Right, I remember it being baked into the build system some time ago... 
Remember, I've been living this dream since **_Day One_**! 
Indeed, omitting shared-macros was my oversight. Thanks for your notes, guys.

alarcher, line 15: No problem. Agreed, I use WS_TOP since my first PR in any subsequent include for the reason that you mentioned: only having to modify the path in one place when the component is relocated. I remember checking that for my first PR: I was not sure whether WS\* variables were defined in the environment as for illumos builds or just as Makefile variables. :)

alarcher:
I am interested in this update so if there is a chance to merge the PR soon-ish ;)

pyhalov:
At least first include should be fixed. And someone should check how inkscape likes it.

alarcher:
OK. BTW I checked the possibility of updating inkscape to 0.91 at some point but it requires gtkmm update together with atkmm, glibmm and giomm.... :S

alarcher:
https://github.com/alarcher/oi-userland/tree/gsl21

pyhalov:
Does inkscape like it? If so, please, also move it to corresponding category

alarcher:
I tried inkscape this afternoon after the modification and I could draw/read nice bubbles and stars. ;)

components/gsl/Makefile
avsej, line 15: @pyhalov so now it is required to define WS_TOP externally?

pyhalov, line 15: whhat do you mean? It was just broken Makefile.

avsej, line 15: but it has been merged https://github.com/OpenIndiana/oi-userland/blob/8b5b0c6c5686e7f219f920eefd19e6b175db49f4/components/gsl/Makefile#L15

pyhalov, line 15: I've fucked up...

alarcher, line 15: BTW the component actually redeclares the FMRI in the manifest as developer/gsl instead of library/math/gsl.
Same applies to the classification.

pyhalov, line 15: good to hear, as it's its actual FMRI

PR #1739, opened 2016-03-09
OpenTTD : fix missed dependencies after renaming of core package

jimklimov:
When renaming of core game package was done, myself and reviewers missed the media packages that referred to a now nonexistent package name... Rectifying that now :)


jimklimov:
On a side note, I verified that the older openttd-1.5.1 package did not include any media, and offered to download some upon startup because otherwise it cannot be played.

PR #1756, opened 2016-03-16
Initial commit of Darwinsys libmagic and file utility ("GNU file")

jimklimov:
File and libmagic (those used in GNU/Linux) at version 5.25

### DIFF
--- a/components/library/libmagic/Makefile
+++ b/components/library/libmagic/Makefile
@@ -1,0 +1,48 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov
+#
+include ../../../make-rules/shared-macros.mk
+
+# Note: The primary component we deliver initially is libmagic here
+# There is already a "file" utility with true Unix legacy in our OS
+# Still, for the sake of completeness and to match GNU/Linux userland
+# we ALSO deliver a "gfile" built from darwinsys sources, as a bonus.
+COMPONENT_NAME=		file
+COMPONENT_FMRI=		library/magic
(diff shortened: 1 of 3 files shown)


components/library/libmagic/libmagic.p5m
```diff
@@ -0,0 +1,40 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/include/magic.h
+file path=usr/share/misc/magic.mgc
+
+file path=usr/share/man/man3/libmagic.3
+# Renamed to avoid conflict with Unix/Sun/Oracle magic.4 and
```
pyhalov: We have /usr/gnu/share/man. Should we use it?

components/library/libmagic/Makefile
```diff
@@ -0,0 +1,49 @@
...
+include ../../../make-rules/shared-macros.mk
+
+# Note: The primary component we deliver initially is libmagic here
+# There is already a "file" utility with true Unix legacy in our OS
+# Still, for the sake of completeness and to match GNU/Linux userland
+# we ALSO deliver a "gfile" built from darwinsys sources, as a bonus.
+COMPONENT_NAME=		file
+COMPONENT_FMRI=		library/magic
+COMPONENT_VERSION=	5.25
+IPS_COMPONENT_VERSION= 5.25.0
```
pyhalov: Why do you add additional .0?

jimklimov: I thought 3-segment versions are required? And note that this is not "5.2.5" - it really is 25th minor in the 5th major series ;)

components/library/libmagic/libmagic.p5m
jimklimov: Might be ok. At least (I checked) `man` does find pages there, if the primary OS location does not override with another file. Which may be a consideration for Unix-vs-GNU variants of `magic.4` even more than for a conflict-less GNU `libmagic.3`...

components/library/libmagic/Makefile
pyhalov: No, they are not required

pyhalov:
Please, merge commits

PR #1776, opened 2016-03-21
Initial commit of recipe for tntdb libraries version 1.3

jimklimov:
tntdb is a sister project for cxxtools and tntnet, which wraps interaction from C++ (web or standalone) applications with SQL databases.


components/library/tntdb/Makefile
```diff
@@ -0,0 +1,61 @@
...
+COMPONENT_FMRI= library/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION= System/Libraries
+COMPONENT_LICENSE = LGPLv2.1
+COMPONENT_LICENSE_FILE = COPYING
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+# Sorry, this listing of our possible mysql implementations (in this repo)
```
pyhalov: We use mariadb 5.5 as default mysql implementation and PG_VERSION as default postgresql version. Please, fix PATH.

pyhalov:
It seems there are too much commits here. Perhaps, you have to rebase on oi/hipster HEAD

components/library/tntdb/Makefile
jimklimov: An OI-Hipster user can also install some other postgresql version (we have
components for 9.3, 9.4 and 9.5 at least), as well as another mysql version
or fork (percona is among them, IIRC?) The actual settings for the build
are taken from configure script's processing of "mysql_config" and
"pg_config" as found in PATH, and for the variants we have in userland
these are mediated somehow.

For the binary API, it "should not" matter which one a consumer is linked
against, and a customer of e.g. percona might be surprised about pulling
mariadb instead to fulfill this package requirements.

Although likely we should then ensure linking against the oldest version
available, or otherwise least common denominator (to only use ubiquitously
available versioned symbols, etc.) - so my theoretical opinion above may be
not that firm in practice.

2016-03-21 19:05 GMT+01:00 Alexander Pyhalov notifications@github.com:

pyhalov: For PostgreSQL we can use the oldest libpq. For MySQL I'm not sure. But I usually use one "default" version for userland. Now it's MariaDB 5.5 and PostgreSQL 9.3. I link with corresponding libraries  and try to ensure that RPATH is encoded in binaries. 
For libraries it's a bit more trickier.

jimklimov: Well... there is not much precedent to look at regarding this - bacula, php and python bindings for MySQL (though python/mysql/Makefile and web/php/php-5_4-ext-mysql and web/php/php-5_4-ext-pgsql seem useful, I'll try to regurgitate that when have time).

jimklimov:
I revised this PR, but mostly by pushing the ability to find "mysql_config" into mediators of mysql variant packages. The original "tntdb" package uses "pg_config" and "mysql_config" outputs to set up its CFLAGS and LDFLAGS.

It may be possible, but probably counter-natural, to try patch-away and override that "configure" logic and enforce some specific paths. Or it might be done as another PR if this state bites someone?

One thing that seems reasonable to do at this moment, and stop at, is to go the bacula Makefile way and set:

```
PATH=$(MYSQL_BINDIR.$(BITS)):$(PG_BINDIR.$(BITS)):/usr/bin:/usr/sbin
```

...which would enforce mariadb5.5 and default postgres without changing the original project. I'll prepare a separate PR built upon this one so you can pick and merge either one.

A further "nice" step for portability might be to decouple dependencies from a particular database variant, by replacing compiled-in references to "/usr/mariadb5.5/lib/..." to become the (already delivered) mediated "/usr/mysql/lib/..." paths.

jimklimov:
For the alternative solution, see https://github.com/OpenIndiana/oi-userland/pull/1786

jimklimov:
By the way, is it correct to SET the initial line of CONFIGURE_OPTIONS=... in the makefile, or only append CONFIGURE_OPTIONS+=... ? I copy-pasted the example from component that had it assigned without the plus. Is that a bug, or might be intentional there?

pyhalov:
If you don't want default CONFIGURE_OPTIONS , you start with =

jimklimov:
PR deprecated in favor of #1786

PR #1785, opened 2016-03-23
MySQL and PostgreSQL mediators updated

jimklimov:
- add an implementation tag for postgres mediators (there are other PG-based distros like enterprisedb)
- add "mysql_config" to mediated "binaries" in common bin dir, including the 64-bit variant, so dependent package configuration can become more simple


pyhalov:
Pleas, bump COMPONENT_REVISION

pyhalov:
First explicit revision is 1

pyhalov:
Merge commits

PR #1787, opened 2016-03-23
Bumped cxxtools with largefile support in 32-bit build

jimklimov:
Just in case, so all packages in the family have the same build settings


pyhalov:
Bump COMPONENT_REVISION

components/library/cxxtools/Makefile
```diff
@@ -17,23 +17,32 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=	cxxtools
 COMPONENT_VERSION=	2.2.1
+COMPONENT_REVISION=	2
```
pyhalov: By default it's 0, so first one you set explicitly is 1.

pyhalov:
Fix COMPONENT_REVISION and merge commits.

PR #1796, opened 2016-03-25
ncurses: install pkg-config .pc files

avsej:


avsej:
@pyhalov but the `make sample-manifest` generated only paths without ARCH64, how I can verify that it works correctly?

avsej:
maybe it will be better to make symlinks?

how I can package to install it locally?

pyhalov:
No, the issue is that 64-bit files rewrote 32-bit. You should find out why this happened and fix it:)

avsej:
ok, I see

components/library/ncurses/Makefile
```diff
@@ -37,9 +37,9 @@ COMPONENT_ARCHIVE_HASH= \
 COMPONENT_ARCHIVE_URL=	http://ftp.gnu.org/pub/gnu/$(COMPONENT_NAME)/$(COMPONENT_ARCHIVE)
 COMPONENT_BUGDB=	library/ncurses
 
-include ../../make-rules/prep.mk
-include ../../make-rules/configure.mk
-include ../../make-rules/ips.mk
+include ../../../make-rules/prep.mk
```
jimklimov: While you're here, fix these relative symlinks to become `$(WS_TOP)/make-rules/...` (so only the topmost include has a changeable explicit path).

components/library/ncurses/patches/05-inject-pkg_config_path.patch
```diff
@@ -0,0 +1,12 @@
+--- ncurses-5.9/configure.orig	2016-04-12 10:42:44.808765427 +0300
++++ ncurses-5.9/configure	2016-04-12 10:42:56.607701646 +0300
+@@ -3621,6 +3621,9 @@
+	echo "$as_me:3621: checking if we should install .pc files for $PKG_CONFIG" >&5
+ echo $ECHO_N "checking if we should install .pc files for $PKG_CONFIG... $ECHO_C" >&6
+
++	if test -n "$PKG_CONFIG_PATH" ; then
++		PKG_CONFIG_LIBDIR=$PKG_CONFIG_PATH
++	fi
+	# Leave this as something that can be overridden in the environment.
```
avsej, line 10: @pyhalov this patch introduces `PKG_CONFIG_PATH` to ncurses

PR #1806, opened 2016-03-29
Optionally integrate ccache to speed up oi-userland (re-)builds

jimklimov:
If `ccache` is installed and found in `PATH`, and the feature is not disabled by `CCACHE_DISABLE=true` (as for e.g. "production builds" where you might want to ensure that nothing interferes), then use `ccache` to speed up rebuilds (e.g. developer iterations, build farm jobs). The effect should be especially noticeable on C++ code that takes longer to parse and compile than plain C.

Review `man ccache` for optional configuration tuning, such as cache size etc.

Alas, the variable-passing in various Makefiles does not allow to just define `CC:=ccache $(CC)` (some code expects space-less single-token values) so a simple wrapper script for each variant is the necessary evil. Alternately, use of `CC=/path/to/ccache` with `export CCACHE_CC=/path/to/gcc-4.9` is a bit unwieldy with our mix of CC/CXX compilers and bitness.

Yet another "simple solution" might be to prefix the `PATH` with `/usr/lib/ccache` (which contains `gcc`, `g++` and similar symlinks to `/usr/bin/ccache`) and then add the chosen GCC version `bin` to the `PATH` and then other directories, but too many component Makefiles re-define the `PATH` for their needs assuming that the `CC` and other similar tools do not need it.

I chose against a more complicated single script that would evaluate its use-case (e.g. according to `basename $0`) in favor of having minimal run-time overhead.


make-rules/shared-macros.mk
```diff
@@ -345,6 +348,15 @@ export PARFAIT_NATIVESUNCXX=$(SPRO_VROOT)/bin/CC
...
 export PARFAIT_NATIVEGXX=$(GCC_ROOT)/bin/g++
 
+# This should evaluate to empty string or a pathname like /usr/bin/ccache
+# depending on current PATH value and "which" implementation.
+# The assignment via ":=" is important, to only do this once in a Makefile,
+# and not on every reference to the value as "=" assignment would result in.
+# Review `man ccache` for optional configuration tuning, like cache size etc.
+# For production builds or suspected errors you can disable this feature by
+# setting CCACHE_DISABLE=true (as an environment or makefile variable value).
+export CCACHE := $(shell if test -n "$(CCACHE)" ; then echo "$(CCACHE)"; else test x"$${CCACHE_DISABLE-}" = xtrue -o x"$(CCACHE_DISABLE)" = xtrue || for F in "$$CCACHE" `which ccache 2>/dev/null | egrep '^/'` /usr/bin/ccache; do test -x "$$F" && echo "$$F" && echo "USING CCACHE FOR OI-USERLAND: $$F" >&2 && break; done; fi)
```
pyhalov: Please, make this more readable... Perhaps, we can rewrite it using less shell and more make expressions, or split it in several steps. At least, split it in several lines...

make-rules/shared-macros.mk
```diff
@@ -367,6 +379,17 @@ CXX.gcc.64 =	$(GCC_ROOT)/bin/g++
...
+ifneq ($(strip $(CCACHE)),)
+CCACHE_WRAP_ROOT   =	$(WS_TOOLS)/ccache-wrap
+export CC_gcc_32  :=	$(CC.gcc.32)
+export CC_gcc_64  :=	$(CC.gcc.64)
+export CXX_gcc_32 :=	$(CXX.gcc.32)
+export CXX_gcc_64 :=	$(CXX.gcc.64)
+CC.gcc.32  :=	$(CCACHE_WRAP_ROOT)/CC.gcc.32
+CC.gcc.64  :=	$(CCACHE_WRAP_ROOT)/CC.gcc.64
+CXX.gcc.32 :=	$(CCACHE_WRAP_ROOT)/CXX.gcc.32
+CXX.gcc.64 :=	$(CCACHE_WRAP_ROOT)/CXX.gcc.64
```
pyhalov, line 428: Are you sure that wrappers are executable after git clone? Perhaps, avoid using scripts? Do we know whic components break when CC is not a single word?

pyhalov:
One general note: perhaps we want to integrate ccache component first?

jimklimov:
RE general note: ccache was one of my first PRs to the project - https://github.com/OpenIndiana/oi-userland/pull/1471 ... this integration was laying around and waiting to be finished (i.e. needed me to recompile the same project a few times too many while iterating the recipes)

make-rules/shared-macros.mk
jimklimov: fixed

make-rules/shared-macros.mk
jimklimov, line 428: Yes, they are committed with POSIX bits in place.

I hoped to avoid using scripts indeed, but see my comment in the PR description regarding the variants I looked at.

As for components which break - I can't point my finger at the moment, and I did not ever rebuild all of this repository, but generally it is any component that does several layers of shell expansions (e.g. at least those that have their own build routines, so we end up with command lines like `./configure CC=/usr/bin/ccache /usr/gcc/4.9/bin/gcc CFLAGS=...` and the program (e.g. `configure` in this example) complains that `/usr/gcc/...` is an unknown parameter.

pyhalov:
Ok, sorry, forgot that we have it :)

pyhalov:
Also I'd disable CCACHE by default and didn't issue warning if it's disabled.

jimklimov:
Done, quietly disabled by default. Reports explicit `CCACHE_DISABLE=true` and enablement via `CCACHE_DISABLE=false`, quiet for all other values (e.g. the default empty value).

Re-tested with both envvar and make-argument settings. The "true" value (to disable), if set anywhere, wins. The user account can add to `~/.profile` lines like this:

```
# To speed up builds (OI-userland and elsewhere)
CCACHE_DISABLE=false
export CCACHE_DISABLE
```

FWIW, I also add these:

```
# For oi-userland tarball storage
USERLAND_ARCHIVES=$HOME/shared/oi-userland/downloads/
export USERLAND_ARCHIVES

# To build oi-userland with deps on external repos, like the local one
#ENCUMBERED=yes
#export ENCUMBERED
```

pyhalov:
Tried in components/check gmake CCACHE=true clean publish : 
configure: error: in `/export/home/alp/srcs/tests/oi-userland/components/check/build/i86':
configure: error: C compiler cannot create executables

configure:3581: checking whether the C compiler works
configure:3603: /export/home/alp/srcs/tests/oi-userland/tools/ccache-wrap/CC.gcc.32 -m32 -O3   -m32 conftest.c  >&5
configure:3607: $? = 0
configure:3645: result: no
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "Check"
| #define PACKAGE_TARNAME "check"
| #define PACKAGE_VERSION "0.10.0"
| #define PACKAGE_STRING "Check 0.10.0"
| #define PACKAGE_BUGREPORT "check-devel at lists dot sourceforge dot net"
| #define PACKAGE_URL ""
| #define PACKAGE "check"
| #define VERSION "0.10.0"
| /* end confdefs.h.  */
| 
| int
| main ()
| {
| 
|   ;
|   return 0;
| }
configure:3650: error: in `/export/home/alp/srcs/tests/oi-userland/components/check/build/i86':
configure:3652: error: C compiler cannot create executables

I suppose, environment is not defined in wrappers

README
```diff
@@ -55,6 +55,25 @@ Building the Bits.
 
     You can build individual components or the contents of the entire gate.
 
+  Integration with ccache to speed up re-builds
```
pyhalov, line 58: Perhaps, this belongs to doc/makefile-variables.txt?

jimklimov:
EWRONGTEST ;)

The `CCACHE` variable is the name (or full path) to the desired binary. So you compiled with "true" and likely got no output file.

To enable/disable the feature, use the inverse `CCACHE_DISABLE` (it is also the envvar that the program itself uses to skip its work, if the user wishes so - thus reused "as is" including the clumsy inverse syntax).

README
jimklimov, line 58: A copy/rephrasing of this - maybe. But I'd keep it in the main README as well to expose visibility so more collaborators can benefit from this feature, should they choose to.

pyhalov:
OK, how can I test it? I mean, I have successfully built ccache with mkdir /tmp/test ;gmake CCACHE_DISABLE=false COMPONENT_BUILD_ARGS="-j4 CCACHE_DIR=/tmp/test" clean publish , but /tmp/test is empty...

jimklimov:
You mean you've built `ccache` itself? and there was no copy of it installed in the `PATH` on your host at the moment?

In this case the makefile may have reported that ccache was not found so it proceeds with the usual procedure.

I tried to verify CCACHE_DIR now, but even though it is passed in envvars into the wrapper script (checked) and is recognized by ccache program (checked), it does not create anything in that directory, not even the cache structure :\

It does somewhat work (at least create structure) in the `$HOME/.ccache` however "Files in cache" remains at zero. Maybe because I used a project that compiles several `*.c` sources at once into a binary, bypassing object files.

pyhalov:
No, I mean I don't see ccache creating any files.

jimklimov:
It seems that passing values into some `configure` driven builds required a bit more care. 

I pushed a few more iterations of my experiments, it seems that now caches are populated, including a symlink in `$HOME/.ccache` and a `CCACHE_DIR=/...` value. It is too cumbersome to produce an exhaustive list of values to extend (ccache has many options, and perhaps the chosen cachedir's config file is a better place to pass them).

jimklimov:
As a test, I rebuilt libsodium. The initial build with ccache took about 1h30m on my laptop (VM, and doing backups at this time of night), but a rebuild got many direct hits into the ccache and was reduced to under 20m.

jimklimov:
Is this good to merge and go?

Having this one in common tree would really simplify and speed up my work on re-spinning other PRs to bring them up to requested standard ;)

make-rules/shared-macros.mk
```diff
@@ -345,6 +348,39 @@ export PARFAIT_NATIVESUNCXX=$(SPRO_VROOT)/bin/CC
...
+# For production builds or suspected errors you can disable this feature by
+# setting CCACHE_DISABLE=true (as an environment or makefile variable value).
+# In fact, if you want to use this feature, you must set CCACHE_DISABLE=false
+# Still, absence of ccache in PATH is not considered a fatal error since the
+# build would just proceed well with original compiler.
+export CCACHE := $(shell \
+    if test -n "$(CCACHE)" ; then \
+        echo "$(CCACHE)"; \
+    else \
+        if test x"$${CCACHE_DISABLE-}" = xtrue -o x"$(CCACHE_DISABLE)" = xtrue ; then \
```
pyhalov: true seems to be not necessary here, as it's disabled by default

pyhalov:
Something still seems to be wrong.
env |grep CC
CCACHE_DIR=/tmp/cache
CCACHE_DISABLE=false

After gmake clean build in library/libmtp /tmp/cache is empty. 
Build system ensures me "USING CCACHE FOR OI-USERLAND"

make-rules/shared-macros.mk
jimklimov: As a ccache setting envvar which we abuse here (and duplicate as a
make-var), it has two valid values: false and true. And empty which is
equivalent to false (e.g. by default if you execute ccache - you want it to
act).

Per your suggestion, for us the default is true (disable and do not even
call ccache unless requested otherwise).

The current logic QUIETLY does nothing if the variable is
empty/unset/invalid, and VERBOSELY does or doesn't (try to) enable ccache
if the value is explicit and valid.

## 

31.03.2016 17:46 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
Trying to reproduce the issue - and can't, my freshly made /tmp/ccache-test directory does get populated.
One possible caveat : are your `CCACHE_*` envvars just defined, or also `export`ed? Bit me a few times ;)
Another caveat: do reconfigure the project if applicable, so it has the tweaked `CC` etc. defined.
Finally, I had issues with a project where all products depended on several sources at once (in short, `gcc *.c -o binprog`) - seems this case is not supported and gets silently skipped. Likewise, files detected to have dynamic contents (TIME macros, etc.) are also not cached.

So, overall, with this branch checked out, I prepared the test like this (assuming `bash`-like syntax):

```
export CCACHE_DISABLE=false
export CCACHE_DIR=/tmp/ccache-test
mkdir -p "$CCACHE_DIR"

export CCACHE_LOGFILE="$CCACHE_DIR/build.log"
### With logfile in place, it is much easier to debug. There is no separate STDERR.

export DEBUG=yes
### With DEBUG, the current branch will call variants of wrapper scripts that
### print out relevant CCACHE envvars and the active settings interpreted
### by the program itself

while : ; do ps -ef | grep -v grep | grep ccache; sleep 1; done | egrep -v '\.log' &
/usr/gnu/bin/tail -F "$CCACHE_LOGFILE" &

cd components/library/libmtp
time gmake clean build

...

time gmake clean build

### compare the two timings
```

The USB-related libs from your test happened to be a bad example (do not build for me, apparently some dependency packages are missing); so I rebuilt with libmagic - got about 2m30s initially, 1m50s cached. Most of this spent in configure script, though... (No idea why they are so slow in Hipster VM, compared to same scripts in another VM with Linux on the same laptop - using the Hipster as NFS/ZFS homedir storage! So it is likely not storage-constrained by speed...) :

```
real    2m29.798s
user    0m50.026s
sys 1m6.195s

real    1m51.117s
user    0m28.411s
sys 0m56.847s
```

doc/makefile-variables.txt
```diff
@@ -107,3 +107,25 @@ that need to be applied to the test output for just this component.
...
+
+If you frequently rebuild the same code, such as when you maintain a build server
+or iterate recipes for the same component, you can benefit by caching the build
+products with "ccache" - so for repeated input conditions you'd get same output
+objects quickly.
+
+* CCACHE_DISABLE is an inverted toggle of the "ccache" program, which our common
+  shared-macros.mk Makefile also honours. By original project's intent, if you
+  execute "ccache" then you want it to do its job. If you are debugging it or
+  unsure if it causes problems, you can easily temporarily disable the "ccache"
```
pyhalov, line 125: Modify description so that it clearly states that this feature is disabled by default, to use it, you should set... Perhaps, invert variable back to "CCACHE_ENABLE" as it's not default?

jimklimov, line 125: Description - re-done.

I'd rather not add the re-inverted variable, because 1) in the future a custom `CCACHE_*` variable might conflict with what the original project offers, and 2) we'd have two different toggles for the same feature whose settings can become conflicting and not apparent to users.

pyhalov:
Works for me now... I'd removed debug versions of tools and renamed var to CCACHE_ENABLE (with default behavior unchanged)

jimklimov:
Cool, we are getting warmer :)

As for the envvar - I wrote above why I'd rather not add complexity, even though the non-inversed values are generally better understood. Reasoning and usage for this case is documented, so I hope it would suffice this way.

As for debug versions - they might be helpful to figure out why things go wrong, like in your case, and should not be a problem by default (ccache disabled) or by second-default (ccache enabled, debug not enabled). Might hide them behind a less-obvious variable name though (`DEBUG_USERLAND_CCACHE` comes to mind) to avoid undesired mis-enabling.

pyhalov:
I'd removed debug versions. If someone digs into this, he'll do one-line change to the scripts. As for envvar name, non-inversed logic is really simpler. Especially when "DISABLE" is default. I suggest rename current setting.

jimklimov:
I think to avoid future namespace clash, this oi-userland variable would be better re-named e.g. `ENABLE_CCACHE` (empty="false"=off by default, "true" upon request). Then if actual CCACHE is detected, it might be or not be neutered by its own `CCACHE_DISABLE` (empty="false"=do the job by default).

>  I'd removed debug versions of tools and renamed var ...
> Why not make it as your subsequent PR then, for me to nit-pick upon? ;)

pyhalov:
sound good

jimklimov:
Wolves not hungry, sheep intact... RIP the shepherd ;)

doc/makefile-variables.txt
```diff
@@ -107,3 +107,40 @@ that need to be applied to the test output for just this component.
...
+disabled by default. See below for hints how to enable and optionally configure it.
+
+* To enable the "ccache" integration in OI-Userland build system you must
+  `export ENABLE_CCACHE=true` in the calling shell, or set the make-variable
+  like this: `gmake ENABLE_CCACHE=true build`
+* CCACHE_DISABLE is an inverted toggle of the "ccache" program, which our common
+  shared-macros.mk Makefile also honours. By original project's intent, if you
+  execute "ccache" then you want it to do its job. If you are debugging it or
+  unsure if it causes problems, you can easily temporarily disable the "ccache"
+  logic by `export CCACHE_DISABLE=true` so it quickly calls the original compiler.
```
pyhalov: At least, make then consistent - (ENABLE|DISABLE)_CCACHE

doc/makefile-variables.txt
```diff
@@ -107,3 +107,40 @@ that need to be applied to the test output for just this component.
...
+  So to enable the caching feature you can also use `export CCACHE_DISABLE=false`.
+* In case of conflicting environtment and/or makefile variables regarding this
+  feature, any explicit request to disable use of "ccache" takes precedence.
+* With default settings (only `CCACHE_DISABLE=false` in place), the cache and
```
pyhalov: ENABLE_CCACHE

make-rules/shared-macros.mk
```diff
@@ -345,6 +348,48 @@ export PARFAIT_NATIVESUNCXX=$(SPRO_VROOT)/bin/CC
 export PARFAIT_NATIVEGCC=$(GCC_ROOT)/bin/gcc
 export PARFAIT_NATIVEGXX=$(GCC_ROOT)/bin/g++
 
+# This should evaluate to empty string or a pathname like /usr/bin/ccache
+# depending on current PATH value and "which" implementation.
+# The assignment via ":=" is important, to only do this once in a Makefile,
+# and not on every reference to the value as "=" assignment would result in.
+# Review `man ccache` for optional configuration tuning, like cache size etc.
+# For production builds or suspected errors you can disable this feature by
+# setting CCACHE_DISABLE=true (as an environment variable value) to tweak
```
pyhalov: Doesn't they behave identically?

doc/makefile-variables.txt
jimklimov: No. "CCACHE_DISABLE" is from ccache program. It is outside our control, and it has the inverted semantics for reasons already told. Even if we don't want to abuse the other project's variable, we can not break it into what we deem prettier.

doc/makefile-variables.txt
```diff
@@ -134,7 +134,14 @@ disabled by default. See below for hints how to enable and optionally configure
   can be helpful to inspect its actual parsing of configuration), and/or use
   `export CCACHE_LOGFILE=/tmp/ccache.log` to trace its activities.
 * You can dedicate a cache directory different from the default `$HOME/.ccache`
-  with `export CCACHE_DIR=/tmp/ccache-dir; mkdir -p $CCACHE_DIR`.
+  for example with `export CCACHE_DIR=/tmp/ccache-dir; mkdir -p $CCACHE_DIR`.
+* Troubleshooting: If no files appear in the cache, verify permissions and space.
+  Also enable the log file and/or inspect configuration with `ccache -p` (see above)
+  to see details about wrapped compilations. In particular, "ccache" might not
+  actually cache the build products if the file types are unsupported (see man
```
pyhalov: man what?

make-rules/shared-macros.mk
```diff
@@ -359,6 +359,8 @@ export PARFAIT_NATIVEGXX=$(GCC_ROOT)/bin/g++
 # is currently the default) to not even define the usage of wrapper in the
 # userland-building makefile system.
 # If you want to speed up your re-builds, you must set ENABLE_CCACHE=true.
+# For legacy reasons, the inverted CCACHE_DISABLE variable (configuration of the
```
pyhalov: I'm still not convinced that we need two variants of the var. We have no "leagacy reasons", code is fresh :)

doc/makefile-variables.txt
jimklimov: Eh, it was near there somewhere... ;)

make-rules/shared-macros.mk
jimklimov: It is ccache's legacy ;)

At the very least, it does not make sense for us to add overhead of calling ccache that would do nothing about caching because we know it won't due to its envvar setting. It might be meaningful for development and debugging of ccache (e.g. does it really do nothing when the setting is present), but not for our build system.

And conversely, if the user jumped through loops to say CCACHE_DISABLE=false when this is already the program's default, it must mean something :)

README
```diff
@@ -55,6 +55,27 @@ Building the Bits
...
+
+    The `ccache` component is available as part of the project repository.
+    Once you have the resulting package installed, you can pass the `make`
+    argument or environment variable `ENABLE_CCACHE=true` to wrap the GNU
+    compiler invocations with the caching program - so that the same inputs
+    would re-produce same outputs quickly.
+
+    You can pre-set this variable in your user account `~/.profile` like this:
+
+````
```
pyhalov: BTW, Readme is not in markup. Avoid using ```.

pyhalov:
It seems ENABLE_CCACHE=true is not enough... 
export ENABLE_CCACHE=true
export CCACHE_DIR=/tmp/ccache-test
doesn't enable ccache, you have explicitly set export CCACHE_DISABLE=false
And something is wrong with latest version, even with 
export ENABLE_CCACHE=true
export CCACHE_DIR=/tmp/ccache-test
export CCACHE_DISABLE=false
after building libmtp /tmp/ccache-test is empty

jimklimov:
Just can not reproduce the issue - works for me: files appear in cache, logs are updated, statistic numbers increase...

Did you try to instrument the ccache-wrap scripts as in that reverted patch? There is a copy in https://github.com/jimklimov/oi-userland/tree/feature/integrated-ccache-wDebug/tools/ccache-wrap-debug ;)

Unfortunately I can not complete building "libmtp" due to "Undefined symbol" and then lots of "usb_*" references. But even so, the attempt works and makes a cache directory with some stuff in it.

```
$ set | grep CCA

(nothing displayed)

$ rm -rf /tmp/ccache-libmtp ; CCACHE_DIR=/tmp/ccache-libmtp ENABLE_CCACHE=true gmake clean build 

... (build partially succeeds)

$ find /tmp/ccache-libmtp -type f | wc -l; CCACHE_DIR=/tmp/ccache-libmtp ccache -s

166
cache directory                     /tmp/ccache-libmtp
primary config                      /tmp/ccache-libmtp/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
cache hit (direct)                     5
cache hit (preprocessed)               0
cache miss                            56
called for link                        3
called for preprocessing              29
compile failed                         7
preprocessor error                     2
bad compiler arguments                 1
autoconf compile/link                 25
no input file                          7
files in cache                       133
cache size                           1.0 MB
max cache size                       5.0 GB

```

pyhalov:
OK, perhaps something with my environment. It's working now.  Just fix Readme

jimklimov:
done

README
```diff
@@ -83,6 +100,17 @@ export USERLAND_ARCHIVES
 
         $ cd (component-dir) ; gmake publish
 
+    If you build several components, and some are required dependencies for the
```
pyhalov: Sorry, haven't seen this. We removed references to ENCUMBERED as setting ENCUMBERED in ~/.profile breaks gmake components.mk in components dir.

jimklimov: ok, i'll remove it (tomorrow probably) and tackle the ability to depend on
local just-built repo separately.

I thought encumbered should be good (didn't break anything for me when
applied to profile after 'gmake setup') - but seems there should appear an
equivalent constrained to compilation or even publishing of just the leaf
components?

## 

05.04.2016 21:12 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov: Makefile has 

```
(WS_TOP)/components/$(ENCUMBERED)components.mk:        # Makefile # $(WS_TOP)/components
        @echo "Generating component list..."
        @$(BASS_O_MATIC) --workspace=$(WS_TOP) --components=paths \
                | sed -e 's;^$(shell pwd)/;;' \
                | sed -f components.ignore \
                | sed -e 's;^;COMPONENT_DIRS += ;g' >$@
```

So setting ENCUMBERED variable breaks components.mk generation. It's supposed to be set including $(WS_MAKE_RULES)/encumbered.mk .

jimklimov: The encumbered.mk is a static usecase for really encumbered components. For
my practical usecase it allows to publish test packages whose dependencies
I built and published earlier - so they are not available in real
oi-userland (your internet repo) yet...

## 

05.04.2016 22:41 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
README fixed back

jimklimov:
FYI: I guess I managed to reproduce your issues with ccache not doing its work.
It seems that if the CCACHE_DISABLE envvar is defined and exported (with any value - empty, "1", "0",  "yes", "no", "true", "false" etc.) then it is treated as "true" and disables ccache.

It is easy to reproduce with:

```
jim@jimoi:~$ unset CCACHE_DISABLE
jim@jimoi:~$ ccache -p | grep -i disable
(default) disable = false
jim@jimoi:~$ export CCACHE_DISABLE
jim@jimoi:~$ ccache -p | grep -i disable
(default) disable = false
jim@jimoi:~$ export CCACHE_DISABLE=""
jim@jimoi:~$ ccache -p | grep -i disable
(environment) disable = true
jim@jimoi:~$ export CCACHE_DISABLE=false
jim@jimoi:~$ ccache -p | grep -i disable
(environment) disable = true
```

I'd treat this as a ccache bug (did not dig if it ever worked or this is a recent regression).
Now that we do not use this envvar and do not suggest it in the docs as an e.g. profile setting, it is not a (recurring) problem for us anymore ;)

This also happens in fresh Linux builds of ccache, so completely not our problem ;)
And what's worse, this causes other booleans to be misinterpreted too. Bug posted to https://bugzilla.samba.org/show_bug.cgi?id=11826

So your hunch to use another toggle was better than mine, in the end ;)

PR #1809, opened 2016-03-29
Initial commit of zeromq-4.1.4 (libzmq ABI version 5)

jimklimov:
Mediated version of zeromq5 (note that sources are 4.1.\* but ABI is marked 5.*)


components/python/pyzmq/Makefile
```diff
@@ -10,15 +10,32 @@
...
+
+# The zeromq components define this directory structure
+ZEROMQ_DIR = $(USRDIR)/zeromq-$(NEEDED_ZEROMQ_VER)
+
+# We define these before including any common structure makefiles
+CXXFLAGS   = -I$(ZEROMQ_DIR)/include
+CPPFLAGS   = -I$(ZEROMQ_DIR)/include
+CFLAGS     = -I$(ZEROMQ_DIR)/include
+LDFLAGS.32 = -L$(ZEROMQ_DIR)/lib
+LDFLAGS.64 = -L$(ZEROMQ_DIR)/lib/$(MACH64)
```
pyhalov: This changes should be in independent PRs as we will not have zeromq 4.1 when pyzmq is rebuilding. Also check that correct version of zeromq is actually chosen when mediator is set to lower version.

jimklimov: Sounds complicated... anyhow a bump of PyZMQ requires ZeroMQ-4.1.2 (or newer hopefully too). Why not do both at once and so ensure we have a working level in the gate.

Though indeed the matrix to test all installable libzmq versions vs build and/or installation of all consumer versions (whether requiring a specific ABI or not), sounds like a large undertaking :\

pyhalov: This is how build server works. It will build all changed components at once, which are zeromq and pyzmq. And pyzmq build will fail because we don't have new zeromq version.

jimklimov: PyZMQ update extracted into https://github.com/OpenIndiana/oi-userland/pull/1847

components/library/zeromq/zeromq5/zeromq.p5m
```diff
@@ -0,0 +1,30 @@
+#
```
pyhalov, line 1: Shouldn't you remove zeromq3/zeromq.p5m ?
Also, shouldn't we depend on ALL zeromq versions to be respectful to packages dependent on older library/c++/zeromq?

components/library/zeromq/zeromq5/zeromq5.p5m
```diff
@@ -0,0 +1,136 @@
...
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_curve.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_inproc.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_ipc.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_null.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_pgm.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_plain.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_tcp.7
+file path=$(COMPONENT_USR_DIR)/share/man/man7/zmq_tipc.7
+
```
pyhalov, line 136: Remove last empty line

components/library/zeromq/zeromq5/zeromq.p5m
jimklimov, line 1: That is an interesting point... I don't really know.

On one hand, in my testing of building and installing all these packages - the copy of `zeromq.p5m` that has the highest IPS_COMPONENT_VERSION (newest zeromq version number) wins and is the one that gets published and installed, and so "pkg install zeromq" pulls the package with highest number. So from limited practice, it does not hurt to have this file supplied in every version, at least as long as I juggle them and they are not all merged into one common repository. Perhaps after integrations of all these "initial zeromq version X" PR's we can follow up with a PR to remove this file from older versions. 

On another hand, shouldn't the packages in this gate that are dependent on older zeromq begin complaining during rebuild (if API/ABI did change incompatibly) and so cause revision of their recipes to either explicitly depend on their zeromq-x.y (packages and libs) or bump if upstream updated the code to work with newest zeromq? 

Hipster is the edge expected to bleed ;-\

pyhalov, line 1: I'd follow what we've done for gcc versions. And for gcc we just preserved old p5m files for runtime. But I would include dependencies on old zeromq versions to be  respectful to packages dependent on older library/c++/zeromq.

jimklimov, line 1: Just to be sure: in this context you refer to all (3 or 4) branches of
zeromq that are under review at the moment, or just the 3.\* one that was
available before I began about a month ago? Should the
backwards-compatibility ensure that versioned /usr/lib/libzmq_.so_ of the
updated zeromq 3.\* are still available at those old paths? (During PR these
ABI-versioned links might be removed from the common locations)

## 

07.04.2016 8:46 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov, line 1: I refer to all versions available before this integration. When you add next version, you'll also add it to corresponding zeromq meta-package.

xen0l:

PR #1814, opened 2016-03-30
(WIP) Initial commit of bleeding-edge zeromq-4.2.0.x (git HEAD)

jimklimov:
The currently stated ABI is version 5 but it is different from 4.1.4 ABI (e.g. malamute does not build with the latest official libzmq release tarball).

I asked the project leader if the ABI version should be bumped, so maybe soon this will boast "zeromq6" and not conflict with the latest stable release.


jimklimov:
Additional question: for this and other packages where we might want something newer than an official release, is there any "protocol" (numbering, naming, ips facet) to mark pre-release packages so they don't conflict with later releases of the same number? e.g. can I use "4.2.0.0.yyyymmddhhmmss" or some-such pattern to tag these bleeding edges, and publish "4.2.0.1.0" when a release tarball is ultimately issued (and then probably continue as "4.2.0.2.yyyymmddhhmmss" for bleeding-edges)?

The ZeroMQ family of projects in particular is a fast-moving bunch, with the project leaders wishing to avoid making releases fixed up at some point in time - rather everyone should use the HEAD of everything in the family and most of the time it is stable. Makes sense for continuous delivery, where their business is...

pyhalov:
We usually use latest release number. But we usually not deliver last release and "last_release_dev" packages simultaneously

jimklimov:
So we can just keep COMPONENT_VERSION=4.2.0 and bump COMPONENT_REVISION as
we randomly pick newer commits, until one day the FMRI points to a build
from the official tarball version rather than from git snapshot?

2016-03-30 14:10 GMT+02:00 Alexander Pyhalov notifications@github.com:

pyhalov:
I think so. The issue arises if next git commit changes ABI.

jimklimov:
Alas, with this project - and especially on unpublished head, this case is
quite possible.

## 

30.03.2016 16:17 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov:
Is zeromq a runtime dependency for software you want to integrate or just build-time? Can we compile libzmq in a component which requires it and avoid delivering dlls?

jimklimov:
I'd say both. There is for example malamute that I want integrated as a
library and tool on its own, as well as a bunch of MQ-related projects at
work (and in examples) which commonly link against -lzmq -lczmq -lmlm and
maybe some more depending on the project. So static linking with private
copies of the libs would be a counter-pattern here. And also C++
compilations are slow enough to want them done just once ;)

My recent burst of activity is about expanding the ability to port and run
further projects, this time around C/C++-based middleware (webapps and
message queues), and so make OI/Hipster less unattractive for newcomers -
since their efforts to run what they need would be smaller and smaller, so
community might grow.

2016-03-30 16:55 GMT+02:00 Alexander Pyhalov notifications@github.com:

pyhalov:
Then deliver links only to the latest .so files to /usr/lib and ensure pyzmq still compiles and sort of works. Also ensure all zeromq consumers explicitly specify which version they want.

jimklimov:
See attached ;) a PoC conversion of a newest ZeroMQ packaging recipe into what looks like what you asked for. Should this pattern be repeated into other versions?

```
root@jimoi:/root# find /usr | grep zmq | grep -v /usr/zeromq
/usr/include/zmq_utils.h
/usr/include/zmq.h
/usr/lib/libzmq.so
/usr/lib/amd64/libzmq.so
/usr/lib/amd64/pkgconfig/libzmq.pc
/usr/lib/pkgconfig/libzmq.pc

root@jimoi:/root# cat /usr/lib/amd64/pkgconfig/libzmq.pcprefix=/usr/zeromq-4.2
exec_prefix=/usr/zeromq-4.2
libdir=/usr/zeromq-4.2/lib/amd64
includedir=/usr/zeromq-4.2/include

Name: libzmq
Description: 0MQ c++ library
Version: 4.2.0
Libs: -L${libdir} -lzmq
Libs.private: -lstdc++
Cflags: -I${includedir}


root@jimoi:/root# cat /usr/lib/pkgconfig/libzmq.pc
prefix=/usr/zeromq-4.2
exec_prefix=/usr/zeromq-4.2
libdir=/usr/zeromq-4.2/lib
includedir=/usr/zeromq-4.2/include

Name: libzmq
Description: 0MQ c++ library
Version: 4.2.0
Libs: -L${libdir} -lzmq
Libs.private: -lstdc++
Cflags: -I${includedir}

```

PyZMQ does build and publish on a system with only this variant of zeromq installed, though with warnings about "/usr/gnu/lib" (32-bit) referenced in 64-bit binaries.

So far I didn't progress on other PRs that regard magic about library linking, so won't endeavour to fix this one either.

pyhalov:
Shouldn't *.pc files set  Libs: "-L${libdir} -R${libdir} -lzmq", not "Libs: -L${libdir} -lzmq"  ?

pyhalov:
You missed "ensure all zeromq consumers explicitly specify which version they want" part. Set -I/path/to/specific/zeromq/include -L/path/to/specific/zeromq/lib  CPPFLAGS/LDFLAGS in pyzmq. Ensure that they are used. It may require patching pyzmq setup.py.

jimklimov:
PyZMQ bump will be addressed in "bump/zeromq5" PR (for PyZMQ update to 14.7.0 which requires zeromq release 4.1.2+) - see https://github.com/OpenIndiana/oi-userland/pull/1809 and https://github.com/OpenIndiana/oi-userland/pull/1847

xen0l:

PR #1818, opened 2016-03-30
Initial commit of augeas tools version 1.4.0

jimklimov:
Library and shell toolkit for config-file inspection and editing


components/augeas/augeas-vim.p5m
```diff
@@ -0,0 +1,33 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# This binding needs ViM to be functional, but we don't want to require
+# installation of ViM if the end-user just installs the grouper package
+# and really only wants augtools, not the editor
+depend fmri=editor/vim	type=optional
```
pyhalov: this dependency doesn't work - https://docs.oracle.com/cd/E53394_01/html/E54820/dependtypes.html#PKDEVglujw

components/augeas/augeas-tools.p5m
```diff
@@ -0,0 +1,39 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=usr.*/man/.+ -> default mangler.man.stability uncommitted>
+
+# Note: the tools require their libs; pkgdepend does this bit automagically :
+depend fmri=$(COMPONENT_FMRI)-libs@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)	type=optional
```
pyhalov: This shouldn't be necessary

components/augeas/augeas-libs.p5m
```diff
@@ -0,0 +1,49 @@
...
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# NOTE: This is an already-available source of "charset.alias" with
+# same content (modulo the comment of what packages use the file).
+depend fmri=pkg:/library/glib2 type=require
```
pyhalov: Perhaps, depend on path?

pyhalov:
Perhaps, move it to library subdirectory and don't split on several packages?

jimklimov:
> Perhaps, move it to library subdirectory and don't split on several packages?

Not quite... there is the core library that does all the heavy work, so for actual use (via CLI, API or binding packages) it needs to be installed along with the lenses. For just building and linking, neither the lenses nor the CLI tools are required, so they are packaged separately.
Finally, ViM support to edit the lenses (IIUC) is an odd bird here, certainly not needed for everybody :)

Also note that so far I had no need to, but someone else can, add recipes for bindings - around 10 languages are suggested on the project's website.

What might make sense, is to reassign component packages to different locations in pkg_fmri taxonomy tree, which I did not quite figure out (what fits where) - so just went with base package name (under library/augeas) and a meaningful suffix. I made a best-guess effort for `info.classification` attributes though.

components/augeas/augeas-lenses.p5m
```diff
@@ -0,0 +1,427 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+### In source dir do this:
+# find lenses -type f |  sed 's,^\(.*\)$,file \1 path=usr/share/augeas/\1,' >> ../augeas-lenses.p5m
+
+file lenses/ssh.aug path=usr/share/augeas/lenses/ssh.aug
```
pyhalov: Perhaps, add $(SOURCE_DIR)/lenses to PKG_PROTO_DIRS and use transform to deliver just path actions? (see apr-util component for example) or some POST_INSTALL action to install them to PROTO area?

components/augeas/Makefile
```diff
@@ -0,0 +1,53 @@
+#
```
pyhalov: Please, move component to some category  - http://wiki.openindiana.org/oi/Categories+layout

jimklimov:
Is it good like this?

pyhalov:
Isn't it replaced by https://github.com/OpenIndiana/oi-userland/pull/1819 ?

jimklimov:
#1819 is based on this changeset, with the second commit delivering the charset files as a separate package split off the same glibc2 sources (so augtools dependency is not on whole glibc2, but on a small package of two short files - and glibc2 depends on that as well).

pyhalov:
Superseded by #1819

PR #1819, opened 2016-03-30
Augtools charset.alias coexistense with glib2

jimklimov:
Improvement over https://github.com/OpenIndiana/oi-userland/pull/1818 to allow either glib2 or augeas-libs to deliver the charset.alias files (glib2 should win in case of conflict).

Addon question: I see that glib2 delivers `*.la` files for libtool... is it a legacy artifact of this particular package (e.g. ok to remove) or do we generally allow these to be delivered after all?


pyhalov:
I don't like to have 2 sources for these files. If you don't like pulling in glib2 as dependency, split it in another package. 
As for *.la files, they are removed by oi-userland's transforms/libtool-drop

jimklimov:
Does this reorganization address your concerns?

components/library/glib/glib2-charset-aliases.p5m
```diff
@@ -0,0 +1,25 @@
+#
```
pyhalov: Rename file to be consistent with package name. Otherwise LGTM

components/sysutils/augeas/Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+#
+# Copyright 2016 Jim Klimov
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=	augeas
+COMPONENT_VERSION=	1.4.0
+COMPONENT_REVISION=	1
```
pyhalov, line 20: Remove COMPONENT_REVISION

pyhalov:
Merge commits or at least fix description of the second commit

components/sysutils/augeas/Makefile
jimklimov, line 20: I fixed it up and remembered now: the revision was intended for the case if you merge #1818 and #1819 separately - the augtools package metadata is changed then.
If you merge only #1819 as the initial delivery of augtools (and just close #1818), the revision tag is indeed not needed.

PR #1823, opened 2016-04-01
Update gobject-introspection to 1.43.3

mtelka:


pyhalov:
Please, move it to some appropriate category (http://wiki.openindiana.org/oi/Categories+layout), perhaps, library/

mtelka:
Moved.

pyhalov:
Please, merge commits.

mtelka:
Why? I prefer to see both changes separated.

pyhalov:
I'd say moving to subcategory is a part of touching component, so shouldn't be separated. I made separate commit with just move step only when basic work on updating wasn't mine.

mtelka:
Move and update are two separate and not related tasks.  I prefer to see two separate commits for that.  If you do not want two commits in one PR, then I'll split this PR to two PRs.

PR #1843, opened 2016-04-05
Zeromq-3.2.5 recipe updated to match PRs for other ZeroMQ versions

jimklimov:
This impacts the "real" package version (zeromq-3.2) and the directory for file deliveries (/usr/zeromq-3.2); the reference-package "zeromq" still remains and should pull the right thing (the highest "zeromq-X.Y" available in repo).

### DIFF
--- a/components/library/zeromq/zeromq3/Makefile
+++ b/components/library/zeromq/zeromq3/Makefile
@@ -20,7 +20,12 @@ COMPONENT_NAME= zeromq
 
 # See configure.ac for ABI version mapping to software version
 COMPONENT_VERSION_ABI= 3
-COMPONENT_VERSION= 3.2.5
+COMPONENT_VERSION_MAJ= 3
+COMPONENT_VERSION_MIN= 2
+COMPONENT_VERSION_PAT= 5
+COMPONENT_VERSION_MAJ_MIN= $(COMPONENT_VERSION_MAJ).$(COMPONENT_VERSION_MIN)
+COMPONENT_VERSION= $(COMPONENT_VERSION_MAJ_MIN).$(COMPONENT_VERSION_PAT)
+
 COMPONENT_SUMMARY= Open source message queue optimised for performance
 COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.gz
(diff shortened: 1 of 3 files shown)


pyhalov:
this PR seems to have one more commit than necessary

components/library/zeromq/zeromq3/Makefile
```diff
@@ -34,31 +39,63 @@ COMPONENT_LICENSE_FILE= COPYING.LESSER
...
 
-PKG_OPTIONS +=	-D COMPONENT_VERSION_MAJ="$(COMPONENT_VERSION_MAJ)"
+COMPONENT_PRE_CONFIGURE_ACTION =	($(CLONEY) $(SOURCE_DIR) $(@D))
 
-COMPONENT_PRE_CONFIGURE_ACTION =	($(CLONEY) $(SOURCE_DIR) $(@D) && cd $(@D) && autoconf )
+# Build docs just once (for one ARCH), parameter varies from version to version
+CONFIGURE_OPTIONS.32 +=	--with-documentation --with-docs
+CONFIGURE_OPTIONS.64 +=	--without-documentation --without-docs
 
-CONFIGURE_OPTIONS +=	--with-documentation
```
pyhalov, line 53: first CONFIGURE_OPTIONS should be = or you set prefix and other default directories twice

pyhalov:
Also it seems you pass some CONFIGURE_OPTIONS twice

jimklimov:
Extra commit - artefact of merge between forced branches between multiple workspaces :(
IMHO the rule on "one merge per PR" should be relaxed for some cases, anyhow you can roll back a whole PR...

The config options are once as I see (different for 32/64 bit, and many explicit path settings due to requested non-"/usr" rooting of versioned zeromq components, like in other PRs).

pyhalov:
I mean you set CONFIGURE_OPTIONS +=  and use CONFIGURE_OPTIONS +=   --prefix="$(COMPONENT_USR_DIR)" . This means you'll get --prefix=/usr (from default options)... --prefix=/usr/zeromq-3.2

jimklimov:
Uhuh... good catch then :-) thanks

## 

05.04.2016 21:16 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
In the end, this un-complicated much of the Makefile: for most things, it suffices to set up just the `CONFIGURE_PREFIX`, and it is expanded to correct values as the variables like `CONFIGURE_BINDIR.32` are actually used in the code (even if defined earlier in time via inclusion of `configure.mk`).

pyhalov:
merge commits

jimklimov:
done

PR #1846, opened 2016-04-06
Configure.mk and common Makefile structure revised

jimklimov:
Here's a big one! ;)

Since verification of this PR would likely need to recompile most of this gate, I added a number of separate changesets that rehaul hundreds of our Makefiles. In the theoretical happy case, if this PR builds successfully, this validates three relatively independent improvements.
- As a separate commit, many Makefiles got trailing-whitespace cleanup (removed spaces and tabs that either comprised the whole line contents, or were following some character other than backslash or equality sign or tab or space)
- Another fix aims to replace all `../(../)*/make-rules/` (except initial shared-macros.mk) and `$(WS_TOP)/make-rules/` references with `$(WS_MAKE_RULES)` so all include-lines now adhere to the same standard, and moving components to other subdirs will be more simple later on.
- Configure.mk extended to pass more options into the "configure" script, covering nearly all standard directory configurators (sysconfdir, localstatedir and libexec being the main glaring omissions that many of our recipes had to define explicitly).
  *\* All components whose Makefile's contained `--.*dir=` were revised to use newly defined variables rather than `CONFIGURE_OPTIONS+=...` for custom values, and/or remove the lines completely for newly-defined standard values (e.g. dozens of `CONFIGURE_OPTIONS+=--sysconfdir=/etc` and `CONFIGURE_OPTIONS+=--localstatedir=/var` and even typical MACH-dependent libexecdir's are no longer needed).
  *\* Care was taken to not break components that re-defined their `CONFIGURE_OPTIONS` from scratch... but this is what PR's are for - work was tedious, errors were possible ;)
  *\* A few components were revised more in-depth (xscreensaver, tigervnc... maybe that's all) to structure their ARCH-dependent builds in a way that seems more correct, which might be moot since they anyway only build 32-bit at the moment... I hope this was a future-proof investment for the case when we want to try them 64-bit.

I tried to build a dozen random components that seemed to have non-trivial configurations and I was unsure if my changes were safe - all got built. But my laptop is not a machine capable of testing this all exhaustively.

PS: You can take this opportunity to populate the `ccache` directory on the build server, now that its integration is merged... make sure to set aside a large-enough cachedir setting (5GB by default) ;)


pyhalov:
At whole this looks sane. But... I know, this is cruel, but I would like to hear success story about "gmake -C /path/to/oi-userland/components clean publish" before integrating changes of such scope :)

jimklimov:
Um... that might take a few months on a laptop in terms of CPU, and I'm not sure it has enough space for "make world"...

And given that at least "libmtp" does not build for me per earlier random tests, I can say in advance that my build results will not authoritatively move this PR forward ;)

Is it possible to organize a trial rebuild of this scale in a checkout of this branch on the Hipster farm?

PS: I've reorganized the patches for easier inspectability, separating whitespace fixes into their own changeset in particular.

pyhalov:
We can try to make trial build when new infrastructure is deployed. @meths  says new hardware will be available in a week or so, but I think @xen0l will need some time to configure it. Perhaps, we could arrange rebuild in a month.

jimklimov:
Sounds cool ;) Just don'tnprotract it too long - or we'd have too many
packages added or revised in the interim that would have to be also changed
to conform to this pattern and not redefine same configure flags twice in
their old-pattern recipes.

## 

07.04.2016 8:40 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
I'll leave breadcrumbs here to help re-sync this branch to codebase du-jour later:

```
git checkout -f oi/hipster && git pull --all && git merge upstream/oi/hipster && git push

# Apply current upstream code into this branch
git checkout -f configure-mk

# Either of these can fail due to source-code incompatibilities to be resolved manually
git merge upstream/oi/hipster
git rebase -i upstream/oi/hipster

# Resolve conflicts with updated and re-based code (if any) manually
git add `git status -s | egrep '^( M|UU) ' | sed 's,^\( M\|UU\) ,,'`
git commit -m 'Resolve merge conflicts from upstream'
git rebase --continue

cd $(WS_TOP)/components

### Fix up make-rules references, in some component or in the root of components tree
find . -type f -name Makefile -exec sed -e 's,^include \(\.\./\..*\|\$(WS_TOP)\)/make-rules/\([^s][^h]\|setup\.py\.mk\|shared-targets\.mk\),include \$(WS_MAKE_RULES)/\2,' -i '{}' \; ; git diff Makefile

# The same effect while updating a Makefile in Midnight Commander editor
# (e.g. outside this PR) can be achieved with regex search-and-replace:
#S# include (\.\./\..*|\$\(WS_TOP\))/make-rules/([^s][^h]|setup\.py\.mk|shared-targets\.mk)
#R# include $(WS_MAKE_RULES)/\2

# Add and commit all tracked modified files
git add `git status -s | egrep '^( M|UU) ' | sed 's,^\( M\|UU\) ,,'`

git commit -m 'Numerous components/*/Makefile unified to take advantage of WS_MAKE_RULES, update '"`date`"

### Optionally flush this change into the original changeset, or keep it separate for easier review
git rebase -i upstream/oi/hipster

# To inspect changes for typical errors and find omissions I used:

### Files that DO NOT contain "shared-macros" but DO contain WS_MAKE_RULES
### If there are any hits - we rewrote something we should not have (or part of a 
### more complex web of makefiles, theoretically)
find . -type f -name Makefile | grep -v /build/ | while read M ; do grep -q shared-macros "$M" || { grep -q "WS_MAKE_RULES" "$M" && echo "$M"; } ; done

### Files that DO contain "setup.py.mk" or "shared-target.mk" but
### DO NOT have them sourced from WS_MAKE_RULES
### If there are any hits - we did not rewrite something we should have
find . -type f -name Makefile | grep -v /build/ | while read M ; do grep -q setup.py.mk "$M" && { egrep -q 'WS_MAKE_RULES.*setup\.py\.mk' "$M" || echo "$M"; } ; done
find . -type f -name Makefile | grep -v /build/ | while read M ; do grep -q shared-target.mk "$M" && { egrep -q 'WS_MAKE_RULES.*shared-target\.mk' "$M" || echo "$M"; } ; done


### Similarly for whitespaces
find . -type f -name Makefile -exec sed -e 's,^\(\|.*[^\\= \t]\)[ \t][ \t]*$,\1,' -i '{}' \; 
git add `git status -s | egrep '^( M|UU) ' | sed 's,^\( M\|UU\) ,,'`
git commit -m 'Numerous components/*/Makefile cleaned up of trailing whitespaces, update '"`date`"
git rebase -i upstream/oi/hipster

### For changes related to configure.mk I searched with Midnight Commander for regex hits:
#Filenames# Makefile
#Content# \-\-(info|man|doc|include|lib|libexec|s?bin|sysconf|localstate|sharedstate)dir=
```

jimklimov:
PR refreshed to solve merge conflicts and process recently added recipes.

jimklimov:
...and how about proceeding with this? :-)

It takes over an hour to resync changes, especially with the frenzy on REQUIRED_PACKAGES colliding with some changes done earlier in this PR...

jimklimov:
Changes pioneered by this PR were re-submitted as smaller PRs referenced above (and done against today's codebase). Roughly one thing not covered by those newer PRs is the manual revision of Makefiles done in commit https://github.com/OpenIndiana/oi-userland/pull/1846/commits/6060747083fdefe6547cd085deff403945f64ba8

pyhalov:
This one stayed open for too long. Sorry, Jim, I don't think we want to do so drastic changes at once. All components are getting updated as they are touched. Perhaps, some parts of configure.mk logic could be integrated as separate PR.

jimklimov:
Make sense, thus parts were resubmitted as separate smaller PRs. In October.

PR #1847, opened 2016-04-06
Bump PyZMQ to 14.7.0 (requires zeromq-4.1.2+)

jimklimov:
Will need https://github.com/OpenIndiana/oi-userland/pull/1809 to build

### DIFF
--- a/components/python/pyzmq/Makefile
+++ b/components/python/pyzmq/Makefile
@@ -10,15 +10,32 @@
 
 #
 # Copyright (c) 2014 Alexander Pyhalov.  All rights reserved.
+# Copyright 2016 Jim Klimov
 #
 
+# See zeromq configure.ac for ABI version mapping to software version
+# NOTE: This requires functionality offered by libzmq-4.1.2 (and hopefully
+# subsequent releases) which we package in a dedicated subtree
+NEEDED_ZEROMQ_VER = 4.1
+
+# The zeromq components define this directory structure
+ZEROMQ_DIR = $(USRDIR)/zeromq-$(NEEDED_ZEROMQ_VER)
+
+# We define these before including any common structure makefiles
+CXXFLAGS   = -I$(ZEROMQ_DIR)/include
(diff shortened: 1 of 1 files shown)


components/python/pyzmq/Makefile
```diff
@@ -10,15 +10,32 @@
 
 #
 # Copyright (c) 2014 Alexander Pyhalov.  All rights reserved.
+# Copyright 2016 Jim Klimov
 #
 
+# See zeromq configure.ac for ABI version mapping to software version
```
pyhalov, line 16: Please, move it below metadata definition, somewhere between "include $(WS_TOP)/make-rules/ips.mk" and  "# common targets".

jimklimov, line 16: As said in the comment and code a few lines lower, this is needed to set up `C*FLAGS` to include and link the desired zeromq version as you asked, if we have the choice of several versions. For that version to have priority over others, these values must be set before we include the oi-userland makefiles and add common include/lib-dirs.

At least this is the intention, which seemed to work for my local build-testing. I am not sure this is the really correct approach to this problem, so am open to suggestions.

pyhalov, line 16: Of course, you can add LDFLAGS/CFLAGS after include $(WS_MAKE_RULES)/ips.mk.  Perhaps, you have to do LDFLAGS += $(LDFLAGS.$(BITS))

jimklimov, line 16: But then it is added at the end of the line. The idea (maybe wrong) is to ensure that the `*FLAGS` lines start with references to the needed versions, in case there are other seemingly matching variants later on (e.g. we presumably need specifically zeromq-4.1 and there is 4.2.*-git in /usr/lib and /usr/include)

pyhalov, line 16: common makefiles  don't add much to CFLAGS/LDFLAGS besides -m$(BITS), you can check

xen0l:

PR #1848, opened 2016-04-07
Moved gdlib into library/ and added publishing of 64-bit gdlib-config

jimklimov:
I wonder if we should also publish MACH64 versions of other binaries (4Gb+ images to process, anyone?) It is a matter of some 200Kb for a total of a dozen binfiles ;)

But at least gdlib-config we do need to help other programs link in any BITS'ness they need.


pyhalov:
I'd say deliver both versions and perhaps use isaexec for binaries intended for users

jimklimov:
Done

PR #1857, opened 2016-04-12
Add libcouchbase 2.5.7

avsej:


components/library/libcouchbase/patches/02-multiarch-pkg_config.patch
```diff
@@ -0,0 +1,12 @@
+--- libcouchbase-2.5.7/packaging/libcouchbase.pc.in.orig	2016-01-12 00:44:09.000000000 +0300
```
pyhalov: Can we use more elegant patch here, for example, refer CMAKE_INSTALL_BINDIR ?

components/library/libcouchbase/Makefile
```diff
@@ -0,0 +1,64 @@
+#
```
pyhalov: use illumos header

components/library/libcouchbase/resolve.deps
```diff
@@ -0,0 +1,3 @@
+library/libev
```
pyhalov: oi-userland doesn't use resolve.deps

components/library/libcouchbase/Makefile
```diff
@@ -0,0 +1,64 @@
...
+CMAKE_OPTIONS+=	-DLCB_NO_TESTS=1
+
+COMPONENT_POST_BUILD_ACTION += \
+	case $(BITS) in \
+	32) $(GSED) -i -e "s:/MACH64::g"         $(BUILD_DIR_32)/packaging/libcouchbase.pc;; \
+	64) $(GSED) -i -e "s:MACH64:$(MACH64):g" $(BUILD_DIR_64)/packaging/libcouchbase.pc;; \
+	esac;
+
+# common targets
+build:		$(BUILD_32_and_64)
```
pyhalov, line 40: Have you compared 32-bit and 64-bit headers? Are they the same?

avsej, line 40: yes, they are identical

avsej:
@pyhalov thanks for pointing, patch has been updated

PR #1866, opened 2016-04-14
Workflow

meths:
Workflow improvements as trying to work with this repo just seems ugly when used to ec-userland.
Keeping git status clean helps dev sanity.


.gitignore
```diff
@@ -1,11 +1,13 @@
+/archives
 /i386
 /sparc
 
-components/components.mk
-components/depends.mk
+components/*.mk
+components/**/build
```
pyhalov, line 6: why should we have have both of this?

make-rules/shared-macros.mk
```diff
@@ -37,15 +37,16 @@ PATH=/usr/bin:/usr/gnu/bin
...
+#export DOWNLOAD_SEARCH_PATH +=	$(INTERNAL_ARCHIVE_MIRROR)
+#export DOWNLOAD_SEARCH_PATH +=	$(EXTERNAL_ARCHIVE_MIRROR)
 
 # The workspace starts at the mercurial root
 ifeq ($(origin WS_TOP), undefined)
 export WS_TOP := \
 	$(shell hg root 2>/dev/null || git rev-parse --show-toplevel)
 endif
 
+USERLAND_ARCHIVES =     $(WS_TOP)/archives/
```
pyhalov, line 49: Is it overwritten by environment?

pyhalov, line 49: I've checked, no it isn't. Should use '?='

pyhalov:
and please, don't merge so much "merged into.." commits in oi-userland :)

.gitignore
meths, line 6: They're both autogenerated content that shouldn't be committed to git and filling git status with them doesn't help anyone who keeps a fairly mature tree around.

make-rules/shared-macros.mk
meths, line 49: Will update.

meths:
Yeah, think I'll just trash this repo and make a clean start so OI/oi-ul doesn't get polluted.

.gitignore
pyhalov, line 6: Comment was wrongly placed, I mean "components/**/build" and "components/meta-packages/history/build".

PR #1867, opened 2016-04-14
Update mbuffer to 20160228

xen0l:


pyhalov:
please, move it to some category, perhaps, shell

xen0l:
Moved to sysutils as mbuffer is more system-related than shell.

PR #1868, opened 2016-04-14
Openssh

pyhalov:


components/meta-packages/auto-install/auto_install.p5m
```diff
@@ -184,6 +184,7 @@ depend type=require fmri=network/ftp@0.5.11-0.151.1.8
 depend type=require fmri=network/ipfilter@0.5.11-0.151.1.8
 depend type=require fmri=network/iscsi/initiator@0.5.11-0.151.1.8
 depend type=require fmri=network/iscsi/iser@0.5.11-0.151.1.8
+depend type=require fmri=network/openssh@7.2.0.2-2015.0.2.0
```
mtelka, line 187: Why there are the version numbers required in this file?

pyhalov, line 187: Honestly, they are not. But this can help you to spot an issue when , for example, A@N can be installed with B@M side by side, and A@N+1 can not.

mtelka, line 187: Hm, I do not understand how the version numbers listed here will help with your scenario (probably lack of pkg knowledge at my side).

Anyway, almost every version number in this file is already obsolete...

PR #1870, opened 2016-04-14
Workflow

meths:
A clean version of the previous PR to cleanup git status to hopefully help dev workflow.

Note: May require intervention on the build server so it doesn't simply break stuff that no longer has a working ARCHIVE_URL.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -46,6 +46,7 @@ export WS_TOP := \
 	$(shell hg root 2>/dev/null || git rev-parse --show-toplevel)
 endif
 
+USERLAND_ARCHIVES ?=	$(WS_TOP)/archives/
 WS_MACH =       $(WS_TOP)/$(MACH)
 WS_LOGS =       $(WS_MACH)/logs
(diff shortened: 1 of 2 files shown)


make-rules/shared-macros.mk
```diff
@@ -37,15 +37,16 @@ PATH=/usr/bin:/usr/gnu/bin
 
 # Default to looking for source archives on the internal mirror and the external
 # mirror before we hammer on the community source archive repositories.
-export DOWNLOAD_SEARCH_PATH +=	$(INTERNAL_ARCHIVE_MIRROR)
-export DOWNLOAD_SEARCH_PATH +=	$(EXTERNAL_ARCHIVE_MIRROR)
+#export DOWNLOAD_SEARCH_PATH +=	$(INTERNAL_ARCHIVE_MIRROR)
+#export DOWNLOAD_SEARCH_PATH +=	$(EXTERNAL_ARCHIVE_MIRROR)
 
 # The workspace starts at the mercurial root
```
avsej, line 43: it is not mercurial anymore, is it?

PR #1895, opened 2016-04-23
Update GNU grep to 2.25

mtelka:


alarcher:
Possible issue ? https://www.illumos.org/issues/6759

XFAIL: equiv-classes
FAIL: multibyte-white-space
XFAIL: triple-backref

PR #1903, opened 2016-04-26
Add libetpan

avsej:
This library is dependency of claws-mail, which I'm going to add next


components/library/libetpan/libetpan.p5m
```diff
@@ -0,0 +1,198 @@
...
+file path=usr/include/libetpan/uidplus_types.h
+file path=usr/include/libetpan/xgmlabels.h
+file path=usr/include/libetpan/xgmmsgid.h
+file path=usr/include/libetpan/xgmthrid.h
+file path=usr/include/libetpan/xlist.h
+file path=usr/lib/$(MACH64)/libetpan.a
+link path=usr/lib/$(MACH64)/libetpan.so target=libetpan.so.17.2.0
+link path=usr/lib/$(MACH64)/libetpan.so.17 target=libetpan.so.17.2.0
+file path=usr/lib/$(MACH64)/libetpan.so.17.2.0
+file path=usr/lib/libetpan.a
```
pyhalov: Do we really need static libs? We usually deliver them only when there's no alternative dynamic library.

avsej: I will remove it

PR #1907, opened 2016-04-27
 emacs: update to 24.5

avsej:


AndWac:
File Changes

```
A components/emacs/.gitignore (1)
```

Are you sure about adding .gitignore?

avsej:
@AndWac I did it previously it helps to keep tree clean and local to this component only

components/emacs/Makefile
```diff
@@ -23,18 +23,18 @@
 include ../../make-rules/shared-macros.mk
```
pyhalov: It would be good also to move emacs to appropriate category

avsej: @pyhalov like what?

avsej: vim and nano on the same level

pyhalov: Either use text/ or create editor/ for them. Don't know.

avsej: moved to editor/

PR #1909, opened 2016-04-27
Add subdir macros

alarcher:
Similar to the macros defined in the x11 gate, to be reused for consistency.


pyhalov:
Where is it going to be used?

alarcher:
This is a macro used in x11 components but also may avoid redefining them in other components.

pyhalov:
I'm asking because of inconsistency between LIB/BIN macroses. So I'm interested in real use cases.

alarcher:
It is basically always the case except for very few uses of isaexec.

pyhalov:
Perhaps, I miss something.
BINSUBDIR32 = $(MACH32) will make BINSUBDIR32=i86 , which we usually don't use except for very few uses of isaexec, so why it's not empty?

alarcher:
Let me check I am on the phone and I replied maybe too fast.

alarcher:
OK I can drop the BINSUBDIR but I'd like to keep LIBSUBDIR which is also named as ARCHLIBSUBDIR for example in xf86-video-*.
I can use ARCHLIBSUBDIR and modify my x11 components to match the existing one.

pyhalov:
Sounds good

PR #1912, opened 2016-04-27
Add Lua 5.3 to oi-userland

xen0l:


pyhalov:
It can't coexist with lua-5.2... We should do something with this. 
1) Is liblua.so compatible? 
2) Does oi-userland lua consumers need patches to work with new lua?

alarcher:
Also VLC depends on some Lua backward compatibility functions, possible breakage should be checked.

xen0l:
Lua 5.3 was packaged as a preparation for bringing in haproxy 1.6, which needs lua 5.3. However, I am afraid other consumers might have issues with newer versions. Besides VLC, what are major lua users in the gate?

pyhalov:
I see 
pkg:/desktop/window-manager/notion-3
pkg:/diagnostic/nmap
pkg:/diagnostic/wireshark/wireshark-common
pkg:/image/gnuplot
pkg:/image/graphviz
pkg:/image/graphviz/graphviz-lua
pkg:/image/library/gegl/workshop
pkg:/image/rrdtool
pkg:/library/e/efl
pkg:/media/vlc
pkg:/web/server/apache-24

pyhalov:
Also affected web/server/apache-22/module/apache-security (doesn't build with 5.3)

components/runtime/lua-53/patches/Makefile.patch
```diff
@@ -0,0 +1,64 @@
...
++INSTALL_EXEC= $(INSTALL) -m 0555
++INSTALL_DATA= $(INSTALL) -m 0444
+ #
+ # If you don't have "install" you can use "cp" instead.
+ # INSTALL= cp -p
+@@ -40,9 +41,10 @@
+ 
+ # What to install.
+ TO_BIN= lua luac
+-TO_INC= lua.h luaconf.h lualib.h lauxlib.h lua.hpp
```
alarcher: Could you keep lua.hpp ?

xen0l: Yes, I can.

xen0l:
I modified this component, so it can be installed next to lua 5.2. In the future, this might be the best approach, as it will allows us to ship multiple lua versions. I based this work on ubuntu's approach. Please review.

components/runtime/lua-53/files/lua-64.pc
```diff
@@ -0,0 +1,11 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${exec_prefix}/lib/MACH64
+includedir=${prefix}/include
```
pyhalov: shouldn't it be ${prefix}/include/lua5.3 ?

xen0l: yes, thanks. Fixed.

PR #1916, opened 2016-04-27
Add xorg-docs 1.7.1 from x-S12

alarcher:


pyhalov:
X-incorporation should be relaxed

alarcher:
Removed xorg-docs from X-incorporation.

PR #1918, opened 2016-04-27
Initial commit of SDL2_mixer version 2.0.1

jimklimov:
NOTE: For MOD file support, PR#1921 should be merged first. Note that the build is not failed otherwise, just the support will be missing, despite the "--enable" flags for configure.

### DIFF
--- a/components/encumbered/SDL2_mixer/Makefile
+++ b/components/encumbered/SDL2_mixer/Makefile
@@ -1,0 +1,78 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2016 Jim Klimov
+#
+
+# NOTE: This recipe supports both "encumbered" and "license-clean"
+# builds (with or without MP3 support). See Makefile.not-encumbered
+# which includes this one as an example that could be delivered as
+# $(WS_TOP)/library/SDL2_mixer/Makefile in some other reality.
+# That file's contents would be just two lines like these:
+### USE_ENCUMBERED=no
+### include ../../encumbered/SDL2_mixer/Makefile
+
+# This file (by default) provides the package with encumbered code
+# or inseparable run-time dependencies:
+USE_ENCUMBERED?=yes
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		SDL2_mixer
+COMPONENT_VERSION=	2.0.1
+COMPONENT_PROJECT_URL=	http://www.libsdl.org/projects/SDL_mixer/
+COMPONENT_SUMMARY=	SDL_mixer is a sample multi-channel audio mixer library
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:5a24f62a610249d744cbd8d28ee399d8905db7222bf3bdbc8a8b4a76e597695f
+COMPONENT_ARCHIVE_URL=	http://www.libsdl.org/projects/SDL_mixer/release/$(COMPONENT_ARCHIVE)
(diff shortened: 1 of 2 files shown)


components/library/SDL2_mixer/Makefile
```diff
@@ -0,0 +1,58 @@
...
+CONFIGURE_OPTIONS +=	--enable-music-cmd
+CONFIGURE_OPTIONS +=	--enable-music-wave
+CONFIGURE_OPTIONS +=	--enable-music-midi
+CONFIGURE_OPTIONS +=	--enable-music-mod-mikmod --enable-music-mod-mikmod-shared
+CONFIGURE_OPTIONS +=	--enable-music-ogg-shared
+CONFIGURE_OPTIONS +=	--enable-music-flac-shared
+
+### Note: currently we do not serve libmad, and MPEG support is reserved for
+### encumbered components
+#CONFIGURE_OPTIONS +=	--enable-music-mp3-smpeg --enable-music-mp3-smpeg-shared
```
pyhalov: Perhaps, you should explicitly disable them, as they are enabled by default if found

jimklimov: fixed better in a recent commit

2016-04-28 0:54 GMT+02:00 Alexander Pyhalov notifications@github.com:

pyhalov:
don't do it...  You deliver the same library/audio/sdl2-mixer with different capabilities both in encumbered and in basic repository. Build server will have only one installed. This sounds like a good way to confuse everyone. Where do you want to use this library?

jimklimov:
It is an open-ended question :) I saw it as a dependency for a number of games that seem to be portable. They sort of do not specify what media they would have, and in general if we want to expand our ecosystem, we should allow building and running the widest possible selection of options, because we can not know in advance what some next game to integrate will need (e.g. might have an in-game playlist like the GTA car-radio that plays whatever you give it, regardless of what media types the game distro provides).

Also there is an evil mesh of libraries around "helping to make games" that have all sorts of optional features, duplicate features, different interfaces to same backends, even two-way dependencies, etc. so there is a whole ecosystem to add as a blob more or less.

My recent burst started with bumping OpenTTD - I saw that our recipe has one such helper library disabled that we can in fact deliver, and that it in turn can do more formats and functions if it is compiled with more third-party libraries (about half a dozen further dependencies listed in the README, and I have yet to see if they are suggestions or requirements), some of which we chose to segregate into encumbered. Alas, not many of them are truly modular, with dynamic links evaluated and fulfilled at run-time on a need-to-use basis (e.g. no MP3 media in a game - no MP3 library needed for this execution, but can do it for another exec with another game) - generally these functionalities are predefined dynamic links :(

I believe that actually for a build server this might not matter which variant is installed - as long as the library is there and the common API is exposed and compilation can be fulfilled (assuming that e.g. routines to work with MP3 for example are still defined, but empty in code-content). Might deal with facets, mediators and other such complications later on - but I guess it would be even less understandable and transparent for end-users.

pyhalov:
Perhaps, this have to be discussed on oi-dev ML. I'd put all such components in encumbered, including games.

jimklimov:
That could be one approach... would this then include games that are not
really encumbered, as well? Just because the frameworks they use are only
built with as many options as they can grab, including encumbered
third-party libs?

2016-04-28 9:40 GMT+02:00 Alexander Pyhalov notifications@github.com:

pyhalov:
it will

jimklimov:
Relocated the makefile into encumbered. Dual-nature structure and sample Makefile.not-encumbered kept to make it possible to deliver both versions, should we later choose to do so.

pyhalov:
please, remove unused Makefile.not-encumbered. I think having USE_ENCUMBERED variable is enough.

jimklimov:
Is this a strict requirement? I'd rather keep it here as an unmistakable
example of how to use this.

## 

04.05.2016 13:28 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov:
I think it's not needed in OI repository. You can add comment to USE_ENCUMBERED variable

alarcher:
I'd rather document it within oi-userland, possibly by giving an example in the templates directory if you really want to make it visible.

jimklimov:
Ok, file removed, comment updated.

alarcher:
Squashy squashy ?

jimklimov:
If LGTY ;) squashing messes up my history and switching workspaces, so i
prefer to delay it until a merge is likely ;)

## 

04.05.2016 18:31 пользователь "Aurélien Larcher" notifications@github.com
написал:

jimklimov:
squishy squashy!

PR #1919, opened 2016-04-27
Initial commit of libmad-0.15.1b

jimklimov:
Let us play some music!

### DIFF
--- a/components/library/libmad/Makefile
+++ b/components/library/libmad/Makefile
@@ -1,0 +1,52 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2015 Alexander Pyhalov (libid3tag recipe)
+# Copyright 2016 Jim Klimov
+#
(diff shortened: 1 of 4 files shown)


alarcher:
What is the difference with the one in encumbered?

pyhalov:
Don't see any differences, so closing this.

jimklimov:
Good question... this one I did not find ;)
Otherwise, the new recipe is more simple structurally and sets/unsets
certain options explicitly.

BTW, what makes it encumbered - dealing with MP3? For example, libid3tag
from the same repo is considered open...

On a more practical note, builds of SDL_mixer and maybe SDL_sound can use
this library. However it is served in a different repository. Should there
be two recipes (in common components and under encumbered) with different
configure-flags (do or do not use libmad, per this example usecase)?

2016-04-28 0:54 GMT+02:00 Aurélien Larcher notifications@github.com:

pyhalov:
This depends on SDL_mixer usage. If you consider it useful for non-encumbered packages, it shouldn't depend on encumbered packages. In other case you should move it (and consumers) to encumbered. 
When we discussed it, we found that tag parsing doesn't deal with patented algorithms, but parsing audio streams do.

jimklimov:
The problem is - these are libraries... They are used to make games in
particular. If the real ultimate consumer (a game) has MP3 media, it will
require an ability to play it; if it has no media like that - this is not
required.

I guess we'll have to make both recipes available, so a user can
install/upgrade with "pkg://hipster-encumbered/library/mad" if needed...

2016-04-28 1:14 GMT+02:00 Alexander Pyhalov notifications@github.com:

PR #1920, opened 2016-04-28
Add a way to NOT remove "corrupt" downloads when they are known new

jimklimov:
Extend userland-fetch with a --keep option to not delete files, and add a "gmake fetch" action to just download and verify (not remove) files.

Save traffic and gain speed as a result, when initially preparing a userland recipe for an archive with yet-unknown checksum.

### DIFF
--- a/README
+++ b/README
@@ -51,6 +51,21 @@ Getting the Bits
     directory and use 'gmake download' from that to only get it's source
     archive.
 
+    Also, when you start to work with a new archive file - update the source
+    version in an existing recipe component, or start a new one from scratch -
+    you can use 'gmake fetch' to download the archive(s) defined in the new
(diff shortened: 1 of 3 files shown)


pyhalov:
Now gmake fetch exits with error on checksum mismatch. Is it an intended behavior? Or should only gmake prep exit with error?

jimklimov:
Yes, I'd say this is intended - a mismatch is there after all. We just do not force removing the "corrupt" file.

There is also a side-effect: by recipe, a file in the download location only depends on the recipe Makefile. So once it is "fetched" (downloaded and not removed), it will not be re-verified - the script is just not called. I'd say this is a moderate problem, since the newly added "fetch" ability is a helper for recipe-makers doing initial archive downloads.

PR #1923, opened 2016-04-28
Add redshift

avsej:


components/x11/redshift/redshift.p5m
```diff
@@ -0,0 +1,77 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/redshift
+file path=usr/bin/$(MACH64)/redshift-gtk
+file path=usr/bin/redshift
+file path=usr/bin/redshift-gtk
+file path=usr/lib/python3.4/site-packages/redshift_gtk/__init__.py
+file path=usr/lib/python3.4/site-packages/redshift_gtk/__pycache__/__init__.cpython-34.pyc
```
pyhalov: Please. move python files to vendor-packages. Also _/**pycache**/_ files shouldn't be mentioned, transforms/autopyc should take care of this.

components/x11/redshift/Makefile
```diff
@@ -0,0 +1,51 @@
...
+    sha256:10e350f93951c0521dd6f103d67a485972c307214f036e009acea2978fe4f359
+COMPONENT_ARCHIVE_URL=	https://github.com/jonls/redshift/releases/download/v$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_LICENSE=	GPL-3
+COMPONENT_LICENSE_FILE=	COPYING
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+PYTHON_VERSION	= 3.4
```
pyhalov, line 34: Why do we need Python 3.4? Does it work with 2.6 or 2.7?

avsej, line 34: redshift requires python 3.2+ https://github.com/jonls/redshift/blob/master/configure.ac#L81
It is completely usable without python (python GTK needed for tray applet)

PR #1934, opened 2016-05-02
Initial commit of LLNL Powerman (2.3.24) lost since SFW days (last v2.3)

jimklimov:
I know it is questionable regarding service integration (so I provide the original init-script in `/etc/poweman` data files)... the original OpenSolaris SFW packaging provided even less (see http://pkg.openindiana.org/hipster/manifest/0/system%2Fmanagement%2Fpowerman%402.3%2C5.11-2013.0.0.0%3A20151027T080243Z for legacy details) ;)
In reality, I currently need the libs for NUT to pass "make distcheck" under OpenIndiana.
So please merge fast if there are no fatal flaws, and leave refinement for other loops and hands ;)

### DIFF
--- a/components/sysutils/powerman/Makefile
+++ b/components/sysutils/powerman/Makefile
@@ -1,0 +1,69 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=	powerman
+COMPONENT_VERSION=	2.3.24
+COMPONENT_SUMMARY=	Powerman - power management software ($(COMPONENT_VERSION))
+COMPONENT_DESCRIPTION=	LLNL CHAOS powerman is free UNIX/Linux software that controls (remotely and in parallel) switched power distribution units
+COMPONENT_SRC=	$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+  sha256:d2f08ee1383d417ed14438443fefbad4954a3912de8e8e574a21089927eda2a0
+COMPONENT_ARCHIVE_URL=	\
+  https://github.com/chaos/$(COMPONENT_NAME)/archive/$(COMPONENT_VERSION).tar.gz
+COMPONENT_PROJECT_URL=	https://github.com/chaos/powerman
+COMPONENT_FMRI=	system/management/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Administration and Configuration
+COMPONENT_LICENSE=	GPLv2+
+COMPONENT_LICENSE_FILE=	COPYING
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+
+COMPONENT_PREP_ACTION =	( cd $(SOURCE_DIR) && ./autogen.sh )
+
+COMPONENT_PRE_CONFIGURE_ACTION =	( $(CLONEY) $(SOURCE_DIR) $(@D) )
+
+CONFIGURE_OPTIONS +=	--disable-static
+CONFIGURE_OPTIONS +=	--enable-shared
+
+CONFIGURE_OPTIONS +=	--sysconfdir=/etc
+CONFIGURE_OPTIONS +=	--with-httppower
+CONFIGURE_OPTIONS +=	--with-snmppower
+CONFIGURE_OPTIONS +=	--with-systemdsystemunitdir=no
+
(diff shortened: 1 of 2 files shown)


alarcher:
Are we going Debian-style with bins, libs, dev, docs, etc... splitting ?

jimklimov:
Not sure about full-way, but imho it makes sense when there are too many
objects and/or megabytes involved in a monolithic package which you only
need on a particular system for its library dependencies.

Another related split-reason for me in some other packages is when the
package delivers a number of related (but not all required) products, some
of which have their certain unique dependencies. My recipe for NUT is
nearing completion, and its drivers (some simple, some usb, and snmp,
netxml, powerman, ipmi) and say its cgi-bin clients all have several
different groups of dependency trees. Not everyone wants or needs to
install half of X11 for gdlib and stuff (used in cgi) just to monitor a
serial UPS - with the split packages they are not required to always pull
all those deps. And in my case, if I work on a NUT-API client project, I
don't even need any bins, drivers and services on this system - only the
libs and headers. Same for clamav with its libs, services, clients etc.
that are not required to even coexist on the same system (think a central
antivirus processing numbercrunching server and several weak clients like
web-proxies or email relays).

So my rule of thumb is that if the package is small and homogenous, it can
be a single package. If it provides nontrivial or separately useable libs,
CLI's and/or services, or separate considerable dependency weight for some
fringe bits - i'd split and provide a grouper to have an easy way to
install a reasonable set (though it is not always "everything there is,
built and packaged").

Hope this clarifies my take on this,

## Jim

03.05.2016 21:58 пользователь "Aurélien Larcher" notifications@github.com
написал:

alarcher:
Sure but maybe this should be discussed to enforce common guidelines ?

jimklimov:
Not a bad idea, feel free to replicate my post to email (likely i won't be
able to contribute there much tonight and tomorrow). I think now that there
was some other PR this or last week where it was suggested to discuss
further on the mailing list, and it sort of fell out of my memory, sorry :(

My recipe experiments of the past few months did start a few "global
policy" as well as technological pinpoint discussions, but these almost
never even took off or did not come to a definitive answer. So for now I
proceed where/how I can, to the best of my thinking and intention - to see
what comes out (and if it is acceptable to me and others)... :-/

## Jim

03.05.2016 22:19 пользователь "Aurélien Larcher" notifications@github.com
написал:

components/sysutils/powerman/Makefile
```diff
@@ -0,0 +1,83 @@
...
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CXXFLAGS.32=-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE
+CFLAGS.32=-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE
+
+CXXFLAGS+=$(CXXFLAGS.$(BITS))
+CFLAGS+=$(CFLAGS.$(BITS))
+
+LDFLAGS.32= -L$(USRLIBDIR) -R $(USRLIBDIR) $(LDFLAGS) -L$(GNULIB) -R $(GNULIB)
```
pyhalov: 1) Why do we need -L$(USRLIBDIR)  -R $(USRLIBDIR) ?
2) Are you sure flags are actually passed and used (I see spaces between -R and $(GNULIB) )

components/sysutils/powerman/Makefile
```diff
@@ -0,0 +1,83 @@
...
+
+# libgenders is another CHAOS project that we don't bother to package for now.
+# It is a static cluster configuration database used for cluster configuration
+# management. It is used by a variety of tools and scripts for management of
+# large clusters.
+CONFIGURE_OPTIONS +=	--without-genders
+
+CONFIGURE_OPTIONS +=	$(CONFIGURE_OPTIONS.$(BITS))
+
+# Temporary for tests
```
pyhalov: Remove it :)

components/sysutils/powerman/powerman-bins.p5m
```diff
@@ -0,0 +1,63 @@
+#
```
pyhalov: Do we really want to split it?

components/sysutils/powerman/powerman-bins.p5m
```diff
@@ -0,0 +1,63 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/powerman
+file path=usr/bin/powerman
+file path=usr/share/man/man1/powerman.1
+
+### Note: "pm" and "powerman" binaries and manpages are identical - I would
```
pyhalov: Does hardlink make sense here or is it of no use?

jimklimov:
Comments above all addressed, including collapse into one package

jimklimov:
Squashed the commits

components/sysutils/powerman/powerman.p5m
```diff
@@ -0,0 +1,136 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov
+#
+
+# Set version of the grouper same as core package
```
pyhalov: It seems comment should be removed

jimklimov:
Point taken, Makefile fixed, commits resquashed :)

jimklimov:
FWIW, here is a link to integration discussion in OpenSolaris (powerman v2.3, Nov 2008) https://www.mail-archive.com/opensolaris-arc@mail.opensolaris.org/msg01825.html

PR #1940, opened 2016-05-03
Include debug information for gcc compiler

avsej:


pyhalov:
Why do you need this? This makes executables large. Sometimes it makes them work incorrectly.

avsej:
Okay. I've got it

PR #1949, opened 2016-05-05
Enable Libiberty in gcc-4.8 and gcc-4.9

jimklimov:
Same gcc sources/versions, just slightly different flags and p5m's, and a bumped revision.


pyhalov:
both of them doesn't deliver 64-bit library

pyhalov:
Perhaps, we just need separate libberty component?

jimklimov:
There does not seem to be a good way to get just its sources - and fetching
whole gcc is kinda wasteful.

## 

05.05.2016 17:14 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov:
where do you need it?

pyhalov:
perhaps, having only 32-bit version is OK?

jimklimov:
At least one practical example is distcc that was desired by community and
required -liberty. This one case may be ok with just a 32-bit build (4gb+
sources are unlikely)? But the general stance was that libs should be
packaged in all bits?

Anyhiw, with default compiler we don't provide any version of the lib :(

## 

05.05.2016 17:21 пользователь "Alexander Pyhalov" notifications@github.com
написал:

perhaps, having only 32-bit version is OK?

jimklimov:
Just to help move this PR forward (and to do get some usable `libiberty.a`, even if just 32-bit, with our default compiler) I removed p5m actions for the 64-bit file which I can't readily get built. That could be a matter for another PR, maybe in another venue (e.g. binutils?)

So far I want to proceed with building some variant of `distcc` so this blocker has to go away :)

jimklimov:
ping?

components/developer/gcc49/Makefile
```diff
@@ -99,6 +99,7 @@ CONFIGURE_OPTIONS+= --without-gnu-ld
 CONFIGURE_OPTIONS+= --with-ld=/usr/bin/ld
```
pyhalov, line 100: Please, bump COMPONENT_REVISION and let's merge it.

jimklimov, line 100: Ok, yay! :-)
wasn't it bumped already? Or did someone else change the recipe while the
PR was posted?

## 

16.06.2016 15:06 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov, line 100: Yes, gcc 49 is already moved to developer and COMPONENT_REVISION was bumped.

jimklimov:
Comments addressed, bump :)

PR #1952, opened 2016-05-05
Add tklib 0.5 (Tk widgets standard library)

avsej:


pyhalov:
Note to myself: requires tcllib

avsej:
Yes, it does

PR #1966, opened 2016-05-08
Add name and notices

alarcher:
Bring back components from the release engineering consolidation:
- name is set to the latest snapshot,
- notices are copied from existing but the version in French is fixed because it had lost all the non-ASCII characters (accents, cedilla, ...),
- version and license are the same as original components,
- dates are updated.


pyhalov:
Can we generate date in /etc/release while publishing the package? Does it sound sane?

alarcher:
Sure.

alarcher:
Updated.

alarcher:
Commit added to retire consolidation.

PR #1967, opened 2016-05-08
Add hwloc 1.11.3

alarcher:


xen0l:
Component could be published, but I wasn't able to test it. The test suite ended with:

<pre>
============================================================================
Testsuite summary for hwloc 1.11.3
============================================================================
# TOTAL: 36
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  36
# XPASS: 0
# ERROR: 0
</pre>

There are unbound variables errors in every test on line 37. It seems to be a check with $libtool_install_magic, which is not defined:

<pre>
# This environment variable determines our operation mode.
if test "$libtool_install_magic" = "%%%MAGIC variable%%%"; then
  # install mode needs the following variables:
  generated_by_libtool_version='2.4.6'
  notinst_deplibs=' /export/home/xenol/code/oi-userland/components/developer/hwloc/build/i86/src/libhwloc.la'
else
  # When we are sourced in execute mode, $file and $ECHO are already set.
  if test "$libtool_execute_magic" != "%%%MAGIC variable%%%"; then
    file="$0"
</pre>

Were you able run the test suite?

xen0l:
Test suite passes, so this seems to be oi-userland related.

pyhalov:
ENV=/usr/bin/env -i fixes this

alarcher:
I admit I ran the test suite outside oi-userland. Thanks for looking into that.

alarcher:
Fixed.

PR #1969, opened 2016-05-08
Set gcc mediator to 4.9 by default

alarcher:
Something like this ?


xen0l:
LGTM

pyhalov:
yes, but you'll have to rebuild gcc 4.9 after this (bump COMPONENT_REVISION)

alarcher:
Updated.

PR #1976, opened 2016-05-09
Add slurm 16.05

alarcher:
Slurm is a utility for scheduling jobs and allowing me to run nice simulations on my workstation.
I patched the latest stable but need help from SMF wizards for packaging it properly.


components/sysutils/slurm/files/getgrouplist.h
```diff
@@ -0,0 +1,96 @@
+/*-
```
pyhalov: Why do we need it? Just to generate patches/11-getgrouplist.patch ? Then there's issue... getgrouplist.c is BSD-licensed, slurm - GPLv2-licensed. We can't mix such code in one file and I'm not sure if we can link such files in one object file.

alarcher: Right, I always forget that the relationship is unidirectional...
The reason is that illumos does not have getgrouplist() maybe I should file a RFE.

alarcher: BTW I took this patch from the Solaris port of KDE 4.

components/sysutils/slurm/patches/11-getgrouplist.patch
```diff
@@ -0,0 +1,214 @@
+--- ./src/slurmd/slurmd/req.c.orig	2016-05-05 22:21:22.550661771 +0200
```
pyhalov: We can't do this... Licensing

alarcher: My understanding was that adding BSD-licensed code patches (or any permissive copyleft) to a GPL licensed code is not an issue while the converse is not possible: GPL code requires that the overall license is GPL.

pyhalov: http://www.gnu.org/licenses/gpl-faq.html#OrigBSD

alarcher: Oh sorry I did not see that it is the 4-Clause version.

alarcher: RFE "Add getgrouplist() to illumos": https://www.illumos.org/issues/7024

alarcher:
Need to deliver example configuration files.

xen0l:
Any news on this?

alarcher:
Yes everything is fine, I fixed all the requests, I bumped to 16.05 while the original was 15.08 and also I fixed the getgrouplist patch issue using Alexander's pointer.
 I need to figure out how which generic default configuration files to deliver and I am not sure how they should look.

pyhalov:
Any news on it?

alarcher:
Some permissions to fix and the PR will be ready to be merged.

alarcher:
Segmentation fault in slurmctld needs love...

pyhalov:
Any news on it ? :) BTW, there's 17.02 version already...

alarcher:
@pyhalov did they mention if there are more segfaults in the new version? ;) Could not figure out a working configuration yet, some weird segfaults.

PR #1990, opened 2016-05-12
Add nodejs-6.1.0 to the runtime category.

AndWac:
Hi,

as I am using JavaScript at work now, I added the latest LTS version of nodejs.
All 1088 tests passed (after copying one file that was referenced from the wrong folder).

regards
Andreas


pyhalov:
we have components/node. How does it relate to it?

AndWac:
components/node is an older version (0.12.7) of nodejs. As far as I know it's the last version of Joyent.
There was a fork when some developers where dissatisfied and later re-unioned. The new version became 4.x (the last LTS version). 5.x was a development version and finally became 6.x which is the newest LTS. I am not sure whether we can/should drop 0.12.x. As far as I know it's still maintained like 4.x.

pyhalov:
It seems `link path=usr/node/6/bin/npm target=../lib/node_modules/npm/bin/npm-cli.js mediator=nodejs mediator-version=6` shouldn't be mediated, it's internal link.

components/runtime/node6/Makefile
```diff
@@ -0,0 +1,53 @@
...
+COMPONENT_ARCHIVE_URL=	http://nodejs.org/dist/v$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+# Include /usr/sbin for dtrace
+PATH=/usr/bin:/usr/gnu/bin:/usr/sfw/bin:/usr/ccs/bin:/usr/sbin
+
+CONFIG_SHELL =	/usr/bin/python
```
pyhalov, line 38: Don't do this, use $(PYTHON) or $(PYTHON.VERSION), as /usr/bin/python is mediated

pyhalov:
Please, also add https://gist.github.com/pyhalov/013edebcc2f70f9397586437718a17fb - fix for compilation with gcc 5.3

AndWac:
The npm link is correct and needed. npm is the package manager for nodejs.

pyhalov:
link path=usr/node/6/bin/npm  shouldn't be mediated (or it will disappear on mediator change)

pyhalov:
why have you closed this?

AndWac:
I am reworking it. I still struggle with git a little bit so it will 
take some time.
This weekend I will be away so I thought it is better to close the PR 
and open a new one later.

Am 14.05.16 um 12:45 schrieb Alexander Pyhalov:

PR #2003, opened 2016-05-14
pidgin: disable gstreamer to protect from crashing because of SIGALRM

avsej:


pyhalov:
Merged https://github.com/OpenIndiana/oi-userland/pull/2006 fix

PR #2013, opened 2016-05-17
Fix tracker build with giflib 5.1

pyhalov:


jimklimov:
Google says it is a popular issue ;)

Care to make the fix dependent on giflib version, like here: https://bugsfiles.kde.org/attachment.cgi?id=87134 or https://github.com/mapserver/mapserver/commit/0101157d4edd5d837d17356c66c7d25af7e020ac ?

Also some precedents show that reasonable values can be passed as the code: https://github.com/matthiaskramm/swftools/pull/2/files

Also you can pass NULL for old behavior it seems : https://github.com/Automattic/node-canvas/issues/440

pyhalov:
Don't care to make it dependent on giflib version, as applications are supposed to build with latest one.

PR #2019, opened 2016-05-19
libcouchbase: update to 2.6.0

avsej:


pyhalov:
Is it ABI-safe?

pyhalov:
Should be.

PR #2022, opened 2016-05-20
nautilus: ABI changes after https://www.illumos.org/issues/6902 integ…

pyhalov:
…ration


pyhalov:
Note, this requires updating build server to the latest osnet-incorporation

PR #2026, opened 2016-05-21
6986 Making vim be more comfortable to use

rtvd:
Adding a system-wide configuration which turns on more feature-rich
mode in VIM. Also turns on syntax highlighting.

https://www.illumos.org/issues/6986

### DIFF
--- a/components/vim/vim.p5m
+++ b/components/vim/vim.p5m
@@ -1636,6 +1636,7 @@ link path=usr/share/man/ru.KOI8-R/man1/rview.1 target=vim.1
 link path=usr/share/man/ru.KOI8-R/man1/rvim.1 target=vim.1
 link path=usr/share/man/ru.UTF-8/man1/rview.1 target=vim.1
 link path=usr/share/man/ru.UTF-8/man1/rvim.1 target=vim.1
+file files/vimrc path=usr/share/vim/vimrc
 
 license vim.license license="VIM License"
 
--- a/components/vim/files/vimrc
+++ b/components/vim/files/vimrc
@@ -1,0 +1,2 @@
+set nocompatible
+syntax on


jimklimov:
As noted on IRC review, I'd rather keep the expected behavior when called as "vi" (that is, old, clumsy, that old-schoolers like me happen to know and can use) and enable convenient new mode when called as "vim". Likely this would require a small patch to be applied against the source code - and according to manpage, the binary already does change its modes of operation based on name ("ex" etc.) so the location to patch should be easy to spot.

Regarding recipe formalities, usually added files (like `vimrc` here) go into the `files/` subdirectory. A few components use `Solaris/` subdirectory as well, mostly for things related specifically to Solaris10+ integration (SMF and RBAC snippets) although I think there are no strict rules on that.

Patches must go into `patches/` subdirectory and be named `*.patch` for the main downloaded source (the framework allows packages built from a mix of sources, where variables and patchfiles are suffixed with `_N` where `N>=1`) - and then they are applied automatically as part of `gmake prep`.

components/vim/Makefile
```diff
@@ -103,6 +103,8 @@ $(BUILD_DIR)/license: $(SOURCE_DIR)/runtime/doc/uganda.txt
 build:		$(BUILD_32)
 
 install:	$(INSTALL_32)
+		@cp $(COMPONENT_DIR)/vimrc $(PROTO_DIR)/usr/share/vim/vimrc
```
jimklimov: Consider using variables from `make-rules/shared-macros.mk` like `$(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)` to ease possible restructuring later in the project evolution ;)

components/vim/vim.p5m
```diff
@@ -1636,6 +1636,7 @@ link path=usr/share/man/ru.KOI8-R/man1/rview.1 target=vim.1
 link path=usr/share/man/ru.KOI8-R/man1/rvim.1 target=vim.1
 link path=usr/share/man/ru.UTF-8/man1/rview.1 target=vim.1
 link path=usr/share/man/ru.UTF-8/man1/rvim.1 target=vim.1
+file path=usr/share/vim/vimrc
```
jimklimov: This could be delivered without changes to `Makefile` `install:` recipe like so:

```
file files/vimrc path=usr/share/vim/vimrc
```

alarcher:
I would certainly be interesting to get vim to work as vim by default.

pyhalov:
@jimklimov suggestion seems to be correct approach here

PR #2028, opened 2016-05-22
6986 Making vim be more comfortable to use

rtvd:
This adds a file with global settings with makes VIM work the way it was intended to work by its authors.

### DIFF
--- a/components/vim/vim-modern.p5m
+++ b/components/vim/vim-modern.p5m
@@ -1,0 +1,27 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Denys Rtveliashvili
+#
+
+set name=pkg.fmri value=pkg:/editor/vim-modern@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="Global settings file to make VIM be actually IMproved by default"
+set name=pkg.description value="Vim is a clone of the Unix editor 'vi'.  By default it runs in an obsolete, compatible mode. These settings make it work in a non-compatible mode, which is assumed to be used in the VIM guide."
+set name=com.oracle.info.description value="Settings for the Vi IMproved text editor"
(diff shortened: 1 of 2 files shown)


components/vim/vim-modern.p5m
```diff
@@ -0,0 +1,27 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Denys Rtveliashvili
+#
+
+set name=pkg.fmri value=pkg:/editor/vim@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
```
jimklimov: The name `editor/vim` is already grabbed ;) Might use `editor/vim/improved` (usually should then rename the P5M to match, as a courtesy for other devs to maintain this), or `editor/vim/modern`...

components/vim/vim-modern.p5m
```diff
@@ -0,0 +1,27 @@
...
+set name=pkg.description value="Vim is a clone of the Unix editor 'vi'.  By default it runs in an obsolete, compatible mode. These settings make it work in a non-compatible mode, which is assumed to be used in the VIM guide."
+set name=com.oracle.info.description value="Settings for the Vi IMproved text editor"
+set name=info.classification value="Development/Editors"
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+file files/vimrc path=usr/share/vim/vimrc
+
+license vim.license license="VIM License"
+
+depend type=require fmri=editor/vim/vim@$(IPS_COMPONENT_VERSION)
```
jimklimov: Depend on the base package `editor/vim@...` (not `vim/vim`) here. Maybe consider the whole suffix =>

```
depend type=require fmri=editor/vim@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) 
```

jimklimov:
Ok, nothing critical pops out now - LGTM :) But I had a completely sleepless night and a 40-hour-long day ;)

pyhalov:
Can we do this with facet? Something like vim-compatible? When it's set, necessary default settings are delivered....

pyhalov:
It seems that making "vim" noncompatible and "vi" compatible is correct aproach

alarcher:
As I explained also on IRC, I like the fact that "vim" would run as not-compatible and that a default configuration file would enforce comfortable defaults optionnally.
I agree that using a facet would be perfect !

xen0l:
The guy is gone, so no point in having open PR. If he comes back, we can reopen.

PR #2029, opened 2016-05-23
Initial recipe for arping-2.16

jimklimov:
How could we not have this? ;)

### DIFF
--- a/components/sysutils/arping/Makefile
+++ b/components/sysutils/arping/Makefile
@@ -53,7 +53,7 @@ CONFIGURE_OPTIONS += $(CONFIGURE_OPTIONS.$(MACH))
 # lazyloading (indirectly enabled by -Bdirect) seems to cause problems with
 # libpcap device enumeration.
 LD_B_DIRECT=
-ASLR_MODE =	$(ASLR_ENABLE)
+#ASLR_MODE =	$(ASLR_ENABLE)
 
 # common targets
 build:		$(BUILD_32_and_64)
--- a/components/sysutils/arping/Makefile
+++ b/components/sysutils/arping/Makefile
@@ -1,0 +1,68 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		arping
+COMPONENT_VERSION=	2.16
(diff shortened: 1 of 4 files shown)


pyhalov:
Should it work in a zone with exclusive IP stack?  

pfexec arping GW
ARPING GW
Timeout
Timeout
Timeout

components/sysutils/arping/Makefile
```diff
@@ -0,0 +1,68 @@
...
+
+# Note: A few options below were replicated from tcpdump and fping recipes
+# fix my alignment issues instead of falling over on sparc
+studio_ALIGN.sparc.64=-xmemalign=16i
+CONFIGURE_OPTIONS.sparc += ac_cv___attribute__=no
+CONFIGURE_OPTIONS += $(CONFIGURE_OPTIONS.$(MACH))
+# lazyloading (indirectly enabled by -Bdirect) seems to cause problems with
+# libpcap device enumeration.
+LD_B_DIRECT=
+ASLR_MODE =	$(ASLR_ENABLE)
```
xen0l: Remove this, we don't have ASLR capabilities on illumos

jimklimov:
It should, and sometimes works for me - though usually only for one
"arping" packet in a long while, e.g. usually the first one and then for a
minute none others pass. I am not sure where in the stack these get lost :(

2016-05-24 23:29 GMT+02:00 Alexander Pyhalov notifications@github.com:

pyhalov:
this issue should be investigated before merge

jimklimov:
From reading similar code (snoop, p0f, nmap, etc.) it seems that some ioctl's are generally in order to force Solarish systems to not-buffer incoming packets. Otherwise you have to wait a while before you see some output from the sniffer.

I see that the main ioctl is there in arping, and others I tried to stuff in there now (to set size and timeout limits for the select() waits to 0) fail with Invalid Parameter (22).

jimklimov:
I did some more research this weekend regarding issues with arping not seeing all traffic - same applies to p0f, and seems to be some change in illumos vs. sol10 where ioctl's to change the STREAMS attributes for the pcap socket do work, and on OI/Hipster they return EINVAL.

The ability to quickly poll incoming packets by a sniffer is there, since snoop (libdlpi) and tcpdump (libpcap) work. "Surprisingly" tcpdump does work on OI, and with same libpcap as does not work for arping/p0f, so one has to dig into code to find differences (who uses which bpf implementation, what ioctl's are used and do work, what flags are used to set up pcap, notably what timeouts are in place for pcap live stream setup and for poll or select - and whether poll or select are used, etc.)

Manpages of libpcap mention that some OSes are worse at poll() of pcap sockets than others (but suggested workarounds did not help me)

jimklimov:
The sniffers apparently need some more debugging on illumos or specifically OI, or maybe just source code inspection  - they are not getting packets, or not many/often, which is seemingly similar to issues with (disabling of) buffering of accumulated data. There are other sniffers in repo that work (nmap, ngrep, tcpdump come to mind first). All are based on libpcap as well.

So there is a need for some quiet time to sit down and compare who does what - and I didn't find enough of that limited resource in months ;(

xen0l:

PR #2033, opened 2016-05-24
Add webkitgtk 2.12.2

pyhalov:


components/library/webkitgtk/Makefile
```diff
@@ -0,0 +1,84 @@
...
+REQUIRED_PACKAGES += library/zlib
+REQUIRED_PACKAGES += system/library/fontconfig
+REQUIRED_PACKAGES += system/library/freetype-2
+REQUIRED_PACKAGES += system/library/g++-4-runtime
+REQUIRED_PACKAGES += system/library/gcc-4-runtime
+REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += x11/library/libx11
+REQUIRED_PACKAGES += x11/library/libxcomposite
+REQUIRED_PACKAGES += x11/library/libxdamage
```
alarcher, line 85: You need to add REQUIRED_PACKAGES += developer/gperf

alarcher:
It is worth noting that pthread_attr_get_np was only added to illumos-gate on May 19, 2016.
Compilation will fail with anterior versions.

pyhalov:
we have suitable version on build server

alarcher:
I just mention that for potential testers.

components/library/webkitgtk/Makefile
pyhalov, line 85: fixed

PR #2036, opened 2016-05-24
Add iperf3

xen0l:


pyhalov:
Do we care about iperf 2 / iperf 3 compatibility?

xen0l:
I personally don't and don't have strong opinion here. But as I have checked, many distributions provide iperf and iperf3 packages. Let's do the same.

PR #2037, opened 2016-05-24
Update pciutils to 3.4.0

xen0l:


xen0l:
Should we drop depend action from p5m file? pci.ids are delivered from illumos-gate. Any comments on this?

pyhalov:
we have pci.ids package, which delivers usr/share/pci.ids.gz. I wonder if we really need it...

pyhalov:
At least pciutils don't need it

xen0l:
pcituils don't need it. I added that line a long time ago, so if we have a querying tools, we should also have a database to query. Should I make it instead dependent on illumos packages or remove it completely?

pyhalov:
perhaps, depend on file?

PR #2038, opened 2016-05-24
Add netperf 2.7.0

xen0l:
Should we split netperf and netserver into two different packages? netserver could also use SMF manifest. What do you think?


pyhalov:
I've looked at Ubuntu, they deliver one package. I don't have strong opinion on this. However, SMF manifest would be good.

PR #2039, opened 2016-05-25
Update screen to 4.3.1

xen0l:


components/sysutils/screen/patches/socket.patch
```diff
@@ -0,0 +1,12 @@
+--- screen-4.3.1/socket.c.~1~	2015-09-28 13:59:32.049275138 -0700
++++ screen-4.3.1/socket.c	2015-09-28 13:59:47.162984046 -0700
+@@ -30,6 +30,9 @@
+ #include <sys/types.h>
+ #include <sys/stat.h>
+ #include <fcntl.h>
++
++#define _XPG4_2
```
pyhalov, line 8: seems ugly, but let it be... We do similar things for php

pyhalov:
finally found something more elegant..... CFLAGS+="-std=c99 -D_XOPEN_SOURCE=600"

PR #2047, opened 2016-05-26
Separate IPS validation from publishing

jimklimov:
During my recipe-development and experiments (including point-backports to another OI release), I've found some use for this change:

Separation of IPS validation from publishing, as a usable command-line target. Now a `make pre-publish` allows to make sure that my package formalities are okay, without taking time to pollute my repo wiith e.g. experimental code, and a `make publish` does the pre-publish and then actually transfers the package bits to repo, as before.

Note that sometimes repetitive calls to either old `publish` or new `pre-publish` re-ran for me without changes to sources. I think this is because of dependency on `$(BUILD_DIR)` somewhere in the tree (around line 333), which causes `pkgmogrify` to re-run up till `depend.res` files. And piling IPS temporary files straight into `BUILD_DIR` seems to cause a sort of loop (for re-invokations of `make`), at least if timestamps are far apart enough for `make` to care (or maybe this only hurts on the first run after a `make clean` - afterwards it is the files being updated/touched, but not the directory object). I did not change anything in this regard, it doesn't happen too often to be a problem so far.

As much as I've tested, the hack is convenient and does not break the ordinary use-cases :)

NOTE: Previously this PR was about two separate changes, but due to their different value and potential desirability for upstream, it was split. The other half is PR #2063.

### DIFF
--- a/make-rules/ips.mk
+++ b/make-rules/ips.mk
@@ -426,7 +426,6 @@ $(MANIFEST_BASE)-%.pre-published:	$(MANIFEST_BASE)-%.depend.res $(BUILD_DIR)/.li
 
 # Push to the repo
 $(MANIFEST_BASE)-%.published:	$(MANIFEST_BASE)-%.pre-published
-	@echo "PUSHING NEW PACKAGE CONTENTS TO THE REPO"
 	$(PKGSEND) $(PKGSEND_PUBLISH_OPTIONS) $<
 	$(PKGFMT) <$< >$@
 
--- a/make-rules/ips.mk
+++ b/make-rules/ips.mk
@@ -174,6 +174,7 @@ MANIFESTS =		$(VERSIONED_MANIFESTS:%=$(MANIFEST_BASE)-%)
 
 DEPENDED=$(VERSIONED_MANIFESTS:%.p5m=$(MANIFEST_BASE)-%.depend)
 RESOLVED=$(VERSIONED_MANIFESTS:%.p5m=$(MANIFEST_BASE)-%.depend.res)
+PRE_PUBLISHED=$(RESOLVED:%.depend.res=%.pre-published)
 PUBLISHED=$(RESOLVED:%.depend.res=%.published)
 
 COPYRIGHT_FILE ?=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).copyright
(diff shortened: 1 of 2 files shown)


jimklimov:
ping?

make-rules/ips.mk
```diff
@@ -412,32 +417,44 @@ FRC:
...
+
+# Do all the hard work that is needed to ensure the package is consistent
+# and ready for publishing, except actually pushing bits to a repository
+$(MANIFEST_BASE)-%.pre-published:	$(MANIFEST_BASE)-%.depend.res $(BUILD_DIR)/.linted-$(MACH)
+	$(CP) $< $@
+	@echo "NEW PACKAGE CONTENTS ARE LOCALLY VALIDATED AND READY TO GO"
+
+# Push to the repo
+$(MANIFEST_BASE)-%.published:	$(MANIFEST_BASE)-%.pre-published
+	@echo "PUSHING NEW PACKAGE CONTENTS TO THE REPO"
```
xen0l: Why are you changing the behaviour by adding text?

jimklimov: Following suit of other actions in this file, many of which are quite
verbose, and to warn about a potentially long operation (e.g. remote repo)
or one that user wants aborted (Ctrl+C) at this time.

From what I gathered on IRC discussion, reaction was "Cool, let's do more
of those" - but maybe that was sarcasm? ;)

2016-06-11 23:06 GMT+02:00 Adam Števko notifications@github.com:

pyhalov:
Do I miss something or there are unnecessary tab/space changes in this PR?

jimklimov:
Finally got to revise the PR - the whitespace changes are not fully unnecessary (it was messy originally, where the recipe lines were not starting with TAB chars, so I fixed up the Makefile code near where I was changing anyway). Though I guess it can be split off into a separate commit/PR as the cleanup is not exactly an integral part of the added feature.

pyhalov:
LGTM, however I'm not sure about adding more noice :) @xenol, are you fine with this PR?

xen0l:
I also don't like adding noise. Either we should agree and add it everywhere (it should have a format, which will clearly distinguish it from the messages from the make targets) or leave it out.

jimklimov:
I think the point is rather moot, given that some existing recipes in this file were already peppered with echo's. I removed one I added in "publish" recipe, but kept one in the new "pre-publish" to mark its success.

PR #2063, opened 2016-05-28
Support a custom macros makefile (optional includes from start and end of shared-macros.mk)

jimklimov:
During my recipe-development and experiments (including point-backports to another OI release), I've found some use for this hack:

While back-porting recipes to an OI-151a8 machine, which generally works for whatever few packages I needed to try there, I needed to override the `GCC_ROOT` and some other stuff. Patching up the `shared-macros.mk` directly is inconvenient (later Git rebases are tricky). So I'd rather optionally include a non-Git-tracked makefile with such customizations.

For example, this snippet allows me to build with gcc-4.4.4 that is available on that box, and happens to work for me:

```
jim@jimoi:~/shared/oi-userland/make-rules$ cat custom-macros-pre.mk 
# Use another GCC
override GCC_ROOT := /opt/gcc/4.4.4
```

As much as I've tested, the hack is convenient for its purpose and does not break the ordinary use-cases :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ tools/python/pkglint/*.pyc
 tools/parfait
 tools/pkglint
 tools/time-*o
+
+make-rules/custom-macros-pre.mk
+make-rules/custom-macros-post.mk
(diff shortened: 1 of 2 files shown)


xen0l:
This was debated on IRC and we came to the conclusion that one can use environmental variables. The risk of allowing customization of the build system is big and people might use it store macros, which are refenrenced in component Makefiles.

PR #2065, opened 2016-05-29
Add Device Driver Utility blob

alarcher:
Just adding back DDU as an in-tree blob for now.
The recipe copies pci.ids from the pciutils/pci.ids package until we have libddudev back in place.


pyhalov:
If we don't have all ddu sources for now, perhaps add it in components/osol/ using the same pattern?

alarcher:
Done.

pyhalov:
I meant fetching from http://pkg.openindiana.org/legacy and patching.... However, if we want to deliver blobs in src/, let it be, but at least it shouldn't be necessary to add to repository *.pyc files.

alarcher:
OK, that's all I can do for now.
I have a repository with some work on libddudev but no time right now.
Let us just do this iteration to fix some needed issues and preserve functionality.

components/osol/ddu/Makefile
```diff
@@ -0,0 +1,59 @@
...
+
+# Using gzipped pci.ids at another location than /usr/ddu/data/ is not possible
+# until all_devices is modified. For now just copy the file as expected.
+
+install:
+	$(MKDIR) $(PROTO_DIR)
+	($(CP) -rpf $(SOURCE_DIR)/srcs/usr $(PROTO_DIR))
+	if [ -f $(PCI_IDS_FILE) ]; \
+	then gunzip -c $(PCI_IDS_FILE) > $(PROTO_DIR)/usr/ddu/data/pci.ids; \
+	fi;
```
pyhalov, line 40: And why do we include srcs/usr/ddu/data/pci.ids?

alarcher, line 40: Just for the sake of keeping original files somewhere in the tree for now.
Without pci.ids the code is useless so I thought it would be good to have a self-contained tree for which some files can be updated.
Once the new git repository with ddu code is set up this will not matter anymore.

PR #2068, opened 2016-05-29
Initial commit: my brew of p0f-3.09b for Solaris

jimklimov:
NOTE: UIDs/GIDs defined for the daemon service are reserved in the OI Wiki.

Note: the use of `p0f` username requires `pkglint` fixed with https://github.com/OpenIndiana/pkg5/pull/14

### DIFF
--- a/components/sysutils/p0f/Makefile
+++ b/components/sysutils/p0f/Makefile
@@ -66,7 +66,9 @@ $(BUILD_DIR)/%/.built: $(SOURCE_DIR)/.prep
 # in the sources, so we do not make a 64-bit build
 build:		$(BUILD_32)
 
-install:	build
+# Note: there is essentially no install recipe in the original project
+# So we fake it here, and the p5m manifest plucks files directly from builddir
+install:	$(BUILD_32)
 	$(TOUCH) $(INSTALL_32)
 
 test:		$(NO_TESTS)
--- a/components/sysutils/p0f/Makefile
+++ b/components/sysutils/p0f/Makefile
@@ -9,8 +9,10 @@
 #
 
 #
-# Fixes to make it compile under Solaris and run under SMF Copyright (c) 2011-2015 by Jim Klimov
-# Build/package procedure ported to OI/Hipster Copyright (c) 2015-2016 Jim Klimov
+# Fixes to make it compile under Solaris and run under SMF -
+#   Copyright (c) 2011-2016 by Jim Klimov
+# Build/package procedure ported to OI/Hipster -
+#   Copyright (c) 2015-2016 Jim Klimov
 #
 
 include ../../../make-rules/shared-macros.mk
--- a/components/sysutils/p0f/Makefile
+++ b/components/sysutils/p0f/Makefile
@@ -1,0 +1,78 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Fixes to make it compile under Solaris and run under SMF Copyright (c) 2011-2015 by Jim Klimov
+# Build/package procedure ported to OI/Hipster Copyright (c) 2015-2016 Jim Klimov
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		p0f
+# For up-to-date latest version see http://lcamtuf.coredump.cx/p0f3/releases/
+COMPONENT_VERSION=	3.09b
+# IPS requires this to be numeric, no alpha, no leading zeroes
+IPS_COMPONENT_VERSION=	3.0.9
+COMPONENT_PROJECT_URL=	http://lcamtuf.coredump.cx/p0f3
+# *Unofficial* Git source at https://github.com/p0f/p0f
+COMPONENT_SUMMARY=	p0f - the passive OS fingerprinter daemon
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tgz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:543b68638e739be5c3e818c3958c3b124ac0ccb8be62ba274b4241dbdec00e7f
+COMPONENT_ARCHIVE_URL=	$(COMPONENT_PROJECT_URL)/releases/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI=		diagnostic/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	Applications/Internet
+COMPONENT_LICENSE=	LGPLv2.1
+COMPONENT_LICENSE_FILE=	docs/COPYING
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+# lazyloading (indirectly enabled by -Bdirect) seems to cause problems with
+# libpcap device enumeration.
+LD_B_DIRECT=
+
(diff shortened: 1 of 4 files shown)


components/sysutils/p0f/Makefile
```diff
@@ -0,0 +1,80 @@
...
+		$(GMAKE) clean)
+	(cd $(@D) && $(ENV) $(COMPONENT_BUILD_ENV) \
+		$(GMAKE) all)
+	$(TOUCH) $@
+
+# Common targets; currently some pointer macros are specifically 32-bit
+# in the sources, so we do not make a 64-bit build
+build:		$(BUILD_32)
+
+install:	build
```
pyhalov: Better do it as 

```
install : $(BUILD_32)
```

or publish will do install and publish every time....

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
```diff
@@ -0,0 +1,859 @@
+Fixes against p0f-3.06b (updated for 3.09b) to have it compile and work
+properly under Solaris/illumos. Also integrates with Solaris 10+ SMF serices.
+Fixes licensed under the terms of MIT License.
```
pyhalov: patches should be licensed by original code license

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
```diff
@@ -0,0 +1,859 @@
...
++//#  warning "SOLARIS=yes"
++#  include <sys/fcntl.h>
++#  include <fcntl.h>
++#  include <sys/ioctl.h>
++#  include <sys/stream.h>
++#  include <sys/dlpi.h>
++#  include <sys/bufmod.h>
++#  include <sys/stropts.h>
++# ifdef SOLARIS_UCB
++#    include "/usr/ucbinclude/sys/file.h"
```
pyhalov, line 389: I suggest the patches were created for older Solaris versions. Perhaps, we could drop cases, which are not relevant on illumos?

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
```diff
@@ -0,0 +1,859 @@
...
++       ### Note we start as "root" above and later drop privileges in script. -->
++      </method_context>
++    </exec_method>
++    <exec_method name='stop' type='method' exec=':kill' timeout_seconds='60'>
++      <method_context/>
++    </exec_method>
++    <exec_method name='refresh' type='method' exec=':kill' timeout_seconds='180'>
++      <method_context/>
++    </exec_method>
++    <property_group name='general' type='framework'>
```
pyhalov, line 342: Do we need this?

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
```diff
@@ -0,0 +1,859 @@
+Fixes against p0f-3.06b (updated for 3.09b) to have it compile and work
+properly under Solaris/illumos. Also integrates with Solaris 10+ SMF serices.
```
mtelka: serVices

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
jimklimov, line 389: I would prefer not to, because it would make maintenance of the patch harder for me (currently I can `git diff master` in a checkout of https://github.com/jimklimov/p0f @ `master-solaris`). Anyhow, the UCB code is by default disabled (in `build.sh`)...

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
jimklimov, line 342: Need what? The service? It is what we need this package for :)

p0f runs on the host (may be in local zone, including both exclusive and shared stacks) and sniffs the communications, such as incoming packets. Then it evaluates what OS fingerprint matches the remote IP address in the dialog, and caches the verdict. Then a p0f-client, such as milter-greylist, or p0f's own command-line tool, comes and asks "what OS is on that IP?" - and the p0f-daemon already knows the answer. And then, in case of e.g. milter, we can be more discriminative against say random Windows desktops directly submitting e-mails to our relay (likely botnet drones).

So yes, by design and purpose it is a long-sunning sniffer service :)

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
jimklimov: Hmmm... that is an interesting point, perhaps the patch-files should be split with varied licensing. E.g. I have no preference for my own code (SMF manifest and wrapper script), so it could be MIT or BSD includable by everybody. There are some fixes to original build script and C code that would indeed follow LGPL2 (although again my fixes posted to the world under MIT license should be includable and redistributable by any-license code including (L)GPL) : http://stackoverflow.com/questions/3902754/mit-vs-gpl-license

The hardest bit is about a few trivial but useful routines snatched from the internet or Solaris sources, concentrated in `p0f.c`: the `flock()` wrapper which originates from NetBSD toolchain (see http://www.perkin.org.uk/posts/solaris-portability-flock.html for details) which should be includable into GPL too - e.g. http://opensource.stackexchange.com/questions/304/can-i-take-bsd-licensed-code-and-distribute-it-under-gpl discusses that.

Then there is also a `strioctl()` wrapper that I took IIRC from Solaris snoop (so maybe it is CDDL) which is IMHO too trivial to be a contention point - the same setup of structure and call to `ioctl()` is often done "inline" (e.g. shown in https://illumos.org/man/7m/timod manpage), it just looks cumbersome and repetitive to do so. 

UPDATE: I did not find it in current `snoop` sources, but there are quite a few examples of similar stuff across illumos-gate code (same name, different args, similar job) in pppoe, bridged, etc.; overall it is a Solaris STREAMS interface ioctl.
Actually, the wording of the bit I have best matches this code layout : http://web.mit.edu/ruggles/Desktop/ruggles/MacData/afs/net/project/attic/ppp/dp/sun4m_51/dp-3.0/conf/strioctl.c from 1992 (which may be a copy of something yet older in history). Licensing of "DP" is uncertain from the project's document (a few parent-dirs up).
Even Linux Kernel has a very distant cousin of this snippet ;) which seems to trace into SGI Unix... https://www.kernel.org/pub/linux/kernel/people/marcelo/linux-2.4/drivers/sgi/char/streamable.c

Example usage of those ioctl's later in the file, that got me started, originates from wpa-supplicant and nmap (both BSD licensed).

So I guess the wording can be changed that I post my patch under BSD license, so there is less confusion (I can't well tell MIT and BSD apart in practice), and it should be bindable with a GPL project. One oddball is the ioctl wrapper, maybe...

components/sysutils/p0f/patches/001-p0f-3.09b-solarisfix.patch
pyhalov, line 342: no, I meant mentioning of solaris.smf.manage.milter authorizaion

components/sysutils/p0f/Makefile
```diff
@@ -0,0 +1,82 @@
...
+	$(TOUCH) $@
+
+# Common targets; currently some pointer macros are specifically 32-bit
+# in the sources, so we do not make a 64-bit build
+build:		$(BUILD_32)
+
+# Note: there is essentially no install recipe in the original project
+# So we fake it here, and the p5m manifest plucks files directly from builddir
+install:	$(BUILD_32)
+	$(TOUCH) $(INSTALL_32)
```
pyhalov, line 72: This line is not necessary here.

components/sysutils/p0f/Makefile
jimklimov: Updated, though I am not sure it matters in practice. The `build` recipe above has the `$(BUILD_32)` as dependency so if the latter does the touch-protection, it applies whatever way you write it.

jimklimov:
NOTE: There are currently some issues with not-sniffing packets for actual work.
They seem technically related to issues in arping, see https://github.com/OpenIndiana/oi-userland/pull/2029

xen0l:

PR #2071, opened 2016-05-31
imagemagick: update to 6.9.4

pyhalov:


pyhalov:
components to rebuild:
xscreensaver/rss-glx
editor/emacs
inkscape

PR #2076, opened 2016-05-31
Update X11 libraries (chunk 1)

alarcher:


components/x11/xcb-util-cursor/Makefile
```diff
@@ -48,3 +49,8 @@ BUILD_PKG_DEPENDENCIES = $(BUILD_TOOLS)
 
 include $(WS_TOP)/make-rules/depend.mk
```
pyhalov: Remove this include

PR #2090, opened 2016-06-01
Update X11 data

alarcher:


pyhalov:
Shouldn't we add dependency on x11/library/libfontenc/encodings to x11/library/libfontenc on this stage?

alarcher:
I have a commit in the branch but could not apply it since libfontenc does not exists in the repo for now. UPDATE: now cherry-picked

PR #2093, opened 2016-06-01
Add Makefile for rebuilding the X11 gate

alarcher:


components/components.ignore
```diff
@@ -33,3 +33,7 @@
 # 
 # Should be delivered by illumos-gate, don't enable
 /^libtecla$/d
+
+# X11 gate is not meant to be built automatically but to allow developers
+# to test fixes and updates
+/^x11-gate$/d
```
pyhalov: is it really necessary? bass-o-matic builds its components list based on p5m files, not makefiles

alarcher: I was not aware of that. Should I remove it or is it better to be conservative?

pyhalov: It's better to be removed

PR #2095, opened 2016-06-01
tuntap grouping and SMF integration

jimklimov:
While running OpenVPN in a local zone [1], I found that the GZ does not provide it the device nodes after a reboot. In fact, the drivers are not loaded at all. And simply `modload`ing them did not work for some reason, while `rem_drv`+`add_drv` did.

So I made this service (disabled by default) to do just that - recreate `/dev/{tun,tap}` after booting before local zones are started.

[1]: The local zone with OpenVPN simply has the devices piped, and it works:

```
  <device match="/dev/tun"/>
  <device match="/dev/tap"/>
```

### DIFF
--- a/components/tuntap/files/tuntap.sh
+++ b/components/tuntap/files/tuntap.sh
@@ -1,0 +1,46 @@
+#!/bin/sh
+
+#
+# Make sure TUN/TAP device nodes are present. This may be not needed
+# if clients such as OpenVPN are running in the global zone and so
+# tickle initialization of the drivers, but seems to be needed when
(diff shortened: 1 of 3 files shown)


xen0l:
We discussed this on IRC that this approach is not a sane one. TUN/TAP driver is installed in another BE, so reboot is needed. If after reboot TUN/TAP is not loaded, device action in p5m file should be investigated. The local zone with OpenVPN has to have access to TUN/TAP devices, but there is nothing we can do with it. If somebody is working using zones for networking purposes, he should have the knowledge.

PR #2133, opened 2016-06-20
Add postfix to oi-userland

xen0l:


components/meta-packages/mta/mta.p5m
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2016 Adam Stevko
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=description value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="org.opensolaris.category.2008:Meta Packages/Group Packages"
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+depend fmri=service/network/smtp/sendmail type=require-any
```
pyhalov: this should look like depend fmri=service/network/smtp/sendmail fmri=mail/postfix type=require-any. And perhaps we should call postfix package service/network/smtp/postfix to be consistent?

PR #2146, opened 2016-06-22
Add dmenu-4.6

avsej:


components/desktop/dmenu/patches/config-mk.patch
```diff
@@ -0,0 +1,32 @@
...
+ # flags
+ CPPFLAGS = -D_BSD_SOURCE -D_XOPEN_SOURCE=700 -D_POSIX_C_SOURCE=200809L -DVERSION=\"${VERSION}\" ${XINERAMAFLAGS}
+-CFLAGS   = -std=c99 -pedantic -Wall -Os ${INCS} ${CPPFLAGS}
+-LDFLAGS  = -s ${LIBS}
++CFLAGS   = ${CC_BITS} -std=c99 -pedantic -Wall -Os ${INCS} ${CPPFLAGS}
++LDFLAGS  = ${LD_BITS} -s ${LIBS}
+ 
+ # compiler and linker
+-CC = cc
++CC = gcc
```
pyhalov: shouldn't this be like CC?=cc so that we can set CC in component's Makefile?

avsej: I'm patching it anyway, but okay, it is easy

components/desktop/dmenu/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+#
+# Copyright (c) 2016 Sergey Avseyev
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		dmenu
+COMPONENT_VERSION=	4.6
+COMPONENT_PROJECT_URL=	http://tools.suckless.org/dmenu/
+COMPONENT_SUMMARY=	dmenu is a dynamic menu for X, originally designed for dwm. It manages large numbers of user-defined menu items efficiently
```
pyhalov: can we preserve COMPONENT_SUMMARY short and move longer description to pkg.description if necessary?

components/desktop/dmenu/Makefile
```diff
@@ -0,0 +1,48 @@
...
+COMPONENT_ARCHIVE_URL=	http://dl.suckless.org/tools/$(COMPONENT_ARCHIVE)
+COMPONENT_LICENSE=	MIT/X
+COMPONENT_LICENSE_FILE=	LICENSE
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_BUILD_ENV=	CC_BITS=$(CC_BITS) LD_BITS=$(LD_BITS)
+COMPONENT_BUILD_ARGS=	CC=gcc
```
pyhalov: CC=$(CC)

PR #2156, opened 2016-06-24
Sendmail

pyhalov:


pyhalov:
Note that this can/should be merged only after removing sendmail from illumos-gate!!

PR #2167, opened 2016-06-27
Initial commit of distcc (at latest Git commit from 20160712)

jimklimov:
NOTES:

1) Currently, the package does not include an init or SMF script like https://github.com/distcc/distcc/blob/master/packaging/RedHat/init.d/distcc since it is questionable with what credentials a host should slave away for compilations from remote clients, and because it is not the only way to get it started (there is also SSH invokation and inetd-style). A service would be nice for e.g. zeroconf (avahi) usecase; the default port is non-privileged so a user may run it. Subsequent PRs to add a service is possible when/if someone has a firmer opinion on what it should be.

2) It is recommended to use same compiler versions (native or cross) on all systems for a distributed build for a particular target platform, so we should have some PR (or manual setup) to expose full compiler names, maybe even with versions (missing now in common cases), into some path that would be usable by distcc. Might be symlinks into /usr/bin, e.g. `i386-pc-solaris2.11-gcc***-4.9.0***`.

3) Integration with CCACHE is a common usecase. Generally the client holds the `.ccache` directory of his build products (can be shared among builders https://ccache.samba.org/manual.html#_sharing_a_cache and buildhosts https://ccache.samba.org/manual.html#_sharing_a_cache_on_nfs), and with an `export CCACHE_PREFIX="distcc"` it spreads the compilation load (if deemed needed) to workers listed in `DISTCC_HOSTS` - e.g. `export DISTCC_HOSTS='--randomize localhost red,cpp,lzo green,cpp,lzo blue,cpp,lzo'`. Not sure how/if this works with pump mode. So far not specially handled nor tested for oi-userland builds in particular (e.g. what level of builds do we want to parallelize - objects inside recipes? many packages with each recipe as a singular-host target?).

4) Did not really get if we can use alien systems to compile for each other (e.g. solarish <-> linux, oi ,-> omnios, etc.) if out-of-the-box compiler versions and system headers are different.

Frankly speaking, the resulting `distcc*` programs are not yet tested in practice (@xen0l - you're welcome ;) )

Some projects and blogs have nice overviews of practical distcc usage:
- https://wiki.gentoo.org/wiki/Distcc
- https://wiki.archlinux.org/index.php/Distcc
- http://blog.kfish.org/2010/06/speeding-up-cross-compiling-with-ccache.html
- http://linux.die.net/man/1/distcc
- http://linux.die.net/man/1/distccd

Also of interest seems to be the DMUCS frontend to list distcc-servers : http://dmucs.sourceforge.net/ which despite a decade-long gap in development seems to be happily used by various internet mailing list members.

### DIFF
--- a/components/developer/distcc/Makefile
+++ b/components/developer/distcc/Makefile
@@ -20,8 +20,8 @@ COMPONENT_VERSION=	3.1
 # Versioning is complicated: samba.org/ftp has 2.x versions as newest,
 # while newest github release is 3.1 from 2008, and there is recent HEAD
 # So we take a randomly newest Git snapshot as of package review:
-COMPONENT_GIT_DATE=	20160427
-COMPONENT_GIT_REV=	544904c
+COMPONENT_GIT_DATE=	20160712
+COMPONENT_GIT_REV=	c8eac97
 IPS_COMPONENT_VERSION=	$(COMPONENT_VERSION).0.$(COMPONENT_GIT_DATE)
 COMPONENT_PROJECT_URL=	http://distcc.samba.org/
 
@@ -31,7 +31,7 @@ COMPONENT_SRC= distcc-distcc-$(COMPONENT_GIT_REV)
 COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.gz
 
 COMPONENT_ARCHIVE_HASH=	\
-  sha256:6f0ae295e2652b44af91086125ca6df254c72bf5283dc5a10a5bf273c45ed05e
+  sha256:13268ca2f6fee14647610e137c68eef14b2686aad251b3c89da81a0593b2698d
 COMPONENT_ARCHIVE_URL= \
   https://codeload.github.com/distcc/distcc/legacy.tar.gz/$(COMPONENT_GIT_REV)
 
--- a/components/developer/distcc/Makefile
+++ b/components/developer/distcc/Makefile
@@ -97,9 +97,8 @@ CFLAGS += -Wno-error=unused-but-set-variable
 CONFIGURE_OPTIONS +=	--with-auth
 
 # Support mDNS "zeroconf" peer discovery/announcement
-# NOTE: May require some source chizeling, skip in first recipe iterations
-#CONFIGURE_OPTIONS +=	--with-avahi
-CONFIGURE_OPTIONS +=	--without-avahi
+REQUIRED_PACKAGES +=	system/network/avahi
+CONFIGURE_OPTIONS +=	--with-avahi
 
 # GUIs for monitoring
 CONFIGURE_OPTIONS +=	--with-gnome
--- a/components/developer/distcc/Makefile
+++ b/components/developer/distcc/Makefile
@@ -86,10 +86,11 @@ CONFIGURE_OPTIONS +=	--sysconfdir=$(ETCDIR)
 CONFIGURE_OPTIONS +=	--enable-rfc2553
 
 # GSS-API mutual auth
-# NOTE: May require some source chizeling, skip in first recipe iterations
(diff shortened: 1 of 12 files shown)


pyhalov:
for first error cast to (struct sockaddr *) would help

jimklimov:
After some late-night glancing, I am not certain whether the two structures are compatible enough to just cast one into another.

pyhalov:
thet are supposed to be, sockaddr_in is a "subclass" of sockaddr

jimklimov:
Thanks, at least as far as compiler and publish are concerned, this worked :)

components/developer/distcc/Makefile
```diff
@@ -0,0 +1,122 @@
...
+
+test:           $(NO_TESTS)
+
+# Auto-generated contents below.  Please manually verify and remove this comment
+REQUIRED_PACKAGES += library/desktop/gtk2
+REQUIRED_PACKAGES += library/desktop/libgnome
+REQUIRED_PACKAGES += library/desktop/libgnomeui
+REQUIRED_PACKAGES += library/glib2
+REQUIRED_PACKAGES += library/popt
+REQUIRED_PACKAGES += runtime/python-26
```
pyhalov: Why both?

jimklimov: Not seen in email, but both pythons i believe? Python3 is needed for pump
server, and any (2.4+?) for the distcctop monitoring client.

## 

27.06.2016 15:57 пользователь "Alexander Pyhalov" notifications@github.com
написал:

In components/developer/distcc/Makefile
https://github.com/OpenIndiana/oi-userland/pull/2167#discussion_r68580297:

> +# common targets
> +# NOTE: Currently GCC libiberty is only published as a 32-bit .a file
> +# so we can only build a 32-bit distcc
> +build:          $(BUILD_32)
> +
> +install:        $(INSTALL_32)
> +
> +test:           $(NO_TESTS)
> +
> +# Auto-generated contents below.  Please manually verify and remove this comment
> +REQUIRED_PACKAGES += library/desktop/gtk2
> +REQUIRED_PACKAGES += library/desktop/libgnome
> +REQUIRED_PACKAGES += library/desktop/libgnomeui
> +REQUIRED_PACKAGES += library/glib2
> +REQUIRED_PACKAGES += library/popt
> +REQUIRED_PACKAGES += runtime/python-26

Why both?

—
You are receiving this because you authored the thread.
https://github.com/OpenIndiana/oi-userland/pull/2167/files/c8103cf97e33e1d97629de4e2cee6f238330803b#r68580297,
or mute the thread
https://github.com/notifications/unsubscribe/ABj5lDNgtF6NryTevVuOQsHDLn_CiOrhks5qP9bFgaJpZM4I-rO4
.

pyhalov: Can we use just python 3.4?

jimklimov: Probably, but then it is a dependency to install for distcc overall, and/or
the monitoring tool should be moved to another subpackage. From what I
gather, distcc can still be used without pump/include_server scripts, so
python3 is pulled by just one subpackage. Just like GUI monitoring tools
that pull X11 onto a build host/wkst are not generally required.

We need someone to actually try and use this (and check what python version
is really needed and sufficient for which bits)... xen0l asked on IRC some
time back if we can build distcc to boost our pipeline in a manner similar
to ccache, so i'm looking at him to give it a shot ;)

## 

27.06.2016 21:51 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov: BTW, if there's choice between Python 2.6 and 2.7, use 2.7 (we are slowly going to make it the default one). If we already require python 3.4 and everything works with python 3.4, it would be good to use just one python version

xen0l:
@jimklimov what's the state of this PR?

PR #2168, opened 2016-06-27
ngrep.p5m / Makefile : added REQUIRED_PACKAGES and manpage transform

jimklimov:
I do not think any of this causes a COMPONENT_REVISION bump. Even if someone has this already installed, the change in e.g. manpages is negligible and can wait till version bump?..

### DIFF
--- a/components/sysutils/ngrep/Makefile
+++ b/components/sysutils/ngrep/Makefile
@@ -37,10 +37,13 @@ COMPONENT_PRE_CONFIGURE_ACTION= \
 CONFIGURE_OPTIONS+= --enable-ipv6
 CONFIGURE_OPTIONS+= --enable-pcre
 CONFIGURE_OPTIONS+= --with-dropprivs-user=nobody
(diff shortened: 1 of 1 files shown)


components/sysutils/ngrep/ngrep.p5m
```diff
@@ -13,6 +13,8 @@
 # Copyright 2016 Adam Stevko
 #
 
+<transform file path=usr.*/man/.+ -> default mangler.man.stability uncommitted>
```
pyhalov: This is not strictly necessary, this transform is already present in transforms/defaults

jimklimov: For some reason, we have it in many manifests... and carry over into new ones too :)

xen0l: If it's already present in transforms/defaults, we should drop it, otherwise LGTM. i forgot about this, thanks for fix!

PR #2174, opened 2016-06-28
Use OI ntp vendor zone in ntp.client and add some security tweaks

xen0l:


pyhalov:
What about delivering default ntp.conf and enabling ntp service by defautl?

pyhalov:
Enabling it by default perhaps is not good, as it's installed by default

AndWac:
Am 29.06.16 um 09:36 schrieb Alexander Pyhalov:

> What about delivering default ntp.conf and enabling ntp service by 
> defautl?
> 
> I like the idea.

pyhalov:
Seems to be ready

PR #2188, opened 2016-06-30
New Component: mate/mate-desktop

Toasterson:


components/desktop/mate/mate-desktop/mate-desktop.p5m
```diff
@@ -0,0 +1,247 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file path=usr/share/locale/([^/]+)(\..+){0,1}(/.+){0,1} -> default facet.locale.%<\1> true>
```
xen0l: We have default transform for this in transforms/locale. It appears to do the same. Can you confirm? If so, please remove it. If not, at least move it under copyright, so it's clear there are transforms used. Otherwise, LGTM.

Toasterson: I dont know the reason why it was added but as all the other mate components get their locale properly cought and this transform looks like a weaker one of the one we have i removed it.

PR #2205, opened 2016-07-03
New Component: mate-session-manager

Toasterson:


components/desktop/mate/mate-session-manager/files/metacity.desktop
```diff
@@ -0,0 +1,7 @@
+[Desktop Entry]
+Type=Application
+Exec=/usr/bin/metacity
```
pyhalov: It's strange that you have to start metacity in such way. As  I understand, macro window manager should be used by default. What's going wrong with it?

components/desktop/mate/mate-session-manager/Makefile
```diff
@@ -0,0 +1,73 @@
...
+
+CONFIGURE_OPTIONS+= --sysconfdir=/etc
+CONFIGURE_OPTIONS+= --libexecdir=/usr/lib/mate
+CONFIGURE_OPTIONS+= --disable-static
+CONFIGURE_OPTIONS+= --disable-scrollkeeper
+CONFIGURE_OPTIONS+= --localstatedir=/var/lib
+CONFIGURE_OPTIONS+= --with-gtk=2.0
+CONFIGURE_OPTIONS+= --with-default-wm
+
+PYTHON_VERSION = 2.7
```
pyhalov: mate-session-manager doesn't check for python, so this line and the next one are not needed here

PR #2215, opened 2016-07-06
New Component: marco

Toasterson:


components/desktop/mate/marco/Makefile
```diff
@@ -0,0 +1,85 @@
...
+
+CFLAGS+=-D__EXTENSIONS__
+CFLAGS+=-std=gnu99
+
+CONFIGURE_OPTIONS+= --sysconfdir=/etc
+CONFIGURE_OPTIONS+= --disable-static
+CONFIGURE_OPTIONS+= --localstatedir=/var/lib
+CONFIGURE_OPTIONS+= --enable-maintainer-mode
+
+CONFIGURE_ENV+= PYTHON="$(PYTHON)"
```
pyhalov: Please, add PYTHON_VERSION=2.7 , so that we have less work during python 2.7 migration, which is expected soon.

PR #2236, opened 2016-07-13
Unship pip from python 2.7

xen0l:


pyhalov:
Not sure about this. At least python-27 depends on pip. And I don't see pip executable in python-27.

pyhalov:
Ubuntu still provides http://packages.ubuntu.com/xenial/python-pip

PR #2243, opened 2016-07-17
Initial commit of sslh

benf33:
Initial commit of sslh
- bumped to latest version v1.18
- added examples in /usr/share/doc/sslh


components/network/sslh/Makefile
```diff
@@ -0,0 +1,55 @@
...
+COMPONENT_CLASSIFICATION= System/Security
+
+
+COMPONENT_LICENSE_FILE= COPYING
+COMPONENT_LICENSE=	GPLv2
+
+PATH=/usr/gnu/bin:/usr/bin:/usr/perl5/bin
+LD_OPTIONS=-lresolv -lsocket -lnsl
+
+include ../../../make-rules/prep.mk
```
pyhalov, line 38: You could use $(WS_MAKE_RULES)/something.mk for this and following makefiles

components/network/sslh/Makefile
```diff
@@ -0,0 +1,55 @@
...
+COMPONENT_BUILD_TARGETS += CC=gcc
+
+# common targets
+build:		$(BUILD_32_and_64)
+
+install:	$(INSTALL_32_and_64)
+
+test:		$(NO_TESTS)
+
+BUILD_PKG_DEPENDENCIES =	$(BUILD_TOOLS)
```
pyhalov, line 51: BUILD_PKG_DEPENDENCIES and depend.mk should be removed. REQUIRED_PACKAGES should be added (you can start from gmake REQUIRED_PACKAGES and add other necessary packages manually)

xen0l:
1) you are specifying authorization in SMF manifest, but you are not delivering any authorization file. Please, do so. Just see the following example: https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/network/openvpn/files/auth_openvpn
2) Create a new user/group "sslh" and allocate yourself a UID/GID 79. See openvpn.p5m file to get user and group IPS actions.

benf33:
By examples, sslh setuid to nobody.
Should I just remove the authorisation in SMF manifest ?

components/network/sslh/Makefile
benf33, line 38: Is WS_MAKE_RULES or WS_TOP that I should use ?

xen0l:
Could you provide a dedicated user for it? It is a good practice to manage a service as a ordinary user and that can be accomplished with authorization. You should leave and ship just needed files.

components/network/sslh/Makefile
xen0l, line 38: WS_MAKE_RULES

components/network/sslh/Makefile
```diff
@@ -0,0 +1,55 @@
...
+#
+# Copyright 2016 John Smith. All rights reserved.
+#
+
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		sslh
+COMPONENT_VERSION=	1.18
+COMPONENT_PROJECT_URL=	http://www.rutschle.net/tech
```
avsej, line 20: it should be http://www.rutschle.net/tech/sslh.shtml

components/network/sslh/Makefile
```diff
@@ -0,0 +1,55 @@
...
+# common targets
+build:		$(BUILD_32_and_64)
+
+install:	$(INSTALL_32_and_64)
+
+test:		$(NO_TESTS)
+
+BUILD_PKG_DEPENDENCIES =	$(BUILD_TOOLS)
+
+include ../../../make-rules/depend.mk
```
avsej, line 53: https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/makefile-targets.txt#L120-L126

benf33:
About the dedicated user, can I find somewhere a standard list of users used by services ?
Should we try to match something done elsewhere in the Opensource world (i.e. *BSD, GNU/Linux) ?
In other word, where the suggested UID 79 comes from and what is the logic behind this choice ?

benf33:
Btw, in openvpn component, Solaris specific files are stored in in a sub directory called "files", some others components provide Solaris specific files from a directory called "Solaris".
Which one is the standard we should use ?

pyhalov:
Oracle tend to use Solaris, all other usually use files. List of users is available here http://wiki.openindiana.org/oi/UIDs+and+GIDs

xen0l:
At first, I wanted you to go and allocate yourself UID/GID, but I wasn't sure if you have the wiki registration, so I did it for you. If you want to pick different UID, just feel free.

PR #2249, opened 2016-07-18
Python 3.5

xen0l:


pyhalov:
Should we update it to 3.6 and retest?

pyhalov:
@xen0l Can we update it to 3.6 and integrate?

PR #2267, opened 2016-07-22
ips.mk : add dependencies on "install" so it is safer to call "gmake -j N"

jimklimov:
Before this, a run like

```
gmake clean; gmake -j 8 pre-publish
```

could error out because it can not `pkgmogrify` before even prepping the source directory. Now it waits for the build to complete.

Note that actually building sources in parallel is a separate issue (and depends on that project's recipe quality), and requires e.g. `COMPONENT_BUILD_GMAKE_ARGS=-j6` to be set before calling `gmake`. So this is the real parallel rebuild:

```
gmake clean; echo =======; COMPONENT_BUILD_GMAKE_ARGS=-j6 gmake -j 3 pre-publish  
```

### DIFF
--- a/make-rules/ips.mk
+++ b/make-rules/ips.mk
@@ -203,7 +203,7 @@ $(GENERATED).p5m:	install
 		cat $(METADATA_TEMPLATE) - >$@
 
 # copy the canonical manifest(s) to the build tree
-$(MANIFEST_BASE)-%.generate:	%.p5m canonical-manifests
+$(MANIFEST_BASE)-%.generate:	%.p5m canonical-manifests install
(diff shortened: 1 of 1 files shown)


xen0l:
Can you remove libvpx commit, please?

jimklimov:
Hm... bugg-o ;) i'll look when get around a computer ;)

## 

25.07.2016 0:53 пользователь "Adam Števko" notifications@github.com
написал:

jimklimov:
fixed

xen0l:
LGTM, @pyhalov what do you think?

pyhalov:
This one was reverted, because it significantly increased oi-userland build time

PR #2269, opened 2016-07-24
sample-manifest should drop empty dirs, .la and .pyc files automatically

xen0l:
This PR adds support for dropping actions, which are usually deleted (e.g. empty dirs, .la and .pyc files). @jimklimov and @pyhalov, please comment.


xen0l:
In St. Petersburg, @pyhalov and I talked about a possibility of storing sample-manifest under $(COMPONENT_DIR)/manifests, which would be stored automatically. The point is that when one wants to determine if any new files are added/deleted before package updates, he has to generate a sample manifest from previous version and run a build for it. This could save some time. What do you think?

jimklimov:
Regarding drops: idea is good. Make sure we do drop .pyc files though -
IIRC we may want to ensure building our own and deliver them, no?

As for sample-manifests - they might be stored indeed (and maybe in
component dir, just with a not-*.p5m name to avoid packaging it), but the
process imho would be different: update recipe, generate new sample, 'git
diff' to see what changed since last delivery.

## 

25.07.2016 0:00 пользователь "Adam Števko" notifications@github.com
написал:

In St. Petersburg, @pyhalov https://github.com/pyhalov and I talked about
a possibility of storing sample-manifest under $(COMPONENT_DIR)/manifests,
which would be stored automatically. The point is that when one wants to
determine if any new files are added/deleted before package updates, he has
to generate a sample manifest from previous version and run a build for it.
This could save some time. What do you think?

—
You are receiving this because you were mentioned.
https://github.com/OpenIndiana/oi-userland/pull/2269#issuecomment-234805230,
or mute the thread

xen0l:
I'm also dropping pyc files from both Python 2 and Python 3. Yes, the process you outlined was intended.

Are there any other files, which should be dropped?

jimklimov:
What about not-dropping directories delivered intentionally (e.g. for data
files, owned by package-delivered user account)? Or do we have some better
mechanism (e.g. a decision about one-time services, an equivalent of
systemd tmpfiles service, etc.)?

## 

25.07.2016 9:11 пользователь "Adam Števko" notifications@github.com
написал:

pyhalov:
*.pyo?

pyhalov:
@xen0l : Note, that not all _.pyc files should be dropped, only usr/lib/python[23]/_.pyc, look at transforms/autopyc

xen0l:
This PR now adds support for creating manifests/ directory and sample-manifest.p5m is going to be saved every time when running <pre>gmake sample-manifest</pre>. If you are happy with the changes, I might add some documentation.

pyhalov:
Looks fine... Another transform which is usually done manually is dropping static libraries if there are corresponding dynamic libraries. However, I'm not sure if it should be (or can be easily) automated.

jimklimov:
Here are also cases where we can want this circumventable, and do
intentionally deliver static libs. One example may be libiberty that's
often delivered as both (if you manage to get an .so built at all).

Another Solarish classic is the root/system shell (/sbin/sh) that used to
be a statically built binary - one noncorrupt file suffices for it to work,
and also no insecurity fooling vs. LD_PRELOAD etc... Maybe it would
nowadays have minimal deps like libc.

## 

26.07.2016 8:04 пользователь "Alexander Pyhalov" notifications@github.com
написал:

make-rules/ips.mk
```diff
@@ -111,14 +111,17 @@ PKG_MACROS +=		USERLAND_GIT_REMOTE=$(USERLAND_GIT_REMOTE)
 PKG_MACROS +=		USERLAND_GIT_BRANCH=$(USERLAND_GIT_BRANCH)
 PKG_MACROS +=		USERLAND_GIT_REV=$(USERLAND_GIT_REV)
 
-PKG_OPTIONS +=		$(PKG_MACROS:%=-D %) -D COMPONENT_SUMMARY="$(COMPONENT_SUMMARY)" -D COMPONENT_CLASSIFICATION="org.opensolaris.category.2008:$(COMPONENT_CLASSIFICATION)" 
+PKG_OPTIONS +=		$(PKG_MACROS:%=-D %) -D COMPONENT_SUMMARY="$(COMPONENT_SUMMARY)" -D COMPONENT_CLASSIFICATION="org.opensolaris.category.2008:$(COMPONENT_CLASSIFICATION)"
```
jimklimov: Nit: just a whitespace fix... and one that will happen to conflict with my PR ;)

jimklimov:
Looked at the PRed code - beside the whitespace nit (that I usually wouldn't care about much - but here it may be a problem for my PR, so I do speak up), the code seems okay and reasonable.

PR #2276, opened 2016-07-27
ips.mk : do not call the "install" dynamic target many times...

jimklimov:
… as found recently, this can invalidate intermediate IPS files and cause needless re-processing and re-publishing; instead, use a common file to fence off installation stage from publishing the installed protodir bits

This hopefully amends PRs #2267  and #2274 and again makes "gmake -jN" safe to use.

### DIFF
--- a/doc/makefile-targets.txt
+++ b/doc/makefile-targets.txt
@@ -75,6 +75,26 @@ install::
   macro used will depend on the component and corresponds to the build target
   macro used.
 
+$(ALL_INSTALLED_STAMP):
+  This variable points to a file in the build directory that can be used by
+  recipe steps which should logically follow the "install" target.  That is,
+  for this file to be touched, the component's generic "install" rule must
+  complete successfully.  This covers doing smaller-scoped install rules of
+  the multiple sub-components, for multiple architectures and bitnesses, as
+  well as custom rules sometimes attached in ultimate recipes to happen after
+  a common "gmake install".  It is also logically different from the several
+  $(BUILD_DIR)/*/.installed files, since those mark only the completion of
+  just the "install" rule for a specific sub-component built for a specific
+  architecture and installed (into component's common $(PROTO_DIR) usually).
+  Rules which intend to run only after all manipulation of the PROTO_DIR has
+  completed, can depend on this file (and so cause its creation) rather than
+  the dynamically re-evaluated "install" rule directly -- both to save some
+  compute resources and to avoid possible invalidation of previously built
+  objects if the "install" rule of a particular recipe always does something
+  actively changing the filesystem contents (even if just touching a file).
+  In particular, the IPS packaging steps should only take place after the
+  installation prototype location has stabilized, so they depend on this file.
+
 pkglint::
   This target is an optional target is automatically defined in each Makefile
   by the inclusion of $(WS_MAKE_RULES)/prep.mk.  It will run the build steps
--- a/make-rules/ips.mk
+++ b/make-rules/ips.mk
@@ -194,12 +194,15 @@ PUBLISH_STAMP ?= $(BUILD_DIR)/.published-$(MACH)
 
 # Do all that is needed to ensure the package is consistent for publishing,
 # except actually pushing to a repo, separately from the push to the repo.
-pre-publish:	build install $(PRE_PUBLISH_STAMP)
+# Note that running "install" rule as a dependency for multiple targets can
+# have adverse effects, so instead we call it at most once by requiring the
+# touched ALL_INSTALLED_STAMP file.
+pre-publish:	$(ALL_INSTALLED_STAMP) $(PRE_PUBLISH_STAMP)
 publish:		pre-publish $(PUBLISH_STAMP)
 
 sample-manifest:	$(GENERATED).p5m
 
-$(GENERATED).p5m:	install
+$(GENERATED).p5m:	$(ALL_INSTALLED_STAMP)
 	[ ! -d $(SAMPLE_MANIFEST_DIR) ] && $(MKDIR) $(SAMPLE_MANIFEST_DIR) || true
 	$(PKGSEND) generate $(PKG_HARDLINKS:%=--target %) $(PROTO_DIR) | \
 	$(PKGMOGRIFY) $(PKG_OPTIONS) /dev/fd/0 $(GENERATE_TRANSFORMS) | \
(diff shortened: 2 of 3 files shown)


make-rules/ips.mk
```diff
@@ -186,24 +186,37 @@ IPS_COMPONENT_VERSION ?=	$(COMPONENT_VERSION)
...
+# The "install" target is not a good one to depend on everywhere,
+# because each evaluation is dynamic - both slow and can change
+# files and invalidate previous steps. So in packaging we call it
+# once and leave a single timestamped file to know it is over.
+# If the stamp exists, we know that the build and install rules
+# (however defined and mutilated by ultimate recipes, but usually
+# depending on "%/.installed" files in all sub-component per-MACH
+# build directories) have passed and a PROTO_DIR has been created.
+$(ALL_INSTALLED_STAMP): build install
+	test -d "$(PROTO_DIR)" || { echo "ERROR: install completed but PROTO_DIR is missing!">&2; exit 2; }
```
pyhalov: don't do this, component can do install: $(BUILD_32)

pyhalov:
if we had good way to test it... 
BTW, initially this will also cause mass republish. But it seems to be unavoidable.

make-rules/ips.mk
jimklimov: ok, fixed

jimklimov:
(OFFTOPIC SARCASTIC RANT): Heh, somebody is abusing the super-power of merge permission, to give favors to their PRs and not to have to resolve the conflicts in their branches ;)

I got a plan to fix this:
1) Get PR merge rights in the project
2) Take over the world

Sounds simple enough, let's do it ;)

doc/makefile-targets.txt
```diff
@@ -75,6 +75,26 @@ install::
   macro used will depend on the component and corresponds to the build target
   macro used.
 
+$(ALL_INSTALLED_STAMP):
+  This variable points to a file in the build directory that can be used by
+  recipe steps thich should logically follow the "install" target.  That is,
```
pyhalov: which?

make-rules/shared-macros.mk
```diff
@@ -420,6 +420,24 @@ $(BUILD_DIR_64)/.tested:		BITS=64
...
+# "install" rule at most once by depending on ALL_INSTALLED_STAMP
+# filename and leave the single timestamped file to know it is over.
+# If the stamp exists, we know that the build and install rules
+# (however defined and mutilated by ultimate recipes, but usually
+# depending on "%/.installed" files in all sub-component per-MACH
+# build directories) have passed and a PROTO_DIR has probably been
+# created (although some components are known to limit themselves
+# to just building code with no actual installation implementation).
+ALL_INSTALLED_STAMP ?= $(BUILD_DIR)/.all-installed
+$(ALL_INSTALLED_STAMP): install $(COMPONENT_INSTALL_TARGETS)
```
pyhalov: don't mention COMPONENT_INSTALL_TARGETS, it's for component's internal use

make-rules/shared-macros.mk
```diff
@@ -728,6 +746,7 @@ IPS2TGZ = 	$(WS_TOOLS)/ips2tgz
 
 INS.dir=        $(INSTALL) -d $@
 INS.file=       $(INSTALL) -m 444 $< $(@D)
+INS.exec=       $(INSTALL) -m 655 $< $(@D)
```
pyhalov: Where do you use it?

make-rules/shared-macros.mk
jimklimov: but doesn't it potentially list additional rules that can change the proto-dir separately of the "install" rule? in that case we certainly should want it done before fencing off the stable protodir from installation steps...

doc/makefile-targets.txt
jimklimov: thanks, fixed

make-rules/shared-macros.mk
jimklimov: Understood via IRC, fixed :-)

make-rules/shared-macros.mk
jimklimov: Nowhere - just found the lines while looking for "install" usecases and adding this seemed to be right for completeness...

Indeed, now I see there are very few hits of original `INS.dir` and `INS.file` in the existing sources... But the idea seems reasonable to hide the CLI arguments of a particular `$(INSTALL)`'s implementation to get a job done.

pyhalov: This can be added, but a) mode should be 555 , b) it doesn't belong to this PR.

make-rules/ips.mk
```diff
@@ -194,12 +194,15 @@ PUBLISH_STAMP ?= $(BUILD_DIR)/.published-$(MACH)
 
 # Do all that is needed to ensure the package is consistent for publishing,
 # except actually pushing to a repo, separately from the push to the repo.
-pre-publish:	build install $(PRE_PUBLISH_STAMP)
+# Note that running "install" rule as a dependency for multiple targets can
+# have adverse effects, so instead we call it at most once by requiring the
+# touched ALL_INSTALLED_STAMP file.
+pre-publish:	$(ALL_INSTALLED_STAMP) $(PRE_PUBLISH_STAMP)
```
pyhalov, line 200: it dependent on build install and now only on install (through ALL_INSTALLED_STAMP). Can it cause any issues?

jimklimov, line 200: I thought about it for a while. Logically, this should not be a problem, if for packaging we depend on stabilizing actually the "install" results. That should call "build", but as an internal implementation detail, IMHO.

But according to https://github.com/OpenIndiana/oi-userland/commit/4221e17e0bba8bf0ee49ce857a6d61200767cbf7 the older state (before introducing pre-publish rule), the "build" step was explicitly required in ips.mk: `-publish:       build install $(PUBLISH_STAMP)`

So I guess I'll add the "build" back to dependencies of ALL_INSTALLED_STAMP where it was initially. At least for now it makes sense, since the main use-case is same ips.mk.

make-rules/shared-macros.mk
jimklimov: ack

pyhalov:
LGTM. @xen0l , what do you think?

pyhalov:
@jimklimov : please, merge commits. Let's merge it and see how it goes :)

jimklimov:
Hopefully squashed without losses ;)

pyhalov:
I had to revert this once again

jimklimov:
Eh... any idea why this ends up disruptive?

## 

30.07.2016 15:02 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
In particular, is the build routine similar to "gmake (clean) publish" or
does it call more intermediate steps explicitly and so different from my
testing?

## 

30.07.2016 15:02 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov:
the build routine looks like 

```
[ -f components/components.mk ] && rm components/components.mk
[ -f components/depends.mk ] && rm components/depends.mk

export PUBLISHER=openindiana.org
export USERLAND_ARCHIVES=/data/userland-archives/
export COMPONENT_BUILD_ARGS=-j4

gmake setup
gmake publish -j8 -k
gmake -C components incorporation
```

PR #2279, opened 2016-07-28
New Component: mate-control-center

Toasterson:
Tested and nimbus works too :)


Toasterson:
Only one fix to add. all other patches in gnome control center add functionality.

PR #2281, opened 2016-07-28
New Component: mate-calc

Toasterson:


components/desktop/mate/mate-calc/Makefile
```diff
@@ -0,0 +1,55 @@
...
+COMPONENT_NAME=		mate-calc
+COMPONENT_VERSION=	1.8.0
+COMPONENT_PROJECT_URL=	http://www.mate-desktop.org
+COMPONENT_SUMMARY=	MATE Calculator App
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.xz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:13d56287430cbb7d60a74666ba752a1a8ea9692125c6a02f97d0cc98d451d99a
+COMPONENT_ARCHIVE_URL=	http://pub.mate-desktop.org/sources/$(COMPONENT_NAME)/1.8/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI=		desktop/mate/$(COMPONENT_NAME)
```
pyhalov, line 26: Perhaps, name it desktop/calculator/mate-calc to match gcalctool? However, I don't have strong preference here

Toasterson, line 26: The question is if we want to keep the upstream projects together or classify per type of application. What of these do we usually do?

pyhalov, line 26: With gnome it's a complete mess - we have gnome components with different fmris - editor/, terminal/, gnome/_, desktop/_

Toasterson, line 26: In that case lets not make a complete mess and keep all components of a project under similar FMRI. eg. library/desktop/mate desktop/mate which also has some base in the file structure of oi-userland. Thus giving contributors some clue where to look for when updating components.

pyhalov, line 26: OK

PR #2286, opened 2016-07-29
Add dconf-editor 3.18.0

pyhalov:


pyhalov:
It doesn't look fine with gtk3, but it's rather useful

PR #2292, opened 2016-07-29
Glibmm prereqs

pyhalov:


pyhalov:
Should be merged when latest gtk3 is installed on the build server

PR #2302, opened 2016-07-31
Add lua.hpp in lua

alarcher:
Header lua.hpp was removed from distribution but is used by several projects (e.g. ptlib).


pyhalov:

pyhalov:
Also, if you touch component, remove BUILD_PKG_DEPENDENCIES and depend.mk and add REQUIRED_PACKAGES

alarcher:
Updated.

PR #2312, opened 2016-08-03
Update tor and add arm

xen0l:


components/python/arm/files/arm
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+if [ "$0" = /usr/bin/arm ]; then
+  arm_base=/usr/lib/python2.6/site-packages/arm/
```
pyhalov: 2.6?

components/python/elixir/Makefile
```diff
@@ -0,0 +1,41 @@
...
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=	https://pypi.python.org/packages/3d/8d/0009e2a623849894131f258529fe3a818c5734f7a9892f8721d99bd5cc31/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH= sha256:a7ef437f25b544e4f74fb3236fc43cd25f5d6feb6037dd7c66931046d75439e9
+COMPONENT_LICENSE=	MIT
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/setup.py.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+PYTHON_VERSIONS= 2.6 2.7
```
pyhalov: have no need in this, as this is default

components/python/arm/files/arm
xen0l: leftover from my tries to provide both 2.6 and 2.7. This will be removed

components/python/elixir/elixir-26.p5m
```diff
@@ -0,0 +1,47 @@
+#
```
pyhalov: Why don't you use PYVER manifest?

components/python/sqlalchemy/Makefile
```diff
@@ -0,0 +1,60 @@
+#
```
pyhalov: Does elixir depend on sqlalchemy?

components/python/sqlalchemy/history
```diff
@@ -0,0 +1,4 @@
+library/python-2/sqlalchemy-26@0.9.8-5.12.0.0.0.86.0
```
pyhalov: we didn't have these packages, so we don't need history for them

components/python/elixir/elixir-26.p5m
xen0l: It was faster and easier to do it this way. And Elixir does not support Python 3.x. So in the future, if we deprecate Python 2.6, we just delete the file.

components/python/sqlalchemy/sqlalchemy-GENFRAG.p5m
```diff
@@ -0,0 +1,25 @@
...
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+# Copyright (c) 2013, 2014, Oracle and/or its affiliates. All rights reserved.
+#
+
+# force the rename with an optional dependency on the old name
+depend type=optional fmri=library/python-2/sqlalchemy@0.7.9,5.12-5.12.0.0.0.41.0
```
pyhalov: I suppose we don't need this

components/python/sqlalchemy/Makefile
xen0l: elixir is a plugin for sqlalchemy. elixir depends on sqlalchemy, not the other way around.

pyhalov: Can they be added in one PR?

components/python/elixir/Makefile
xen0l: In the future, when we add 3.x to PYHTON_VERSIONS, it might be needed to have it explicitly there (even I am not using PYVER).

components/python/sqlalchemy/Makefile
xen0l: Yes

PR #2331, opened 2016-08-13
Fix component consistency and rename clang-34

alarcher:
Not sure about the renaming process.


pyhalov:
You can use history file to rename package, look, for example, at components/python/gnome-python/history. Not sure that we should deal with this ancient clang. I'd rather integrated 3.8 and deprecated 3.4

alarcher:
I do not know if anyone relies on clang 3.4 but just in case maybe only deprecate 3.4 when we are sure 3.8 does the job?

pyhalov:
Of course, so I'd do it in one PR :)

xen0l:
note to myself: to further avoid such PRs, we should write some oi-userland best practices and start with naming components.

alarcher:
OK. So I will update this when I can access my build machine again.

xen0l:
Can this be merged prior to clang-38?

alarcher:
This is independent but could be done.
I cannot spend too much time on that though and my build machine is being moved to another room right now.

alarcher:
@pyhalov erf... I screwed up clang-38's name ... I named it clang-3.8 ....

pyhalov:
Good to know, I'm not the only one... Time to fix it?

alarcher:
@pyhalov Done.

PR #2332, opened 2016-08-14
7293 Sun Secure Shell is neither

pyhalov:


pyhalov:
Should be merged after Sun SSH is nuked from illumos-gate

xen0l:
Seems reasonable.

PR #2335, opened 2016-08-15
Add illumos-closed component

xen0l:
This PR adds component "illumos-closed", which packages closed bits needed to build illumos-gate and derivates. It was also added to build-essential.


pyhalov:
1) If we add components/openindiana, will we move components/illumos/\* there?
2) Should we modify illumos-gate Makefile to use these bits?

xen0l:
1) I added openindiana subcategory, to start the move, so yes - we will move the rest there slowly (PR per component)
2) yes, next round of changes. Will update illumos-gate makefile in another commit

pyhalov:
Do we want to move it slowly or in one git mv? :)

PR #2337, opened 2016-08-16
Add libcdr 0.1.3

alarcher:


components/library/libcdr/Makefile
```diff
@@ -0,0 +1,60 @@
...
+COMPONENT_LICENSE_FILE= $(COMPONENT_NAME).license
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+CONFIGURE_OPTIONS+=	--enable-shared
+CONFIGURE_OPTIONS+=	--disable-static
+
+build: $(BUILD_32)
```
pyhalov, line 45: Why only 32-bit?

alarcher, line 45: Because all the other libraries from LibreOffice for dealing with file formats are 32bit only and there are no other consumers for them than LibreOffice and inkscape, as such there is no point in having 64bit.

xen0l:
LGTM

pyhalov:
I don't understand something... https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libcidr

PR #2364, opened 2016-08-25
openvpn: build with --enable-password-save

jeffpc:


pyhalov:
should have bumped COMPONENT_REVISION...

PR #2365, opened 2016-08-26
OpenSSH: salvage ssh-askpass from SunSSH so we have one; fix the download URL for tarball

jimklimov:
It was a sad surprise to update (and get new OpenSSH instead of SunSSH) and no longer see my `.xsession`'s `ssh-add` popup to enter the SSH key before I spread my tentacles to different machines. Now I grieve no more, the popup is back :)


components/network/openssh/network-ssh.p5m
```diff
@@ -43,6 +43,7 @@ file path=usr/bin/scp
 file path=usr/bin/sftp
 file path=usr/bin/ssh
 file path=usr/bin/ssh-add
+file files/ssh-askpass path=usr/lib/ssh/ssh-askpass mode=0555
```
pyhalov: please, move it between lines 40 and 42, so that all additional files were in one group.

components/network/openssh/files/ssh-askpass
```diff
@@ -0,0 +1,31 @@
...
+
+#
+# Copyright 2009 Sun Microsystems, Inc.  All rights reserved.
+# Use is subject to license terms.
+#
+
+prompt=$(echo $1 | sed s/_/__/g)
+ICON=/usr/share/pixmaps/blueprint-keyring.png
+exec /usr/bin/zenity --entry --title "ssh-askpass" \
+    --text="$prompt" --hide-text --window-icon=$ICON
```
pyhalov: Should it have some fallback when zenity is not found?

jimklimov: Hi, while there is merit in both proposals, one is cosmetic and another is
an improvement without which this variant of askpass lived for years as
part of SunSSH. (I tried to build a few others, including contribs that go
with openssh-portable, and the old x11-ssh-askpass from 2004 or so - they
didn't work on Hipster out of the box).

I suggest to solve problems one at a time, and for now - quickly and simply
restore the 'status quo' (return the useful and habitual functionality into
the world of living) by merging this PR if there are no other objections. I
am on vacation, so someone else (or myself much later) can tackle other
bits of evolution in separate PRs.

## Jim

26.08.2016 7:11 пользователь "Alexander Pyhalov" notifications@github.com
написал:

pyhalov: At least fix formatting

xen0l:
Why use ssh-askpass from SunSSH and not OpenSSH? SunSSH is dead, please deal with it. ssh-askpass is provided by OpenSSH and we should just stick to it.

components/network/openssh/files/ssh-askpass
jimklimov: This is a ditto copy from older illumos-gate - what's wrong with formatting
here? Make it single-line?

Note even this will take uncertain time until I get to a computer with
editor and git and internet all at once, and something that anyone could
fix in another PR ;)

## 

27.08.2016 7:57 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov: Additionally, changing the script against a well-known baseline state (from
old SunSSH likely available in same form in other repos for years) is
_reasonable_ to do in separate changesets so patches/diffs can be ported
elsewhere more easily.

## 

27.08.2016 8:05 пользователь "Jim Klimov" jimklimov@gmail.com написал:

jimklimov:
If you manage to build it - feel free to push a PR to replace it with a
functional equivalent - I'd merge it if I could. I just need something to
fill this new void ;)

The trick is that neither of those two contribs works, so you're welcome to
waste a bit more time than I had. One does not build, the other segfaults
in maybe gtk. I looked at others that also did not build or run, and turned
to 'maybe there's a shell or zenity script for this' - and voila, the first
google hit was the sunssh script which happens to still work with openssh
ssh-add ;)

Jim

## 

27.08.2016 8:02 пользователь "Adam Števko" notifications@github.com
написал:

xen0l:
Unfortunately, researching ssh-askpass problem is out of scope of my interest and options right now. This adds dependency to openssh on zenity and I don't think it is correct. Can we make the file deliver only if zenity is installed? @pyhalov What do you think?

pyhalov:
If it delivered in separate package, ssh can depend on it, if zenity is installed

jimklimov:
Makes sense, i'll look at it. Although it borders on 'ssh depends on x11 for any ssh-askpass gui naturally, do we want that?' sort of question.

However such separate packaging allows for different implementations to be provided over time, at users preference which one to install.

Is there a pkg(5) way to have a package installed if a set of prerequisites is present (e.g. we want this ssh-askpass if user has both ssh, x11 and zenity)? Or should this go via usual optional dependencies in e.g. ssh metapackage which gets e.g. a server, a client, and the optional tools and gui stuff, pam modules, etc?

pyhalov:
I'd say depend on it if user has zenity. Both mate_install and gui_install depend on it, so for desktop it will be installed.

components/network/ssh-askpass/Makefile
```diff
@@ -0,0 +1,42 @@
...
+include $(WS_MAKE_RULES)/ips.mk
+
+# common targets
+build:
+
+install:
+
+test:		$(NO_TESTS)
+
+REQUIRED_PACKAGES += gnome/zenity
```
pyhalov: We don't need one for bulding

components/network/ssh-askpass/network-ssh-askpass.p5m
```diff
@@ -0,0 +1,35 @@
+#
```
pyhalov, line 1: Do we really want two ssh-askpass packages?
Other issue: please, use new CDDL header in your files.

jimklimov: Hi, I'm back from vacation. Nice of you to roam through so many PRs while I was away ;p

As for several packages - it is an option which is possible, and may be attractive to some people. For example, if x11-ssh-askpass worked here (I didn't bother too much trying to make it build), it could be set up to securely block the X11 session until you enter the passphrase. I guess some people like that; for me the simple zenity requestor suffices. And extending the way to set up probable multiple packages later on, instead of preparing now, seems like bad foresight.

components/network/openssh/network-ssh.p5m
```diff
@@ -65,3 +65,5 @@ license openssh.license license="BSD, BSD-like (gsskex)" \
 
 depend type=conditional fmri=pkg:/x11/session/xauth \
     predicate=pkg:/x11/library/libxau
+
+depend type=optional fmri=pkg:/network/ssh-askpass
```
pyhalov: optional dependency on non-versioned package is no-op

jimklimov: ohs$#@ right, thanks :)
Hope to get to a coding laptop soon ;)

components/network/ssh-askpass/network-ssh-askpass.p5m
jimklimov: CDDL headers done

components/network/openssh/network-ssh.p5m
jimklimov: Hopefully the new choice of dependency will be better (conditional to get ssh-askpass "dispatcher" if an x-server is installed here)

components/network/ssh-askpass/Makefile
jimklimov: done

jimklimov:
Squashed to two commits - one keeps the initial state of the script's code (copypaste from opensolaris legacy, tracked to Darren Moffat's 2004 blog https://blogs.oracle.com/darren/entry/zenity_tools_a_simple_gui - same code without (C) headers) in a separate changeset, and subsequent improvement as another.

PR #2366, opened 2016-08-26
add developer/cscope

jeffpc:


pyhalov:
Conflicts with developer/sunstudio12u1/cc

jeffpc:
On Fri, Aug 26, 2016 at 03:37:26 -0700, Alexander Pyhalov wrote:

> Conflicts with developer/sunstudio12u1/cc

Hrm... do you have any suggestions how to handle this conflict?

pyhalov:
Is Studio cscope somehow used, for example, in illumos development?  If not (as I suppose), I'd removed usr/bin/cscope and usr/share/man/man1/cscope.1  links from Studio package.

jeffpc:
illumos-gate has support for generating a cscope database (see usr/src/tools/scripts/xref.mk), but as far as I can tell it's using the code that gets built from usr/src/tools/cscope-fast (and installed in /opt/onbld/bin/$(ARCH)/cscope-fast) instead of relying on the ancient binaries that come from studio.

pyhalov:
So you are welcome also to fix Studio in this PR :)

jeffpc:
On Mon, Aug 29, 2016 at 23:48:49 -0700, Alexander Pyhalov wrote:

> So you are welcome also to fix Studio in this PR :)

Do you have a preference: drop it by commenting out stuff in the .p5m or
by using a transform?

## 

All parts should go together without forcing.  You must remember that the
parts you are reassembling were disassembled by you.  Therefore, if you
can’t get them together again, there must be a reason.  By all means, do not
use a hammer.
        — IBM Manual, 1925

pyhalov:
drop by commenting it out in p5m file.

pyhalov:
I see 

```
In file included from /export/home/alp/srcs/tests/oi-userland/components/developer/cscope/cscope-15.8b/src/global.h:43:0,
                 from /export/home/alp/srcs/tests/oi-userland/components/developer/cscope/cscope-15.8b/src/vpopen.c:37:
../config.h:256:0: warning: "_FILE_OFFSET_BITS" redefined
 #define _FILE_OFFSET_BITS 64
```

Shouldn't we add CFLAGS+=-D_FILE_OFFSET_BITS=64

PR #2369, opened 2016-08-27
Ekiga 4

alarcher:


components/desktop/ekiga/Makefile
```diff
@@ -12,25 +12,33 @@
...
 include $(WS_TOP)/make-rules/ips.mk
 
 PATH=/usr/gnu/bin:/usr/bin:/usr/perl5/bin
 
+COMPONENT_PREP_ACTION = ( cd $(@D) && \
+          libtoolize --force && \
+          aclocal -I. && \
+          automake -a -f -c &&\
+          autoconf )
+
```
pyhalov: Please, drop depend.mk and add REQUIRED_PACKAGES

alarcher:
Tested with webcam and calls through voip provider.

components/ekiga/patches/ekiga-02-openldap-location.patch
```diff
@@ -1,51 +0,0 @@
---- ekiga-3.2.5/configure.ac.orig	2009-07-07 03:09:34.000000000 +0800
```
pyhalov, line 1: Does it detect openldap automagically now?

alarcher: LDAP was already disabled in the component so I did not test it.

pyhalov:
So,  can we merge it (with necessary libraries)?  Or somehow call for testing?

alarcher:
The last thing is I do not know what ekiga-06-t140.patch fixed and if it is any relevant.
With my use of the past days I see no regression so we could merge it and then people could test until October.

pyhalov:
Evidently, added T.140 support, https://en.wikipedia.org/wiki/Text_over_IP . Ekiga still misses it. But I don't know how widely it's used (just have read about it for the first time).

alarcher:
I should have worded it differently: I do not know how relevant it is but I can update the patch if you like.

pyhalov:
I don't care much about it, as I've just heard about it for the first time. If you update patch, you should somehow test it. Or find user of this feature on ML/IRC.

alarcher:
I looked at it and while most changes are easy, interfaces in opal have changed so it needs some work.
I would say merge it.

PR #2383, opened 2016-09-03
openjdk-8: fix PKCS11 (7227) https://www.illumos.org/issues/7227

fronald:
Following the correction of bug # 7227. OpenJDK 8 only has this behavior in Solaris and Illumos, why the Solaris Cryptographic Framework does not support SHA-256 hash. This Solaris Cryptographic Framework interface in Java in OpenIndiana was enabled by default, but the right is the opposite. Applications tested on other operating systems, by default use the PKCS11 interface Java itself, not native.

Please give me access to remove the bug triage status or enter me in the group "OI-userland" in Redmine. I will select another task to contribute.

This repository belongs to the company that I am co-founder. We will officially help develop OpenIndiana. The idea is to deploy on all servers of our new customers.

Please visit our website: http://www.wiselabs.com.br

### DIFF
--- a/components/runtime/openjdk-8/Makefile
+++ b/components/runtime/openjdk-8/Makefile
@@ -16,6 +16,7 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= oppenjdk-8
+COMPONENT_REVISION= 1
 
 OPENJDK_BUILD_NUMBER=92
 COMPONENT_VERSION=1.8.$(OPENJDK_BUILD_NUMBER)
--- a/components/runtime/openjdk-8/patches/patch-pkcs11.patch
+++ b/components/runtime/openjdk-8/patches/patch-pkcs11.patch
@@ -1,0 +1,22 @@
+$OpenIndiana: Fix https://www.illumos.org/issues/7227 - Franklin Ronald <franklin@wiselabs.com.br>$
+
+--- jdk/src/share/classes/sun/security/jca/ProviderConfig.java.1	2016-05-06 09:11:17.000000000 -0300
++++ jdk/src/share/classes/sun/security/jca/ProviderConfig.java		2016-09-02 15:32:30.826383860 -0300
+@@ -105,5 +105 @@
+-                File file = new File("/usr/lib/libpkcs11.so");
+-                if (file.exists() == false) {
+-                    return Boolean.FALSE;
+-                }
+-                if ("false".equalsIgnoreCase(System.getProperty
++                if ("true".equalsIgnoreCase(System.getProperty
+@@ -111 +107,4 @@
+-                    return Boolean.FALSE;
++                    File file = new File("/usr/lib/libpkcs11.so");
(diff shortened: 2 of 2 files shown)


xen0l:
@alarcher can you look at the Redmine thing?

xen0l:
@fronald what repository are you referring to?

xen0l:
@pyhalov change looks good to me. What do you think? 
@fronald please change the commit message, so it contains fixed problem and put bug issue in brackets after the message, e.g. "openjdk-8: fix PKCS11 (7227)"

pyhalov:
We can stick to illumos-gate commit message format here "BugID BugTitle". 
COMPONENT_REVISION should be bumped. The patch seems to be an acceptable workaround, but not a fix. I'm fine with it. However it would be interesting to find out what causes the original issue, why illumos libpkcs11.so can't be used by OpenJDK...

ptribble:
This doesn't look quite right to me. If you're going to disable pkcs11, then don't you simply remove it from the provider list in jre/lib/security/java.security?

Alternatively, modify sunpkcs11-solaris.cfg by taking the sunpkcs11-solaris.cfg file from openjdk7 (7u79, for example). Although you should be able to safely remove CKM_MD2 and CKM_SHA_1 from the disabledMechanisms list, as I think those are the 2 mechanisms that our libpkcs11.so does support.

fronald:
@ptribble I not disable pkcs11. The security tests passed and PKCS11 certificate in NetBeans is working. The JDK already have a pure Java solution thats works on  all operating systems including openindiana. Modify sunpkcs11-solaris.cfg was the first thing that i do, but the test (i post test project on bugs.openindiana.org) never pass. To make PKCS11 works native, i try change sunpkcs11-solaris.cfg pointing to opencryptoki, but, this native feature is only needed on extremely specific cases.

This issues are too reproducible on:
- NetBeans
- IntelliJ IDEA not complete licensing.

Resuming: I not disable security. I disabled the NATIVE OBSOLETE and NOT MORE NEEDED for OpenJDK the integration with Solaris Cryptographic Framework.

We can think about migrate from Solaris Cryptographic Framework to another compatible solution like OpenCryptoki. The Solaris Cryptographic Framework don't have any significant evolution since Sun Microsystems. Please read: http://www.oracle.com/technetwork/articles/javase/crypt-framework-149884.pdf

SHA-256 is really not supported.

fronald:
@xen0l I've bee changed commit description.
Thanks.

pyhalov:
Please, bump COMPONENT_REVISION and merge commits

fronald:
@pyhalov Sorry about my doubt. But, how i "bump COMPONENT_REVISION and merge commits"?

xen0l:
Add COMPONENT_REVISION= 1 to Makefile, so you trigger rebuild. And by merging commit he meant squahing them, so there is only one commit.

fronald:
Thanks @xen0l. All done now.

PR #2390, opened 2016-09-09
Add vobcopy 1.2.0

alarcher:
Contributed by Apostolos Syropoulos


components/encumbered/vobcopy/patches/Makefile
```diff
@@ -0,0 +1,59 @@
+
```
pyhalov: What is this?

components/encumbered/vobcopy/patches/vobcopy.patch
```diff
@@ -0,0 +1,144 @@
+Author: Apostolos Syropoulos <asyropoulos@yahoo.com>
```
pyhalov: Please, convert it to unified diff

alarcher: Apostolos updated the patch.

PR #2394, opened 2016-09-12
Add Time Slider support to caja

alarcher:
Adapted piggy-back of ZFS snapshot exploration to caja.

TODO:
- Check "Explore versions" functionality (DONE)
- Fix VFS init call in ZFS bar (WORKAROUND)
- Adapt old SUN localization and patch caja's to add time-slider specific messages (DROPPED for now).


pyhalov:
I like it... The issue I not iced: when you run "take snapshot" from snapshot view interface, you receive "Snapshot created successfully" and right after this "Failed to run /usr/lib/timeslider-snapshot..." Seems to be time-slider issue. Yes, https://github.com/OpenIndiana/time-slider/pull/3

pyhalov:
BTW, if we want to upstream it, there's an issue - zfs still doesn't have stable API. libzfscore should have solved this issue, but it's also not stable.

components/desktop/mate/caja/caja.p5m
```diff
@@ -247,11 +247,35 @@ file path=usr/share/man/man1/caja-autorun-software.1
 file path=usr/share/man/man1/caja-connect-server.1
 file path=usr/share/man/man1/caja-file-management-properties.1
 file path=usr/share/man/man1/caja.1
+#file path=usr/share/mime/XMLnamespaces
```
pyhalov, line 252: Why do you leave this commented files?

alarcher: They were not in the manifest but appeared when I generate it... no idea if they should be in there.

pyhalov:
Could you merge two diffs for caja-1.14.1/src/caja-window-manage-views.c in one?

pyhalov:
As for localization files, it's ungrateful and boring work - to find all places in c sources where these messages are used and renumber them in po-sun/*. I'd just dropped them. BTW, they plainly doesn't work now due to files renames.

alarcher:
The patch is updated.
Let us leave the *.po to make sure they are not lost but I commented the prep action in the Makefile.
I commented in the manifest about the mime files that are packaged elsewhere.

alarcher:
I removed a useless button in the preferences... I had added the button back but it is connected to nothing...

pyhalov:
One more time slider issue - https://github.com/OpenIndiana/time-slider/pull/4

PR #2412, opened 2016-09-16
mod_wsgi: update, drop python 2.6 version and add 3.4 version

pyhalov:


pyhalov:
Needs rebuilt python 3.4

PR #2414, opened 2016-09-16
Remove libbe

pyhalov:


pyhalov:
This should be merged with IPS changes, when illumos-gate will be switched to python 2.7

PR #2417, opened 2016-09-22
Move Python 2.6 to history

pyhalov:


pyhalov:
Firstly needs updated illumos-gate with https://www.illumos.org/issues/5969 integrated, updated IPS with dependencies on python 2.7 only

PR #2429, opened 2016-09-28
Add scandir to oi-userland

xen0l:


components/python/scandir/Makefile
```diff
@@ -0,0 +1,38 @@
...
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=	https://github.com/benhoyt/scandir/archive/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH= sha256:006ff89c427b4b8269515a476d2b08881bb2c587d18aebc6a85d6d574d01ae84	
+COMPONENT_LICENSE=	BSD
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/setup.py.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+PYTHON_VERSIONS= 2.7
```
pyhalov, line 32: Do we need it for Python 3.4?

components/python/scandir/Makefile
```diff
@@ -0,0 +1,38 @@
...
+# Copyright 2016 Adam Stevko
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		scandir
+COMPONENT_VERSION=	1.3
+COMPONENT_SUMMARY=	Better directory iterator and faster os.walk(), now in the Python 3.5 stdlib
+COMPONENT_PROJECT_URL=	https://github.com/benhoyt/scandir
+COMPONENT_FMRI=		library/python-2/scandir-27
```
pyhalov: We are changing python fmris to library/python/module-PYV

components/python/scandir/scandir-27.p5m
```diff
@@ -0,0 +1,29 @@
+#
```
pyhalov: Isn't it better to use PYVER manifests?

components/python/scandir/Makefile
xen0l: It is compatible with 3.4, but was made part of python 3.5, so I would ship it just for 2.7. Once python 3.5 is integrated, there really is no reason to use 3.4.

components/python/scandir/Makefile
xen0l: I'll change it

components/python/scandir/scandir-27.p5m
xen0l: I want to ship package for python 2.7 only

pyhalov: You can still do this with PYVER, just set PYTHON_VERSIONS in Makefile. But this allows in future easily add one more python version.

PR #2435, opened 2016-09-29
Update cmake to 3.5.2

AndWac:
This builds all our cmake dependent packages despite openal which also doesn't build in my env by our ancient cmake.


components/developer/cmake/Makefile
```diff
@@ -96,4 +102,4 @@ test:		$(TEST_32)
 
 BUILD_PKG_DEPENDENCIES =	$(BUILD_TOOLS)
 
-include $(WS_TOP)/make-rules/depend.mk
+include $(WS_MAKE_RULES)/depend.mk
```
pyhalov: Please, remove BUILD_PKG_DEPENDENCIES, depend.mk.
Add REQUIRED_PACKAGES (gmake REQUIRED_PACKAGES).

PR #2438, opened 2016-09-30
add librecad recipe

cgrzemba:
Please add librecad ripe to oi-userland.

I hope the other "noise" don't confuse you, but you can pull it too.

in userland.py I refined the license check because with the missing attribute in license action I had a problem on package uninstall.


components/desktop/librecad/Makefile
```diff
@@ -0,0 +1,114 @@
...
+COMPONENT_SUMMARY =	LibreCAD is a 2D CAD drawing tool based on the community edition of QCAD.
+COMPONENT_CLASSIFICATION = Applications/Graphics and Imaging
+
+COMPONENT_REVISION = $(shell cd $(COMPONENT_SRC); git rev-list HEAD --count)
+IPS_COMPONENT_VERSION = $(COMPONENT_VERSION)
+
+$(SOURCE_DIR)/.downloaded: $(ARCHIVES:%=$(USERLAND_ARCHIVES)%)
+	@echo "target .download"
+	[ -d $(SOURCE_DIR) ] || \
+	$(GIT) clone -b $(GIT_TAG) --single-branch --depth 1 $(GIT_REPO) $(SOURCE_DIR) && \
```
pyhalov: Why do we need to base our component on git head of LibreCAD and not on published tarball?

components/desktop/librecad/Makefile
```diff
@@ -0,0 +1,114 @@
...
+REQUIRED_PACKAGES += system/library/security/libgcrypt
+REQUIRED_PACKAGES += x11/library/libx11
+REQUIRED_PACKAGES += x11/library/libxi
+REQUIRED_PACKAGES += x11/library/libxtst
+REQUIRED_PACKAGES += library/boost
+REQUIRED_PACKAGES += library/muparser
+
+include ../../../make-rules/depend.mk
+
+BUILD_PKG_DEPENDENCIES += library/boost
```
pyhalov: depend.mk and BUILD_PKG_DEPENDENCIES are not needed. If packages are required to build librecad, they should be added to REQUIRED_PACKAGES

components/desktop/librecad/Makefile
```diff
@@ -0,0 +1,114 @@
...
+BUILD_PKG_DEPENDENCIES += library/muparser
+
+download:: $(SOURCE_DIR)/.downloaded
+
+show-var:
```
pyhalov: Remove this

components/desktop/nvidia/Makefile
```diff
@@ -0,0 +1,59 @@
+#
```
pyhalov: What is this doing here? Please, rebase your branch on oi/hipster HEAD

components/desktop/librecad/Makefile
```diff
@@ -0,0 +1,114 @@
...
+
+# publish: update-publish
+include ../../../make-rules/ips.mk
+
+GIT = git
+GIT_REPO = https://github.com/LibreCAD/LibreCAD.git
+GIT_TAG = $(COMPONENT_VERSION)
+GIT_BRANCH = master
+
+tolower = $(shell echo $(1) | tr '[A-Z]' '[a-z]')
```
pyhalov: Below snipplet looks too ugly... Perhaps, better to patch librecad's Makefile to do correct thing on gmake install?

components/desktop/librecad/Makefile
cgrzemba: download:: target is neccessary because we hav not tar ball

components/desktop/librecad/Makefile
cgrzemba: There isn't any.

components/desktop/librecad/Makefile
cgrzemba: Hm, 
Makefile will generated by qmake, so it does not exist at prep-patch time.

components/desktop/nvidia/Makefile
cgrzemba: Right, missplaced here, I remove.

components/desktop/librecad/Makefile
cgrzemba: I remove.

components/desktop/librecad/Makefile
xen0l: Github can generate tarball on the fly for everything (specific commit, tag, version, branch). You should switch to it as it will simplify the Makefile. For more examples, have a look in other components.

components/desktop/librecad/Makefile
pyhalov: I mean show-var

components/desktop/librecad/Makefile
cgrzemba: Can you point me to a example for this. I tried some URL's with no success.
Which I found for this is:
https://api.github.com/repos/LibreCAD/LibreCAD/tarball/2.1.3
But this gives me a file 2.1.3
Is there something like a COMPONENT_POST_DOWNLOAD_ACTION for rename the downloaded archive?

cgrzemba: found the solution:
COMPONENT_ARCHIVE =     $(COMPONENT_SRC).tar.gz
COMPONENT_ARCHIVE_HASH =        \
        sha256:5d0bc37efa1006742785c06ec30737c7866b0366048553c5560e06a3e442f7b3
COMPONENT_ARCHIVE_URL=  https://api.github.com/repos/$(COMPONENT_NAME)/$(COMPONENT_NAME)/tarball/$(COMPONENT_VERSION)
UNPACK_ARGS = --relocate-to=$(COMPONENT_SRC)

components/desktop/librecad/Makefile
cgrzemba: removed.

PR #2450, opened 2016-10-02
X 201610 Part 5: libvdpau

alarcher:
- Adding libvdpau in /usr/lib/xorg to avoid clash with Nvidia driver.
- Nvidia driver's symlinks in /usr/include are removed as libvdpau has priority over it.
- Removed glamor 0.6.0 from the PR.

### DIFF
--- a/components/openindiana/nvidia/driver-graphics-nvidia.p5m
+++ b/components/openindiana/nvidia/driver-graphics-nvidia.p5m
@@ -243,7 +243,6 @@ link path=usr/X11/lib/modules/NVIDIA/libnvidia-wfb.so \
 link path=usr/X11/lib/modules/extensions/NVIDIA/$(MACH64)/libglx.so \
     target=libglx.so.1
 link path=usr/X11/lib/modules/extensions/NVIDIA/libglx.so target=libglx.so.1
-link path=usr/include/vdpau target=../X11/include/NVIDIA/vdpau
 link path=usr/lib/$(MACH64)/libnvidia-glcore.so.1 \
     target=../../X11/lib/NVIDIA/$(MACH64)/libnvidia-glcore.so.1
 link path=usr/lib/$(MACH64)/libnvidia-tls.so target=libnvidia-tls.so.1
--- a/components/x11/libvdpau/Makefile
+++ b/components/x11/libvdpau/Makefile
@@ -1,0 +1,47 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2016 Aurelien Larcher
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		libvdpau
+COMPONENT_VERSION=	1.1.1
+COMPONENT_SUMMARY=	VDPAU libaries and headers
+COMPONENT_PROJECT_URL=	http://freedesktop.org/wiki/Software/VDPAU/
+COMPONENT_FMRI=		library/graphics/libvdpau
+COMPONENT_CLASSIFICATION=	System/Libraries
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.bz2
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:857a01932609225b9a3a5bf222b85e39b55c08787d0ad427dbd9ec033d58d736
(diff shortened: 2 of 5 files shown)


alarcher:
@pyhalov Since you already bumped nvidia , should we merge this PR without glamor or wait until Ken provides his fix?

pyhalov:
if you think that it's fine

pyhalov:
@alarcher What's the current status of this PR? What was/is wrong with glamor (sorry, don't remember)?

alarcher:
@pyhalov Requires patching one line in glamor but Ken did not send it. We can remove glamor for now and merge libvdpau. Tell me if I should change anything in the libvdpau commits. I will remove my Nvidia bump from the PR.

pyhalov:
Why do we need glamor? What are known consumers?

alarcher:
@pyhalov xorg-video-ati and xorg-video-intel can use it to accelerate 2D rendering.

pyhalov:
Note, that if you integrate it before https://github.com/OpenIndiana/oi-userland/pull/2669 , we can remove ffmpeg dependency on nvidia

pyhalov:
BTW, can we remove nvidia's vdpau and leave only this one? Are they compatible?

tomww:
Nvidia writes in a release doc that vdpau will be removed from the driver package.
While this version looks like still including the  wrapper library:
http://us.download.nvidia.com/solaris/352.79/README/vdpausupport.html
Then this version writes aout the external available wrapper library:
http://us.download.nvidia.com/solaris/375.26/README/vdpausupport.html
(not necessarily at the version in the URL above removed the vdpau wrapper lib)

Which version of the Nvidia driver will be in the repo? Depending on this version we need the external wrapper library jumping in.

If OI should include an older driver with the wrapper, then it could be an idea to remove vdpau files from the OI delivered nvidia package at the time when the new vdpau wrapper goes in. Including a new required dependency in nvida driver would be necessary.
If users install their own copy of the nvidia driver (some need to for older Chips), then they would need to install the OI -vdpau package on their own. I would not make a dependency in the consuming packages like vlc.

alarcher:
@pyhalov @tomww As a first step we should keep libvdpau with 34x drivers and mediate the symlinks in the common directories. The NVIDIA libraries do not bother anyone as they live in their own directory, only the symlinks are to blame.
The new libvdpau is already installed in another prefix together with other Xorg libs.
We already discussed that we want to leave the possibility to install 34x even after 36x because fairly common Quadro adapters are not supported by 36x

components/openindiana/nvidia/driver-graphics-nvidia.p5m
```diff
@@ -243,7 +243,6 @@ link path=usr/X11/lib/modules/NVIDIA/libnvidia-wfb.so \
 link path=usr/X11/lib/modules/extensions/NVIDIA/$(MACH64)/libglx.so \
     target=libglx.so.1
 link path=usr/X11/lib/modules/extensions/NVIDIA/libglx.so target=libglx.so.1
-link path=usr/include/vdpau target=../X11/include/NVIDIA/vdpau
```
pyhalov, line 246: Please, bump COMPONENT_REVISION for nvidia  component

components/x11/libvdpau/Makefile
```diff
@@ -0,0 +1,47 @@
...
+CONFIGURE_LIBDIR.64 = /usr/lib/xorg/$(MACH64)
+
+CONFIGURE_OPTIONS+= --sysconfdir=$(ETCDIR)
+
+build:		$(BUILD_32_and_64)
+
+install:	$(INSTALL_32_and_64)
+
+test:		$(NO_TESTS)
+
```
pyhalov, line 47: REQUIRED_PACKAGES?

pyhalov (changes requested):

pyhalov:
You spoke about mediated symlinks. Are you going to add them in this PR or later?

alarcher:
I am working on the links for this PR to be able to merge before the ffmpeg PR as you suggested earlier.

pyhalov:
@alarcher : what's the current status of this PR?

components/x11/libvdpau/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,44 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=etc/vdpau_wrapper.cfg
+file path=usr/etc/vdpau_wrapper.cfg
```
pyhalov, line 26: usr/etc/ ? Is something misconfigured?

pyhalov:
Superseded by https://github.com/OpenIndiana/oi-userland/pull/2732

PR #2456, opened 2016-10-04
top: fix wrong cpustats and kernel stats

pyhalov:


pyhalov:
@alarcher , can you try this? Don't know if it's the best fix, but it seems to work for me.
Another question: with 32-bit top I have 

```
ARC:    -1251721K Total, -1846191K MRU, 453M MFU, 1716K Anon, 8696K Header, 118M Other
```

Seems to be overflow of long. Perhaps, it's time to ship only 64-bit version?
Or do counters long long?

PR #2465, opened 2016-10-05
ibus-anthy: update to 1.2.7, switch to python 2.7

pyhalov:


pyhalov:
depends on ibus python 2.7 modules

PR #2467, opened 2016-10-05
Add OI loader branding

pyhalov:


tsoome (approved):
seems good, I trust you have tested on the screen;)

xen0l (approved):

minikola:
Distribution name is Openindiana and not hipster.
It could be good to see on screen it is hipster rolling-release development Openindiana,
but I don't think it is appropriate to have "hipster" on screen instead of "Openindiana" or "Openindiana hipster", unless you are planning registering hipster.org or something.
Leaving hipster there instead of Openindiana would lead to confusion about illumos distribution name.

PR #2469, opened 2016-10-06
zsvcadm : initial import of the script and packaging recipes for it

jimklimov:
Another nice pearl from Darren Moffat - a GUI for svcadm.


alarcher (changes requested):
Are you able to add the COMPONENT_\* variables in the Makefile instead of specifying metadata in the manifest?
Just to comply with the usual style.

PR #2472, opened 2016-10-06
Thunderbird: update to 45.3.0

pyhalov:


pyhalov:
I have to recheck REQUIRED_PACKAGES

pyhalov:
Also have to recompile with -O2...

PR #2474, opened 2016-10-08
Libjpeg macros

alarcher:
We kept IJG's libjpeg6 as default for backward compatibility but libjpeg-turbo is the default of major Linux distributions.
I did some simple testing [1] on images linked from [2] which shows a minimum x2.7 decompression performance (about ~x3 average).
The PR is to add a bunch of macros for selecting the default libjpeg for compilation.

Ref: https://www.illumos.org/issues/7391

[1] https://github.com/gkbrk/ImageBench
[2] http://www.libjpeg-turbo.org/About/Performance


xen0l (approved):

xen0l (changes requested):
Bump COMPONENT_REVISION, so this gets rebuilt when picked up by the build server

alarcher:
Done.

PR #2449, opened 2016-10-02
X 201610 Part 6: Xorg 1.18.4 and TurboVNC

alarcher:
NOTE: #2451 and #2971 should be merged immediately after publication of this PR on the build server.

PR #2475, opened 2016-10-08
Bump libXfont to 1.5.2

alarcher:
Requires #2449


alarcher:
@pyhalov This guy can be merged as well.

pyhalov:
Will it break ABI?

PR #2477, opened 2016-10-08
FIX: top broken on amd64.

fronald:
Fixed the bug on amd64 build of **top**.
The error was caused by me. I left a residue of a test I did. Apologies for that. It will not be repeated.


xen0l:
I reverted this fix because it was fixed by @pyhalov.

fronald:
Ok. Thanks.

PR #2492, opened 2016-10-12
Move shmux from SFW to oi-userland (Predrag Zečević)

alarcher:
Contributed by Predrag Zečević.

NOTE: /usr/share/shmux/mcmd.sh is provided as example and should not be executable.

### DIFF
--- a/components/shell/shmux/shmux.p5m
+++ b/components/shell/shmux/shmux.p5m
@@ -1,0 +1,31 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2016 Predrag Zečević
+#
+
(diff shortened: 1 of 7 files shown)


components/shell/shmux/shmux.p5m
```diff
@@ -0,0 +1,30 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/shmux
+file path=usr/bin/shmux
+file path=usr/share/man/man1/shmux.1
+file path=usr/share/shmux/mcmd.sh
+# Optional?
```
xen0l: is fping needed for proper functioning / building?

alarcher: From what I understood, I do not think so but I do not know whether it hinders any important feature.
I sent a message to Predrag.

3eka: Well, 
according to http://web.taranis.org/shmux/:
"To take full advantage of shmux you will need to have fping installed."

I would say, yes make it required. 
But as you have noticed, it can be also optional (it is used to detect if node is alive and is much faster/better then normal ping, IMHO)

Regards.

alarcher: Thanks, I updated the comment.

xen0l (approved):

PR #2493, opened 2016-10-13
userland.py: extend license check action

cgrzemba:
This extends pkglint license action check for presence of license attribute. With a license action where license attribute is missing, I got a error on 'pkg **un**install'.


tools/python/pkglint/userland.py
```diff
@@ -492,9 +496,9 @@ def component_check(self, manifest, engine, pkglint_id="001"):
 #			engine.error( _("missing ARC data (org.opensolaris.arc-caseid)"),
 #				msgid="%s%s.0" % (self.name, pkglint_id))
 
-	component_check.pkglint_dest = _(
+	component_check.pkglint_desc = _(
 		"license actions and ARC information are required if you deliver files.")
-
+
```
alarcher: Nitpicking but can you remove this space?

alarcher (commented):
A small git pull --rebase and it seems we are good to go.

tools/python/pkglint/userland.py
cgrzemba: Done.

PR #2497, opened 2016-10-16
OI-Docs: Add openindiana-docs

alarcher:


components/openindiana/docs/Makefile
```diff
@@ -0,0 +1,69 @@
...
+    $(CP) -r site/* $(PROTO_DIR)$(OI_DOCS_DIR) && \
+    $(CP) openindiana-docs.desktop $(PROTO_DIR)/usr/share/applications;
+
+build: $(BUILD_32)
+
+install: $(INSTALL_32)
+
+test: $(NO_TESTS)
+
+$(PRE_PUBLISH_STAMP): $(GENERATED).p5m
```
pyhalov, line 67: I dislike this automation... This means repository p5m file will be changed on next rebuild... Can we checkout the exact revision, so that we don't have to modify p5m file? Or move p5m file to oi-docs?

components/openindiana/docs/Makefile
```diff
@@ -0,0 +1,69 @@
...
+COMPONENT_PROJECT_URL= https://github.com/OpenIndiana/oi-docs
+COMPONENT_ARCHIVE_URL= https://github.com/OpenIndiana/oi-docs
+COMPONENT_SRC=      $(COMPONENT_NAME)
+COMPONENT_LICENSE= PDL
+
+GIT_REPO=git://github.com/OpenIndiana/oi-docs.git
+GIT_BRANCH=master
+GIT_CHANGESET=HEAD
+
+include prep-git.mk
```
pyhalov, line 30: Well, I see two ways: 1) make prep-git.mk common for several components which rely on fetching sources from git/github, 2) make this component more like existing components which already checkout sources from github. Perhaps, we even don't want to checkout latest sources, just exact revision (and just can fetch it as usual archive).

components/openindiana/docs/Makefile
alarcher: Then we have a problem ;) This was exactly my intent to be able to republish fresh documentation in a similar fashion as illlumos-gate.

components/openindiana/docs/Makefile
alarcher: Yes, I wanted to raise the question if we want to have a file with a generic git recipe.
I think this would be useful but it should be improved.

makruger:
This PR might well satisfy (and exceed) the spirit of https://www.illumos.org/issues/7333

minikola:
Including any documentation without versions, releases and review, other that before inclusion, where it is not expected of OI maintainers to read every line, is somewhat troubling as an idea.
Documentation is not the code.
As I see there is no way to review it before inclusion. Publishing in a form of release archive that gets reviewed and refreshed from time to time, sounds much more pleasant also for maintaining (what's new in between releases etc). It is also a separate project.

makruger:
@minikola 

There have been QA reviews and changes resulting from them. However, I agree having more people engaged in the review process (as well as the authoring process) would be helpful.

While I agree, the documentation should match the current state of OI, I am not sure versioned releases makes a whole lot of sense at this point. After all, the distribution is a rolling release with periodic snapshots. So, the documentation would also follow this model.

minikola:
Documentation can't be rolling release because it is in the nature of documentation of documenting release, that is used in production, is released and needs documenting to use.

In all good will I have not see any QA review nor even read single line of those documents in available form for editing. 
Aether reviews are done somewhere I don't know by someone I don't know or they are just rolled out as is?
I don't think that updated content should be ever be updated directly as it is edited, it should go through review that culminates with releasing versioned archive, as suggested, to be manageable by Openiindiana hipster. 
Wiki could match current state of OI hipster, being easier to edit and contribute,
while docs could be periodically reviewed, grown, versioned and published.

alarcher:
@minikola @makruger This is a place for discussion about a _component_ for fetching text files, calling mkdocs and publishing not for discussing _documentation_ _policy_. This is off-topic here.

components/openindiana/docs/Makefile
pyhalov: If we consider to ship precise documentation version, it's fine to have fixed p5m file. Perhaps, it's fine. Otherwise, we have to find a way to generate manifest without touching manifest in git repository on each gmake clean publish

xen0l: We don't have a fixed documentation as we don't have a release policy for that in place. The idea is to have most up to date bits available in the rolling release scenario. As for finding a way without touching a manifest in git - @alarcher already did that. What practical problems are we talking about here?

pyhalov: I see the following problem - when documentation is changed and new pages are added or reorganized, they should be included in p5m file. As documentation is a very dynamic content, it is likely to be broken in this respect.

PR #2500, opened 2016-10-17
add smp_utils version 0.98

grueni:
Please check the handling of the license. I copied the type of the license from smp_utils.spec and the content from COPYING.


components/sysutils/smp_utils/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+PKG_PROTO_DIRS += $(COMPONENT_DIR)/files
+
+build:		$(BUILD_32)
+
+install:	$(INSTALL_32)
+
+test:	 $(TEST_32)	
+
+REQUIRED_PACKAGES += system/library
```
xen0l, line 47: did you run gmake REQUIRED_PACKAGES, or you added this just by hand?

grueni: I run make REQUIRED_PACKAGES.

components/sysutils/smp_utils/Makefile
```diff
@@ -0,0 +1,54 @@
...
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CFLAGS += -I$(COMPONENT_DIR)/$(COMPONENT_SRC)/include
+
+CONFIGURE_OPTIONS += CFLAGS="$(CFLAGS)"
+CONFIGURE_OPTIONS += --bindir ${CONFIGURE_SBINDIR.32}
+
+# Move the section 8 man pages into section 1m.
+COMPONENT_POST_INSTALL_ACTION = \
```
pyhalov: This looks ugly. If component is automake/autoconf-based, it can be rather simple to patch Makefile.am

alarcher: Or just use a transform.

grueni: This is a copy of the same proceeding in sg3_utils. Same author, same attitude.
configure has no options to define the target for the man files.
In folder doc I find the files with extension ".8" and the Makefile.am using this files.
Shall I rename *.8 to *.1 and change the content of Makefile.am?
I fear this won't be look nicer.

grueni: transform sounds interesting, will try it, give me some time

grueni:
Squash the 2 commits in one?

PR #2524, opened 2016-10-21
GPG signature verification support

xen0l:
This PR adds support for verifying GPG signatures for component downloads. This increases security. I did following changes:
- made userland-fetch PEP8 compliant and code should be more readable
- ported the functionality from userland-gate and adapted to oi-userland
- added COMPONENT_SIG_URL for some components

Workflow is rather simple:
- add COMPONENT_SIG_URL and run gmake download. Usually, these suffixes are standard:
  **asc**, **sign** or **sig**.
- check if corresponding public key has been imported into pubring.gpg:

```
gpg2 --homedir=$(git rev-parse --show-toplevel)/tools/.gnupg --list-keys
```
- commit Makefile change and $(git rev-parse --show-toplevel)/tools/.gnupg/pubring.gpg

With this functionality it is possible to omit component tarball checksum and use GPG verification for downloaded component. We should add it where possible once merged.


tools/userland-fetch
```diff
@@ -28,246 +28,371 @@
...
+    # configuration and keyring.
+    proc = subprocess.Popen(["git", "rev-parse", "--show-toplevel"], stdout=subprocess.PIPE)
+    proc.wait()
+    if proc.returncode != 0:
+        return False
+    out, err = proc.communicate()
+    gpgdir = os.path.join(out.strip(), "tools", ".gnupg")
+
+    # Skip the permissions warning: none of the information here is private,
+    # so not having to worry about getting mercurial keeping the directory
```
pyhalov: s/mercurial/git/ :)

xen0l: Fixed.

tools/userland-fetch
```diff
@@ -28,246 +28,371 @@
...
+import subprocess
+import re
+import gzip
+import bz2
 from urllib import splittype
 from urllib2 import urlopen
 from urllib2 import Request
 import hashlib
 
+
```
pyhalov, line 44: please, clean up this new line

tools/userland-fetch
```diff
@@ -28,246 +28,371 @@
...
+        yield name
 
-	return urls
 
 def usage():
-	print "Usage: %s [-a|--user-agent (user-agent)] [-f|--file (file)] " \
-		"[-l|--link] [-h|--hash (hash)] [-s|--search (search-dir)] " \
-		"-u|--url (url)" % (sys.argv[0].split('/')[-1])
-	sys.exit(1)
+    print "Usage: %s [-f|--file (file)] [-l|--link] [-h|--hash (hash)] " \
```
pyhalov: where -a has gone?

xen0l: Oops, I will add it back.

tools/userland-fetch
xen0l: This is PEP8 specific, so I would rather leave it there. (PEP8 mandates to use two blank lines between imports and function/class definitions).

tools/userland-fetch
```diff
@@ -28,246 +28,371 @@
...
+                # the target file if the signature was valid.
+                if not sig_valid:
+                    try:
+                        os.remove(name)
+                    except OSError:
+                        pass
+
+                    continue
+
+        sys.exit(0)
```
pyhalov: what if there's no signature and hash invalid? Will this line be reached (hard to follow intendation :)

tools/userland-fetch
pyhalov: OK

PR #2526, opened 2016-10-21
Bump freetype-2 to 2.7

alarcher:
Implementation was changed in 2.6.4 according to
https://sourceforge.net/projects/freetype/files/freetype2/2.6.4/
but compatibility seems to be retained.


alarcher:
Tested.

PR #2532, opened 2016-10-22
Fix libxcb component

alarcher:


components/x11/libxcb/Makefile
```diff
@@ -22,38 +22,60 @@
...
 test: $(NO_TESTS)
 
-BUILD_PKG_DEPENDENCIES =    $(BUILD_TOOLS)
-
-include $(WS_TOP)/make-rules/depend.mk
-
-
 REQUIRED_PACKAGES += system/library
 REQUIRED_PACKAGES += x11/library/libxau
 REQUIRED_PACKAGES += x11/library/libxdmcp
```
pyhalov, line 80: Does REQUIIRED_PACKAGE change after enabling all these options?

alarcher: No, it does not change.

PR #2536, opened 2016-10-26
Add python-flamegraph to oi-userland

xen0l:


components/python/flamegraph/Makefile
```diff
@@ -0,0 +1,38 @@
+# This file and its contents are supplied under the terms of the
```
pyhalov, line 2: Missed first '#' in header

xen0l: Fixed.

PR #2537, opened 2016-10-26
Add abi-compliance-checker 1.99.25+

pyhalov:


pyhalov:
As far as I understand, it works correctly only when it can compile all library headers. To do it, it needs specific configuration.

pyhalov:
it's not so useful without possibility to examine binaries... For this it needs abi-dumper. And abi-dumper depends on elfutils, which AFAIK were never ported to Solaris. So, until someone wants to look at it more attentively, closing it.

PR #2541, opened 2016-10-27
OpenVPN bump to 2.3.12 and improved support for multi-instance setups

jimklimov:


components/network/openvpn/files/openvpn.xml
```diff
@@ -1,8 +1,69 @@
 <?xml version="1.0" ?>
 <!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
+
```
xen0l: Move these bits to other location, e.g. /usr/share/doc/openvpn/README.openindiana and just reference that file here.

components/network/openvpn/files/openvpn.xml
```diff
@@ -1,8 +1,69 @@
...
+You might also want to enable some permissions in the service manifest below,
+or per-instance, if e.g. you want one to listen on privileged ports, or to
+chroot(), or to use configuration files and device nodes owned by another user.
+Remember that under SMF a "root" is not almighty! For more details, see the
+commented "method_credential" suggestions below.
+-->
 <service_bundle type="manifest" name="openvpn">
-    <service name="network/openvpn" type="service" version="0">
-	<create_default_instance enabled="false"/>
+    <service name="network/openvpn" type="service" version="0.01">
```
xen0l: Use 1, instead of 0.01

components/network/openvpn/files/openvpn.xml
```diff
@@ -12,30 +14,51 @@
             <service_fmri value="svc:/system/filesystem/local"/>
         </dependency>
 
-        <method_context>
+        <method_context project=':default' resource_pool=':default' working_directory='/etc/openvpn'>
```
xen0l: project and resource_pool are something we shouldn't user unless we have support for it in IPS. As for working directory, won't it cause the issue, if people use different location for config file?

components/network/openvpn/files/openvpn.xml
```diff
@@ -1,8 +1,10 @@
 <?xml version="1.0" ?>
 <!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
+<!-- See the non-trivial-setup notes and recommendations for the OpenVPN SMF
+     service integration in /usr/share/doc/openvpn/README.openindiana -->
 <service_bundle type="manifest" name="openvpn">
-    <service name="network/openvpn" type="service" version="0">
-	<create_default_instance enabled="false"/>
+    <service name="network/openvpn" type="service" version="0.01">
```
xen0l: 0.01 -> 1

components/network/openvpn/openvpn.p5m
```diff
@@ -25,6 +25,7 @@ license openvpn.license license='GPLv2'
 file files/auth_openvpn path=etc/security/auth_attr.d/openvpn
 file files/prof_openvpn path=etc/security/prof_attr.d/openvpn
 file files/openvpn.xml path=lib/svc/manifest/network/openvpn.xml restart_fmri=svc:/system/manifest-import:default
+file files/README.openindiana path=usr/share/doc/openvpn/README.openindiana
 
 dir  path=etc
```
xen0l: Can you add dir action for /var/log/openvpn and optionally remove other dir actions, so p5m file is consistent with others?

components/network/openvpn/files/openvpn.xml
```diff
@@ -12,30 +14,51 @@
...
 
-	<property_group name='config' type='application'>
-	    <propval name="config_file" type='astring' value='/etc/openvpn/openvpn.conf'/>
-	    <propval name='value_authorization' type='astring' value='solaris.smf.value.openvpn'/>	
-	</property_group>
+        <property_group name='config' type='application'>
+            <!-- Full path to config file is required -->
+            <propval name="config_file" type='astring' value='/etc/openvpn/openvpn.conf'/>
+            <!-- Note that the log location can be overridden inside the config file -->
+            <propval name="default_logfile" type='astring' value='/dev/stderr'/>
```
xen0l: What about using /var/log/openvpn/%i.log instead of /dev/stderr?

jimklimov: not in the variable at least - so the string with macro will get hardcoded to exec line instead.. but doable...

jimklimov: not in the variable at least - so the string with macro will get hardcoded to exec lien instead.. but doable...

components/network/openvpn/files/openvpn.xml
jimklimov: The workdir should not influence an explicitly specified "atomic" config file, and should help with files referenced from it by relative locations - if they are all in /etc/openvpn per defaults. Although with the other bit (--cd) - the command-line option wins.

pyhalov:
@xenol, @jimklimov  - what's is the current status of this PR?

jimklimov:
Never an end to perfection there is, but interruptions there may be :-)
Other ideas = other PRs? ;)

PR #2549, opened 2016-10-28
drm: don't try to attach to unsupported adapters

pyhalov:


pyhalov:
Need to gather more complete list of unsupported hardware

alarcher:
@pyhalov @gwr I started listing chipsets and will update information on current i915 kernel driver support:
https://wiki.openindiana.org/oi/Intel+KMS+driver

alarcher:
@pyhalov Could you list the issue when removing an id so that we understand what should be fixed? e.g. your Ironlake should be supported.

PR #2554, opened 2016-10-29
configure.mk : extend support for common autoconf "configure" script args

jimklimov:
- optionally update PROTO*DIR definitions where applicable
- shared-macros.mk : added comment about relation to configure.mk

This PR is splintered from #1846 as a smaller, better digestible, piece ;)

Note that after this one is merged, the optimizations from https://github.com/OpenIndiana/oi-userland/pull/1846/commits/a31693a70c88165ca825898469c3a3b83d9a8151 and https://github.com/OpenIndiana/oi-userland/pull/1846/commits/06bcc2c82b53dae6d4c1d7e23038dda85418b8f3 should be next on the line

### DIFF
--- a/make-rules/configure.mk
+++ b/make-rules/configure.mk
@@ -54,21 +54,97 @@
 #	COMPONENT_BUILD_TARGETS, COMPONENT_INSTALL_TARGETS
 #	COMPONENT_TEST_TARGETS
 #
+# Common values for USRDIR BINDIR BINDIR LIBDIR MACH64 and so on are defined
+# in shared-macros.mk - so it should be included in your recipe too.
+#
 
-CONFIGURE_PREFIX =	/usr
+CONFIGURE_PREFIX =	$(USRDIR)
 
-CONFIGURE_BINDIR.32 =	$(CONFIGURE_PREFIX)/bin
-CONFIGURE_BINDIR.64 =	$(CONFIGURE_PREFIX)/bin/$(MACH64)
-CONFIGURE_LIBDIR.32 =	$(CONFIGURE_PREFIX)/lib
-CONFIGURE_LIBDIR.64 =	$(CONFIGURE_PREFIX)/lib/$(MACH64)
-CONFIGURE_SBINDIR.32 =	$(CONFIGURE_PREFIX)/sbin
-CONFIGURE_SBINDIR.64 =	$(CONFIGURE_PREFIX)/sbin/$(MACH64)
-CONFIGURE_MANDIR =	$(CONFIGURE_PREFIX)/share/man
-CONFIGURE_LOCALEDIR =	$(CONFIGURE_PREFIX)/share/locale
+# Note: definitions below specifically rely on $(CONFIGURE_PREFIX) rather
+# than e.g. $(USRBINDIR64) in order for certain components to be able to
+# deliver their whole trees into dedicated subdirectories for each version
+# (databases, compilers, etc.) by just redefining CONFIGURE_PREFIX once in
+# their recipe.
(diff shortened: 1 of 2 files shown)


pyhalov:
I'm a bit scared by this change. I'd have no questions if all additional options were turned off. But when they are on, we need mass rebuild to test what can possibly go wrong.

jimklimov:
The point of this change is to allow making leaf Makefiles shorter by not repeating the same lines over and over (advantage of this is taken in PR #2555), as well as having less surprise when making new recipes (e.g. "why can't it assume sysconfdir=/etc as it is default in 99% cases?") - and enforcing some same distro standard along the way.

So yes, these by design are enabled by default (and since there are recipes like MPI which deliver completely independent directory trees, there are toggles to turn it off for certain cases).

I am not fully sure indeed how this change alone (without the fixed Makefiles) would or would not pass the builds - given that many Makefiles would define same arguments twice, although most often setting them to same values. In this regard, #2555 is more consistent to merge as one atomic unit - with fixes on both sides of the equation. But someone on IRC was afraid of large changes - so the old assumed-consistent PR was split into several smaller-step PRs with no idea if they are as consistent ;)

In either case, assumptions are one thing, practice is another - whether you merge the recipe-infrastructure changes like this via one PR or by 10 PRs, you'd have to recompile (most of) the gate. Eggs must be broken to make an omelette.

Of course, for the peace of mind we can amend this PR into disabling these changes by default (and enabling for PR #2555 along with the fixed Makefiles) - but then there is little point in it: adding changes... off by default... how do we know if they work or explode? :)

pyhalov:
@xenol: Where can we test mass rebuild???

jimklimov:
@xen0l @pyhalov : "where" - that is the difficult part :-)

Technically, I have a new laptop now for work so the weaker old one is unused at the moment - but it has OI/Hipster set up. I also have an OmniOS Bloody box with an OI Hipster zone that "should" suffice for userland stuff.

These boxes might now be assigned to churn for days or whatever it takes with little impact on interactive humans, to see if the PR is ok (and then iterate some) ;)

How much disk space does a "rebuild world" take, and what would the procedure be - spawn a zone, git clone oi-userland, make setup + make env-prep + make all from workspace root?

Alternately, do EC resources allow such experimental things? Or do we only always build the production master branch there?

jimklimov:
>  I'd have no questions if all additional options were turned off

Regarding this point, last night I wondered if the PR #2555 could be amended with a commit (later to be reverted, so a standalone unsquashed commit) to indeed disable the new magic by default and explicitly enable it in roughly 1400 Makefiles touched by that PR to use the new feature. This would be "safer" regarding other Makefiles that are not touched (did not estimate how many or few those are), but it feels like an effort of several hours of semimanual work done to be essentially wasted time in the end. So technically it seems possible, but I'm reluctant to go this way.

A mass rebuild (or probably several, as statistics tells us to expect), iron out the wrinkles, and merge when it's ok sounds a lot better. But then it should get merged to get it into production quickly - re-merging with changing upstream, especially as components move around in classification, takes again hours of manual revision.

But as far as incrementals go, this PR #2554 can be amended to define just the new variables and disabled-by-default new magic (so at least the vars can be used from this point), and PR #2555 would reenable the magic and change hordes of Makefiles.

jimklimov:
Update: I went the way detailed above - PR #2554 does not enable the new magic, and PR #2555 includes and reverts this commit to do enable the magic by default and amend many Makefiles.

Of course, by iterative Git logic this makes the small added commit and its reversion unsquashable so we don't go mad with this ;)

PR #2555, opened 2016-10-29
Makefiles which consume the (improved) configure.mk - revised

jimklimov:
Splintered from PR #1846 and building on PR #2554 - this commit simplifies existing Makefile's to take advantage of the smarter `configure.mk` which now sets the common arguments.


jimklimov:
As detailed in PR #2554, as of now it does not enable the new magic, and this PR #2555 includes and then reverts the added commit to do enable the magic by default and amend many Makefiles.

Of course, by iterative Git logic this makes the small added commit and its reversion unsquashable so we don't go mad with this ;)

PR #2561, opened 2016-10-30
Initial commit of tools/mfstyle script

jimklimov:
Automation of machine-friendly routines from PR #1846 (implying that there were manual revisions to many makefiles' markup as well).


jimklimov:
@xen0l , @pyhalov : bump? At least, this PR alone should not break anything - and it is a foundation for some other works in the queue...

PR #2564, opened 2016-10-30
Initial commit of sysstat version 20151012

jimklimov:
Another useful tool from the author of mbuffer

Note : `sysstat` includes a daemon mode (`sysstatd`) which is currently provided "as is" but without an SMF manifest - this piece is left for another PR if anyone wants to figure out how and why to run it at all, what permissions are needed, etc. ;)

Assumption from glancing over docs is is that multiple hosts may run their daemons and an admin workstation can collect their data over LAN (broadcasts) - this implies privacy, security, firewall and other concerns that are a lot of research out of scope for initial recipe which just adds the utility.

### DIFF
--- a/components/sysutils/sysstat/Makefile
+++ b/components/sysutils/sysstat/Makefile
@@ -1,0 +1,54 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016, Jim Klimov
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		sysstat
+COMPONENT_VERSION=	20151012
+COMPONENT_FMRI=		diagnostic/sysstat
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_SUMMARY=	sysstat - tool for quick overview of various Solaris stats
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tgz
+COMPONENT_ARCHIVE_HASH=	sha256:4ec664956b557f7c3a01eb644f731314f976e6ff6b83afe55c2d6e8f73ae0736
+COMPONENT_ARCHIVE_URL=	http://www.maier-komor.de/sysstat/$(COMPONENT_ARCHIVE)
+COMPONENT_PROJECT_URL=	http://www.maier-komor.de/sysstat.html
+COMPONENT_LICENSE=	GPLv3
+COMPONENT_LICENSE_FILE=	LICENSE
+# Note: the website says the project is under GPLv2, but sources provide a
+# copy of GPLv3 explicitly. Probably the page was not updated in this regard.
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+# WARNING: non-autoconf "configure" script here
+COMPONENT_PRE_BUILD_ACTION = \
+	( cd $(@D) && \
+		./configure \
+			-c "$(CC)" \
+			-m "$(BITS)" \
+			-P "$(MACH$(BITS))" \
+			-K "$(MACH$(BITS))" \
+			--prefix "$(PROTOUSRDIR)" \
+	)
+
+build:		$(BUILD_32_and_64)
+
+install:	$(INSTALL_32_and_64)
+
+REQUIRED_PACKAGES += library/ncurses
+REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/math
--- a/components/sysutils/sysstat/sysstat.p5m
+++ b/components/sysutils/sysstat/sysstat.p5m
@@ -1,0 +1,46 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016, Jim Klimov
+#
+
+<transform file path=usr.*/man/.+ -> default mangler.man.stability volatile>
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value="$(COMPONENT_PROJECT_URL)"
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+### Client part
+file path=usr/bin/$(MACH64)/sysstat
+file path=usr/bin/$(MACH32)/sysstat
+file path=usr/share/man/man1m/sysstat.1m
+
+# Should run the binary matching the kernel bitness
+hardlink path=usr/bin/sysstat target=../lib/isaexec pkg.linted=true
+
+### Server part, required for client even locally,
+### even if not networked via the -l CLI switch
+file files/auth_sysstatd path=etc/security/auth_attr.d/sysstatd
+file files/prof_sysstatd path=etc/security/prof_attr.d/sysstatd
+file files/sysstatd.xml path=lib/svc/manifest/system/diagnostic-sysstatd.xml
+
+file path=usr/sbin/$(MACH64)/sysstatd
+file path=usr/sbin/$(MACH32)/sysstatd
+# Note: No manpage for the daemon is provided in the sources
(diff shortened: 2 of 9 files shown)


components/sysutils/sysstat/Makefile
```diff
@@ -0,0 +1,48 @@
+#
```
pyhalov, line 1: As always, category better to correspond to fmri

components/sysutils/sysstat/Makefile
```diff
@@ -0,0 +1,48 @@
...
+COMPONENT_LICENSE_FILE=	LICENSE
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION =  \
+		($(CLONEY) $(SOURCE_DIR) $(@D))
+
+# non-autoconf "configure" script here
```
pyhalov: perhaps, use configure.mk and CONFIGURE_OPTIONS =  (I mean, completely overwrite them?)

jimklimov: Not sure it makes sense here to pretend that the script is compatible to autoconf configure (as other parts of that makefile can assume now or later) when it is an arbitrary script with the same name. As for FMRI - this and xjobs recipes were copied from mbuffer which got into this classification :)

pyhalov: @xen0l - what do you think about it? I still feel inconvenient about this PRE_BUILD_ACTION. 
Also, as script doesn't use configure.mk, COMPONENT_PRE_CONFIGURE_ACTION is not used.

xen0l:
We should either deliver SMF manifest for sysstatd or drop it for now. Also, separating these into two packages would be good.

components/shell/sysstat/Makefile
```diff
@@ -0,0 +1,48 @@
...
+#
+
+#
+# Copyright 2016, Jim Klimov
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		sysstat
+COMPONENT_VERSION=	20151012
+COMPONENT_FMRI=		shell/sysstat
```
xen0l: shell/sysstat -> diagnostic/sysstat, also ship it in sysutils/

xen0l:
Also, please drop mbuffer changes.

components/shell/sysstat/sysstat.p5m
```diff
@@ -0,0 +1,35 @@
...
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value="$(COMPONENT_PROJECT_URL)"
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/sysstat
+file path=usr/bin/$(MACH32)/sysstat
+file path=usr/sbin/$(MACH64)/sysstatd
```
xen0l: Separate sysstatd into separate package. I expect people will use sysstat alone.

components/shell/sysstat/sysstat.p5m
```diff
@@ -0,0 +1,35 @@
...
+file path=usr/sbin/$(MACH32)/sysstatd
```
xen0l: No manpage for sysstatd?

xen0l (changes requested):

components/shell/sysstat/sysstat.p5m
jimklimov: None, and few mentions in the docs. Source file and content (somewhat) is different, though.

components/shell/sysstat/sysstat.p5m
jimklimov: It seems (from readme) that `sysstat -d` provides similar functionality. But ok, separated if you like so.

components/shell/sysstat/Makefile
jimklimov: done

jimklimov:
> We should either deliver SMF manifest for sysstatd or drop it for now. Also, separating these into two packages would be good.

You did not read the PR header, did you? Or chose to ignore it? TL;DR version is: you want a feature - merge the foundation (if there's no fatal flaws) and vote with your PRs on top - this gets around faster, and the commonly satisfactory part of the recipe gets merged earlier and does not rot as part of a dormant PR ;)

jimklimov:
@xen0l : So... how is the latest PR version?

Most of your comments were addressed, except for an SMF service. I am not ready now to look at what it requires to run and how to do it securely, so would rather leave the exercise for somebody who would be first (next) to use this to monitor several OI/Hipster installations from one console ;)

xen0l:
I am fine with the changes, but I don't like shipping a service package without SMF manifest. @pyhalov your take on this?

pyhalov:
If we deliver daemon which can be usable in any way, we should provide SMF manifest.

jimklimov:
At this time I'm not responsibly ready to wrap the service - I am not
certain what it does, what rights it needs and what insecurities it brings
- or carefully does not and so successfully avoids.

We can deliver just the daemon binary, so that somebody else who ever (if
ever) needs this feature, researches and SMFizes it correctly for their
system, and upstreams the change to us. We can also drop delivering the
daemon and stiffle innovation in this area ;)

If you already have a strong opinion about how this should be SMFized -
feel free to merge the PR with code+recipe and follow up with your PR with
a manifest shortly.

2016-11-09 19:11 GMT+01:00 Alexander Pyhalov notifications@github.com:

xen0l:
I don't have time for investigating SMF service right now. Would you be able to provide even basic SMF service? RBAC integration is just an added value in many cases.

jimklimov:
Some manifest added (untested) :)

xen0l:
What does svccfg validate say? Can you at least run the service and application works?

pyhalov:
Did you check service? Can you update Makefile to use configure.mk?

components/sysutils/sysstat/Makefile
jimklimov: Thanks, good catch with this one - removed the action with no losses :)

jimklimov:
Ok, so I experimented with this a bit.

1) The daemon is required in any case - if a service is not running, the client launches it as a sub-process to monitor the local machine, using the same multicast strategy (by default)

2) If the daemon is started manually, or via SMF, the client connects to it
2.1) If I then kill the daemon, client stops refreshing

3) not checked - the client should be able to track a farm reported by sysstatd's on each node.

It seems the impact of networking daemon is moderate: systems can not be managed over the net, only monitored. It does however expose system stats so anyone can subscribe with a sysstat client or even sniff them (unencrypted plaintext), which may be undesirable in shared networks but okay in managed LANs.

The data are sent out as multicasts, about 4kb/sec, it depends on amount of hosts and router/switch support (and imposed limits) whether the traffic is considered large or not, and would reach the monitoring station in same subnet or LAN. Note the packets will likely be fragmented UDP (per comment in code, there was a legacy typo with chunk size not equal to common MTU, and now for backwards-compatibility the project chose to keep it). Probably it won't be an issue on Jumbo VLANs though :)

components/sysutils/sysstat/sysstat.p5m
```diff
@@ -0,0 +1,46 @@
...
+# Client part
+file path=usr/bin/$(MACH64)/sysstat
+file path=usr/bin/$(MACH32)/sysstat
+file path=usr/share/man/man1m/sysstat.1m
+
+# Should run the binary matching the kernel bitness
+hardlink path=usr/bin/sysstat target=../lib/isaexec pkg.linted=true
+
+# Server part, required for client even locally, even if not networked via the -l CLI switch
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
```
pyhalov: Remove second license action

components/sysutils/sysstat/Makefile
```diff
@@ -0,0 +1,45 @@
...
+COMPONENT_PROJECT_URL=	http://www.maier-komor.de/sysstat.html
+COMPONENT_LICENSE=	GPLv3
+COMPONENT_LICENSE_FILE=	LICENSE
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+# non-autoconf "configure" script here
+COMPONENT_PRE_BUILD_ACTION = \
```
pyhalov, line 37: Please, convert it to using configure.mk, like nginx.

jimklimov: It is NOT using autoconf configure. So IMHO the recipe integration for configure.mk is a poor choice here. We have many (112 ATM) recipes for projects with justmake.mk, including projects who poorly chose the name for their build-setup scripts "configure" without following the autoconf CLI-API.

Of course, anything can be twisted into anything else (e.g. by overriding proper autoconfigure with pieces for this custom one), but I assume it is a dark path to follow ;)

Conversely: is there something that does not work up to spec with recipe as is? Is there a practical reason to twist it?

components/sysutils/sysstat/Makefile
jimklimov: Perhaps, but

1) There is no "diagnostic/" directory
2) There are many packages with "diagnostic/" FMRI, mostly sniffers and local load analyzers
3) The closest equivalent to sysstat is "diagnostic/top" which is why I picked this FMRI patern

pyhalov: I'm fine with this category

components/sysutils/sysstat/Makefile
pyhalov: @xenol: Are you OK with it?

xen0l: @pyhalov yes. @jimklimov thanks for providing the testing results.

xen0l (approved):

xen0l:
This change has been marked for inclusion. Please, squash commits.

jimklimov:
Thanks. Tomorrow, i guess.

--

08.12.2016 20:56 пользователь "Adam Števko" <notifications@github.com>
написал:

components/sysutils/sysstat/Makefile
pyhalov: Perhaps, at least split it in several (2?) lines to fit in 80 lines?

jimklimov:
Squashed

PR #2580, opened 2016-11-06
Add nodejs-7 to oi-userland

AndWac:


xen0l (commented):
What about also mediating npm.1?

AndWac:
we could do that,
but we did not do that for node and node6 yet. Furthermore there are several other npm man pages (man1, 5, and 7). What should be done with them?

xen0l:
@pyhalov what do you think about mediating npm manpages? If we didn't do it, I'd say to leave it as it is.

pyhalov:
As we mediate npm, it will be good to mediate its manpage

xen0l:
As discussed yesterday, @AndWac will provide npm man page mediation in subsequent PRs. I mark this for merging.

PR #2581, opened 2016-11-06
Fix permissions on powerdns-recursor config file

xen0l:
When I installed pdns-recursor, I was not able to modify config file (had to chmod +w it). @pyhalov review when you have time.


xen0l:
SMF manifest isn't imported after install and it seems is some issue. However, looking at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/transforms/actuators#L27 I don't see anything why it shouldn't work..

PR #2067, opened 2016-05-29
Bump of libgd2, libwebp and libvpx, and rehaul of make-rules for better sub-component support

jimklimov:
Recently we mentioned libvpx... and there were a few more repos in the project that were not stagnant, not for Windows, and low-hanging fruit to build. So why not? ;)

PR #2583, opened 2016-11-07
Better support for recipes with sub-components

jimklimov:
Modernized version of infrastructural foundation part of PR #2067 

These changes streamline support for recipes with subcomponents (many sources and builddirs, one component for overall packaging) so more of the typical magic is "guessed", and ultimate component recipes can be much more simple - and focused on the software they build, not the voodoo rituals of the gate.

### DIFF
--- a/make-rules/lint-libraries.mk
+++ b/make-rules/lint-libraries.mk
@@ -36,33 +36,70 @@
 # llib-l{libname}
 #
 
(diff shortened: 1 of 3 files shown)


jimklimov:
Left until better days, especially as this changes the mechanics of package-building, so we'd have to rebuild much of the gate with these new rules to verify nothing gets broken by them.

ENEW_BUILD_SYSTEM_NEEDED

PR #2585, opened 2016-11-07
Initial commit of libwebm-1.0.0.27

jimklimov:
Splintered from PR #2067 

As subsequent discussion has shown, it is questionable if the WebM library is at all intended for packaging and redistribution as a binary. The scope of PR is likely to get reduced to delivering just the few command-line tools (and then maybe from a newer Git snapshot, as the upstream reported many bugfixes in changelog over the past year - https://chromium.googlesource.com/webm/libwebm origin / https://github.com/webmproject/libwebm/ replica), at least until we find that a shared `libwebm.so` library *can* be consumed by any other software - and then add it with another PR.


components/library/libwebm/libwebm.p5m
```diff
@@ -0,0 +1,36 @@
...
+file path=usr/bin/$(MACH64)/vttdemux
+file path=usr/bin/$(MACH64)/webm2pes
+
+file usr/bin/sample path=usr/demo/libwebm/sample
+file path=usr/bin/sample_muxer
+file path=usr/bin/vttdemux
+file path=usr/bin/webm2pes
+
+file path=usr/lib/$(MACH64)/libwebm.so
+file path=usr/lib/libwebm.so
```
pyhalov: I have one question - how this library is used and what are its (possible) consumers?
Strange to see dynamic library without headers.

components/library/libwebm/libwebm.p5m
```diff
@@ -0,0 +1,36 @@
...
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file usr/bin/$(MACH64)/sample path=usr/demo/libwebm/$(MACH64)/sample
+file path=usr/bin/$(MACH64)/sample_muxer
+file path=usr/bin/$(MACH64)/vttdemux
+file path=usr/bin/$(MACH64)/webm2pes
+
+file usr/bin/sample path=usr/demo/libwebm/sample
+file path=usr/bin/sample_muxer
```
pyhalov, line 31: Should it also live in usr/demo/libwebm ?

components/library/libwebm/libwebm.p5m
jimklimov: According to http://www.webmproject.org/ "The WebM Project is dedicated to developing a high-quality, open video format for the web that's freely available to everyone" with its alliance ecosystem "including Cisco, Google, Intel, Microsoft, Mozilla and Netflix".

It is part of the same family of projects as libvpx (vp8/vp9 video codec) and libwebp (still image files). From what I gather, WebM is a container format for video, audio and text streams similar to Matroska. This libwebm apparently allows developers to manipulate it :)

Some more here:
- https://www.webmproject.org/about/faq/
- https://en.wikipedia.org/wiki/WebM
- https://www.macxdvd.com/mac-dvd-video-converter-how-to/webm-vs-mp4.htm

Headers is a good catch... I mostly had it "make install" and packaged what it delivered into the proto dir. Not sure even which binaries are tools or samples, nor how developers would use this. IIRC it got pulled as dependency I needed for something else while preparing the grounds for games and multimedia programs to be at all buildable on OI/Hipster.

components/library/libwebm/libwebm.p5m
jimklimov: Not sure even which binaries of this project are tools or samples... but in this context "sample" may refer to audio/video sampling - not "example software". But gotta make sure if we must...

components/library/libwebm/libwebm.p5m
jimklimov: Looking more at this, e.g. https://groups.google.com/a/webmproject.org/forum/#!topic/webm-discuss/7ztiZTH8xBA I'm less convinced we do need this as a separate library. For example, libvpx includes and apparently builds its own copy to embed. However the command-line tools may be useful (and so the library for them - hence maybe it is shared but with no headers). Strange project :)

components/library/libwebm/libwebm.p5m
jimklimov: There are almost no other recipes for packaging this, which may be an ominous sign ;)
But one I did find included `sample_muxer` in `bin/` as a tool. Sources (and the usage text) suggest it is a tool indeed.

components/library/libwebm/libwebm.p5m
pyhalov: If we want just to have tools, I'd rather compiled library as static one and avoided delivering its files. Can you recall the software which required external libwebm?

jimklimov: I guess the point becomes more moot with every comment ;-\

I'd place this PR/branch on hold until I again come across that software (and make sure if it wants or even _can_ use external libwebm rather than using or even customizing a built-in copy). At least the parts regarding shared library... For now I'll excise the webm commits from remains of PR #2067 which was started to integrate it and update related things ;)

It seems the copy in libvpx for example is larger, and might be an import from later development in Google Git (sources are organized and named differently, and there's more code/larger source files).

jimklimov:
Status: planning to reduce the recipe and PR scope to deliver just the command-line tools from the newly introduced libwebm component, probably built statically. Delivery of shared library and headers(to consume from other software) might be elected and done later, with or without rebuilding these tools to use it dynamically (would save some 300-400Kb per binary).

Maybe also take a look at https://github.com/webmproject/webm-tools/ too (has no releases or tags, only Git HEAD - and there was some code-moving from this one to Git HEAD of libwebm repo)...

xen0l:
@pyhalov it would be possibly the best option to bring prep-(git|svn|hg).mk from userland-gate at this time into oi-userland. With that, @jimklimov can use it to fetch from Git repository in a sane manner.

jimklimov:
For now, reduced scope to build static lib and package only the WebM tools from that release. Recipe is left "dirty" (commented away lines, renamed patch files) to facilitate in-place rollback and experimentation, for now...

pyhalov:
Sorry, Jim, but looking at this package I have one question. Do we really need it???

jimklimov:
Not certain either, especially as pieces are chipped off this PR ;-\
Initially I thought this may be required for some multimedia and browser packages, but later found that the main usage of this "library" is to include an arbitrary version of its code into a project that uses it. As such, maybe a package of it is not needed (and we disabled the .so generation earlier in reviews). So one thing that remain is the set of tools to inspect and manipulate webm containers. Which may be useful to someone working with multimedia on OI :)

PR #2586, opened 2016-11-07
Add requests to oi-userland

xen0l:
Pre-requisite for pip 9.0 update

### DIFF
--- a/components/python/requests/Makefile
+++ b/components/python/requests/Makefile
@@ -1,0 +1,59 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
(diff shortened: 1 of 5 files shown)


components/python/requests/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+# common targets
+build:		$(BUILD_NO_ARCH)
+
+install:	$(INSTALL_NO_ARCH)
+
+# Tests require access to AWS services & the Internet
+test:		$(NO_TESTS)
+
+system-test:	$(NO_TESTS)
```
pyhalov: REQUIRED_PACKAGES += runtime/python-27
REQUIRED_PACKAGES += runtime/python-34

pyhalov: Also we usually don't have system-test target in components (however, there are some exceptions) as we miss SYSTEM_TEST rules.

PR #2588, opened 2016-11-08
Geany plugins

3eka:
Separate PR for Geany Plugins package


pyhalov:
it seems you've removed geany in this commit... You shouldn't have done this ;)

3eka:
Hi,
maybe I did not understood it properly. Shall I bring it back?
Regards.

pyhalov:
You should have oi/hipster branch + geany-plugins changes

alarcher:
When he squashes the commits this will not show anyway.

alarcher:
@3eka What @pyhalov means is that you should have a single commit for geany-plugins.
This requirement implies that you need to rewrite your commit history before the PR is merged.
You need to distinguish two phases:
1) working on the issue/feature: committing as you like, privately on your branch,
2) preparing the work for merging: cleaning up the commit history such that each commit addresses a logical unit.

Let us say you worked on your branch and have, in reverse order:
commit 3 : remove geany
commit 2 : add plugins
commit 1: add geany

The PR for geany-plugins can be merged when only commit 2 remains in the branch.

You need to use `git rebase -i HEAD~3` to modify the last 3 commits.
In this case you have 2 options giving the same result:
1) drop commit 1 and 3: they will be ignored such that only 2 remains
2) squash all 3 commits into one: since everything added in commit 3 is removed in commit 1 then nothing will be added.

The reason why I suggested initially to drop the geany update commit in the 'geany_plugins' branch was to avoid having to remove the files manually in another commit.
I guess now you understand why :)

3eka:
squashed

pyhalov:
Don't see this. PR still has 8 commits.

PR #2605, opened 2016-11-17
Initial commit for libp11 version 0.4.2

jimklimov:
Helper library to wrap PKCS#11 manipulation via openssl.

Openconnect wants this, along with p11-kit which we have already.

### DIFF
--- a/components/library/libp11/Makefile
+++ b/components/library/libp11/Makefile
@@ -1,0 +1,70 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		libp11
+COMPONENT_VERSION=	0.4.2
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.gz
+COMPONENT_FMRI=		library/security/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Security
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	https://github.com/OpenSC/$(COMPONENT_NAME)/wiki
+COMPONENT_ARCHIVE_URL=	https://github.com/OpenSC/$(COMPONENT_NAME)/releases/download/$(COMPONENT_NAME)-$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:e2c3614a314b452a9b57e2914252df3ffee59e262dfb75b4fc73a2247f8ddebe
+COMPONENT_LICENSE=		LGPLv2.1+
+COMPONENT_LICENSE_FILE= COPYING
+COMPONENT_DESCRIPTION=	libp11 is a library implementing a thin layer on top of PKCS11 API to make using PKCS11 implementations easier
+COMPONENT_SUMMARY=	libp11 is a library to make using PKCS11 implementations easier
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PREP_ACTION = \
+        (cd $(@D) &&\
+        autoreconf -fiv &&\
+        $(RM) config.h )
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+        ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+INDIR.32=/usr/bin
+BINDIR.64=/usr/bin/$(MACH64)
+LIBDIR.32=/usr/lib
+LIBDIR.64=/usr/lib/$(MACH64)
+
+# OpenSSL engine directories; default is for files, engines is for symlinks
+ENGINESDIR_32 = /lib/openssl/engines
+ENGINESDIR_64 = /lib/openssl/engines/$(MACH64)
+DEFENGINESDIR_32 = /lib/openssl/default/engines
+DEFENGINESDIR_64 = /lib/openssl/default/engines/$(MACH64)
+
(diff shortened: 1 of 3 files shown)


components/library/libp11/Makefile
```diff
@@ -0,0 +1,65 @@
...
+COMPONENT_FMRI=		library/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Security
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	https://github.com/OpenSC/$(COMPONENT_NAME)/wiki
+COMPONENT_ARCHIVE_URL=	https://github.com/OpenSC/$(COMPONENT_NAME)/releases/download/$(COMPONENT_NAME)-$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:e2c3614a314b452a9b57e2914252df3ffee59e262dfb75b4fc73a2247f8ddebe
+COMPONENT_LICENSE=		LGPLv2.1+
+COMPONENT_LICENSE_FILE= COPYING
+COMPONENT_DESCRIPTION=	libp11 is a library implementing a thin layer on top of PKCS#11 API to make using PKCS#11 implementations easier
```
pyhalov: # should be escaped

components/library/libp11/libp11.p5m
```diff
@@ -0,0 +1,44 @@
...
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/include/libp11.h
+
+link path=usr/lib/$(MACH64)/libp11.so target=libp11.so.2.4.2
+link path=usr/lib/$(MACH64)/libp11.so.2 target=libp11.so.2.4.2
+file path=usr/lib/$(MACH64)/libp11.so.2.4.2
+
+file path=usr/lib/$(MACH64)/pkcs11.so
+link path=usr/lib/$(MACH64)/libpkcs11.so.libp11 target=pkcs11.so
```
pyhalov: What is this?

components/library/libp11/Makefile
```diff
@@ -0,0 +1,65 @@
+#
```
pyhalov: Please, use new illumos CDDL header

pyhalov, line 1: Please, use new style illumos header (like in templates/configure.mk)

jimklimov:
No good idea about 
`link path=usr/lib/$(MACH64)/libpkcs11.so.libp11 target=pkcs11.so` ,
- it puzzled me too when conflicts popped up during package installation ;)

Apparently this lib registers itself as an engine to openssl, and maybe we 
have a namespace collision when we deliver those not in `/usr/lib/engines`
as expected by default, but into `/usr/lib` directly. Anyhow, for now I
decided to deliver it _somehow_ so just renamed the symlink noticeably.

## 

17.11.2016 8:28 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
Comments should be addressed now

components/library/libp11/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+        ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+INDIR.32=/usr/bin
+BINDIR.64=/usr/bin/$(MACH64)
+LIBDIR.32=/usr/lib
+LIBDIR.64=/usr/lib/$(MACH64)
+
+# Normally this expects /usr/lib/engines - but our openssl does not deliver that
```
pyhalov: But we have /lib/openssl/engines/ . And libpk11.so is already delivered there. But I know almost nothing about openssl engines

jimklimov: Hm... curious... And is the `/lib` location captured by the illumos-gate or by userland packages? (don't both still have a clashing definition of openssl - for OS and for consumers?)

UPDATE it is packaged openssl from oi-userland:

```
root@jimoi:/root# pkg search -r libpk11.so
INDEX      ACTION VALUE                                        PACKAGE
basename   link   lib/openssl/default/engines/amd64/libpk11.so pkg:/library/security/openssl@1.0.2.10-2016.0.1.0
basename   link   lib/openssl/default/engines/libpk11.so       pkg:/library/security/openssl@1.0.2.10-2016.0.1.0
basename   link   lib/openssl/engines/amd64/libpk11.so         pkg:/library/security/openssl@1.0.2.10-2016.0.1.0
basename   link   lib/openssl/engines/libpk11.so               pkg:/library/security/openssl@1.0.2.10-2016.0.1.0
```

And also, this package is about libp11 (not libp_k_11) ;)

I'll see if it can deliver there for clarity, and if openconnect would rather use that too...

pyhalov: If you deliver it there, can OpenSSL somehow use it?

jimklimov: You keep expecting that the recipe committer knows everything about the packaged bits. I was just unwrapping dependencies wanted by another project to compile ;)

No firm idea about "can" question above, beyond the `README.md` of the project saying it can be used by openssl (transparently with p11-kit, or with extra configuration files if directly).

From the README:

> This code repository produces two libraries:
> - libp11 provides a higher-level (compared to the PKCS#11 library) interface to access PKCS#11 objects.  It is designed to integrate with applications that use OpenSSL.
> - pkcs11 engine plugin for the OpenSSL library allows accessing PKCS#11 modules in a semi-transparent way.
> 
> The wiki page for this project is at https://github.com/OpenSC/libp11/wiki
> and includes a bug tracker and source browser.

jimklimov: So, after some unraveling, it seems that:
- the `/usr/lib/libpkcs11.so*` files are delivered by `pkg:/system/library` wherever it takes the sources.
- the opessl-1.0.2 recipe builds and delivers its own (+patches) code, including the `/lib/openssl` directory with engines, some scripts, etc. and the `openssl` binary to which `/usr/bin/openssl` is just a symlink. It does not deliver a `libpkcs11.so*` although refers to it in the configure options of the build recipe.
- the libp11 makes (from different sources) a `libp11.so*` set of libraries that it installs into a common libdir (e.g. /usr/lib) and a `pkcs.so` for the engine (according to source-file set) which it puts into `--engine-dir=...` location (/usr/lib/engine by default), and makes a symlink to it as the `./libpkcs11.so` wherever it installed this piece.

I have no idea if the two variants of `libpkcs11.so` (from `system/library` in the common location and the engine from libp11) are related by source origins, or even by intention (libp11 one is very likely an openssl engine, no idea what the system one intends to be). But at least if we deliver the engine shared object into engine location, the conflict seems avoided too ;)

jimklimov:
Fixed packaging to take into account openssl engines location (and also avoid a filename clash conflict), and adjusted the FMRI to point out library/_security_

PR #2616, opened 2016-11-20
Initial recipe for groovy-2.4.7 language

jimklimov:
Just IPS packaging in our common directory layout for the upstream binary (JAR) distribution.

And a PoC for other Java-based distro integrations (think netbeans, glassfish, etc.)

### DIFF
--- a/components/developer/groovy/Makefile
+++ b/components/developer/groovy/Makefile
@@ -32,9 +32,6 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/justmake.mk
 include $(WS_MAKE_RULES)/ips.mk
 
-LN_S ?= $(LN) -s
-LN_SYM_REL = $(GNUBIN)/ln -s -r
-
 # This is an archive with prebuilt binaries, nothing to build
 $(BUILD_$(BITS)): $(SOURCE_DIR)/.prep
 	$(MKDIR) $(BUILD_DIR_$(BITS))
@@ -42,23 +39,11 @@ $(BUILD_$(BITS)): $(SOURCE_DIR)/.prep
 
 build: $(BUILD_32)
 
-$(INSTALL_32): $(BUILD_32) $(SOURCE_DIR)/.prep
-	$(MKDIR) $(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)-$(COMPONENT_VERSION) $(PROTOUSRBINDIR)
-	cd $(SOURCE_DIR) && \
-		$(CP) -pr \
-bin \
-conf \
-embeddable \
-grooid \
-indy \
-lib \
-LICENSE \
-licenses \
-NOTICE \
-		$(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)-$(COMPONENT_VERSION)
-	$(RM) $(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)-$(COMPONENT_VERSION)/bin/*.bat
-	cd $(PROTOUSRBINDIR) && \
-		$(LN_SYM_REL) $(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)-$(COMPONENT_VERSION)/bin/* .
+# There are many "intimate" variables to pass into the other Makefile,
+# so rather than run a sub-make we include it into our context.
+include $(COMPONENT_DIR)/files/Makefile
+
+$(INSTALL_32): install-groovy
 	$(TOUCH) $@
 
 install: $(INSTALL_32)
--- a/components/developer/groovy/Makefile
+++ b/components/developer/groovy/Makefile
@@ -1,0 +1,69 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Jim Klimov
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		groovy
+# Commit timestamp and previous known release (none here)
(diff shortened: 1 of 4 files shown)


components/developer/groovy/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+#
+# Copyright 2016 Jim Klimov
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		groovy
+# Commit timestamp and previous known release (none here)
+COMPONENT_VERSION=	2.4.7
+COMPONENT_REVISION=	1
```
pyhalov: Please, remove COMPONENT_REVISION for initial component version

components/developer/groovy/Makefile
```diff
@@ -0,0 +1,70 @@
...
+COMPONENT_ARCHIVE=	apache-$(COMPONENT_NAME)-binary-$(COMPONENT_VERSION).zip
+COMPONENT_FMRI=		developer/$(COMPONENT_NAME)
```
pyhalov, line 20: Do we want so split out runtime/groovy (for example, for /usr/bin/groovy)?

components/developer/groovy/Makefile
```diff
@@ -0,0 +1,70 @@
...
+$(BUILD_$(BITS)): $(SOURCE_DIR)/.prep
+	$(MKDIR) $(BUILD_DIR_$(BITS))
+	$(TOUCH) $@
+
+build: $(BUILD_32)
+
+$(INSTALL_32): $(BUILD_32) $(SOURCE_DIR)/.prep
+	$(MKDIR) $(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)-$(COMPONENT_VERSION) $(PROTOUSRBINDIR)
+	cd $(SOURCE_DIR) && \
+		$(CP) -pr \
```
pyhalov: Looks ugly... Perhaps, just set some transforms and pick up files out of $(SOURCE_DIR)? or cp -r $(SOURCE_DIR)/* $(PROTOUSRSHAREDIR)/$(COMPONENT_NAME)-$(COMPONENT_VERSION) ? 
@xen0l : what do you think?

components/developer/groovy/Makefile
jimklimov: Fixed

components/developer/groovy/Makefile
jimklimov: Not sure about this... In particular, I don't really know the inner workings of the project, like what is the minimally sufficient set of JARs to be the runtime. And it is not that large for it to really matter (although yes, delivering 6Mb vs. 30Mb is yummy).

components/developer/groovy/Makefile
jimklimov: There is some build-system cruft in the SOURCE_DIR (the touched dot-files). So I'd rather deliver into prepared tree only the FS objects known to be the project deliverables.

pyhalov: Well, let's at least make it look more reasonable. Perhaps, create separate Makefile for grovy with just empty build and necessary install targets, put it under files/ and copy at PREP step? In such way we remove install logic from component's Makefile... Like in components/fonts/baekmuk-ttf...

jimklimov:
Good idea, I'll look at that, but some time later.

## 

22.11.2016 7:21 пользователь "Alexander Pyhalov" notifications@github.com
написал:

jimklimov:
Revised and updated, thanks for the comments. In the end, dedicating a standalone Makefile was too much hassle (passing same vars to keep our standard layout rather than hardcode arbitrary strings, etc.) so I made it an includable snippet. Main recipe looks simpler, installation logic is offloaded into another file nearly "as is".

xen0l (approved):

components/developer/groovy/files/Makefile
```diff
@@ -0,0 +1,44 @@
...
+# This Makefile installs groovy into location specified by variables below,
+# as defined in common oi-userland framework (this file is included from
+# main recipe Makefile).
+# List of FS objects below coresponds to groovy-2.4.7 layout.
+#
+
+LN_S ?= $(LN) -s
+LN_SYM_REL = $(GNUBIN)/ln -s -r
+
+$(BUILD_DIR_$(BITS))/.installed-groovy: $(BUILD_$(BITS)) $(SOURCE_DIR)/.prep
```
pyhalov, line 25: I'm still not sure about this additional complexity. You can use it as independent Makefile... 
Set PATH=/usr/gnu/bin:/usr/bin in top level Makefile and you can do something like:
```
SUBDIRS=bin conf embeddable grooid indy lib
GROOVY_VERSION ?= 2.4.7
all: build

build: 

install: 
       mkdir -p $(DESTDIR)/usr/share/grovy-$(GROOVY_VERSION)
       mkdir -p $(DESTDIR)/usr/bin
       cp -rf $(SUBDIRS) $(DESTDIR)/usr/share/grovy-$(GROOVY_VERSION)
       ln -rs $(DESTDIR)/usr/share/grovy-$(GROOVY_VERSION)/bin/* $(DESTDIR)/usr/bin
```
I think this is a bit prettier as Makefile is independent and shouldn't have any knowledge of our build system.

jimklimov:
I considered this and have a PoC lying around, but in this case the
"usr/..." paths become hardcoded (unlike configure'd projects where our
build system tells the project where to install itself). Here we can of
course pass it via INSTALL_ENV or somesuch, rather than including. But that
does add more complexity, I've checked ;)

Anyhow, do we expect to use this as an independent Makefile, outside the
build system that defines it at all?

2016-11-25 6:45 GMT+01:00 Alexander Pyhalov <notifications@github.com>:

pyhalov:
No, we don't expect it to be used outside of build system

jimklimov:
Then I think it does not matter much that the file can be used standalone
or is more intimate and simple to invoke ;)

2016-11-25 12:02 GMT+01:00 Alexander Pyhalov <notifications@github.com>:

pyhalov:
But at least it's easier to read ;)

pyhalov:
See https://github.com/OpenIndiana/oi-userland/pull/2744

PR #2621, opened 2016-11-24
add bundler 1.13.6

grueni:
use ruby 2.2 in package
use gemspec created from archive content
path for binaries of bundler, bundler corrected


components/ruby/bundler/Makefile
```diff
@@ -0,0 +1,58 @@
...
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+# Copyright 2016 Grüninger GmbH, Andreas Grüninger (grueni)
+#
+include ../../../make-rules/shared-macros.mk
+
+RUBY_VERSION = 2.2
+RUBY_LIB_VERSION = 2.2.0
```
xen0l, line 26: RUBY_LIB_VERSION, RUBY.2.2 and RUBY definitions should be moved to shared-macros.mk, as they are useful for other projects as well. I don't consider it a blocker. @pyhalov should we move these?

xen0l (approved):

components/ruby/bundler/Makefile
pyhalov: I think that 
```
diff --git a/make-rules/shared-macros.mk b/make-rules/shared-macros.mk
index 8df28d3..d34fe8e 100644
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -567,9 +567,14 @@ F77 =              $(F77.$(COMPILER).$(BITS))
 FC =           $(FC.$(COMPILER).$(BITS))
 
 RUBY_VERSION =  1.9
-RUBY_LIB_VERSION =      1.9.1
+RUBY_LIB_VERSION.1.9 = 1.9.1
+RUBY_LIB_VERSION.2.2 = 2.2.0   
+RUBY_LIB_VERSION.2.3 = 2.3.0
 RUBY.1.9 =      /usr/ruby/1.9/bin/ruby
+RUBY.2.2 =     /usr/ruby/2.2/bin/ruby
+RUBY.2.3 =     /usr/ruby/2.3/bin/ruby
 RUBY =          $(RUBY.$(RUBY_VERSION))
+RUBY_LIB_VERSION = $(RUBY_LIB_VERSION.$(RUBY_VERSION))
``` 
is not so a big change to avoid doing it in this PR, especially when component can leverage it.

grueni: Should I do something or just wait for the change of shared-macros.mk?

pyhalov: I suppose you incorporate this in your commit and modify component's Makefile accordingly.

pyhalov:
Modified version merged as https://github.com/OpenIndiana/oi-userland/commit/c8d494d7b8a712a3cfbdc74c0d90448111649cc0

PR #2622, opened 2016-11-24
pcs: use ruby 2.2

grueni:
add dependencies for bundler 
use environment variables for RUBY


components/cluster/pcs/Makefile
```diff
@@ -14,6 +14,12 @@
 
 include ../../../make-rules/shared-macros.mk
 
+RUBY_VERSION = 2.2
+RUBY_LIB_VERSION = 2.2.0
+RUBY.2.2 = /usr/ruby/2.2/bin/ruby
+RUBY = $(RUBY.$(RUBY_VERSION))
+RUBY_VERSIONS = $(RUBY_LIB_VERSION)
```
pyhalov: @xenol I think we should add part of these definitions to shared-macros.mk, and here use just RUBY_VERSION=2.2

components/cluster/pcs/Makefile
```diff
@@ -14,6 +14,12 @@
...
+
 COMPONENT_NAME= pcs
 COMPONENT_VERSION= 0.9.154
```
pyhalov: @grueni Please, bump COMPONENT_REVISION

components/cluster/pcs/Makefile
xen0l: yes, we should move it to sharec-macros.mk

components/cluster/pcs/Makefile
xen0l: @grueni COMPONENT_REVISION has to be bumped, so build server picks up the build

pyhalov:
Should it be RUBY 2.2 or does it work with 2.3 ?

pyhalov:
Switch to ruby 2.3 was merged as https://github.com/OpenIndiana/oi-userland/commit/e4475440970198870c1761897e9b2661f3a965d7 . Additional changes (if necessary) should be submitted as separate PR.

PR #2636, opened 2016-11-30
7957 lightdm switches console to VT1 too early

pyhalov:


tsoome (approved):
Based on testing, it does seem to behave much better - I get the console restored every time I disable lightdm so far. The source review on patch file is impossible however. So I set "approve" based on test results;)

PR #2648, opened 2016-12-02
graphviz: ruby module is linked to libruby 1.9

pyhalov:


pyhalov:
To be merged after swig update

PR #2654, opened 2016-12-05
Remove missing files from manifests

pyhalov:


alarcher:
Do you know which package is missing on the build server?

pyhalov:
Likely, apache fop is required. But we don't have it....

alarcher:
I do not have extra packages on the machine I used for bumping the version and generating the manifests so this comes as a surprise.

pyhalov:
can you recheck that gmake clean publish actually works? Perhaps, you had files left from old build?

alarcher:
I just checked and they publish correctly, I will try to understand the difference.

pyhalov:
I see 
'configure: WARNING: fop not found - documentation targets will be skipped' while running  configure

alarcher:
Yes I have the same, this does not seem to concern the missing files.

alarcher:
@pyhalov In any case, thank you for pushing the fixed manifests. :)

PR #2680, opened 2016-12-15
Bump libXpm to 3.5.12

alarcher:


xen0l (approved):

alarcher:
@pyhalov @xen0l thanks, I think that except libXfont which requires some care we are pretty much up-to-date when it comes to Xorg libraries.

PR #2687, opened 2016-12-20
Bump opus to 1.1.3

alarcher:
- ABI compatible.
- All tests passed.


AndWac:
COMPONENT_LICENSE occurs twice.

PR #2711, opened 2017-01-05
mutt: enable debugging support

jeffpc:
This doesn't change any behavior, it simply enables support for the -d
command line option.  This option is very useful for debugging assorted mutt
issues.


xen0l:
To trigger rebuild, please add COMPONENT_REVISION= 1, otherwise LGTM.

pyhalov:
Do we really need debug enabled in shipped binary? How does it impact runtime performance? What about binary size?

pyhalov:
From IRC conversation: 
```
if you run it the same way as before, I don't see any perf difference
if you turn on some of the debugging stuff by giving it the -d argument, it can get slower
```

PR #2737, opened 2017-01-19
Add ck 0.5.2

alarcher:
- Patched configure to allow 32-bit build on amd64.
- Header should be split between MARCH32/MACH64 due to specific values, same as Circonus/OmniOS.
- Tests do not work due to unsupported out-of-source build, do not throw error, checked manually 100% passed.

### DIFF
--- a/components/library/ck/Makefile
+++ b/components/library/ck/Makefile
@@ -1,0 +1,47 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Aurelien Larcher
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=           ck
+COMPONENT_VERSION=        0.5.2
+COMPONENT_SUMMARY=        ck - Concurrency Kit
+COMPONENT_DESCRIPTION=    Concurrency primitives, safe memory reclamation mechanisms and non-blocking data structures for the research, design and implementation of high performance concurrent systems.
+COMPONENT_PROJECT_URL=    http://concurrencykit.org/
+COMPONENT_FMRI=           library/ck
+COMPONENT_CLASSIFICATION= System/Libraries  
+COMPONENT_SRC=            $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=        $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=    http://concurrencykit.org/releases/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH= \
+  sha256:5cf44b33f9279c653ec9b2b085d628c86336e4da18897be449f074283e5c5b3a
+COMPONENT_LICENSE=        BSD
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CONFIGURE_OPTIONS+= --includedir=$(USRINCDIR)/$(COMPONENT_NAME)
+CONFIGURE_OPTIONS+= --environment=$(BITS)
+
+COMPONENT_TEST_TARGETS= regressions
+
+build:    $(BUILD_32_and_64)
+
+install:  $(INSTALL_32_and_64)
+
+# Tests are not supported out-of-source for now
+test:    $(TEST_32_and_64)
+
+REQUIRED_PACKAGES += system/library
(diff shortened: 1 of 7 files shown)


components/library/ck/Makefile
```diff
@@ -0,0 +1,50 @@
...
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Aurelien Larcher
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=						ck
```
pyhalov: Please, fix formatting, it looks really ugly :)

alarcher: I switched to tabs to conform with most Makefiles... shoud I use spaces instead?

components/library/ck/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_ARCHIVE_HASH= \
+	sha256:5cf44b33f9279c653ec9b2b085d628c86336e4da18897be449f074283e5c5b3a
+COMPONENT_LICENSE=				BSD
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+USRINCDIR.32=$(USRINCDIR)
+USRINCDIR.64=$(USRINCDIR)/$(MACH64)
```
pyhalov: Don't do it... We don't ship /usr/include/$(MACH64).
Possible sollutions: 
- /usr/lib/ck/include, /usr/lib/$(MACH64)/include for platform-dependent headers (with corresponding pkgconfig patches);
 - deliver /usr/include/header-32/64.h and generate /usr/include/header.h , which includes necessary headers - like it's done in python (it would be good to have only one such header with platform-specific definitions);
- patch headers as needed (perhaps, even post-configure, like openssl does).

components/library/ck/Makefile
```diff
@@ -0,0 +1,50 @@
...
+CONFIGURE_OPTIONS+= --environment=$(BITS)
+
+COMPONENT_TEST_TARGETS= regressions
+
+build:		$(BUILD_32_and_64)
+
+install:	$(INSTALL_32_and_64)
+
+# Tests are not supported out-of-source for now
+test:		$(TEST_32_and_64)
```
pyhalov: Will something like in postgresql work? I mean clone or copy source to build dir in pre-configure hook and use CONFIGURE_SCRIPT=$(@D)/configure.

components/library/ck/Makefile
alarcher: FYI Circonus/OmniOS does like this and /usr/include/amd64 already exists for sys.
Which option do you prefer?

components/library/ck/ck.p5m
```diff
@@ -0,0 +1,319 @@
...
+file path=usr/include/spinlock/fas.h
+file path=usr/include/spinlock/hclh.h
+file path=usr/include/spinlock/mcs.h
+file path=usr/include/spinlock/ticket.h
+file path=usr/lib/$(MACH64)/libck.a
+link path=usr/lib/$(MACH64)/libck.so target=libck.so.$(COMPONENT_VERSION)
+file path=usr/lib/$(MACH64)/libck.so.$(COMPONENT_VERSION)
+link path=usr/lib/$(MACH64)/libck.so.0 target=libck.so.$(COMPONENT_VERSION)
+file path=usr/lib/$(MACH64)/pkgconfig/ck.pc
+file path=usr/lib/libck.a
```
pyhalov: Do we need static libs?

components/library/ck/ck.p5m
```diff
@@ -0,0 +1,319 @@
...
+link path=usr/lib/libck.so target=libck.so.$(COMPONENT_VERSION)
+file path=usr/lib/libck.so.$(COMPONENT_VERSION)
+link path=usr/lib/libck.so.0 target=libck.so.$(COMPONENT_VERSION)
+file path=usr/lib/pkgconfig/ck.pc
+file path=usr/share/man/man3/CK_ARRAY_FOREACH.3.gz
```
pyhalov: Does illumos man understand compressed manuals? I thought it didn't.

components/library/ck/Makefile
pyhalov: I don't know, perhaps, wrong tab count, but usually it looks better on github

components/library/ck/Makefile
pyhalov: Does all headers really differ? If only several, we could extract these definitions in /usr/include/header-$(BITS).h. If all, perhaps it's not the best option.

components/library/ck/ck.p5m
alarcher: Well I just get that the file is not in gzip format... :S

components/library/ck/Makefile
alarcher: Not much actually. I can piggy patch the offending header.

alarcher: I just usually avoid messing with code that is not mine.

alarcher: Added header with _LP64 ifdef.

pyhalov: seems better, don't forget to remove unnecessary definitions from Makefile

pyhalov:
It would be good to ship uncompressed man pages

PR #2745, opened 2017-01-24
Add GCC 6.3.0

alarcher:
Tested packages:
- boost


alarcher:
@pyhalov This could be merged as-is to allow broader testing.

PR #2747, opened 2017-01-24
Add zoltan 3.83

alarcher:
Please look at the naming scheme and the mpi-macros.mk solution retained for building a component with multiple MPI implementations.

### DIFF
--- a/make-rules/mpi-macros.mk
+++ b/make-rules/mpi-macros.mk
@@ -18,11 +18,12 @@
 #
 # CDDL HEADER END
 #
-# Copyright (c) 2013, Aurelien Larcher. All rights reserved.
+# Copyright (c) 2013-2017, Aurelien Larcher. All rights reserved.
 #
 
 MPI_IMPLEMENTATIONS_LIST = mpich openmpi
 
+# Define compilers
 MPI.mpich =         mpich
 
 MPICC.mpich =       mpicc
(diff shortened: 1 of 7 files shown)


pyhalov:
Overall seems fine, but:
a) Do we need openmpi/gcc and mpich/gcc? I mean, does someone care about studio? Even in such case, will be resulting object files incompatible, if compiled with Studio?  
b) Should these really be static libs?
c) Your component doesn't touch $(BUILD_DIR) on subsequent installs, this can cause troubles (Test three rebuilds. (https://wiki.openindiana.org/oi/oi-userland+-+best+practices ,  rule of three publishes :))

alarcher:
a) the whole stack should be preferably compiled with the same compiler (to avoid weird MPI issues) and we may use clang in our projects as well. 
b) for scientific computing static libs are the norm, so a number ship and expect only static libs (dynamic linking may be troublesome on supercomputers)
c) thank you for pointing it out, I will address the issue :)

alarcher:
c) I do not understand your point with touching $(BUILD_DIR): the targets are generated for each MPI impl and the component is not republished at the 3rd gmake publish. Could you point me to the specific rules?

pyhalov:
Look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/ips.mk#L345 - mogrified manifest will not be regenerated if build dir is not touched (and as result component will not be published).

alarcher:
I looked at it this afternoon and I fail to understand how it differs from the Python modules with different versions...

alarcher:
I do see the problem though.

alarcher:
Still does not republish at 3rd publish, and now does rebuild+republish if Makefile is touched.

make-rules/mpi-macros.mk
```diff
@@ -41,18 +42,17 @@ MPIFC.openmpi =     mpif90
...
-MPI_PREFIX.64 =     $(USRLIBDIR64)/$(MPI_IMPLEMENTATION)/$(MPI_COMPILER)
-MPI_PREFIX =	    $(MPI_PREFIX.$(BITS))
-
 MPICC =             $(MPICC.$(MPI_IMPLEMENTATION))
 MPICXX =            $(MPICXX.$(MPI_IMPLEMENTATION))
 MPIF77 =            $(MPIF77.$(MPI_IMPLEMENTATION))
 MPIFC =             $(MPIFC.$(MPI_IMPLEMENTATION))
 
-CONFIGURE_DEFAULT_DIRS=no
+# Define immplementatio specific paths
```
pyhalov, line 50: Will it affect existing components, using mpi-macros.mk?

pyhalov:
It seems there's still an issue with component.
```
gmake publish; gmake publish; gmake publish # Here everything is as expected, component is republished twice
# After this:
touch Makefile; gmake publish
# Component is reinstalled, but not republished

PR #2748, opened 2017-01-25
Add Qt 5.8.0

alarcher:


components/qt5/Makefile
```diff
@@ -0,0 +1,99 @@
...
+COMPONENT_NAME=           qt
+COMPONENT_VERSION=        5.8.0
+COMPONENT_VERSION_MJR=    5.8
+COMPONENT_FMRI=           library/qt5
+COMPONENT_PROJECT_URL=    https://www.qt.io/
+COMPONENT_SUMMARY=        Qt cross-platform application and UI framework
+COMPONENT_CLASSIFICATION= Desktop (GNOME)/Libraries
+COMPONENT_SRC=		$(COMPONENT_NAME)-everywhere-opensource-src-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+#COMPONENT_ARCHIVE_HASH=	\
```
pyhalov: why commented out?

components/qt5/Makefile
```diff
@@ -0,0 +1,99 @@
...
+CONFIGURE_OPTIONS+= -disable-reduce-exports
+CONFIGURE_OPTIONS+= -qt-xcb
+CONFIGURE_OPTIONS+= -bindir $(QT5_PREFIX)/bin/$(QT_ARCHSUBDIR.$(BITS))
+CONFIGURE_OPTIONS+= -libdir $(QT5_PREFIX)/lib/$(QT_ARCHSUBDIR.$(BITS))
+CONFIGURE_OPTIONS+= -plugindir $(QT5_PREFIX)/plugins/$(QT_ARCHSUBDIR.$(BITS))
+CONFIGURE_OPTIONS+= -importdir $(QT5_PREFIX)/imports/$(QT_ARCHSUBDIR.$(BITS))
+CONFIGURE_OPTIONS+= -platform $(QT_PLATFORM.$(BITS))
+CONFIGURE_OPTIONS+= $(CONFIGURE_OPTIONS.$(BITS))
+CONFIGURE_OPTIONS+= -I$(MYSQL_INCDIR)
+CONFIGURE_OPTIONS+= -L$(MYSQL_LIBDIR.$(BITS)) -R$(MYSQL_LIBDIR.$(BITS))
```
pyhalov: Does configure really handle -L/-R/-I options?

pyhalov:
Why there's no p5m manifest provided? It doesn't reach install target? :)

AndWac (commented):
Shouldn't this go into a category eg. runtime or developer?

components/qt5/Makefile
alarcher: It is Qt's configure not autotools.

alarcher:
@pyhalov The manifest is generated but had to re-run the build for now. This is WIP to allow people working on the Qt patches to follow the changes.
@AndWac The category is the same as for Qt4. If it should be changed then both should change: not sure it is worth the trouble.

pyhalov:
of course, both components should be moved to corresponding category (likely, library).

alarcher:
@AndWac @pyhalov sorry I answered too fast and did not get that you meant category subdirectory.
Of course I will move them.

components/qt5/Makefile
alarcher: userland-fetch  throws a memory error:
File "/usr/lib/python2.7/gzip.py", line 338, in _add_read_data
    self.extrabuf = self.extrabuf[offset:] + data
MemoryError

alarcher:
Need to fix the memory error during checksum validation... I guess the tarball is too big.

alarcher:
Everything seems fine so the PR could be merged shortly.

components/library/qt5/qt5.p5m
```diff
@@ -0,0 +1,10273 @@
...
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2017 <contributor>
```
pyhalov: Missing your name

components/library/qt5/qt5.p5m
```diff
@@ -0,0 +1,10273 @@
...
+file path=usr/lib/qt/5.8/doc/global/template/style/icomoon.svg
+file path=usr/lib/qt/5.8/doc/global/template/style/icomoon.ttf
+file path=usr/lib/qt/5.8/doc/global/template/style/icomoon.woff
+file path=usr/lib/qt/5.8/doc/global/template/style/list_arrow.png
+file path=usr/lib/qt/5.8/doc/global/template/style/list_expand.png
+file path=usr/lib/qt/5.8/doc/global/template/style/offline-simple.css
+file path=usr/lib/qt/5.8/doc/global/template/style/offline.css
+file path=usr/lib/qt/5.8/doc/global/template/style/online.css
+file path=usr/lib/qt/5.8/doc/global/template/style/theqtcompany.png
+file path=usr/lib/qt/5.8/include/Qt3DCore/$(COMPONENT_VERSION)/Qt3DCore/private/aspectcommanddebugger_p.h
```
pyhalov, line 184: Have you compared 32-bit and 64-bit headers? Do they differ?

alarcher: @jgueytat Could you comment on this?

alarcher:
@pyhalov I have rebuilt again against gstreamer 1.0, I could confirm that there is no explicit link to gst-plugins so that it does not generate a dependency to encumbered packages.

components/library/qt5/qt5.p5m
alarcher: @jgueytat Ping?

jgueytat: Just tried 32 and 64 bits => Same files

components/library/qt5/Makefile
```diff
@@ -0,0 +1,117 @@
...
+REQUIRED_PACKAGES += library/print/cups-libs
+REQUIRED_PACKAGES += library/zlib
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/fontconfig
+REQUIRED_PACKAGES += system/library/freetype-2
+REQUIRED_PACKAGES += system/library/g++-4-runtime
+REQUIRED_PACKAGES += system/library/gcc-4-runtime
+REQUIRED_PACKAGES += system/library/libdbus
+REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += x11/keyboard/xkeyboard-config
```
pyhalov: pkg install: The following pattern(s) did not match any allowable packages.  Try
        /x11/keyboard/xkeyboard-config

components/library/qt5/qt5.p5m
jgueytat: There are configuration files that "should" be the same in 32 and 64 bits. We might lack some dependencies or packages to get the same configuration result in 32 and 64 bits. Checking it now.

pyhalov: just compared myself, 32/64-bit headers are the same

jgueytat: I have two builds with 32 and 64 bits on alarcher's computer that are slightly different. I need to double check that. It's about those config files. I'll come back as soon as I'll have checked.

alarcher:
@pyhalov @jgueytat Last difficulty is to understand why Qt multimedia wants to link directly against gst-plugins-bad.

pyhalov:
Any progress here?

alarcher:
I do not know how to avoid linking to gstreamer-bad's libphotography.

pyhalov:
BTW, the second part of the change triggers some 'bad runpath' issues, for example, for lcms2:
```
ERROR userland.action001.3        bad RUNPATH, 'usr/bin/transicc' includes '/usr/lib/libjpeg8-turbo/lib'
```

alarcher:
Good to go.

components/library/qt5/qt5.license
```diff
@@ -0,0 +1,9355 @@
+#
```
pyhalov: This is clearly not a license :)

components/library/qt5/qt5.p5m
```diff
@@ -0,0 +1,10312 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# Ignore pkglint error with runpath including $ORIGIN
+<transform file path=.*.so.* -> default pkg.linted.userland.action001.3 true >
```
pyhalov, line 26: Do we want /usr/bin/* mediators for qt 5 binaries? If we do, do we want to set mediator-priority=vendor for qt4?

alarcher:
I would feel more confortable if we do that in another stage because we'll have to change the qt4 component since .pc files are linked (non-mediated) in /usr/lib/pkgconfig. Otherwise this would mean that qtconfig and .pc file will not match if qt5 is chosen.

PR #2752, opened 2017-01-29
Bump mesa to 12.0.6

alarcher:


pyhalov:
I woudn't touched manifest file and sample-manifest, as they didn't actually change, just hardlink/file action generation is not determenistic.

alarcher:
OK I will revert these changes then

alarcher:
Reverted the manifest changes.

PR #2753, opened 2017-01-29
Add experimental illumos-gcc patches to GCC 5.4

alarcher:


pyhalov:
Do we want to have one 1000+ line patch or set of smaller patches? How does it look in dilos-userland-gate?

alarcher:
@pyhalov this patchset is a one-shot as it will most likely not be updated so I'd rather not spend time on this one. I will split the 6.3 patchset for better clarity.

pyhalov:
OK

pyhalov:
Should check that at least mongodb (which uses 5.x compiler) built with this compiler doesn't immediately dump core...

alarcher:
@pyhalov MongoDB compiled && tested. I ran the 'dbtests' test suite, over a thousand tests passed until one failed (errno: 24 Too many open files) unrelated to the topic at hand.

PR #2760, opened 2017-01-30
7824 Fix guile, /usr/bin/guile-snarf path to gcc hard-coded

alarcher:


components/library/guile/manifests/sample-manifest.p5m
```diff
@@ -5,12 +5,12 @@
 # 1.0 of the CDDL.
```
pyhalov, line 5: hm...???

components/library/guile/Makefile
```diff
@@ -20,37 +20,40 @@
...
 
-include ../../make-rules/prep.mk
-include ../../make-rules/configure.mk
-include ../../make-rules/ips.mk
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
 
-# Built in SFW with gcc and builds cleanly now only with gcc
-COMPILER = gcc
```
pyhalov, line 40: What do you want to achieve with this change?

alarcher: These lines are obsolete. I added CPP="/usr/bin/gcc -E" so that the CPP hardcoded in guile-snarf is not dependent on the default gcc in oi-userland.

pyhalov: OK, got it.

pyhalov: Should have read bug report. Missed the fact that we have one :)

PR #2776, opened 2017-02-05
Align templates with prototypes from illumos-gate

mtelka:


transforms/manifest-metadata-template
```diff
@@ -5,7 +5,7 @@
 # 1.0 of the CDDL.
 #
 # A full copy of the text of the CDDL should have accompanied this
-# source. A copy of the CDDL is also available via the Internet at
+# source.  A copy of the CDDL is also available via the Internet at
```
pyhalov, line 8: Is there something wrong with spaces?

transforms/copyright-template
```diff
@@ -1,13 +1,14 @@
...
 # This file and its contents are supplied under the terms of the
-# Common Development and Distribution License ("CDDL"). You may
-# only use this file in accordance with the terms of the CDDL.
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
 #
 # A full copy of the text of the CDDL should have accompanied this
-# source. A copy of the CDDL is also available via the Internet at
+# source.  A copy of the CDDL is also available via the Internet at
```
pyhalov, line 8: Why do we have 2 spaces here?

alarcher: FYI I removed the double space in our template copyrights when I bumped to 2017. I thought it was a typo on our end.

mtelka: https://en.wikipedia.org/wiki/Sentence_spacing#Digital_age
http://src.illumos.org/source/xref/illumos-gate/usr/src/prototypes/prototype.Makefile#8

alarcher: OK. Previously we had a mix-and-match of both (first sentence had a single space) so I modified what thought was a typo. I do not mind as long as it is consistent.

alarcher (approved):

PR #2777, opened 2017-02-06
[WIP/DISCUSS] Dependency-tracking scripts

jimklimov:
Hi, following discussion on FOSDEM, I took a deeper look at the building blocks we need for building in zones, more reliable parallelization of component-building (considering dependencies), etc. - and particularly at bass-o-matic script.

The latter seems very... strange :) and most of its logic is not even used in our recipes (and perhaps for a good reason - I don't see how those bits could work). First of all it seems VERY unfinished or in a middle of re-work. It took a bit of pounding to just get it to do anything reasonable beside listing directories, and for the past tens of minutes it loops over my workspace with approx. 1402 found components so it would just report on what depends on what (`-c depends` codepath) - I hope it will finish today :)

I think its idea was to produce a series of dependencies as a file that can be fed into gmake, so when I ask it to 'make component_X' it will know which ones to refresh, if sources were also changed...
At least this sounds as a reasonable approach we could adopt, even if not intended by original authors :)

Optimization can be added in the form of caching the called commands' outputs `gmake print-(packaged|required)-(named|paths)`, to re-run quickly at least; this can be done with Make recipes that depend on current component's `Makefile`, maybe `$WS_TOP/make-rules/*.mk` (for common required-packages and/or required-paths which seem to not be set anywhere in our recipes), and maybe `*.p5m` in its dir (for listing of paths provided by package)

These listings could be possibly git-tracked and maybe even required, similar to sample-manifests and REQUIRED_PACKAGES we ask for today when components are added or bumped.

Also I *think* it currently lacks a recursive mode, to produce a chain of dependencies for requested component(s) with one command, and possibly minimizing the number of iterations and forks it does, and certainly the `--make` and zone support modes are utterly placeholders :) UPDATE: I see now that one usage of `--make` is present in the `components/Makefile` which my initial PR state might break, I'll look at it a bit later.

Also I *think* it currently is not able to provide an answer "what directory holds the Makefile responsible for certain PKG_FMRI?" which is a critical request for unrolling the dependency chain

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -1175,7 +1175,13 @@ endif
 
 include $(WS_MAKE_RULES)/environment.mk
 
-# A simple rule to print the value of any macro.  Ex:
+# Print the required packages as a listing, line by line
+print-required-packages::
+	@for fmri in $(REQUIRED_PACKAGES) ; do \
+		echo $$fmri ; \
+	 done
+
+# A simple rule to print the value of any macro (as MACRONAME=value).  Ex:
 #    $ gmake print-REQUIRED_PACKAGES
 # Note that some macros are set on a per target basis, so what you see
 # is not always what you get.
--- a/tools/bass-o-matic
+++ b/tools/bass-o-matic
@@ -234,10 +234,12 @@ def main():
         # TODO: Should this loop over component_paths to build each (if several given)?
         # TODO: Should there be an order of dependencies/parallelism for several paths?
         # TODO: Should this use or pass the template_zone value somehow?
-        if component_arg:
-            proc = subprocess.Popen(['gmake'] + args, cwd = workspace + subdir + component_arg)
-        else:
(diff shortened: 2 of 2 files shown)


jimklimov:
To ease readability, I'd refrain from updating older comments but would post new ones ;)

As of now, the `-c deps` mode works and does produce a long list of which source directory depends on which other source directories. I tee'd the output - it is 2.7Mb of text with over 25k lines, of which 18k are the 1:1 dependency delcarations (dir on dir style). Took 64 minutes to produce on my high-end laptop, of which I believe most time went into calling 4*1402 `gmake`s to produce series of empty or not outputs, so caching and other optimizations of this procedure (especially to repeat it iteratively in Jenkins) are very welcome.

In an inefficient way, this generated list already gives a way to run a helluvalot `make`s against known dependencies (and it is also recursable to a point) - usage of which can be greatly optimized by subsequently converting this output into a real `Makefile` that would be evaluated and stat'ed by one `gmake` process rather than hundreds independently.

PR #2780, opened 2017-02-07
GCC 6 fixes

alarcher:
- Missing recent patch "16 update cmn_err format specifier" from https://github.com/illumos/gcc/commits/il-4_4_4
- Fixed manifest, generated sample-manifest missed files.
- Added customized bug url and pkg version.


tsoome (approved):
seems good:)

PR #2384, opened 2016-09-05
Gstreamer1.0

alarcher:
Introspection is broken for multilib builds... needs work.

PR #2782, opened 2017-02-08
Add gst-plugins-base 1.10.3

alarcher:
Requires #2384

### DIFF
--- a/components/multimedia/gst-plugins-base1/Makefile
+++ b/components/multimedia/gst-plugins-base1/Makefile
@@ -1,0 +1,82 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016 Aurelien Larcher.  All rights reserved.
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME= gst-plugins-base1
+COMPONENT_VERSION= 1.10.3
+COMPONENT_SUMMARY= GNOME streaming media framework plugins
+COMPONENT_CLASSIFICATION= System/Multimedia Libraries
+COMPONENT_FMRI= library/audio/gstreamer1/plugin/base
+COMPONENT_SRC_NAME= gst-plugins-base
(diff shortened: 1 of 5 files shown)


components/desktop/gst-plugins-base1/Makefile
```diff
@@ -0,0 +1,81 @@
+#
```
pyhalov: Please, move them to components/multimedia to be consistent with other gstreamer* components.

components/desktop/gst-plugins-good1/Makefile
```diff
@@ -0,0 +1,91 @@
...
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PREP_ACTION = ( cd $(SOURCE_DIR) && aclocal -I ./m4 -I./common/m4 &&\
+                          autoheader &&\
+                          automake -a -f -c --gnu &&\
+                          autoconf )
+gcc_OPT = -O2
```
pyhalov: Is -O2 still necessary in these Makefiles? I've added it to gstreamer as I got really strange crashes with -O3 and some multimedia applications (now don't remember details).

components/desktop/gst-plugins-good1/Makefile
```diff
@@ -0,0 +1,91 @@
...
+
+PATH = /usr/gnu/bin:/usr/bin
+
+CFLAGS += -I/usr/sfw/include -I/usr/X11/include
+
+CONFIGURE_OPTIONS += --sysconfdir=/etc
+CONFIGURE_OPTIONS.32 += --libexecdir=/usr/lib
+CONFIGURE_OPTIONS.64 += --libexecdir=/usr/lib/$(MACH64)
+CONFIGURE_OPTIONS += --enable-external
+CONFIGURE_OPTIONS += --disable-sunaudio
```
pyhalov: Why sunaudio is disabled? I see it doesn't build on Solaris (as userland-gate comment states). Are we also affected?

components/desktop/gst-plugins-base1/Makefile
alarcher: Done.

components/desktop/gst-plugins-good1/Makefile
alarcher: I do not know but I'd rather be safe than sorry if it is known to work that way.

components/desktop/gst-plugins-good1/Makefile
alarcher: Last time I checked it did not compile and was somehow considered deprecated.

alarcher:
Just figured orc dependency should be bumped to be used because ours is too old.

PR #2794, opened 2017-02-16
Bump cirrus driver and set default depth to 16 bpp

Mno-hime:
Current xf86-video-cirrus X11 driver does not work with Qemu emulated
GD5446 graphics adapter (on Linux Qemu v2.6.2 at least) - opening
terminal results in distorted, visally unstable window.

This is now fixed with setting default depth to 16 bpp (was 24 bpp).

Also the driver was bumped to version 1.5.3 and `NoAccel` set as a
default. All changes are from upstream. Two OpenIndiana-specific
patches were updated.

Tested with current OpenIndiana on Linux Qemu KVM v2.8.0 with `-vga
cirrus`.

### DIFF
--- a/components/x11/xf86-video-cirrus/Makefile
+++ b/components/x11/xf86-video-cirrus/Makefile
@@ -11,18 +11,18 @@
 
 #
 # Copyright 2015 Alexander Pyhalov
+# Copyright 2017 Michal Nowak
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= xf86-video-cirrus
-COMPONENT_VERSION= 1.5.2
-COMPONENT_REVISION=1
+COMPONENT_VERSION= 1.5.3
 COMPONENT_SUMMARY= xf86-video-cirrus - Cirrus Logic video driver for the Xorg X server
 COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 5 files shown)


components/x11/xf86-video-cirrus/Makefile
```diff
@@ -11,18 +11,19 @@
...
 # Copyright 2015 Alexander Pyhalov
+# Copyright 2017 Michal Nowak
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= xf86-video-cirrus
-COMPONENT_VERSION= 1.5.2
+COMPONENT_VERSION= 1.5.3
 COMPONENT_REVISION=1
```
alarcher: You can remove COMPONENT_REVISION as the COMPONENT_VERSION was bumped.

Mno-hime:
@alarcher FYI.

alarcher (approved):

components/x11/xf86-video-cirrus/Makefile
alarcher: Stupid page just shifted when I clicked, so just fix the component revision and it should be fine.

alarcher (changes requested):
Just remove COMPONENT_REVISION and this is good.

Mno-hime:
`COMPONENT_REVISION` removed.

alarcher (approved):

alarcher:
@Mno-hime Thanks. Compiled and published locally.

PR #2797, opened 2017-02-18
build with the hipster default jpeg implementation

timmooney:
Update terminal/mrxvt to build with libjpeg8-turbo.

@pyhalov and @alarcher , please review my changes carefully.

I can't get either pkg://openindiana.org/terminal/mrxvt@0.5.4-2016.0.0.0:20160730T005926Z or my updated version that uses libjpeg8-turbo to load a background JPEG pixmap, but at least mrxvt seems to be functioning as it did before.

mrxvt had no REQUIRED_PACKAGES specified; I've only added image/library/$(JPEG_IMPLEM), not any of the other missing packages.

If these changes appear to be correct and are merged, I will update [the illumos issue tracker issue 7391](https://www.illumos.org/issues/7391) and then try some of the other "easy" components.

During testing, I was not able to install my updated terminal/mrxvt because it was blocked by userland-incorporation.  tomww from irc oi-dev suggested using a version unlock on terminal/mrxvt, which worked.  If that's preferred over uninstalling userland-incorporation, I'll also create a pull request for the docs to mention that testing updates for some components may require you to version unlock.

### DIFF
--- a/components/mrxvt/Makefile
+++ b/components/mrxvt/Makefile
@@ -16,6 +16,7 @@ include ../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		mrxvt
 COMPONENT_VERSION=	0.5.4
+COMPONENT_REVISION=     1
 COMPONENT_PROJECT_URL=	http://materm.sourceforge.net/wiki/pmwiki.php
 COMPONENT_SUMMARY=	Mrxvt - lightweight multi-tabbed terminal emulator
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
@@ -34,6 +35,11 @@ include $(WS_TOP)/make-rules/ips.mk
 
 PATH=/usr/gnu/bin:/usr/bin
 
+# build with the system default libjpeg
+CFLAGS            +=	$(JPEG_CPPFLAGS) $(JPEG_CFLAGS)
+CXXFLAGS          +=	$(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS)
+LDFLAGS           +=	$(JPEG_LDFLAGS)
+
 CONFIGURE_OPTIONS +=	--sysconfdir=/etc
 CONFIGURE_OPTIONS +=	--enable-everything
 CONFIGURE_OPTIONS +=	--disable-debug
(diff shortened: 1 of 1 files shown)


pyhalov:
Please, bump COMPONENT_REVISION and specify all REQUIRED_PACKAGES. For me decision to relax userland-incorporation or to uninstall it depends on the scale of issue. For example, I have it uninstalled on my build and test hosts, where I have to install different versions of many packages and have it relaxed on workstations, where I install one or another package to test sometimes.

alarcher:
Hi, very good!
I am surprised that gmake REQUIRED_PACKAGES does not detect any dependency.
As for the version-lock, I do the same thing as Alexander: it depends if the environment is a build zone or a machine I use for work.
I will test your branch by tomorrow.
Thanks.

timmooney:
I've added COMPONENT_REVISION=1 and incorporated the results of REQUIRED_PACKAGES into the Makefile, as requested by @pyhalov .

@alarcher , in my original comments I just meant that I hadn't attempted to add any of the other missing dependencies, not that the REQUIRED_PACKAGES target fails to find any dependencies.  I was trying to make the minimal amount of changes to accomplish the goal of converting it to libjpeg-turbo8.  With the feedback from Alexander, I see that I should have just fixed the missing dependencies too.

Note that when I 'gmake publish', I get warnings from pkglint.action005.1 about obsolete dependency check skipped for each of the packages that were added to the REQUIRED_PACKAGES lines in the Makefile.

Please review and let me know if additional changes are required.

alarcher:
@timmooney Sorry for the late reply, I have been stuck with work. Thanks for your help! As you did it with this package the steps will always be the same if you modify a package: 1) bump COMPONENT_REVISION 2) apply the modifications 3) gmake REQUIRED_PACKAGES to regenerate the list of *runtime* dependencies. Make sure libjpeg8-turbo appears in the list: if not it means the the components build system needs modification to pick the libjpeg we want.

Finally you can make sure that the right library is picked by using ldd on the binaries and libraries:

```
alarcher@helios:/home/alarcher$ ldd /usr/bin/mrxvt 
        libXft.so.2 =>   /usr/lib/libXft.so.2
        libfontconfig.so.1 =>    /usr/lib/libfontconfig.so.1
        libXrender.so.1 =>       /usr/lib/libXrender.so.1
        libjpeg.so.8 =>  /usr/lib/libjpeg8-turbo/lib/libjpeg.so.8
...
```

Kind regards,

Aurélien

PR #2812, opened 2017-02-28
Bump sessreg to 1.1.1

alarcher:
Fresh from this morning: https://lists.x.org/archives/xorg-announce/2017-February/002778.html

### DIFF
--- a/components/x11/sessreg/manifests/sample-manifest.p5m
+++ b/components/x11/sessreg/manifests/sample-manifest.p5m
@@ -10,7 +10,7 @@
 #
 
 #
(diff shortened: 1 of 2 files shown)


components/x11/sessreg/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2016 <contributor>
+# Copyright 2017 <contributor>
```
AndWac: You should insert your name here.

PR #2822, opened 2017-03-01
Bump libjpeg-turbo to 1.5.1

alarcher:


pyhalov:
What about ABI compatibility?

alarcher:
@pyhalov The JPEG ABI is specified for each version v6b, v7, and v8. When the ABI is broken -- like it has been for ARM at some point --, this is a bug. I flagged the PR for testing to avoid such bad surprise.
Since we aim for ABI 8 as default, should I place symlinks from /usr/lib to the *.pc files to allow auto-detection?

pyhalov:
Likely...

alarcher:
@pyhalov Updated.

pyhalov (approved):
Please, merge commits before merging into oi/hipster.
I hope you've tested that at least some libjpeg-dependent apps work (and build) with new libjpeg-turbo.

alarcher:
@pyhalov Squashed. So far the only apps depending on libjpeg8-turbo are the ones migrated away from IJG's (3 only) and VLC depends on libjpeg6-turbo.
For VLC the dependency was not found before and now listed as duplicate, I will push the modification to the manifest.
All apps were all tested to build and run.

PR #2831, opened 2017-03-03
7479 pulseaudio can't manage OSS

pyhalov:


alarcher (approved):

alarcher:
Actually I think I kept a similar patch for 8.0 on my workstation.

PR #2836, opened 2017-03-04
Make Java 8 the default one

pyhalov:


jimklimov:
Didn't a higher mediator version fulfill the "defaultness"? IMHO removing the ability to mediate is not a good step.

pyhalov:
Removing the ability to mediate /usr/java is by design. Services will be surprised when it's JDK 7.

jimklimov:
services are in packages? can they depend on (and so toggle) a particular mediation?
Also, they can just set a particular JAVA_HOME in their settings (e.g. env via SMF).

PR #2870, opened 2017-03-12
pkg: Added REQUIRED_PACKAGES

mtelka:


alarcher (approved):

pyhalov:
If we do it, can we add all required packages?
At least there's developer/opensolaris/pkg5

mtelka:
I do not know how to identify other required packages (gmake REQUIRED_PACKAGES does not work for this component).  I added system/zones/internal only because I found that this package only is missing from my system and gmake env-check didn't complain about it.  If you (or somebody else) knows about other missing REQUIRED_PACKAGES feel free to add them.

pyhalov:
At least, add developer/opensolaris/pkg5 while touching it

mtelka:
For what purposes is developer/opensolaris/pkg5 needed? I do not have installed this package and the component was built successfully.

pyhalov:
It's a meta-package, "official" list of dependencies required to build ips - look at https://github.com/OpenIndiana/pkg5/blob/oi/src/pkg/manifests/developer:opensolaris:pkg5.p5m#L32

mtelka:
Ah, okay.  Makes sense then.

PR #2879, opened 2017-03-14
system/library/math/header-math was renamed

mtelka:


components/meta-packages/build-essential/build-essential.p5m
```diff
@@ -43,7 +43,7 @@ depend fmri=developer/sunstudio12u1 type=require
 
 # Headers
 depend fmri=system/header type=require
-depend fmri=system/library/math/header-math type=require
+depend fmri=system/library/math type=require
```
pyhalov: Please, if you add system/library/math, add it to 'Useful libraries' section. Also note that you don't need it as system/library depends on system/library/math

pyhalov:
I hope you've rebuilt all of them to find out that they are still buildable? :)

mtelka:
Yes. I rebuilt all components I touched.  Initially, I just replaced system/library/math/header-math by system/library/math, but I found that clang and gcc builds are failing with that because the system/library/math dependency is twice there, so I had to simply remove system/library/math/header-math for such components.

components/meta-packages/build-essential/build-essential.p5m
mtelka: You are right.  The system/library/math is not needed since it is already included by indirect dependencies.

PR #2880, opened 2017-03-14
pygobject-3: add python 3.4 version

pyhalov:


pyhalov:
Should be merged when all pycairo fixes are built and installed on the build server

PR #2888, opened 2017-03-16
X 201704 input

alarcher:
Extracted input driver updates from X_S12_2016_7.


pyhalov:
Have you checked that this builds with our ancient xorg? BTW, why did you have to add patch to xf86-input-mouse ?

alarcher:
I built and installed them in my VirtualBox image with Xorg 1.14, and on my workstations with 1.18, and 1.19. Without the patches the fixed components only build with 1.14.

PR #2900, opened 2017-03-22
Add rest 0.8.0

pyhalov:


jimklimov:
Why is component 0.8.0 delivering paths with 0.7 all over the place? Typo or intentional?

pyhalov:
Because it delivers 0.7 all over the place :) Rechecked twice

PR #2910, opened 2017-03-23
gtk+3: The textiter test passes in 32 bit

mtelka:


pyhalov:
Hm... On my build zone I see:
```TEST: textiter... (pid=29261)
  /TextIter/Search Empty:                                              OK
  /TextIter/Search Full Buffer:                                        OK
  /TextIter/Search:                                                    OK
  /TextIter/Search Caseless:                                           OK
  /TextIter/Forward To Tag Toggle:                                     OK
  /TextIter/Word Boundaries:                                           OK
  /TextIter/Visible Word Boundaries:                                   OK
  /TextIter/Cursor Positions:                                          OK
  /TextIter/Visible Cursor Positions:                                  OK
  /TextIter/Sentence Boundaries:                                       **
ERROR:textiter.c:644:check_sentence_boundaries: assertion failed (inside_sentence == gtk_text_iter_inside_sentence (&iter)): (0 == 1)
FAIL
GTester: last random seed: R02S22f72ba1643553f6f40a9f12d2f31b2b
(pid=29275)
FAIL: textiter
```
But they really pass on build server.  Need to find 10 differences.

pyhalov:
Updated build zone - the same result...

pyhalov:
Checked different locale settings, x11 availability... still no luck, have to actually read the code :)

pyhalov:
The things I've noticed. When we run test once on one Xvfb instances, it always succeeds. If we ran something before it,. it fails. The failure so far is reproducible only on my build server. If we add usleep(10) between steps of the test, it also succeeds (even if something was run before it on the same Xvfb instance).

pyhalov:
With -O0 the test passes...

pyhalov:
Also test fails only when it's run under gtester control.

mtelka:
Maybe you use a different gcc version?

pyhalov:
I use pkg://openindiana.org/developer/gcc-49@4.9.4-2017.0.0.1:20170315T101457Z . The issue is reproducible if I compile gtk3 with gcc-5 as well.

pyhalov:
https://github.com/OpenIndiana/oi-userland/pull/2916

pyhalov:
There are strange issues which can lead to test failures when they run under gtester. Compiling without -zignore heals them in the build environment of the component. But still if test is compiled separately with -O3 it fails. So we add -O2 just to be sure... Some strange ld/gcc issue...

PR #2911, opened 2017-03-24
time-slider: The library/python-2/python-notify-27 package was renamed

mtelka:


components/openindiana/time-slider/Makefile
```diff
@@ -21,7 +21,7 @@ COMPONENT_VERSION=	0.2.101
 COMPONENT_LICENSE=	CDDL
 COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
 
-COMPONENT_REVISION=$(shell cd $(COMPONENT_SRC); git rev-list HEAD --count)
+COMPONENT_REVISION=$(shell cd $(COMPONENT_SRC); git rev-list HEAD --count).1
```
pyhalov: Is it something local? Please, remove it. I know, we usually bump COMPONENT_REVISION, but here it is generated on the base of component's git revision, so this component is a bit special.

mtelka: It is just to uprev the package version/revision so it is installed on the next update.  Strictly speaking it is not needed, but we would need to wait for the next push to the time-slider repo.

The other option to force the package version update is to make a fake push to the time-slider repo.

Or, should I just update the COMPONENT_VERSION to 0.2.102 or 0.2.101.1?

pyhalov: I'd just let it distinguish only by timestamp

mtelka: Changed timestamp is not enough to get the package updated.  If it is, then we do not need to change COMPONENT_REVISION in every package we touch.

pyhalov: It is. We change COMPONENT_REVISION to achieve 2 goals:
1) clearly distinguish different package versions (at least for diagnostic purposes if not for fixing with userland-incorporation),
2) always trigger rebuild on component change.

mtelka: Yes.  This is exactly what I want to achieve.  Both, updated incorporated version in the userland-incorporation, and to trigger a rebuild of the component.

pyhalov: This will get us time-slider@0.2.101-2017.0.0.107.1 version instead of time-slider@0.2.101-2017.0.0.107. Seems a bit strange. As for rebuild, I'll just do gmake clean on build server. And we don't want rebuild for this change, just republishing. 
Sorry, we don't do such things for components like illumos-gate, gfx-drm and so on. If it was really important, we could bump UPDATENUM , but it's just a trifle.

mtelka: And the republishing needs the version increment.  Both in the component and in the userland-incorporation.  For illumos-gate it is not a big deal since the gate changes very often.  If we do some packaging changes for gfx-drm (outside the gfx-drm gate) we should increment the gfx-drm version too.

But as I understood you hates incorporations, so I'll remove the version change.  Better to have this fixed with the same version number, than not fixed at all.

And, there is nothing strange with time-slider@0.2.101-2017.0.0.107.1 :-).

mtelka:
Here it is...

PR #2916, opened 2017-03-26
gtk+3: textiter test can fail sporadically

pyhalov:


pyhalov:
There are strange issues which can lead to test failures when they run under gtester. Compiling without -zignore heals them in the build environment of the component. But still if test is compiled separately with -O3 it fails. So we add -O2 just to be sure... Some strange ld/gcc issue...

PR #2930, opened 2017-03-29
4678 lms and heci can probably be EOF'd

pyhalov:


pyhalov:
Waiting for illumos-gate commit

PR #2934, opened 2017-03-30
8012 Add PATH symbols to shared-macros.mk

jgmills:
This change has no effect by itself, but once the component Makefiles are changed to use the symbols to set PATH, a change in shared-macros.mk will be propagated to all the component Makefiles.


jimklimov:
Neat :)

My equivalent PR fell under the rug somewhere and got lost ;)

PR #2939, opened 2017-03-31
bass-o-matic: refactor

xen0l:
This PR remove functionality, which was never finished by Oracle and we don't need it (yet). We also do following changes:
- switch from getopt to argparse
- use scandir.walk if available
- user proper debug logging

Another big feature is support for incremental builds (more details will be posted later).

### DIFF
--- a/tools/bass-o-matic
+++ b/tools/bass-o-matic
@@ -27,36 +27,48 @@
 #  on dependency related information.
 #
 
+from __future__ import print_function, absolute_import
+
 import os
 import sys
 import re
 import subprocess
+import argparse
+import logging
+
+try:
+    from scandir import walk
+except ImportError:
+    from os import walk
+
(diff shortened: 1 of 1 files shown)


jimklimov:
@xen0l : see also #2777 and #2940 for more inspiration
I like your PR's style, it seems cleaner ;)

xen0l:
@jimklimov thanks, I got stranded by 'print package-required-paths' not working and reworked it to 'print-required-packages' when I got basically something that look like your code. 

My main focus right now is to finish '--changed' logic, so we can speed up jenkins build time from 4h to something more reasonable. I will consider adding 'print-required-packages' as a part of this PR, but that is not a finished thing as getting a list of required packages is one part of the problem. Solving dependency graph issues and transform that into correct build order (and do correct parallelization) is another issue.

PR #2940, opened 2017-03-31
Speed up bassomatic search for component dirs

jimklimov:
Instead of walking the FS, run with the git index

Also does not give irrelevant hits e.g. for removed components with remaining build areas, which the older walk routine did return.

### DIFF
--- a/tools/bass-o-matic
+++ b/tools/bass-o-matic
@@ -35,13 +35,13 @@ import subprocess
 # Locate SCM directories containing Userland components by searching from
 # from a supplied top of tree for .p5m files.  Once a .p5m file is located,
 # that directory is added to the list and no children are searched.
-def FindComponentPaths(path, debug=None, subdir='/components'):
+def FindComponentPaths_walk(path, debug=None, subdir='/components'):
     expression = re.compile(".+\.p5m$", re.IGNORECASE)
 
     paths = []
 
     if debug:
-        print >>debug, "searching %s for component directories" % path
+        print >>debug, "searching %s for component directories by walking FS" % path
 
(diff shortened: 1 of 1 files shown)


jimklimov:
Note for related question: to search for Makefiles (which is what triggers rebuild of components) changed since last build (commit ID assumed stashed somewhere or git-tagged), you can do something like

    git diff oldcommitid --name-only | egrep '/Makefile$'

pyhalov:
I like this idea, but a) for users builds we need a mode where bass-o-matic lists files, which are not in git;
b) I'd left full review to @xen0l .

jimklimov:
Yes, this is why the dispatcher routine at original func-name, and code for original solution, are there. Just someone with better python skills has to bind them together with some logic or config option :)

And since it seems @xen0l spearheads the improvements here, I'd better leave such reformatting and integration nuances to him so it is done properly :)

xen0l:
@jimklimov I like the idea, but @pyhalov's comment are relevant, we need to have a way how to switch between these (Ideally with some clear magic!).

pyhalov:
Based on working in jenkins or regular environment? Our jobs, for example, export JENKINS_BUILD_NUMBER.

jimklimov:
Regarding files in a developer's workspace that are not yet in Git - one approach might be to require git-tracking all the development :-) and another would be to call `git ls-files` with additional options like `--cached --others` to show untracked files as well...

Jenkins env sounds good, maybe a more generic solution would be some setting in the root Makefile (so part of each workspace setup). Note that we seem to require GNU make and it has optional include keywords `sinclude` or `-include` which do not error out if the file is missing - that could be used to provide the customizations in specific use-cases.

PR #2942, opened 2017-04-01
Fix and move zziplib to archiver directory

alarcher:


components/archiver/zziplib/manifests/sample-manifest.p5m
```diff
@@ -88,6 +88,9 @@ link path=usr/lib/$(MACH64)/libzzipmmapped-0.so.13 \
     target=libzzipmmapped-0.so.13.0.62
 file path=usr/lib/$(MACH64)/libzzipmmapped-0.so.13.0.62
 link path=usr/lib/$(MACH64)/libzzipmmapped.so target=libzzipmmapped-0.so.13.0.62
+link path=usr/lib/$(MACH64)/libzzipwrap-0.so.10 target=libzzipwrap-0.so.13
```
xen0l, line 91: Why these changes?

xen0l (changes requested):

components/archiver/zziplib/manifests/sample-manifest.p5m
alarcher: It seems they were missing.

PR #2955, opened 2017-04-05
fix ~~ expansion in tcsh

hrosenfeld:
tcsh tries to expand ~~ and fails, which according to its own manpage it shouldn't.

This is a simple fix that really only fixes the case of ~~, not any other combinations of ~ with special characters that aren't valid starting characters of user names. This fix has been integrated into tcsh today and will be in the next release.


pyhalov:
Please add COMPONENT_REVISION=1

hrosenfeld:
Done.

pyhalov:
will merge with next brasero-related changes

PR #2958, opened 2017-04-05

jgmills:
This PR contains only the one commit that I intended in the first place.

### DIFF
--- a/components/desktop/gnome2/gnome-games/Makefile
+++ b/components/desktop/gnome2/gnome-games/Makefile
@@ -9,6 +9,7 @@
 #
 
 #
+# Copyright 2017 Gary Mills
 # Copyright (c) 2014 Alexander Pyhalov
 #
 
@@ -29,7 +30,8 @@ include $(WS_TOP)/make-rules/prep.mk
 include $(WS_TOP)/make-rules/configure.mk
 include $(WS_TOP)/make-rules/ips.mk
 
-PATH=/usr/gnu/bin:/usr/bin
+# Need automake 1.11
+PATH = $(PATH.gnu)
 
 COMPONENT_PREP_ACTION =        ( cd $(@D) && \
 					libtoolize --automake --copy --force && \
--- a/components/python/jinja2/Makefile
+++ b/components/python/jinja2/Makefile
@@ -11,11 +11,12 @@
 #
 # Copyright (c) 2014 Alexander Pyhalov.  All rights reserved.
 # Copyright 2015, PALO Richard.
+# Copyright 2017 Gary Mills
 #
 
 include ../../../make-rules/shared-macros.mk
 
-PATH=/usr/bin:/usr/gnu/bin:/usr/sbin
+PATH = $(PATH.ill):/usr/sbin
 
 COMPONENT_NAME=		Jinja2
 COMPONENT_VERSION=	2.8
--- a/components/desktop/gnome2/at-spi/Makefile
+++ b/components/desktop/gnome2/at-spi/Makefile
@@ -10,6 +10,7 @@
 #
 
 #
+# Copyright 2017 Gary Mills
 # Copyright 2015 Alexander Pyhalov
 #
 
@@ -35,7 +36,7 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 
-PATH=/usr/gnu/bin:/usr/bin
+PATH = $(PATH.gnu)
 
 COMPONENT_POST_UNPACK_ACTION = ( cd $(SOURCE_DIR) && cp -r pyatspi pyatspi-26)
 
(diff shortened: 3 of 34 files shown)


components/desktop/gnome2/gnome-games/Makefile
```diff
@@ -29,7 +30,8 @@ include $(WS_TOP)/make-rules/prep.mk
 include $(WS_TOP)/make-rules/configure.mk
 include $(WS_TOP)/make-rules/ips.mk
 
-PATH=/usr/gnu/bin:/usr/bin
+# Need automake 1.11
```
pyhalov, line 33: Why do you add these comments? We use automake 1.15, all components which required automake 1.11, as far as I know, were patched.

pyhalov (commented):
Sorry, I don't understand the point of doing this. Why do we want it?

components/desktop/gnome2/gnome-games/Makefile
xen0l: I would clarification on this as well.

components/python/jinja2/Makefile
```diff
@@ -11,11 +11,12 @@
 #
 # Copyright (c) 2014 Alexander Pyhalov.  All rights reserved.
 # Copyright 2015, PALO Richard.
+# Copyright 2017 Gary Mills
 #
 
 include ../../../make-rules/shared-macros.mk
 
-PATH=/usr/bin:/usr/gnu/bin:/usr/sbin
+PATH = $(PATH.ill):/usr/sbin
```
xen0l, line 19: can we change this to PATH.illumos, so it reflects the full implementation name? We use "gnu" for GNU implementation, so we should use "illumos", not "ill". Besides, PATH.ill is also a little bit weird as ill also has a different meaning.

xen0l (changes requested):

xen0l:
@jgmills thanks for working on this. I have few comments:

- we should change PATH.ill to PATH.illumos, so we consistent in implementaiton naming (we don't abbreviate "gnu", so let's no do that for illumos)
- Remove "need automake 1.11" comments
- consider adding /usr/perl5/bin to variable as well, so we can unite PATH configuration in everywhere

components/desktop/gnome2/gnome-games/Makefile
jgmills: I presume you are objecting to the comment.  Indeed, it is misleading and incorrect in most cases.  I'll remove all of those comments and produce a new PR.

components/python/jinja2/Makefile
jgmills: I used PATH.ill because I wanted to keep it the same length as PATH.gnu .  I could change it to PATH.illumos or anything else, but it has to match the symbol already defined in make-rules/shared-macros.mk .  This would need another commit.  I'd also have to change about 50 component Makefile to match it, and feed them through in separate commits.  If we want to change it, we have to decide right now.

jgmills: I didn't define a symbol for /usr/perl5/bin because that's only one of many binary directories that appear in the PATH of some component Makefiles.  I only did so for the two most common ones, which is all that we need in practice.  My use was to add two directories for hand-built and hand-installed software, when it was not available as packages.

components/desktop/gnome2/gnome-games/Makefile
alarcher: Just to make clear that you do not need to open a new PR: modify the branch and push -f to overwrite the content, the PR will be updated accordingly.

components/python/jinja2/Makefile
pyhalov: Can we go with a more generic PATH.illumos and PATH.gnu, which would cover 90%cases? I think we could include /usr/sbin (commonly used for dtrace) and /usr/perl5/bin in both of them.

jgmills: I examined 101 component Makefiles.  11 of them contained /usr/sbin in the PATH, and six of them contained /usr/perl5/bin .  There were no others.  I can append /usr/sbin:/usr/perl5/bin to the two PATH symbols in shared-macros.mk .  I'll also change the name of PATH.ill to PATH.illumos in that same file.  Fixing that will be one commit.  The second commit will update a batch of component Makefiles.

pyhalov: Sounds good

pyhalov:
Why did you close it? Just add necessary changes, ammend (or rebase -i ) commits and force push changes to this branch.

jgmills:
I didn't specifically close it.  I deleted the branch on my github fork of oi-userland .  I used that branch both for the first commit that updated shared-macros.mk and for the second commit that proposed updating a batch of Makefiles.  I assume that github closed the PR as a side effect.  In any case, I have to go back to the first commit now, the one that updated shared-macros.mk only.  The change is in oi-userland now, but not in my fork.  I need to get it back in my fork now.  Very likely I'm doing this all wrong, but I am following the recommended procedure for contributing to hipster.

PR #2962, opened 2017-04-06
exiv2: add to oi-userland

cgrzemba:
Please add to oi-userland, with exiv2 gthumb is able to handle exif data of pictures.
but check if the subdirectory 'library' the right place for the recipe, for me it is a library but it ships also a CLI command


components/library/exiv2/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_ARCHIVE_URL= \
+  http://www.exiv2.org/$(COMPONENT_ARCHIVE)
+
+COMPONENT_PROJECT_URL = http://www.exiv2.org
+COMPONENT_FMRI = image/library/exiv2
+COMPONENT_LICENSE = GPLv2
+COMPONENT_LICENSE_FILE = COPYING
+COMPONENT_CLASSIFICATION = System/Multimedia Libraries
+
+include $(WS_TOP)/make-rules/prep.mk
```
xen0l: $(WS_TOP)/make-rules -> $(WS_MAKE_RULES)

components/library/exiv2/exiv2.p5m
```diff
@@ -0,0 +1,90 @@
...
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/exiv2
+file path=usr/bin/exiv2
+file path=usr/include/exiv2/basicio.hpp
+file path=usr/include/exiv2/bmpimage.hpp
+file path=usr/include/exiv2/config.h
```
xen0l: You are compiling both 32bit and 64bit. Are the headers the same?

components/library/exiv2/Makefile
```diff
@@ -0,0 +1,50 @@
...
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+     $(CLONEY) $(SOURCE_DIR) $(@D)
+
+# linking 64bit '-m64' option is missing
+LDFLAGS += -m$(BITS) -lnsl -lsocket
```
pyhalov: We have LDFLAGS =$(LD_BITS) by default. Why is it needed to add -m$(BITS)?

components/library/exiv2/Makefile
pyhalov: Please, switch to $(WS_MAKE_RULES)

xen0l (changes requested):
Library category is fine by me. What abotu REQURIED_PACKAGES? Aparat from these comment reviews, LGTM.

components/library/exiv2/Makefile
```diff
@@ -0,0 +1,50 @@
...
+include $(WS_TOP)/make-rules/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+     $(CLONEY) $(SOURCE_DIR) $(@D)
+
+# linking 64bit '-m64' option is missing
+LDFLAGS += -m$(BITS) -lnsl -lsocket 
+
+# using mkdir option '-v'
+COMPONENT_BUILD_ARGS += PATH="/usr/gnu/bin:$(PATH)"
```
pyhalov: Perhaps, better do it on top level? Just set 
PATH=$(PATH.gnu)
in the beginning (somewhere around 37 line).

components/library/exiv2/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_INSTALL_ARGS += PATH="/usr/gnu/bin:$(PATH)"
+
+build: $(BUILD_32_and_64)
+
+install: $(INSTALL_32_and_64)
```
pyhalov: Add test target (at least with $(NO_TESTS), but if library has some tests, better to use them).

components/library/exiv2/exiv2.p5m
```diff
@@ -0,0 +1,90 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/exiv2
+file path=usr/bin/exiv2
```
pyhalov: Why deliver both? Perhaps, just deliver 64-bit copy in /usr/bin?

components/library/exiv2/exiv2.p5m
```diff
@@ -0,0 +1,90 @@
...
+file path=usr/include/exiv2/basicio.hpp
+file path=usr/include/exiv2/bmpimage.hpp
```
pyhalov: Have you compared include files, delivered by 32 and 64 bits build? If they are different, this should be accounted for (for example as it's done in python - look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/python/python27/Makefile#L102 an component's manifest.

components/library/exiv2/Makefile
pyhalov: Also component misses REQUIRED_PACKAGES

components/library/exiv2/Makefile
cgrzemba: no idea, on linking with g++ it is missing

components/library/exiv2/Makefile
cgrzemba: This construct exists nowhere, but I can set 
PATH:=$(GNUBIN):$(PATH)

components/library/exiv2/exiv2.p5m
cgrzemba: do we have something like INSTALL_64_LIBS_ONLY?

components/library/exiv2/exiv2.p5m
cgrzemba: The headers are all the same

components/library/exiv2/exiv2.p5m
cgrzemba: or BUILD_32_LIBS_ONLY

components/library/exiv2/Makefile
pyhalov: I've removed it and it still works

components/library/exiv2/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+# mkdir option '-v' is used
+PATH:=$(GNUBIN):$(PATH)
+
+# g++ linking 64bit '-m64' option is missing
+LDFLAGS += -m$(BITS) -lnsl -lsocket 
+
+build: $(BUILD_32_and_64)
+
+test: $(TEST_32_and_64)
```
pyhalov: 1) style - test target follows install target
2) gmake test fails (as exiv2 doesn't support check target), so  change it to test: $(NO_TESTS)

cgrzemba:
Done.

components/library/exiv2/Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+     $(CLONEY) $(SOURCE_DIR) $(@D)
+
+# mkdir option '-v' is used
+PATH:=$(GNUBIN):$(PATH)
```
pyhalov: One small issue: in spite of recent @jgmills changes you can use just 'PATH=$(PATH.gnu)'

pyhalov (changes requested):

pyhalov:
ehm? Why does this PR contain > 250 commits?

cgrzemba:
I fight against the rebase nightmare

alarcher:
If you git pull --rebase origin oi/hipster there should not be any issue.

PR #2969, opened 2017-04-09
Bump boost to 1.64.0

alarcher:
Time to bump boost and move it to the library directory.
Opening a PR right away to allow for testing.
We will need to bump the component revision of dependent packages when this PR is ready to be merged.


pyhalov:
An issue here is that we don't know what components require it as build dep. Often only headers are used... Need to check 
a) mysqls 
b) ekiga
who knows what else...

alarcher:
Only ekiga has a dependency on boost libraries, other packages depend on headers only.

alarcher:
MariaDB packages rebuild correctly with the new version.

PR #2972, opened 2017-04-10

jgmills:
This commit contains all the changes suggested by the reviewers.


components/sysutils/lsof/Makefile
```diff
@@ -30,6 +31,9 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/justmake.mk
 include $(WS_MAKE_RULES)/ips.mk
 
+# Need GNU version of yes
+PATH = $(PATH.gnu)
+
```
pyhalov: Builds for me with PATH=$(PATH.illumos) 
Do I miss something?

pyhalov:
If you touch component, please, move it to corresponding category (for components under components/), change $(WS_TOP)/make-rules to  $(WS_MAKE_RULES)/ and add REQUIRED_PACKAGES.

components/sysutils/lsof/Makefile
jgmills: That comment is a warning for other developers.  The illumos version of yes, at least the one supplied with oi_151a8, ignores SIGPIPE.  This causes yes to never terminate.  Here's part of my truss:

write(1, "\n\n\n\n\n\n\n\n\n\n\n\n".., 5120)    Err#32 EPIPE
    Received signal #13, SIGPIPE [ignored]
write(1, "\n\n\n\n\n\n\n\n\n\n\n\n".., 5120)    Err#32 EPIPE
    Received signal #13, SIGPIPE [ignored]
write(1, "\n\n\n\n\n\n\n\n\n\n\n\n".., 5120)    Err#32 EPIPE
    Received signal #13, SIGPIPE [ignored]

This only happens when `gmake build' is run from the components directory.  It works correctly when only that component is built.

Perhaps yes has been fixed in later versions of illumos.  Perhaps hipster is already running the GNU version of yes.

My solution was the change PATH so that the GNU version is always selected.

pyhalov: Is it related to https://www.illumos.org/issues/3927 ? If so, I don't think that we need workaround for bug fixed 3 years ago.

jgmills: Yes, it sounds like the same bug.  As I mentioned, I'm building under oi_151a8, the most recent OI version available on the SPARC hardware platform.  I do need the workaround.  People building under hipster for x86 hardware don't need it, but that workaround shouldn't cause any problems for them.  Shall I make that comment more specific?

jgmills:
I don't understand your `if you touch component' statement.  I've never seen it before on the wiki.  What's the purpose of doing this?  I have tested all of the component Makefiles I've changed, and have gotten successful builds for all of them.

pyhalov:
I understand... Well, we are in process of migrating to https://wiki.openindiana.org/oi/Categories+layout and some style cleanup.  As we don't want to make one big 'fix style' commit, we do it on as-we-go basis...

components/sysutils/lsof/Makefile
pyhalov: OK, please, make comment more specific.

pyhalov:
As for wiki articles - I've tried to summarize some common issues in https://wiki.openindiana.org/oi/oi-userland+-+best+practices

components/sysutils/lsof/Makefile
jgmills: Okay, I've done that.  The comment now reads:

# In oi_151a8, illumos version of yes ignores SIGPIPE, never exits

pyhalov:
Please:
- [ ]  change $(WS_TOP)/make-rules to $(WS_MAKE_RULES);
- [ ]  move components to appropriate category;
- [ ] add REQUIRED_PACKAGES where they are missing.

jgmills:
Is the category the first word of the COMPONENT_FMRI in the component Makefile?  For example, the first file changed of the 33 of this commit is components/desktop/gnome2/at-spi/Makefile .  The COMPONENT_FMRI is gnome/accessibility/at-spi .  So, the component should be moved to components/gnome/gnome2/at-spi .  Is that correct?  Or, should it be components/gnome/accessibility/at-spi ?

alarcher:
For reference: https://wiki.openindiana.org/oi/Categories+layout

at-spi is at the correct location.
We only limit ourselves to the first level for categories, with the exception of desktop which is also (sub)categorized by desktop environment.

jgmills:
Okay, the only ones I need to move are the ones without their own subdirectory, like these ones:

components/gnome-mag/Makefile
components/gnome-media/Makefile
components/gnome-nettool/Makefile
components/gnome-utils/Makefile

They should all go under components/desktop I assume.  There's also these ones:

components/libcanberra/Makefile
components/libchamplain/Makefile
components/libunique/Makefile

They should all go under components/library .  Is that all correct?

pyhalov:
Please, move components/gnome* to components/desktop/gnome2, libraries - to components/library

PR #2976, opened 2017-04-11
Bump efl to 1.18

alarcher:
Post-snapshot goodie: back from underworld, death by epoll.
Moved to 64bit only.


pyhalov:
can we update and merge it?

PR #2988, opened 2017-04-15
Ant Revision - Add support to script target - Fixed

fronald:
Fixed manifest. The cause of build fail.

### DIFF
--- a/components/developer/ant/Makefile
+++ b/components/developer/ant/Makefile
@@ -20,13 +20,15 @@
 #
 
 #
-# Copyright (c) 2011, 2012, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c), 2011, 2012, Oracle and/or its affiliates. All rights reserved.
+# Copyright 2017, Wiselabs Software Ltda - http://www.wiselabs.com.br. All rights reserved.
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		ant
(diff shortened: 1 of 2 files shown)


AndWac:
Then where is the fixed manifest? Did you miss to add it to the commit?

pyhalov:
Why initially these files appeared into manifest? Why there were a lot of errors during javadoc generation?

fronald:
@AndWac My mistake was update Manifest with my build environment. I run gmake sample-manifest again and the p5m maintains equals to original revision. I do a "git squash", then my fix on p5m remains equals to original version and disappear of the commit.

fronald:
@pyhalov All errors in generating JavaDoc occurred by "optional tasks" thats depends of Jars in CLASSPATH to be compiled.
My only fix in this revision was put "commons-bsf" into classpath to enable "script" tags into build.xml, just to satisfy the property "bsf.present" (needed to compile jEdit).
Ant have a lot of optional tasks, but the JavaDoc run like all is satisfied.
I can make a revision to enable ALL optional modules, and not only the module that i need to compile jEdit. Today, this ANT compilation in oi-userland is very very basic, many tasks (like FTP) is not enabled too.

fronald:
@pyhalov I think that is better pause my jEdit port and work to enable all modules into Apache Ant. If interest. With this, all javadoc errors disappear and any software using any ant feature can be compiled.

PR #2990, opened 2017-04-17

jgmills:
This time I've moved seven components to new locations.  I've also changed all the Makefiles to use $(WS_MAKE_RULES) instead of $(WS_TOP)/make-rules .  As well, I've added REQUIRED_PACKAGES by running `gmake REQUIRED_PACKAGES' where it was missing.  In three cases, this failed with an error that some files could not be found.  In all cases, the build succeeded.  Is this adequate now?


components/desktop/gnome2/gnome-media/Makefile
```diff
@@ -64,4 +65,4 @@ test:		$(NO_TESTS)
 
 BUILD_PKG_DEPENDENCIES = $(BUILD_TOOLS)
 
-include $(WS_TOP)/make-rules/depend.mk
```
pyhalov, line 67: please, remove depend.mk and BUILD_PKG_DEPENDENCIES

jgmills: Do you mean from all of the files, or just this one?  I see five Makefiles that contain at least one of these lines.  Will this removal facilitate building `gmake REQUIRED_PACKAGES'?

pyhalov: It was a part of outdated logic to facilitate finding required packages, which was once used in userland-gate, but was never used in oi-userland (and was replaced by REQUIRED_PACKAGES rules). So, we remove these lines and eventually will remove make-rules/depend.mk.

PR #2992, opened 2017-04-17
vala: move to category runtime and adapt Makefile and manifest

AndWac:
I have changed the COMPONENT_FMRI from developer/vala to runtime/vala and removed the superfluous COMPONENT_DESCRIPTION (copy of COMPONENT_SUMMARY).


alarcher:
You need to add a history file to trigger the rename, see https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/developer/clang-38/history

components/runtime/vala/Makefile
```diff
@@ -19,21 +19,28 @@
...
+# Copyright (c) 2017 Andreas Wacknitz
 #
-include ../../make-rules/shared-macros.mk
+include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		vala
 COMPONENT_VERSION=	0.20.1
+COMPONENT_REVISION=	1
+COMPONENT_LICENSE=	LGPL2.1
+COMPONENT_LICENSE_FILE=	vala.license
```
alarcher: You can remove COMPONENT_LICENSE_FILE.

components/runtime/vala/Makefile
```diff
@@ -45,6 +52,6 @@ install:	$(INSTALL_32_and_64)
 
 test:		$(TEST_32_and_64)
 
-BUILD_PKG_DEPENDENCIES =	$(BUILD_TOOLS)
-
-include ../../make-rules/depend.mk
+REQUIRED_PACKAGES += SUNWcs
```
alarcher: Are there build dependencies that are not auto-generated? (e.g. header-only packages)

components/runtime/vala/vala.p5m
```diff
@@ -19,21 +19,22 @@
 # CDDL HEADER END
 #
 # Copyright (c) 2013 David Hoeppner. All rights reserved.
+# Copyright (c) 2017 Andreas Wacknitz
 #
 
 # pull the manpages out of the component dir
 <transform file path=usr/share/man/(man3/.+$) -> set action.hash %<\1> >
-
 <transform file path=usr.*/man/.+ -> default mangler.man.stability uncommitted>
```
alarcher: I think this one is now a default transform.

alarcher (changes requested):
History file required.

components/runtime/vala/Makefile
```diff
@@ -19,21 +19,28 @@
...
 #
-include ../../make-rules/shared-macros.mk
+include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		vala
 COMPONENT_VERSION=	0.20.1
+COMPONENT_REVISION=	1
+COMPONENT_LICENSE=	LGPL2.1
+COMPONENT_LICENSE_FILE=	vala.license
+COMPONENT_FMRI=		runtime/vala
```
pyhalov: Please, don't change FMRI without really good reason

components/runtime/vala/Makefile
AndWac: I had checked this (following your receipt): dependencies detected via grep in configure.ac are glib-2.0, gobject-2.0, gmodule-2.0 and gthread-2.0.
I concluded that I don't have to add other dependencies than the generated.

components/runtime/vala/vala.p5m
AndWac: Ok, seems you are right. I've found
defaults:<transform file path=usr.*/man/.+ -> default mangler.man.stability uncommitted>
in the transformations folder.

components/runtime/vala/Makefile
AndWac: Ok, I can revert this. I thought it was a good idea because I moved vala to the runtime folder. Maybe I should move it to developer?

components/runtime/vala/Makefile
alarcher: Great, thank you.

components/runtime/vala/Makefile
alarcher: Yes just move it to developer, as long as it is consistent it is fine.

alarcher (approved):

PR #3008, opened 2017-04-23
move stardict to desktop folder

AndWac:
I'm not sure this package is worth it (it's not satisfactorily working for me).
The check for additional build dependencies (via grep PKG_CHECK_MODULES configuration.ac) didn't reveal anything.


AndWac:
I could update to 3.0.4 (new archive url at sf; one patch upstreamed) but that would remove stardict-editor. What do you think?

pyhalov:
I honestly don't know, haven't used stardict for ages and have never used its editor.

AndWac:
Ok, how to proceed? If you merge this as is nothing should be lost. Or should I close this PR and we will delete this package?

alarcher:
@AndWac Just for information, you can update this page whenever you add/fix REQUIRED_PACKAGES: https://wiki.openindiana.org/oi/Build+dependencies

PR #3009, opened 2017-04-24
Bump OpenBLAS to 0.2.19

alarcher:


pyhalov:
Why did you add 'pending' label here? Are there any known issues?

alarcher:
Need to fix the test suite and test with numpy 1.12.2.

PR #3016, opened 2017-04-25
Bump bash to 4.4p12

alarcher:


pyhalov:
1) Do we need 32-bit bash?
2) Please, remove resolve.deps

pyhalov:
3. Please, preserve Solaris-specific bash patch descriptions.

pyhalov:
4. solaris-userland has more Solaris-specific patches. I suppose we'd likely want to import them

components/shell/bash/patches/solaris-009.configure.patch
```diff
@@ -10,7 +10,7 @@
  ## M4sh Initialization. ##
  ## -------------------- ##
  
-+export LD_OPTIONS="-lc -lm -lgen -lsocket -lnsl -lresolv -lncurses -z ignore"
++export LD_OPTIONS="-lc -lm -lgen -lsocket -lnsl -lresolv -lncurses "
```
pyhalov: Interesting, if -zignore is default in latest S11?

alarcher:
@pyhalov @xen0l @tsoome If you have time to test. I left 32bit as tsoome and mnowak use 32bit images.

components/shell/bash/Makefile
```diff
@@ -107,20 +114,16 @@ CONFIGURE_OPTIONS  += 		--enable-restricted
...
-CONFIGURE_OPTIONS  += 		--disable-strict-posix-default
+CONFIGURE_OPTIONS  +=			--enable-static-link
 CONFIGURE_OPTIONS  += 		--enable-usg-echo-default=yes
 CONFIGURE_OPTIONS  += 		--enable-xpg-echo-default=yes
-CONFIGURE_OPTIONS  += 		--enable-mem-scramble
-CONFIGURE_OPTIONS  += 		--disable-profiling
-CONFIGURE_OPTIONS  += 		--enable-static-link
-CONFIGURE_OPTIONS  += 		--enable-largefile
-CONFIGURE_OPTIONS  += 		--enable-nls
+CONFIGURE_OPTIONS  +=			--infodir=$(CONFIGURE_INFODIR)
```
pyhalov: something seems to be wrong with formatting

components/shell/bash/patches/solaris-009.configure.patch
```diff
@@ -1,95 +1,32 @@
...
-+#!/bin/bash
- # From configure.ac for Bash 4.3, version 4.063.
- # Guess values for system-dependent variables and create Makefiles.
- # Generated by GNU Autoconf 2.69 for bash 4.3-release.
-@@ -15,6 +15,8 @@
- ## M4sh Initialization. ##
- ## -------------------- ##
- 
-+export LD_OPTIONS="-lc -lm -lgen -lsocket -lnsl -lresolv -lncurses -z ignore"
+--- configure.ac.orig	2017-05-08 03:23:15.044220076 +0200
```
pyhalov, line 1: Need to add autoreconf step to PREP action

components/shell/bash/Makefile
```diff
@@ -107,20 +114,16 @@ CONFIGURE_OPTIONS  += 		--enable-restricted
...
+CONFIGURE_OPTIONS  +=			--localstatedir=/var
 CONFIGURE_OPTIONS  += 		--with-bash-malloc=yes
 CONFIGURE_OPTIONS  += 		--with-curses
 CONFIGURE_OPTIONS  += 		--with-installed-readline=no
-CONFIGURE_OPTIONS  +=		--infodir=$(CONFIGURE_INFODIR)
 
-COMPONENT_TEST_DIR =		$(BUILD_DIR_32)
+COMPONENT_TEST_DIR =		$(BUILD_DIR_64)
```
pyhalov: This breaks 'gmake test' in clean environment. If we want to test only 64-bit version, perhaps, just do so. 
There's no much sense in testing 32-bit version, as isaexec is used and in real world 64-bit version will be used.
There's no much sense to have 32-bit bash, but that's another story :)

alarcher: OK so I am re-reverting to 64-bit only then :)

pyhalov: Any news on this? BTW, I'm using bash 4.4.12 in my build zone, so far without issues.

alarcher: Just that autoreconf fails since all the AC_DEFINE occurences should be fixed and I had no motivation to spend time on this...

components/shell/bash/patches/solaris-009.configure.patch
alarcher: Alright, autoreconf cannot work because autoheader does not work.
I spent some time trying to fix it but apparently it is not meant to be used since most declarations are missing. I guess they just use the config.h as is. Even using aclocal breaks the build. See if the latest revision of the PR works for you.

components/shell/bash/Makefile
```diff
@@ -73,54 +75,59 @@ LIBS =	-lc -lsocket -lgen
...
-CONFIGURE_OPTIONS  += 		--with-installed-readline=no
-CONFIGURE_OPTIONS  +=		--infodir=$(CONFIGURE_INFODIR)
-
-COMPONENT_TEST_DIR =		$(BUILD_DIR_32)
+TEST_PATH = "PATH=$(BUILD_DIR_64):$(GNUBIN):$(USRBINDIR)"
+
+COMPONENT_PREP_ACTION = ( cd $(@D); autoconf -f -i )
+
+CONFIGURE_ENV += LIBS="$(LIBS)"
+CONFIGURE_ENV += LD_OPTIONS="-lc -lm -lgen $(LIBSOCKET) $(LIBNSL) -lresolv $(LD_Z_IGNORE)"
```
pyhalov: does it require -lsocket or not? I just don't see where $(LIBSOCKET) is defined. Ah, it's used from $(LIBS)...

pyhalov (approved):

PR #3017, opened 2017-04-25
xscreensaver: show current xkb layout

pyhalov:


pyhalov:
This is WIP for xscreensaver layout indication. It works as is in my environment, but 
a) need to test it in cases when xkb is not available;
b) need to update it so that it catches layout changes (for now it only checks layout on dialog init).

pyhalov:
One more issue - xscreensaver seems somehow to mangle xkb data. I see two layouts, but both of them are 'English (US)'...

pyhalov:
b) is fixed;
c) is fixed by using libxkbfile;
When xmodmap is used instead xkb, it sees it as us layout (at least with IIIM).
The only remaining concern - is testing it on something which doesn't support xkb... 
Xsun?

alarcher:
I'll test tomorrow.

PR #3025, opened 2017-04-28
8124 sysutils/ipmitool build stops with a compiler options error

jgmills:
Old versions of sys/feature_tests.h do not recognize _XOPEN_SOURCE larger than 600 or POSIX_C_SOURCE larger than 200112L, resulting in the error message:

Compiler or options invalid for pre-UNIX 03 X/Open applications and pre-2001 POSIX applications

This patch fixes that problem and permits the build to succeed.  It also makes the component more flexible.

### DIFF
--- a/components/sysutils/ipmitool/Makefile
+++ b/components/sysutils/ipmitool/Makefile
@@ -10,6 +10,7 @@
 #
 
 #
+# Copyright 2017 Gary Mills
 # Copyright 2016 Alexander Pyhalov
 #
 
@@ -17,6 +18,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= ipmitool
 COMPONENT_VERSION= 1.8.17
+COMPONENT_REVISION= 1
 COMPONENT_SUMMARY= IPMI management tool
 COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.bz2
@@ -30,9 +32,9 @@ COMPONENT_LICENSE = BSD
 COMPONENT_LICENSE_FILE = COPYING
 COMPONENT_FMRI = system/management/ipmitool
 
-include $(WS_TOP)/make-rules/prep.mk
-include $(WS_TOP)/make-rules/configure.mk
-include $(WS_TOP)/make-rules/ips.mk
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
 
 CPPFLAGS += -D__EXTENSIONS__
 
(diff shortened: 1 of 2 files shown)


pyhalov:
Fails to build with 

ipmitool-1.8.17/lib/ipmi_main.c:38:
/usr/include/sys/feature_tests.h:400:2: error: #error "Compiler or options invalid; UNIX 03 and POSIX.1-2001 applications       require the use of c99"
 #error "Compiler or options invalid; UNIX 03 and POSIX.1-2001 applications \

pyhalov:
of course, it should be std=c99...

jgmills:
I just tried a compile with `-std=c99' and got the error:

Compiler or options invalid for pre-UNIX 03 X/Open applications and pre-2001 POSIX applications

I conclude that there is no combination of GCC options that will work for both the old and new versions of the sys/feature_tests.h header.  Therefore, I'm going to withdraw this PR, and add it to my list of components that won't build under oi_151a8 .  It's a very short list.  You can close this PR.

alarcher:
@jgmills Did you try to add _XOPEN_SOURCE=600 as it is expected with c99?

jgmills:
I just tried adding `-D_XOPEN_SOURCE=600' to CFLAGS.  The build still failed, but it pointed out the problem:

...ipmitool-1.8.17/lib/ipmi_main.c:32:0: warning: "_XOPEN_SOURCE" redefined
 #define _XOPEN_SOURCE 700

The file ipmitool-1.8.17/lib/ipmi_main.c contained this peculiar code:

`#define _XOPEN_SOURCE 700
#define _BSD_SOURCE || \
        (_XOPEN_SOURCE >= 500 || \
        _XOPEN_SOURCE && _XOPEN_SOURCE_EXTENDED) && \
        !(_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600)
`
At least I know where to fix the problem now.  Thanks.

alarcher:
@jgmills Great! Any time you need c99, _XOPEN_SOURCE=600 should be defined at minimum, possibly __EXTENSIONS__ if some non-POSIX/SUS (e.g GNU extensions) are involved.
My usual strategy is to define the macro and look at the output to see if the code mixes up standards.

jgmills:
I've updated this PR with a completely new patch.  sysutils/ipmitool now builds correctly on my SPARC machine.  I haven't tested it on x86 hardware, but will do this in a couple of days.

The cause of the initial build failure was that two .c files contained these lines:

#define _XOPEN_SOURCE 700

The patch changes them to `600'.  Obviously, the code does not require `700' because it compiles with `600'.  I didn't need my previous patch at all.

pyhalov:
Please, change commit message to more appropriate

pyhalov:
I've included you into OI developers group in redmine, so you'll be able to fix bug title.

jgmills:
I'm pleased to report that the sysutils/ipmitool component with patches/07-xopen.patch builds successfully on x86 hardware with the current hipster distribution.  This change should be ready to go.

PR #3040, opened 2017-05-03
Some GCC 6 fixes

alarcher:


components/developer/bcc/Makefile
```diff
@@ -58,6 +57,3 @@ install build:		$(BUILD_32)
 
 test:		$(NO_TESTS)
```
pyhalov, line 59: Are there any REQUIRED_PACKAGES?

PR #3041, opened 2017-05-03
Some GCC 6 fixes

alarcher:


xen0l (approved):

components/library/libgnomeui/patches/libgnomeui-04-gcc6.patch
```diff
@@ -0,0 +1,33 @@
+--- libgnomeui-2.24.5/libgnomeui/gnome-dateedit.c.orig	Fri Apr 28 19:30:02 2017
```
xen0l: Patch is added, so we need to bump component revision

components/library/wxwidgets/patches/01-gcc6.patch
```diff
@@ -0,0 +1,107 @@
+
```
xen0l, line 1: Bump component_revision, so patch is applied.

xen0l (changes requested):

components/library/wxwidgets/patches/01-gcc6.patch
alarcher: Done.

components/library/libgnomeui/patches/libgnomeui-04-gcc6.patch
alarcher: Done.

components/library/libgnomeui/Makefile
```diff
@@ -13,12 +13,12 @@
 # Copyright 2014 Alexander Pyhalov.  All rights reserved.
 #
 
-include ../../make-rules/shared-macros.mk
```
pyhalov: perhaps it's better to move gnome2 libraries to desktop/gnome2/ ?

alarcher: I initially put them in desktop/gnome2 but the category layout specifies library/* to go to library. What do you prefer?

pyhalov: do it the same way with mate libs

alarcher:
Updated.

PR #3046, opened 2017-05-03
bass-o-matic: incremental builds

xen0l:


xen0l:
I am testing this right now on jenkins. I also modified Makefile, where I removed --component and --template-zone. As those were never implemented, I dropped them.

Please also not that this turns the incremental magic automatically on if BASS_O_MATIC_MODE is set to incremental. It tries to guess everything from environment variables. If the variable is not set, it will behave like it has until now. Let me know what you think.

pyhalov:
Looks interesting. Have you checked that it works with encumbered components?

jimklimov:
I am slowly moving to make the script to manage golden-template and disposable zones, and run commands in one, so maybe the option to use those for a build would come back in a while ;)

Otherwise, the code looks neat :)

xen0l:
@pyhalov I haven't check if it works with encumbered, but it should as the changes are made to bass-o-matic itself and not Makefile.

xen0l:
@jimklimov I am fine with adding it back once we have the functionality. Until then I would like to keep it minimal. At the moment, the code doesn't do anything.

jimklimov:
Yes, right - good with me.

pyhalov:
Cool. Thanks.

PR #3057, opened 2017-05-07
8172 Fine-grained optimization needed for the gcc compiler

jgmills:
This PR adds four optimization settings for the gcc compiler to the make-rules/shared-macros.mk file.  These are useful when optimization needs to be different for certain types of builds.  GCC optimization settings already in component Makefiles are not affected.


alarcher (approved):

alarcher:
I checked and there is no clash in existing components: there is no gcc_OPT.$(MACH).$(BITS) in the tree.

PR #3062, opened 2017-05-08
Fixes

pyhalov:


alarcher:
If doxygen was not published libwpg and libwps will fail again next round.
HTLM documentation generation is different.

alarcher:
I forgot that the gcc option is gcc-7 specific sorry :S

PR #3064, opened 2017-05-08
Add Openbox window manager

dsendkowski:


components/desktop/openbox/openbox.p5m
```diff
@@ -0,0 +1,178 @@
...
+file path=usr/include/openbox/3.6/obt/xqueue.h
+link path=usr/lib/$(MACH64)/libobrender.so target=libobrender.so.32.0.0
+link path=usr/lib/$(MACH64)/libobrender.so.32 target=libobrender.so.32.0.0
+file path=usr/lib/$(MACH64)/libobrender.so.32.0.0
+link path=usr/lib/$(MACH64)/libobt.so target=libobt.so.2.0.2
+link path=usr/lib/$(MACH64)/libobt.so.2 target=libobt.so.2.0.2
+file path=usr/lib/$(MACH64)/libobt.so.2.0.2
+file path=usr/lib/$(MACH64)/pkgconfig/obrender-3.5.pc
+file path=usr/lib/$(MACH64)/pkgconfig/obt-3.5.pc
+file path=usr/libexec/openbox-autostart
```
alarcher: Executables in /usr/libexec need "mode=0555" added at the end of the line.

alarcher (changes requested):
Tested on my machine.
You need to fix the 0555 mode but otherwise good.
Thank you.

components/desktop/openbox/openbox.p5m
dsendkowski: Like this?:
```
file path=usr/libexec/openbox-autostart mode=0555
file path=usr/libexec/openbox-xdg-autostart mode=0555
```

alarcher: Yes!

alarcher: For reference, default transforms set the modes at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/transforms/defaults#L105

dsendkowski: I added missing `python` and `python-xdg` dependencies, which emerged after specifying python version for `openbox-xdg-autostart`.

xen0l (approved):

alarcher (approved):

alarcher:
@dsendkowski I just noticed that files were delivered in /usr/libexec. We do not use this folder: depending on the type of file either we put the binaries in /usr/bin or we put libraries in /usr/lib/$component/.

pyhalov:
Usually we put them in /usr/lib. /usr/lib/$component was usually used for mate to avoid intersections with gnome.

alarcher:
In this particular case libexecdir may be set to /usr/bin to be consistent with other components;
```
file path=usr/libexec/openbox-autostart
file path=usr/libexec/openbox-xdg-autostart
```

dsendkowski:
OK, but first I'll check whether moving these files requires any additional changes.

alarcher:
Sure. This is not a priority, no rush. It would just be nice to fix it at some point.

PR #3072, opened 2017-05-11
exiv2: add exv_conf.h to package

cgrzemba:
The header is needed to build other OSS components


pyhalov:

PR #3079, opened 2017-05-15
8211 The ncurses library is not found in older OI versions

jgmills:
This PR updates the Makefile of five oi-userland components so that they build on older OI versions where the ncurses library is installed in /usr/gnu/lib .  Two of the components needed to be moved to a suitable category directory.  I also made the usual changes to all of the Makefiles.

Two other files are changed because of unrelated problems.  The sysutils/gptfdisk component used the wrong install command, ignoring the INSTALL macro already defined by one of the included makefiles.  The change to the files/Makefile.solaris file corrects this problem.

The sysutils/watch component did a 64-bit compile during a 32-bit build, causing a wrong ELF type error.  It also ignored the CFLAGS macro set by the component Makefile.  Both of these errors are corrected by the changes to patches/Makefile.patch .


pyhalov:
Why just don't use newer ncurses component?

jgmills:
I forgot to mention that I built those five components on SPARC running oi_151a8, and also on x86 running a recent hipster version.  The newer ncurses component is installed on the x86 machine, but it's not available for the SPARC machine.  The component Makefiles adapt to either installation.

components/text/hunspell/Makefile
```diff
@@ -41,6 +42,13 @@ CPPFLAGS += -I/usr/include/ncurses
 CFLAGS += -D_XOPEN_SOURCE=600
 CFLAGS += -std=c99
 
+CPPFLAGS +=	-I/usr/include/ncurses
+
+# Check location of ncurses libraries
+ifeq ($(shell test -f /usr/gnu/lib/libncurses.so && echo y),y)
```
alarcher, line 48: Actually we have a problem, our current component still refers to /usr/gnu/lib:
helios> pkg-config --libs ncurses
-L/usr/gnu/lib -lncurses

I am not a big fan of this change, better patch the build systems to rely on pkg-config.

pyhalov: Note that it doesn't have -R/usr/gnu/lib

pyhalov:
So far I don't see the necessity of this change. Why just not update ncurses on SPARC system?

components/text/hunspell/Makefile
alarcher: Of course, but if there is something that can be fixed, it is not by hacking the flags like in this component.

jgmills: There's no hacking of flags.  That's part of the design of the oi-userland build system.  Many macros are defined in make-rules/shared-macros.mk so that they can be added to compile flags in the component Makefiles.  About 10% of them take advantage of this feature.  Yes, pkg-config would be useful, but only if it were correct, and only if it handled 32 and 54-bit builds.

jgmills:
This change enables the components to build regardless of the location of the ncurses libraries.  Updating ncurses on my SPARC system doesn't help because the only package available installs the libraries in /usr/gnu/lib .  I could, of course, install a newer ncurses outside of the package system, but I'm trying to minimize how much of that I do.  I'm engaged in building all of the oi-userland components for a new repository of SPARC packages.  This will be an iterative process.  I'm still at the first phase of this process, using oi_151a8 as the build system.

pyhalov:
What do you mean? I suggest you to build updated ncurses from oi-userland, install it on build server and forget about issue? Why it doesn't work?

jgmills:
Of course, I can build it that way.  Of course, it will work.  However, that's part of the second phase.  I'm still on the first phase.  What I'll do instead is to mark all five components as not buildable in this phase.

Feel free to close the bug report and close the PR.

PR #3081, opened 2017-05-15
Add Eigen3

dsendkowski:


components/scientific/eigen3/Makefile
```diff
@@ -0,0 +1,39 @@
...
+#
+
+#
+# Copyright 2017 Dariusz Sendkowski
+#
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=           eigen
+COMPONENT_VERSION=        3.3.3
+COMPONENT_FMRI=           library/eigen
```
alarcher: Should the package name be 'library/eigen3'? It seems that other systems have one package per major version.

components/scientific/eigen3/Makefile
```diff
@@ -0,0 +1,39 @@
...
+include $(WS_MAKE_RULES)/cmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+
+
+build:		$(BUILD_32)
+
+install:	$(INSTALL_32)
+
+test:		$(NO_TEST)
```
pyhalov, line 39: Any specific build  requirements?

dsendkowski: Actually, there is no build. This is a bunch of header files to be installed.

components/scientific/eigen3/Makefile
dsendkowski: Yeah, I've changed it.

PR #3097, opened 2017-05-19
8252 developer/libtool always invokes /usr/bin/automake

jgmills:
This PR fixes 8252 for the developer/libtool component of oi-userland.  It builds cleanly on both SPARC and x86 hardware, correctly searching PATH for the automake command.  Only the Makefile needed to be changed.

### DIFF
--- a/components/developer/libtool/Makefile
+++ b/components/developer/libtool/Makefile
@@ -20,6 +20,7 @@
 #
 
 #
+# Copyright 2017 Gary Mills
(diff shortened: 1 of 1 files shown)


pyhalov:
why just not remove AUTOMAKE=...

jgmills:
Okay, I removed that phrase.  It was the default.  The component still builds on both SPARC and x86 hardware.

PR #3103, opened 2017-05-21
8258 library/protobuf build requires the symbol SOLARIS_64BIT_ENABLED

jgmills:
This PR fixes 8258 for the library/protobuf component of oi-userland.  It does this by defining the SOLARIS_64BIT_ENABLED symbol for 64-bit builds.  Note that the CADD.32 macro is undefined, so that it expands to an empty string.  This is the desired behavior for 32-bit builds.

### DIFF
--- a/components/library/protobuf/Makefile
+++ b/components/library/protobuf/Makefile
@@ -18,6 +18,7 @@
 #
 # CDDL HEADER END
 #
+# Copyright 2017 Gary Mills
 # Copyright (c) 2015, Predrag Zecevic. All rights reserved.
 # Copyright (c) 2015, Aurelien Larcher. All rights reserved.
 #
@@ -25,6 +26,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		protobuf
 COMPONENT_VERSION=	2.6.1
+COMPONENT_REVISION=	1
 COMPONENT_FMRI=		library/c++/protobuf
 COMPONENT_CLASSIFICATION=System/Libraries
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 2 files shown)


pyhalov:
SOLARIS_64BIT_ENABLED should be defined in config.h. Needs investigating why it's not done on SPARC.

jgmills:
It's not done on SPARC because of a bug in this version of protobuf.  The recommended workaround is to define the SOLARIS_64BIT_ENABLED symbol.  I can change the Makefile to define this symbol only for 64-bit SPARC builds, instead of for all 64-bit builds.  Would that be sufficient?

pyhalov:
Does this patch fixes the issue? https://github.com/google/protobuf/commit/97fa4ca1565d216d102af9510b17966c28c7a52a Perhaps, we should just import it?

jgmills:
I can't tell if that patch fixes the problem, but it is for version 3.3.0.  The protobuf in oi-userland is version 2.6.1.  The patch may not fit.  It looks as if even 3.3.0, the latest version, has problems with SPARC hardware.  The workaround for 2.6.1 does work, and seems to be adequate.  We should keep the x86 and SPARC components at the same version level.  Any version upgrade should be done by somebody who can test the software.

pyhalov:
As I see, patch to headers applies cleanly, perhaps, with some minor modifications - http://buildzone.oi-build.r61.net/sparc/97fa4ca1565d216d102af9510b17966c28c7a52a.patch , and  fixes build on 64-bit SPARC.

jgmills:
Okay, I agree.  With your patch, the component builds on both SPARC and x86 hardware, for both 32 and 64 bits.  I'll submit a revised PR soon.

jgmills:
The revised PR is available now.

pyhalov:
Please, rename patch to something like 01-fix-sparc-build.patch and bump COMPONENT_REVISION

jgmills:
Okay, I've made the changes you requested.  The component still builds on both SPARC and x86 hardware.

PR #3111, opened 2017-05-23
8261 library/openssl/openssl-1.0.2 build uses cc options with gcc

jgmills:
This PR fixes 8261 for the library/openssl/openssl-1.0.2 component of oi-userland.  The changes only affect the SPARC build of this component.  The x86 build is completely unaffected.  Patch Configure.gcc.patch is changed to include SPARC configurations.  These are selected by the CONFIGURE_OPTIONS32_sparc and CONFIGURE_OPTIONS64_sparc lines in the Makefile.  In addition, the engines/pkcs11/e_pk11.c file is changed, again only for SPARC, to define two symbols if these happen to be missing from the /usr/include/sys/auxv_SPARC.h header.

As a test, I built this component on x86 hipster.  This build was successful too.

### DIFF
--- a/components/library/openssl/openssl-1.0.2/Makefile
+++ b/components/library/openssl/openssl-1.0.2/Makefile
@@ -18,6 +18,7 @@
 #
 # CDDL HEADER END
 #
+# Copyright 2017 Gary Mills
 # Copyright (c) 2011, 2015, Oracle and/or its affiliates. All rights reserved.
 #
 include ../../../../make-rules/shared-macros.mk
@@ -26,7 +27,7 @@ COMPONENT_NAME =	openssl
 # When new version of OpenSSL comes in, you must update both COMPONENT_VERSION
 # and IPS_COMPONENT_VERSION.
 COMPONENT_VERSION =	1.0.2k
-COMPONENT_REVISION =	1
+COMPONENT_REVISION =	2
 # Version for IPS. It is easier to do it manually than convert the letter to a
 # number while taking into account that there might be no letter at all.
 IPS_COMPONENT_VERSION = 1.0.2.11
@@ -45,7 +46,7 @@ include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 include $(WS_MAKE_RULES)/lint-libraries.mk
 
-PATH=$(GCC_ROOT)/bin:/usr/bin:/usr/gnu/bin:/usr/perl5/bin
+PATH=$(GCC_ROOT)/bin:$(PATH.illumos)
 
 # OpenSSL does not use autoconf but its own configure system.
 CONFIGURE_SCRIPT = $(SOURCE_DIR)/Configure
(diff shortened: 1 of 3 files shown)


components/library/openssl/openssl-1.0.2/Makefile
```diff
@@ -44,8 +45,6 @@ include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 include $(WS_MAKE_RULES)/lint-libraries.mk
 
-PATH=$(GCC_ROOT)/bin:/usr/bin:/usr/gnu/bin:/usr/perl5/bin
```
pyhalov, line 48: Please, preserve $(GCC_ROOT)/bin in front of PATH, as OpenSSL uses compiler as just 'gcc'

components/library/openssl/openssl-1.0.2/engines/pkcs11/e_pk11.c
```diff
@@ -3707,6 +3708,15 @@ hw_aes_instruction_set_present(void)
 #if defined(__amd64) || defined(__i386)
 		present = (ui & AV_386_AES) > 0;
 #elif defined(__sparc)
+/*
+ * These symbols should already be defined from /usr/include/sys/auxv_SPARC.h
+ */
+#ifndef AV_SPARC_AES
+#define AV_SPARC_AES 0
```
pyhalov, line 3715: DilOS used different values for these constants. As I understand, they were coming from Solaris 11. Have you checked what Solaris 11 uses now?

components/library/openssl/openssl-1.0.2/Makefile
jgmills: Yes, I can do that.  It worked without it because /usr/bin/gcc was a symlink to the actual compiler.

components/library/openssl/openssl-1.0.2/engines/pkcs11/e_pk11.c
jgmills: I run Solaris 11.3 on my desktop.  However, we can't use those values from /usr/include/sys/auxv_SPARC.h because that file is copyright by Oracle.  Setting them to zero has the effect of disabling the hardware AES instructions for CPUs that have them.  That's our best option just now.  Our copy of that header file comes from illumos.  Maybe we could ask the illumos developers to update it.

pyhalov: OK, let it be. Please, fix PATH so that this PR was a no-op for x86.

jgmills:
Okay, I've fixed the PATH and added COMPONENT_REVISION to Makefile .  Now this file has a conflict.  What should I do now?

pyhalov:
Update your oi/hipster branch, rebase commit on top of it and force push.

jgmills:
Okay, the conflict seems to be gone now.

PR #3112, opened 2017-05-23
Bump zziplib to 0.13.66

alarcher:
Need to check about ABI compatibility.


pyhalov:
As I see, we don't have consumers in oi-userland. Haven't found anything about zziplib ABI guarantees.

PR #3131, opened 2017-05-26
8120 developer/beanshell requires a specific Java compiler version

jgmills:
This PR fixes 8120 for the developer/beanshell component of oi-userland.  The only change is the addition of a patch called javac.patch .  Of course, COMPONENT_REVISION is also incremented in Makefile .  With this change, the component builds on both SPARC and x86 hardware platforms.

### DIFF
--- a/components/developer/beanshell/Makefile
+++ b/components/developer/beanshell/Makefile
@@ -18,13 +18,14 @@
 #
 # CDDL HEADER END
 #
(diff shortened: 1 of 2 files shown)


xen0l:
@pyhalov what do you think? To me it looks OK, but don't know what are the consequences.

alarcher:
@fronald Any comment on this?

PR #3132, opened 2017-05-26
Bump freetype to 2.8

alarcher:


components/library/freetype/Makefile
```diff
@@ -49,6 +48,10 @@ CONFIGURE_OPTIONS+= --with-png
 CONFIGURE_OPTIONS+= --with-harfbuzz
 CONFIGURE_OPTIONS+= --with-pic
 
+COMPONENT_PREP_ACTION = ( cd $(@D); \
+	$(GSED) -ri 's:.*(AUX_MODULES.*valid):\1:' modules.cfg && \
+	$(GSED) -r 's:.*(\#.*SUBPIXEL_RENDERING) .*:\1:' \
+		-i include/freetype/config/ftoption.h )
```
pyhalov, line 54: Will it affect existing applications or behavior?

alarcher: It enables better hinting of fonts. In any case the update of freetype itself, without this, affects how fonts are rendered.

alarcher: The subsequent fontconfig PR defines a configuration to make the default rendering look nicer.

pyhalov: should applications, using freetype, be recompiled?

alarcher: No it is not required. We may want to recompile cairo to support the additional hinting at some point, but that's all.

PR #3138, opened 2017-05-27
8287 awesome window manager is missing lgi library for Lua 5.3

dsendkowski:
According to the readme for the latest (0.9.1) `lgi`:
_... LGI is tested and compatible with standard Lua 5.1, Lua 5.2 and LuaJIT2.  Compatibility with other Lua implementations is not tested yet. ..._

In order to fix this issue (8287), I suggest forcing `CMake` to build the component for `lua-52`, at lease until `lgi` for `lua-53` is ready.

### DIFF
--- a/components/desktop/awesome/Makefile
+++ b/components/desktop/awesome/Makefile
@@ -14,6 +14,7 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         awesome
+COMPONENT_REVISION=     1
 COMPONENT_VERSION=      4.1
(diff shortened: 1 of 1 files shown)


xen0l (approved):

Mno-hime (commented):
https://github.com/pavouk/lgi/pull/152 suggests that latest version of lgi is compatible with Lua 5.3, it just perhaps was not know at the time of the release and found out subsequently. lgi-53 package seems to me more appropriate then.

PR #3147, opened 2017-05-28
8287 Add Lua 5.3 LGI

Mno-hime:
awesome has problems with redrawing in `xterm` and elsewhere (but not in Firefox), I thought it might be due to not using Lua 5.3 specific LGI library.

[awesome-xsession-errors.txt](https://github.com/OpenIndiana/oi-userland/files/1034174/awesome-xsession-errors.txt)

This patch adds LGI for Lua 5.3 and dependency on ImageMagick as build was failing on missing `convert`.

Currently I am not able to publish awesome due to CMake not being able to find (installed) XCB:

```
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.28") 
-- Found Lua: /usr/lib/64/liblua5.3.so;/usr/lib/64/libm.so (found version "5.3.2") 
-- Not generating manpages. Missing: asciidoc xmlto
-- Using version from /ws/mn/oi-userland/components/desktop/awesome/awesome-4.1/.version_stamp: v4.1
-- Checking for module 'xcb>=1.6'
--   
CMake Error at /usr/share/cmake-3.5/Modules/FindPkgConfig.cmake:363 (message):
```

[awesome-build-failure.txt](https://github.com/OpenIndiana/oi-userland/files/1034190/awesome-build-failure.txt)

### DIFF
--- a/components/desktop/awesome/Makefile
+++ b/components/desktop/awesome/Makefile
@@ -14,6 +14,7 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         awesome
+COMPONENT_REVISION=     2
 COMPONENT_VERSION=      4.1
 COMPONENT_PROJECT_URL=  https://awesomewm.org
 COMPONENT_SUMMARY=      A highly configurable, next generation framework window manager for X.
@@ -44,11 +45,13 @@ install:	$(INSTALL_64)
 test:		$(NO_TEST)
 
 REQUIRED_PACKAGES += library/desktop/cairo
+REQUIRED_PACKAGES += image/imagemagick
 REQUIRED_PACKAGES += library/desktop/gdk-pixbuf
 REQUIRED_PACKAGES += library/desktop/startup-notification
 REQUIRED_PACKAGES += library/glib2
 REQUIRED_PACKAGES += library/xdg/libxdg-basedir
 REQUIRED_PACKAGES += runtime/lua-53
+REQUIRED_PACKAGES += library/lua/lgi-53
 REQUIRED_PACKAGES += SUNWcs
 REQUIRED_PACKAGES += system/library
 REQUIRED_PACKAGES += system/library/libdbus
--- a/components/desktop/awesome/Makefile
+++ b/components/desktop/awesome/Makefile
@@ -14,7 +14,6 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         awesome
-COMPONENT_REVISION=     1
 COMPONENT_VERSION=      4.1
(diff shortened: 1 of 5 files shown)


alarcher:
Why not bump lgi to 0.9.1 which is compatible with Lua 5.3?

dsendkowski:
On my build machine CMake finds `xcb`:

```
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.28") 
-- Found Lua: /usr/lib/64/liblua5.3.so;/usr/lib/64/libm.so (found version "5.3.2") 
-- Not generating manpages. Missing: asciidoc xmlto
-- Using version from /projects/oi-userland/components/desktop/awesome/awesome-4.1/.version_stamp: v4.1
-- Checking for module 'xcb>=1.6'
--   Found xcb, version 1.12
-- Checking for module 'glib-2.0'
--   Found glib-2.0, version 2.46.2
...
```

I also installed `lgi-53` (0.9.1) earlier, but had `relocation errors` in `libcorelgilua53.so` during awesome startup.

Mno-hime:
I was missing `libpthread-stubs` package and because of that `pkg-config --atleast-version=1.6 xcb` was always returning failure. I guess I need to add `libpthread-stubs` as a requirement to `developer/build/pkg-config`.

alarcher:
Are you sure that you mean `developer/build/pkg-config`?

Mno-hime:
@dsendkowski I can see the same relocation error with 0.9.1 as you do... Not sure what to do about it.

@alarcher I believe I do: W/o the stubs package binary `pkg-config` binary did not work. Or?

alarcher:
@Mno-hime I do not think the dependency should be in pkg-config. We should add dependency 'require' to libpthread-stubs in libxcb to force the install of libpthread-stubs since xcb.pc has "Requires.private: pthread-stubs"

alarcher:
@Mno-hime When you add the dependency in libxcb's manifest, add it to REQUIRED_PACKAGES as well and remove the resolve.deps file.

pyhalov:
If awsome build depends on lua-lgi, the PR should be split in two parts.

dsendkowski:
@Mno-hime  `lgi-53` will have `relocation errors` because 
```
lua_replace
lua_insert
lua_remove
```
are not functions anymore but macros in `lua-53`. A path to the `lua-53` header file needs to be specified in the `Makefile` explicitly.

Mno-hime:
@dsendkowski It seem that you know, what should be done (I don't and don't want to block possible awesome fix). Would you mind taking over? Feel free to take code from this PR.

dsendkowski:
@Mno-hime Sure, no problem.

PR #3164, opened 2017-05-29
8141 desktop/xscreensaver build stops with gcc option error

jgmills:
This PR fixes 8141 for the desktop/xscreensaver component of oi-userland.  With this change, the component builds on both SPARC and x86 hardware.  The new patch removes the AC_NO_OBJECTIVE_C statements from the configure.in file.  These were used, perhaps incorrectly, to add an Apple-specific option to gcc.

### DIFF
--- a/components/desktop/xscreensaver/Makefile
+++ b/components/desktop/xscreensaver/Makefile
@@ -10,6 +10,7 @@
 #
 
 #
+# Copyright 2017 Gary Mills
 # Copyright 2017 Alexander Pyhalov
 #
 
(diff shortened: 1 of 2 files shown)


alarcher:
xscreensaver could be bumped to 5.36

jgmills:
Are you saying that version 5.36 fixes the bug that I reported?

alarcher:
No I am just saying that minor bump could be also done. :)

jgmills:
I don't have the facilities to do a proper test of a new version on either hardware platform.  I assume that version 5.35 is well tested by oi-userland users.  I'd like to stay with that version with my change for bug 8141.  I prefer to be cautious.

jgmills:
What's the verdict on this PR?  I can withdraw it and submit another PR.  I have three more left to go.

alarcher:
I do not mind this one if it builds, I just could not build it for other reasons.

PR #3167, opened 2017-05-30
Mongodb

pyhalov:


pyhalov:
This should fix gcc 7 compilation. 
@alarcher : Should we also switch used gcc version from 5 to 7 at this point?

alarcher (approved):
Yes, if it works fine just move it to gcc-7.

PR #3168, opened 2017-05-30
Add GNU Data Language 0.9.7

dsendkowski:


components/scientific/gdl/patches/07-objects.cpp.patch
```diff
@@ -0,0 +1,15 @@
+--- gdl-0.9.7/src/objects.cpp.orig	Mon May 29 11:48:57 2017
++++ gdl-0.9.7/src/objects.cpp	Mon May 29 11:50:09 2017
+@@ -935,6 +935,12 @@
+     {
+       return default_num_threads;
+     }
++#elif defined(__sun__)
```
pyhalov: why avoid using getloadavg(3C) ?

dsendkowski: I just wanted to keep the original _style_ and be least intrusive, if you know what I mean.

pyhalov: Better use function, as /usr/bin/uptime format differs from /usr/gnu/bin/uptime . And anyway you avoid doing several exec syscalls.

dsendkowski: OK.

pyhalov:
```
CMake Error at CMakeLists.txt:851 (message):
  EIGEN3 is required but was not found.
```
while trying to build
Do you miss some REQUIRED_PACKAGES?
Yes, library/eigen3

dsendkowski:
I'm afraid I did.

pyhalov:
Please, sort the REQUIRED_PACKAGES list :)

alarcher (approved):

alarcher:
@dsendkowski Ah I see that your commit message has leftover history, could you amend it so that it only contains "Add GNU Data Language 0.9.7".

PR #3189, opened 2017-06-01
Bump cmake to 3.8.2

alarcher:
All tests passed.
Tested with a few components in oi-userland.


alarcher:
@pyhalov I have 3.8.2 ready actually.

PR #3196, opened 2017-06-02
sudo: update to 1.8.20p2

pyhalov:


pyhalov:
Unfortunately, audit in sudo doesn't work due to https://www.illumos.org/issues/8309
So I enabled it, but disabled warning about au_preselect failure

PR #3208, opened 2017-06-03
nodejs-8: add new package for version 8.0.0

AndWac:
This is the newest nodejs version. It will become the LTS version later this year.
This one uses gcc-7.


components/runtime/node8/Makefile
```diff
@@ -0,0 +1,58 @@
...
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		node
+COMPONENT_FMRI=		runtime/nodejs-8
+COMPONENT_SUMMARY=	Evented I/O for V8 JavaScript.
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+COMPONENT_PROJECT_URL=	http://nodejs.org
+COMPONENT_VERSION=	8.0.0
+COMPONENT_LICENSE=	BSD-like, MIT, Apache
+COMPONENT_LICENSE_FILE=	node.license
```
alarcher: You can remove the license file line.

components/runtime/node8/Makefile
```diff
@@ -0,0 +1,58 @@
...
+
+CONFIGURE_SCRIPT =	$(@D)/configure
+
+CONFIGURE_OPTIONS= --prefix=/usr/node/8
+
+build:		$(BUILD_32)
+
+install:	$(INSTALL_32)
+
+test:       $(NO_TESTS)
```
alarcher, line 50: Are there no runnable tests?

AndWac: There are tests but they haven't been used in the past (node, node6, node7) and I only run them manually yet. There is still some work left to use them.
Maybe in a later PR?

alarcher: OK, fine by me.

components/runtime/node8/Makefile
AndWac: done

xen0l:
@AndWac thanks for bringing in this update! One question though, is GCC 7 really required?

AndWac:
@xen0l: No, gcc-7 is not necessary. As I said on IRC I can remove it. Shall I?

xen0l:
@AndWac yes, please. gcc 7 is not yet used as a official compiler and I wouldn't like to introduce it when we don't need it (yet) or there is a very good rationale for doing so.

AndWac:
done

xen0l (approved):

PR #3231, opened 2017-06-05
Bump libsigc++ to 2.10.0

alarcher:
ABI compatible: https://abi-laboratory.pro/tracker/timeline/libsigc++/


components/library/libsigc++/test/results-32.master
```diff
@@ -0,0 +1,39 @@
+PASS: test_accum_iter
```
pyhalov: Why should we ship 2 files if results match?

alarcher: I can merge them if you want.

pyhalov: it would be better

pyhalov: Otherwise, LGTM

alarcher: Fixed.

PR #3246, opened 2017-06-06
Bump librsvg to 2.39.0

alarcher:
Do not merge for now.
2.40.17 is ABI compatible but dropped gtk2 support.


alarcher:
Been using this for 3 weeks now.

components/library/librsvg/Makefile
```diff
@@ -13,24 +13,30 @@
...
-include ../../make-rules/configure.mk
-include ../../make-rules/ips.mk
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
 
-PATH=/usr/gnu/bin:/usr/bin
+PATH=$(PATH.gnu)
+
+CPPFLAGS+= -P
```
pyhalov: How is it used?

alarcher: If I recall the build fails with gcc 7.1 due to line numbers during file generation.

pyhalov: But where CPPFLAGS are passed to configure?

alarcher: Good question... I need to check again in the gcc 7 zone if there was a problem. Another question is why are they not in general?

pyhalov: As all update of build logic, which could affect many components, adding it to default list should be tested. (How will current components behave when they will get CPPFLAGS="" in configure env / parameters?)

alarcher: I am asking if there a reason why this was not in the original recipe

pyhalov: don't think there was exact reason

alarcher: Also originally this recipe was for 2.40.17, could be that it was needed for this version.
In any case I will remove.

alarcher:
Updated.

PR #3257, opened 2017-06-08
Bump tmux to 2.5

alarcher:


components/sysutils/tmux/Makefile
```diff
@@ -15,30 +15,34 @@
...
+COMPONENT_LICENSE_FILE=	COPYING
+COMPONENT_LICENSE=	BSD
 
 include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 
 CFLAGS +=	-I/usr/include/ncurses
-LDFLAGS +=	-L/usr/gnu/lib -R/usr/gnu/lib
+LDFLAGS+=	-L/usr/gnu/lib -R/usr/gnu/lib
```
pyhalov: wait... We build 64-bit tmux and link it to 32-bit ncurses? Check if it works without these LDFLAGS.

alarcher: Now we have ncurses in /usr/lib/foo so I doubt it cares these flags anymore. I thought I had removed them.

alarcher:
Fixed.

PR #3272, opened 2017-06-10
Bump gd to 2.2.4

alarcher:
ABI compatible: https://abi-laboratory.pro/tracker/timeline/libgd/

With the broken CMake build system this has been rather tedious...


components/library/gd2/patches/000-cmake-soname.patch
```diff
@@ -0,0 +1,170 @@
+--- libgd-gd-2.2.4/CMakeLists.txt.orig	2017-06-10 01:26:44.907495342 +0000
```
pyhalov, line 1: Where is this patch coming from?

alarcher: It comes from their 2.1 branch. For some reason their 2.2 branch is behind when it comes to CMake files. So basically the CMake files were not updated for the 2.2.4 release and the older 2.1 branch has better support for CMake.

alarcher: Basically I imported their patchset for soname from the 2.1 branch, I fixed the generation of pkg-config files with CMake and patched to keep the gdlib-config script (being deprecated) to ensure proper detection by older components.

pyhalov (approved):

PR #3273, opened 2017-06-11
Vagrant: Support host IO cache on SATA, and, optionaly, SSD-aware fea…

Mno-hime:
…tures

Host IO cache: trade IO speed for security on e.g. host power failure,
more here: https://www.virtualbox.org/manual/ch05.html#iocaching.

`--nonrotational` (optional): This switch allows to enable the
non-rotational flag for virtual hard disks. Some guests (i.e.
Windows 7+) treat such disks like SSDs and don't perform disk
fragmentation on such media.

`--discard` (not implemented): This switch enables the auto-discard
feature for the virtual hard disks. This specifies that a VDI image
will be shrunk in response to the trim command from the guest OS.

https://www.virtualbox.org/manual/ch08.html#vboxmanage-storageattach

Tested on openSUSE Leap 42.2 with Vagrant 1.9.5 from upstream RPM.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -60,6 +60,20 @@ Vagrant.configure("2") do |config|
 
     v.customize ["modifyvm", :id, "--memory", mem]
     v.customize ["modifyvm", :id, "--cpus", cpus]
+    v.customize ["storagectl", :id, "--name", "SATA Controller", "--hostiocache", "on"]
+    # Enable following line, if oi-userland directory is on non-rotational
+    # drive (e.g. SSD). (This could be automated, but with all those storage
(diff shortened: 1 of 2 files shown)


xen0l:
@Mno-hime changes look good to me. Do you think it would be possible for you to add support for turning this on automatically on macOS? It should be easy as majority of Macbooks have SSDs inside them (default for few years now).

Mno-hime:
Sure, thanks for suggestion. I'll try to add it in next couple of days.

Mno-hime:
Done. Re-tested on Linux-only.

xen0l:
@Mno-hime I tested this on my laptop and it works as expected. Thanks!

PR #3293, opened 2017-06-17
Add missing dependencies to xorg-server

alarcher:
We need to check which dependencies are required to make xorg function if only xorg is installed.


ILLUXA:
Here is a list of packages that you need to install to start X, tested on a fresh install (text installation):
> xorg
> xserver-common
> xorg-core
> xproto
> data-xkb
> xkbcomp
> font-util
> mesa
> libepoxy
> xinit
> xorg-input-mouse
> xorg-input-keyboard
> system/font/xorg/iso8859-1
> system/font/xorg/xorg-core
> xorg-video-vesa

Also, you need to install "/x11/server/xorg/driver/*", if you don't use nvidia.

But after these manipulations, X-server crashes with an error
> Cannot open "/var/run/xkb" to write keyboard description
> XKB: Failed to compile keymap
> Keyboard initialization failed. This could be a missing or incorrect setup of xkeyboard-config.

It seems that this error appears because of /var/run/xkb" permissions:
`% stat /var/run/xkb`
>   File: '/var/run/xkb'
>   Size: 189             Blocks: 8          IO Block: 4096   directory
> Device: 8580003h/139984899d     Inode: 1254718540  Links: 2
> Access: (0775/drwxrwxr-x)  Uid: (    0/    root)   Gid: (    3/     sys)
> Access: 2017-06-18 06:52:28.674374539 +0300
> Modify: 2017-06-18 06:52:28.674636188 +0300
> Change: 2017-06-18 07:28:57.583611615 +0300

**/var/run/xkb** dir has read-write-execute permissions only for _root_ and _sys_ group,
but by default, user, created with text installation, is a member of **staff** group only,
so to fix X-server errors we need to execute:
`# chown :staff /var/run/xkb`
(`chmod` doesn't work, it restores to "0775" after every `startx`).

Then X starts fine, but keyboard input doesn't work... :_(
If I'll find the way how to fix it, I'll post it here.

**EDIT:**
Cannot determine which package helped (it seems that it was system/keyboard/kbd_mode), because I've installed many packages, which installed many dependencies, and package manager doesn't show what dependencies it's installing... But anyway, after reboot keyboard works fine. Here is my full package list, if it helps, — https://pastebin.com/7N5d73GN

jimklimov:
Note that /var/run should be ephemeral, on tmpfs, so repopulated by services as they start up. Probably here 2 services (or service and some copy of the FS from preinstalled image) have different opinions on the same path?

ILLUXA:
By the way, just noticed, that "/var/run/xkb" dir changed its permissions
`% stat /var/run`
>  File: '/var/run/xkb'                                                                                                                                                                   
>  Size: 117             Blocks: 8          IO Block: 4096   directory
>Device: 8580003h/139984899d     Inode: 1257021748  Links: 2
>Access: (0775/drwxrwxr-x)  Uid: (    0/    root)   Gid: (    0/    root)                                                                                                                 
>Access: 2017-06-18 18:35:26.735122253 +0300                                                                                                                                              
>Modify: 2017-06-18 19:00:06.859389530 +0300
>Change: 2017-06-18 19:00:06.859389530 +0300
> Birth: -

but all is working fine, without any issues. It seems that it is "system/keyboard/kbd_mode" package
helped, so it needs to be added to dependency list too. 
P.S: What I really don't like in Openindiana, is its package manager... It is very spartan.

alarcher:
@ILLUXA The package manager should have a simple interface... it is very powerful but I would say that it exposes too much of the underlying mechanisms to the user... Anyway the original target audience of pkg(5) is system administrators and developers.

ILLUXA:
I am "system administrator" :)
IMO every app should be simple, handy and comfortable to use, even for admins and developers, to maximize results of work.
(But maybe it is hard to work with "pkg(5)" for me, because I'm a novice in "Solaris" world, and I do not understand it correctly?)
It seems it's even missing "autoremove"? So it is impossible to delete orphaned packages in automatic mode, am I right?
If so, than it is even concedes to OpenBSD package managment (that is worst in BSD world IMO), because `# pkg_delete -a` is available on that system.
And is "verbose" mode is available in "pkg(5)", to always print an applications list, that it is installing?

Also I like NetBSD very much, and its pkg manager — pkgin,
IMO it will be great if Openindiana will use it as a main and only one pkg manger,
along with pkgsrc ports tree preinstalled. pkg(5) can be renamed to something like
"oi-update", and can be used to upgrade base system only... it seems I'm dreaming about FBSD now :)

Also some of Solaris "coreutils" in OI — grep, ls, etc are outdated, it should be replaced
with BSD utils, or GNU. But of course it is not critical, because apps from "/usr/gnu/bin" are already
preinstalled, and it is possible to use them by default (you need to to add "/usr/gnu/bin" at the beginning
of "$PATH", or add an alias-es in your shell configuration file:
`alias ls    '/usr/gnu/bin/ls -F --color'`).

alarcher:
@ILLUXA Well pkg(5) is file-based (RCS-like approach), so compared to other "fetch a tarball" package managers it has other qualities. Comparing the overal quality of package managers based on a specific CLI question seems a bit unfair, but this is just my opinion. I do agree that the command line interface could be improved to provide "shortcuts" to simple commands that are used often. If you have a list of such "easy to use" shortcuts, this would help improve the situation. Thanks for you help!

ILLUXA:
@alarcher Not at all! Thank you! Despite everything, Openindiana is very interesting
and useful **UNIX** operating system, with not bad driver support (for example, proprietary nvidia video driver
is available only on FreeBSD in BSD world, if you're using nvidia graphic card and want to run real UNIX OS, then you're able to use shitty "nv" driver only, on OpenBSD, NetBSD or DragonFlyBSD, for example).
I like it much!

alarcher:
@ILLUXA I really appreciate your feedback ;) It is nice to have your perspective so we can improve our tools :)

pyhalov:
@alarcher : is it ready?

alarcher:
@pyhalov Not yet. Need to complete the list.

PR #3294, opened 2017-06-17
Initial commit of recipe for Jenkins CI server

jimklimov:
This includes the core (war file), common (user, group and OS objects) and a simple standalone service packages, and lays groundwork for other webapp-server integrations (those would probably put a copy or link of the WAR file into corresponding auto-deploy location), and for LTS support (not relevant at the moment).

The `jenkins-common` package can also be a dependency for subsequent integrations like `jenkins-slave` and `jenkins-swarm` to make prepared or even self-configuring executors for the master server, using a common user account (registered in the OI Wiki) for interaction and an FS location for scratch data. Note that these might be complicated to get right for full automation, because the server side needs to be configured too, but we can at least help developers in this adventure.

This component directory passes `gmake clean publish sample-manifest` but so far it was not tested much further.

One issue that grieves me is that there is considerable storage overhead (due to oi-userland default concepts) where we just copy around the same web-archive file. Maybe symlinks could be a smart solution instead. UPDATE: Symlinks worked for me in local builds with out-of-workspace directory for archives, so added this bit to the PR.

Another is that since this is about weeklies, the package manifest should be updated regularly. And also revisit this when illumos-gate is updated in regard to ZFS... perhaps, dogfooding by running OI Jenkins with this service in an OI machine would help catch discrepancies quickly ;)


components/developer/jenkins/files/jenkins-standalone.xml
```diff
@@ -0,0 +1,107 @@
...
+        <exec_method type='method'
+                   name='start'
+                   exec='/usr/bin/java %{java_startup_options} -jar %{jenkins_war_file} --httpPort=%{httpPort} %{jenkins_startup_options}'
+                   timeout_seconds='240'>
+          <method_context>
+            <method_environment>
+              <envvar name='JENKINS_HOME' value='/var/lib/jenkins/home'/>
+              <envvar name='LIBZFS4J_ABI' value='openzfs'/>
+            </method_environment>
+            <method_credential user='jenkins' group='jenkins' privileges='basic,net_privaddr'/>
```
alarcher: Thanks for the nice PR :)
svccfg complains that method_credential should be before method_environment

jimklimov:
Thanks for the catch, fixed.

components/developer/jenkins/files/jenkins-standalone.xml
```diff
@@ -0,0 +1,108 @@
...
+                   timeout_seconds='60'>
+          <method_context>
+            <method_credential user='jenkins' group='jenkins' />
+          </method_context>
+        </exec_method>
+
+        <property_group name='httpd' type='application'>
+            <stability value='Evolving' />
+            <propval name='httpPort' type='count' value='8080' />
+            <propval name='java_startup_options' type='astring' value='-DsessionTimeout=1440' />
```
alarcher: How does it work if you want to pass multiple options separated by a whitespace? It seems that spaces are escaped at token expansion.

jimklimov:
Maybe... perhaps then we should drop this part in favor of fixed args,
until better research and a next PR?

2017-06-18 22:18 GMT+02:00 Aurélien Larcher <notifications@github.com>:

components/developer/jenkins/LICENSE
```diff
@@ -0,0 +1,23 @@
+MIT License
```
xen0l: Not a hard requirement, but why LICENSE and not jenkins.license?

components/developer/jenkins/Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+# This, and the symlinks in the custom Makefile, save quite a bit
+# of storage overhead (do not need to copy the WAR 3 extra times)
+COMPONENT_COPY_ACTION = ( $(CLONEY) $(SOURCE_DIR)/ $(@D)/ )
+
+COMPONENT_INSTALL_ENV += JENKINS_VERSION=$(COMPONENT_VERSION)
+COMPONENT_INSTALL_ENV += JENKINS_RELEASE=weekly
+
+
+#COMPONENT_BUILD_ACTION = true
```
xen0l: Drop this if not needed

components/developer/jenkins/Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+#COMPONENT_INSTALL_ACTION = true
```
xen0l: Drop this if not needed

components/developer/jenkins/Makefile
```diff
@@ -0,0 +1,88 @@
...
+build: $(BUILD_32)
+
+install: $(INSTALL_32)
+
+test: $(NO_TESTS)
+
+# Note: java8+ is required since 2017
+# This may be JRE8 for the server, though JDK8 also suffices :)
+REQUIRED_PACKAGES += SUNWcs
+#REQUIRED_PACKAGES += developer/java/openjdk8
```
xen0l: Why is this commented out?

components/developer/jenkins/files/Makefile
```diff
@@ -0,0 +1,11 @@
+JENKINS_VERSION ?= 2.65
+JENKINS_RELEASE ?= weekly
+all: build
+
+build: 
+
+install: 
+	mkdir -p $(DESTDIR)/usr/share/jenkins/war
+	mkdir -p $(DESTDIR)/usr/share/jenkins/war-$(JENKINS_RELEASE)
```
xen0l: We should ship this under /usr/share/jenkins/war/weekly or /usr/share/jenkins/weekly. Does this setup mimics some practices in other distro?

components/developer/jenkins/files/jenkins-standalone.xml
```diff
@@ -0,0 +1,108 @@
+<?xml version='1.0'?>
+<!DOCTYPE service_bundle SYSTEM '/usr/share/lib/xml/dtd/service_bundle.dtd.1'>
+<service_bundle type='manifest' name='jenkins-standalone'>
+  <service
+    name='network/http'
```
xen0l: I would call it just application/jenkins.

components/developer/jenkins/jenkins-common.p5m
```diff
@@ -0,0 +1,52 @@
...
+dir group=jenkins owner=jenkins mode=0755 path=var/lib/jenkins/backup
+dir group=jenkins owner=jenkins mode=0755 path=var/lib/jenkins/build_record
+dir group=jenkins owner=jenkins mode=0755 path=var/lib/jenkins/build_workspace
+dir group=jenkins owner=jenkins mode=0755 path=var/lib/jenkins/syshome
+dir group=jenkins owner=jenkins mode=0700 path=var/lib/jenkins/syshome/.ssh
+
+# Note: java8+ is required since 2017
+# This may be JRE8 for the server, though JDK8 also suffices :)
+depend fmri=__TBD pkg.debug.depend.file=usr/bin/java type=require
+#depend fmri=__TBD pkg.debug.depend.file=usr/bin/javac type=require
```
xen0l: Drop this if not needed

components/developer/jenkins/jenkins-core-weekly.p5m
```diff
@@ -0,0 +1,34 @@
...
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# Note: java8+ is required since 2017
+# This may be JRE8 for the server, though JDK8 also suffices :)
+depend fmri=__TBD pkg.debug.depend.file=usr/bin/java type=require
+#depend fmri=__TBD pkg.debug.depend.file=usr/bin/javac type=require
```
xen0l: Drop this if not needed

components/developer/jenkins/jenkins-core-weekly.p5m
```diff
@@ -0,0 +1,34 @@
...
+
+file path=usr/share/jenkins/war-weekly/$(COMPONENT_NAME)-$(COMPONENT_VERSION).war
+link path=usr/share/jenkins/war/jenkins.war \
+     target=../war-weekly/$(COMPONENT_NAME)-$(COMPONENT_VERSION).war \
+     mediator=jenkins-core-weekly \
```
xen0l: Does it make sense to mediate this? Normally, people will run either LTS or the latest version. Also what happens when you use plugins, but they are not compatible with the version configured in the mediator? Do we care about those problems? To me, it would make sense just to have updates shipped by "pkg update"

components/developer/jenkins/Makefile
```diff
@@ -0,0 +1,88 @@
...
+# for "weather report" (user votes on good/bad state of the weekly)
+# At some point it may make sense to deliver packaged LTS releases,
+# and the initial recipe supports that notion already just in case,
+# but no LTS issued so far includes updated libzfs.jar (v0.8+ needed).
+COMPONENT_NAME=		jenkins
+COMPONENT_MAJOR_VERSION=	2
+COMPONENT_MINOR_VERSION=	65
+COMPONENT_VERSION=	$(COMPONENT_MAJOR_VERSION).$(COMPONENT_MINOR_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).war
+COMPONENT_FMRI=		developer/jenkins-core-weekly
```
xen0l, line 32: I see the name and it is clearly thought out and well reflects the jenkins release model. However, could we create a package called 'developer/jenkins', so people trying to do 'pkg install jenkins' won't be annoyed? That package should also ship mediator, so people are able to choose mediator implementation: 'weekly' or 'core'. What do you think about this?

xen0l (changes requested):
Thanks for working on this, it is very much appreciated!

Few nits:
* Do you also plan to provide LTS package?
* developer/jenkins metapackage would be userful
* Rename the SMF service to application/jenkins
* SMF method script would be nicer in the future, but let's go with MVP right now.

I will provide example nginx configuration file with best practice configuration, so people can just copy it and run it via nginx. 

Otherwise, very good job man!

components/developer/jenkins/Makefile
pyhalov: Isn't it an over-engineering? Just deliver one version and call it developer/jenkins?

components/developer/jenkins/LICENSE
jimklimov: Reasonable point. Copied from some other recipe, one of those this was stitched from :) Probably that one pulled a file from GNU-style sources.

components/developer/jenkins/Makefile
jimklimov: Unfortunately no - some people apparently use the stale but predictable LTS releases :)

Regarding mediation - feel free to add the "developer/Jenkins", it would make sense perhaps. I am not that good in the theory and practice of mediators, so would prefer to leave it for subsequent PRs by others :)

But note that one bulky matter is delivery of a WAR file, and another is of deploying and running it in some container - or standalone as done in this initial package. Other packages may follow if there's interest and a reliable idea how to implement it. Maybe just deliver and kick a service to re-deploy into a corresponding appserver like tomcat, tomcat8, glasshish/payara... this is not in scope for this PR anyhow.

So installing just "Jenkins" may be not what they intended, for certain users. On another hand, even if they get the standalone service ability, it is not enabled by default - so not much troble here either...

components/developer/jenkins/Makefile
jimklimov: Per comment above.

I don't think the `REQUIRED_PACKAGES` allow to specify a "require-any" dependency on e.g. "JDK or JRE". Thinking of it, for the "build" we don't need any Java, just a curl/wget :)

components/developer/jenkins/jenkins-core-weekly.p5m
jimklimov: I think it may make sense to mediate. Most plugins are compatible with way older "core" versions than those of a latest weekly or LTS. The web-gui will usually complain if some config syntax is out of date, etc. - but it is up to the admins (our end-users) to figure out if they want to switch releases.

Also regarding your question in another post - as written in the `Makefile` comments, currently we can not reasonably deliver an LTS because the updated `libzfs.jar` was not included, or so it seems (not in changelog at least).

components/developer/jenkins/Makefile
jimklimov: I added a small package that just requires (or requires-any) the other packages. It may be that this will just work, updating whatever the user has already installed on the system (initially through this package, or manually at some point), and would not add other stuff. We'll see :)

components/developer/jenkins/LICENSE
jimklimov: fixed

components/developer/jenkins/files/jenkins-standalone.xml
jimklimov: fixed

components/developer/jenkins/Makefile
jimklimov: fixed

components/developer/jenkins/Makefile
jimklimov: fixed

components/developer/jenkins/Makefile
jimklimov: fixed

components/developer/jenkins/jenkins-core-weekly.p5m
jimklimov: fixed

components/developer/jenkins/jenkins-common.p5m
jimklimov: fixed

components/developer/jenkins/files/jenkins-standalone.xml
jimklimov: Well, it does present an HTTP server when run standalone, and end-users might want to be aware of that. In particular, some orgs did have to close down public access to Jenkins farms back in the day, and just republish build results and logs as static data etc. when they suspected or found vulnerabilities in Jenkins app itself. Also, using a relatively common port of 8080, this one can conflict on systems that do not run services in separate zones...

components/developer/jenkins/files/Makefile
jimklimov: No, not mimicking, will change dash to slash.

jimklimov:
@alarcher @xen0l @pyhalov : I think I fixed or commented about all the currently existing nits, please review whether the results are suitable :)

jimklimov:
Also, just so this bit from IRC discussions does not slip away from memory: the fixes done so far for libzfs.jar included in recent weekly Jenkins just ensure that it does^h^h^h^h has a way to not crash when using native `libzfs.so` ABI on various OpenZFS systems. This does not necessarily mean that the ZFS-related features function well :) That needs to be looked into separately. But at least we have a way to deploy and update the core WAR more reliably, portably (across illumos-gate updates) and simply than before.

Oh, and now keep in mind that openzfs improvements that fall on us in the future may require that the `LIBZFS4J_ABI_*` fine-tuning settings in app-servers or standalone SMF service get updated, for the WAR to still not core-dump its JVM upon startup. See https://github.com/kohsuke/libzfs4j (README and referenced sources) for details, and possible PRs to add support for new ABIs as they come.

components/developer/jenkins/jenkins.p5m
```diff
@@ -0,0 +1,50 @@
...
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
```
pyhalov: please, use illumos CDDL header

components/developer/jenkins/jenkins.p5m
```diff
@@ -0,0 +1,50 @@
...
+
+# TODO: At some point this package should help mediate between weekly
+# or LTS releases. Maybe it will do so automagically just by virtue of
+# pulling newer updates of already-installed "require-any" dependencies?
+depend fmri=pkg:/developer/jenkins-common@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend type=require-any \
+    fmri=pkg:/developer/jenkins-core-weekly \
+    fmri=pkg:/developer/jenkins-core-lts
+depend type=require-any \
+    fmri=pkg:/runtime/java/openjdk8 \
```
pyhalov: if you require-any, just depend on runtime, as jdk requires jre

jimklimov:
Doesn't JDK include JRE? At least in Sun/Oracle tarballs, this is so: you unpack JDK and have JRE + DK tools, and directory layouts of the two tarballs differ a bit. Is it different with IPS for OpenJDK?

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jimklimov:
Sorry for heading - must be an artifact of copypasting from another component.

If only you'd accept PRs that fix up such mess everywhere once and for all! ;)

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

pyhalov:
jdk package depends on jre package, so no need to depend on one of them.

jimklimov:
Bump? ;)

Back from vacation, addressed the nits...

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

components/developer/jenkins/jenkins-standalone.p5m
```diff
@@ -0,0 +1,41 @@
...
+depend fmri=pkg:/developer/jenkins-common@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend type=require-any \
+    fmri=pkg:/developer/jenkins-core-weekly \
+    fmri=pkg:/developer/jenkins-core-lts
+depend type=require \
+    fmri=pkg:/runtime/java/openjdk8
+
+file \
+    files/jenkins-standalone.xml \
+    path=lib/svc/manifest/application/jenkins-standalone.xml \
```
xen0l: The proposal was to keep nework/http as a fmri name, but we are shipping manifest in lib/svc/manifest/application. Shouldn't we be consistent here? This is not a blocker, just asking.

xen0l:
Am I doing something wrong?
<pre>
xenol@external jenkins (jenkins) % pfexec pkg install jenkins                       
           Packages to install:  4
           Mediators to change:  1
            Services to change:  1
       Create boot environment: No

Completed                                4/4           6/6    68.1/68.1    0B/s

Installing new actions                         75/75
Updating package cache                           3/3 
xenol@external jenkins (jenkins) % svcs -a | grep jenkins
disabled        2:03:11 svc:/network/http:jenkins
xenol@external jenkins (jenkins) % svcadm enable jenkins                             
xenol@external jenkins (jenkins) % svcs -p svc:/network/http:jenkins
STATE          STIME    FMRI
maintenance     2:03:47 svc:/network/http:jenkins
xenol@external jenkins (jenkins) % tail $(svcs -L svc:/network/http:jenkins)
[ Jul 20 02:03:11 Disabled. ]
[ Jul 20 02:03:11 Rereading configuration. ]
[ Jul 20 02:03:47 Enabled. ]
[ Jul 20 02:03:47 Could not expand method tokens in "/usr/bin/java -DsessionTimeout=1440 -jar %{jenkins_war_file} --httpPort=%{httpPort}": property "jenkins_war_file" not found. ]
xenol@external jenkins (jenkins) % ls -l /usr/share/jenkins/war/jenkins.war                                                                                          
ls: cannot access '/usr/share/jenkins/war/jenkins.war': No such file or directory
xenol@external jenkins (jenkins) % pkg list -a | grep jenkins                 
developer/jenkins                                 2.65-2017.0.0.0            i--
developer/jenkins-common                          2.65-2017.0.0.0            i--
developer/jenkins-core-weekly                     2.65-2017.0.0.0            i--
developer/jenkins-standalone                      2.65-2017.0.0.0            i--
xenol@external jenkins (jenkins) % pkg mediator              
MEDIATOR     VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
apache       vendor    2.2     vendor     
automake     system    1.15    system     
clang        system    3.8     system     
gcc          vendor    4.9     vendor     
java         vendor    8       vendor     
jenkins-core system    2.65    system     weekly
mta          system            system     mailwrapper
mysql        system    5.5     system     mariadb
openssl      vendor            vendor     default
perl         local     5.24    system     
php          system    7.0     system     
python       local     2.7     vendor     
python3      system    3.4     system     
ruby         vendor    2.3     vendor     
ssh-askpass  system            system     ssh-askpass-zenity
system-perl  vendor    5.22    vendor 
</pre>

EDIT:
I found an issue in SMF manifest, it's supposed to be %{httpd/jenkins_war_file} and %{httpd/httpPort}. I fixed it and still go similar error:
<pre>
[ Jul 20 02:13:34 Executing start method ("/usr/bin/java -DsessionTimeout=1440 -jar /usr/share/jenkins/war/jenkins.war --httpPort=8080"). ]
Error: Unable to access jarfile /usr/share/jenkins/war/jenkins.war
</pre>

components/developer/jenkins/files/jenkins-standalone.xml
```diff
@@ -0,0 +1,111 @@
...
+            type='service'>
+            <service_fmri
+            value='svc:/system/filesystem/autofs:default'/>
+        </dependency>
+
+<!-- TODO: Wrap this in a script? There are many possible CLI options to consider,
+     both for JAVA and the JAR itself... Or keep it simple? -->
+        <exec_method type='method'
+                   name='start'
+                   exec='/usr/bin/java -DsessionTimeout=1440 -jar %{jenkins_war_file} --httpPort=%{httpPort}'
```
xen0l: %{jenkins_war_file} -> %{httpd/jenkins_war_file}
%{httpPort} -> %{httpd/httpPort}

xen0l (changes requested):

components/developer/jenkins/jenkins-core-weekly.p5m
```diff
@@ -0,0 +1,38 @@
...
+# This may be JRE8 for the server, though JDK8 also suffices :)
+depend fmri=__TBD pkg.debug.depend.file=usr/bin/java type=require
+
+file path=usr/share/jenkins/war/weekly/$(COMPONENT_NAME)-$(COMPONENT_VERSION).war
+link path=usr/share/jenkins/war/weekly/jenkins.war \
+     target=$(COMPONENT_NAME)-$(COMPONENT_VERSION).war \
+     mediator=jenkins-core \
+     mediator-implementation=weekly \
+     mediator-version=$(COMPONENT_VERSION)
+link path=usr/share/jenkins/war/jenkins.war \
```
xen0l: This doesn't seem to work correctly.

<pre>
xenol@external weekly % ls -l /usr/share/jenkins/war/jenkins.war
ls: cannot access '/usr/share/jenkins/war/jenkins.war': No such file or directory
xenol@external weekly % ls -l /usr/share/jenkins/war/weekly/jenkins.war
lrwxrwxrwx 1 root root 16 Jul 20 02:03 /usr/share/jenkins/war/weekly/jenkins.war -> jenkins-2.65.war
</pre>

xen0l:
I tried to run Jenkins itself, but come into few issues:
1) SMF service never gets out of "offline*" state
2) There is a dependency on system/font/truetype/dejavu as it will result in https://wiki.jenkins.io/display/JENKINS/Jenkins+got+java.awt.headless+problem (I don't know if it can be different font, I just followed their wiki)

I was able to get service working with these steps:
1) pkg install system/font/truetype/dejavu
2) I had to modify XML manifest: https://gist.github.com/xen0l/5fe613e1ab47f8bd8ed6ff389385a385

With these changes I was able to get web interface up and running and configure it.

jimklimov:
Cool, thanks for testing and proposed fixes. I'll try to tend to it soon.

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jimklimov:
Posting these (and a few other) fixes, and hope to get the symlink mediation addressed. Seems to be written same way as we have for e.g. MySQL variants... :\

components/developer/jenkins/files/jenkins-standalone.xml
```diff
@@ -0,0 +1,128 @@
...
+// Common Development and Distribution License ("CDDL"), version 1.0.
+// You may only use this file in accordance with the terms of version
+// 1.0 of the CDDL.
+//
+// A full copy of the text of the CDDL should have accompanied this
+// source.  A copy of the CDDL is also available via the Internet at
+// http://www.illumos.org/license/CDDL.
+//
+//
+// Copyright 2016-2017 by Jim Klimov, fixes 2017 added by Adam Stevko
```
pyhalov, line 15: Please, make it sane. Either remove @xen0l if he is not against it or put him on separate line.

xen0l: I don't mind, just let's fix this ASAP.

xen0l (approved):

PR #3302, opened 2017-06-20
Bump gc to 7.6.0

alarcher:
Tested with inkscape.
libatomic_ops is now in a separate tarball.


pyhalov:
perhaps, patch out /usr/lib/lwp RPATH?

alarcher:
Updated.

PR #3311, opened 2017-06-25
mrtg: Move to sysutils and add REQUIRED_PACKAGES

AndWac:
The .p5m file contains a dependency to rrdtool@1.4.7.
Is this necessary? The newest rrdtool version 1.4 is 1.4.9.
The newest rrdtool version is 1.7.0, though.


pyhalov:
Don't remember, if rrdtool is really necessary, but of course, it's useful with mrtg

AndWac:
It's an optional dependency as far as I understand. My question was more: is the fixed version (@1.4.7) in the .p5m file really necessary? If we change the version of rrdtool we need to (remember and) change the p5m file of mrtg.

pyhalov:
no, we don't `depend type=require fmri=A@V` means "we need package A at version V1>=V"

PR #3352, opened 2017-07-21
8471 enlightenment: missing execute permissions on some files

alarcher:
https://www.illumos.org/issues/8471


pyhalov:
transform on line 26 was intended to do it, but tranform is incorrect (as pattern is regex, not shell pattern), better fix transform and check that it works

alarcher:
Updated.

PR #3356, opened 2017-07-21
Bump to Jenkins weekly 2.77 and restructure the recipe a bit

jimklimov:
This PR separates the tracking of `Jenkins-core-weekly` into a separate component directory with its separate `Makefile` and versioning, because the OS scaffolding lives a life independent from WAR file bumps, most of the time. Also bump the Jenkins WAR version to this week's one.


jimklimov:
I tried a series of pkg update's (from my local repo) and it seems to pass nicely, including restart of the Jenkins standalone server.

Note that compared to 2.65 packaging, now the standalone WAR is unpacked into a dedicated directory `/var/lib/jenkins/standalone-war`, so a `/var/lib/jenkins/home/war` from 2.65 remains to be wiped out (if present at early adopters) to reduce disk footprint.

components/developer/jenkins-core-weekly/files/Makefile
```diff
@@ -1,4 +1,19 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016-2017 Jim Klimov
+#
+
+JENKINS_VERSION ?= 2.70
```
xen0l: Could we make this value go away and read from somewhere? It's not nice if you have to bump in two different places.

xen0l (commented):
More feedback after I build and test these changes

jimklimov:
The ?= assignment sets a default, if not set yet. It should suffice if we only set it in main recipe makefile (this assumption should be easy to test ;))

The idea was that this same helper makefile text (or a symlink to this one) can be used for also LTS eventually. 

_Sent from my Xiaomi Redmi Note 4 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jimklimov:
The 2.71 came out in the meanwhile, so I updated only the main Makefile and the package with new archive got built and delivered correctly. So the `?=2.70` works as expected (presence of some fallback value) and does not have to be bumped. I think it can stay in the code to serve as documentation of expected value format or to aid other experimentation, but would not have effect in standard builds.

components/developer/jenkins-core-weekly/files/Makefile
jimklimov: To clarify on the replies posted earlier - this is not read, but passed by caller via `COMPONENT_INSTALL_ENV` in main recipe `Makefile`.

jimklimov:
Also, verified that the SMF properties can be changed to use the privileged port numbers like 80 (as expected via the service privileges).

Also contributed a download link and initial-setup chapter to https://github.com/jenkins-infra/jenkins.io/blob/master/content/doc/book/getting-started/installing.adoc ;)

components/developer/jenkins-core-weekly/Makefile
```diff
@@ -0,0 +1,90 @@
...
+COMPONENT_VERSION=	$(COMPONENT_MAJOR_VERSION).$(COMPONENT_MINOR_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).war
+COMPONENT_FMRI=		developer/jenkins-core-weekly
+COMPONENT_CLASSIFICATION=	Development/Java
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	https://jenkins.io/changelog/
+COMPONENT_ARCHIVE_URL=\
+	http://mirrors.jenkins-ci.org/war/$(COMPONENT_VERSION)/jenkins.war
+# See $(COMPONENT_ARCHIVE_URL).sha256 e.g. for current weekly release, run
+#   wget -O - http://mirrors.jenkins-ci.org/war/latest/jenkins.war.sha256
```
xen0l: Is there a way to validate this, e.g. gpg?

components/developer/jenkins/files/jenkins-standalone.xml
```diff
@@ -95,15 +101,36 @@
           </method_context>
         </exec_method>
 
+<!-- At the bare minimum, we want to restart and redeploy the WAR file and
+     its Jetty container when the service is refreshed (e.g. package update).
+     So we just kill it (if running) and hope SMF restart would do the rest.
```
pyhalov, line 106: I dislike this. It seems we have no means to do this intelligently without restart script.

components/developer/jenkins-core-weekly/files/Makefile
```diff
@@ -1,10 +1,26 @@
...
 JENKINS_RELEASE ?= weekly
 all: build
 
 build: 
 
 install: 
+	$(RM) -r $(DESTDIR)/usr
```
pyhalov: Why do we need this? I understand this Makefile will be used only by control of oi-userland, but still this looks scary.

pyhalov:
One thing why I'm reluctant to merge this immediately is introducing new component just to ship supplementary files....

jimklimov:
Do you mean the weekly? On one hand, sooner or later I hope there will be LTS with (updated) OpenZFS support too - which of course will be two different packages with different versioning and update schedules. Also, the WAR file of either type can be used in a number of use-cases (standalone or not), and it makes little sense to re-deliver it separately bundled with different appservers, right?

components/developer/jenkins-core-weekly/files/Makefile
jimklimov: To make sure rebuilds are clean - that we remove artifacts from previous build in the same workspace. I had them layering over, so sample-manifest counted all sub-versions I've built in this repo so far ;)

components/developer/jenkins/files/jenkins-standalone.xml
jimklimov: Feel free to improve iteratively. This is a bare minimum which does work, hands-off. Shutdown is graceful (JVM has a SIGTERM and processes it to stop jobs, etc.) though fast (probably does not wait for jobs to finish running, if anything is building).

Jenkins2 has a mode with workers/nodes/agents/slaves/whatever the PC name is nowadays, where the master can be restarted while remote jobs are running - and it reattaches to manage them. So this is not very disruptive.

Also there is some balance to be had about updating packages and not blocking `pkg` for hours wating for the service to restart leisurely (can untie this somehow else though, if needed in practice). And system shutdowns of different urgency have their say too. Going down for maintenance window, well prepared, is one thing. Powering off safely because UPS is ONBATT is another.

All this can be handled by a method script, but really is out of scope for weekly bumps of the core and can be handled by a separate PR to improve this proposed method, hanging in discussion for a few more weeks :)

pyhalov: Of course... Sorry, but if it was only version bump, I'd merged it much faster :)

components/developer/jenkins-core-weekly/files/Makefile
pyhalov: do gmake clean before gmake sample-manifest :) I'm a bit worried about fact that if by some chance DESTDIR is not defined, it'll try to remove /usr.

pyhalov: do gmake clean before gmake sample-manifest :) I'm a bit worried about fact that if by some chance DESTDIR is not defined, it'll try to remove /usr.

components/developer/jenkins-core-weekly/jenkins-core-weekly.p5m
```diff
@@ -26,7 +26,13 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 # This may be JRE8 for the server, though JDK8 also suffices :)
 depend fmri=__TBD pkg.debug.depend.file=usr/bin/java type=require
 
-file path=usr/share/jenkins/war/weekly/$(COMPONENT_NAME)-$(COMPONENT_VERSION).war
+# Note: as we deliver the new WAR file, we tell the appservers to refresh
+# using the known (or expected soon) SMF FMRIs. Depending on integration,
+# this would restart a standalone server or bump and redeploy in container.
+file path=usr/share/jenkins/war/weekly/$(COMPONENT_NAME)-$(COMPONENT_VERSION).war \
+     refresh_fmri=svc:/network/http:jenkins \
+     refresh_fmri=svc:/network/http:jenkins-tomcat \
```
pyhalov: Please, leave only existing services, other can be added when they appear

components/developer/jenkins-core-weekly/Makefile
```diff
@@ -0,0 +1,90 @@
...
+COMPONENT_INSTALL_ENV += JENKINS_VERSION=$(COMPONENT_VERSION)
+COMPONENT_INSTALL_ENV += JENKINS_RELEASE=weekly
+
+build: $(BUILD_32)
+
+install: $(INSTALL_32)
+
+test: $(NO_TESTS)
+
+# Note: this recipe just downloads and packages stuff.
```
xen0l, line 88: If it doesn't do anything, I would move this and drop REQUIRED_PACKAGES completely.

components/developer/jenkins-core-weekly/files/Makefile
xen0l: I am worried about this as well. Rather add dependency on clean target

components/developer/jenkins/Makefile
```diff
@@ -14,76 +14,36 @@
...
 
 COMPONENT_NAME=		jenkins
-COMPONENT_MAJOR_VERSION=	2
-COMPONENT_MINOR_VERSION=	65
-COMPONENT_VERSION=	$(COMPONENT_MAJOR_VERSION).$(COMPONENT_MINOR_VERSION)
-COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).war
-COMPONENT_FMRI=		developer/jenkins-core-weekly
-COMPONENT_CLASSIFICATION=	Development/Java
-COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_VERSION=	2.65
```
xen0l, line 22: Shouldn't this be bumped as well?

components/developer/jenkins/Makefile
```diff
@@ -14,76 +14,36 @@
...
 
-install: $(INSTALL_32)
+install:
+	$(MKDIR) $(PROTO_DIR)
 
 test: $(NO_TESTS)
 
-# Note: this recipe just downloads and packages stuff.
-# No real build, no magic tools, no java even, should be needed!
+# Note: this recipe just defines packages.
```
xen0l: If no magic is done here, please drop REQUIRED_PACKAGES.

xen0l (changes requested):

components/developer/jenkins-core-weekly/files/Makefile
jimklimov: Added a sanity check for the value. Not sure if dependency on `clean` is good - can loop (because it cleans the builddir too, with the touch-file that we're ready to install...)

components/developer/jenkins/Makefile
jimklimov: No, preferably it should restart from 1.0 but I guess it is too late :)
This is in fact version of the OS-integration bits like jenkins-common and jenkins-standalone, which change on a much different schedule than the weekly.

components/developer/jenkins-core-weekly/Makefile
jimklimov: No idea, I don't use that - have to copy-paste instead :(

Also, wasn't the idea behind caching the SHAs in our Makefiles to protect against upstream re-publish of source+checksum while we are not aware of the change?

components/developer/jenkins-core-weekly/jenkins-core-weekly.p5m
jimklimov: done

components/developer/jenkins-core-weekly/Makefile
jimklimov: done

components/developer/jenkins/Makefile
jimklimov: done

xen0l (approved):

PR #3359, opened 2017-07-23
enlightenment: update to 0.21.8, fix reboot/poweroff

optarg:
Enlightenment expects the SUID bit to be set on a few files, so I took the liberty of restoring it. This also enables the 'Reboot' and 'Power Off' options which were previously greyed out.

### DIFF
--- a/components/desktop/e/enlightenment/enlightenment.p5m
+++ b/components/desktop/e/enlightenment/enlightenment.p5m
@@ -22,9 +22,10 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-<transform file path=usr/lib/(.*)/module.so$ -> default mode 0555>
 <transform file path=usr/lib/$(MACH64)/enlightenment/utils/(.*)$ -> default mode 0555>
 
+file path=etc/enlightenment/sysactions.conf mode=0644 preserve=true
+file path=etc/xdg/menus/e-applications.menu
 file path=usr/bin/emixer
 file path=usr/bin/enlightenment
 file path=usr/bin/enlightenment_filemanager
@@ -32,8 +33,6 @@ file path=usr/bin/enlightenment_imc
 file path=usr/bin/enlightenment_open
 file path=usr/bin/enlightenment_remote
 file path=usr/bin/enlightenment_start
-file path=usr/etc/enlightenment/sysactions.conf
-file path=usr/etc/xdg/menus/e-applications.menu
 file path=usr/include/enlightenment/e.h
 file path=usr/include/enlightenment/e_Efx.h
 file path=usr/include/enlightenment/e_about.h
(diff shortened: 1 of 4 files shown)


pyhalov:
@alarcher : Can you look at it? 
Can we make enlightenment behave without SUID files? 
https://sourceforge.net/p/enlightenment/mailman/enlightenment-devel/thread/20081215172635.GA3359@black/

FreeBSD has SUID bit set just for freqset https://svnweb.freebsd.org/ports/head/x11-wm/enlightenment/pkg-plist?revision=422814&view=markup

optarg:
FreeBSD's approach is to inform users to set the SUID bit themselves:
https://svnweb.freebsd.org/ports/head/x11-wm/enlightenment/files/pkg-message.in?view=markup

It might be possible to use dbus+consolekit, but that's not something I'm particularly familiar with.

I would also point out that Enlightenment's intended mechanism for managing rights is sysactions.conf. If it is desirable to restrict users from being able to perform certain actions by default, one option is to simply change what is delivered in this file.

alarcher:
@pyhalov I will ask.

alarcher:
@pyhalov Enlightenment developers point out that the SUID bit is a reasonable approach for this application perimeter and any other solution would ultimately rely on a daemon running with same permissions. If people run multiuser systems they could restrict permissions as @optarg explained.
i would personally go with the PR as-is for now.

components/desktop/e/enlightenment/enlightenment.p5m
```diff
@@ -22,7 +22,6 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-<transform file path=usr/lib/(.*)/module.so$ -> default mode 0555>
 <transform file path=usr/lib/$(MACH64)/enlightenment/utils/(.*)$ -> default mode 0555>
 
 file path=usr/bin/emixer
```
pyhalov: At least, while you are here, fix usr/etc directory location, likely application needs sysconfdir

pyhalov:
Also, we don't want to grant so wide permissions to all these groups by default (look at etc/enlightenment/sysactions.conf). I'd be happy if we could grant them only to Console User and Primary Administrator... If it's too complex, allow everything just to root by default and set preserve=true on etc/enlightenment/sysactions.conf

optarg:
As far as I can gather, there's currently no support for anything besides traditional users and groups inside sysactions.conf. Judging from a cursory look at the source files, adding support for something like rights profiles is certainly possible, but maybe better left for a future commit. For now, I simply commented out the swathe of group allowances that were the default inside sysactions.conf.

alarcher:
@optarg Thank you for your contribution! Very much appreciated :) I do not have much time right now for Enlightenment. I did try EFL 1.19.x but there are a few issues. I have a patch for one of them (mkdir path on auto-mounted dirs). Thanks!

PR #3377, opened 2017-07-26
libgphoto2: update to 2.5.14

pyhalov:


pyhalov:
ABI-incompatible, fixes https://www.illumos.org/issues/8483

PR #3387, opened 2017-07-28
Initial recipe for payara-server-4.1.2.172.0 (evolution of glassfish)

jimklimov:
While working on this recipe, I found that the original GF project woke up from several years in slumber. So just in case we wrap it too at some point, I mediated the links.

Note that wrappers like `asadmin` did not prove to robustly find proper paths (unless I make a `/usr/glassfish` which is an option), so I did not make a `/usr/bin/asadmin` and users have to type its pathname in full.


pyhalov:
Can we separate logs to /var directory? 
Do we really need all these mediated links?

pyhalov:
Does it need separate user?

jimklimov:
I am not sure where there was a separate user (might be an artefact of borrowing from other recipes), but I think not required per packaging - a `webservd:webservd` should do.

Separating logs might be complicated: domains, nodes etc. in Glassfish are deployed and replicated and migrated in clusters as a big unit (each), essentially whatever is in dedicated tree under a subdirectory. It was a feat to not have the default domains pre-installed in the same dir as software but rather under /var, because usually the tarball is expanded under /opt and domains live there too - with deployed webapps, configs, logs and whatnot. Also it would make little sense as admins and tools using this technology also inherently expect such layout.

Mediating is similar usecase to mariadb/mysql/... as the original Glassfish project woke up from slumber and we might provide it too (Payara fork appeared because GF future competing with Oracle's own appserver products was uncertain and had this project in limbo for too long). In the same vein as mysql and forks, the particular vendor/build is mediation detail with the rest looking the same to generic user, including (should be) migratability of deployed domains to be run by that or other appserver version.

The choices of names/paths in mediation are open for discussion or improvement though. I expect to have the default domain in some predictable location so e.g. an integration package to deploy or upgrade jenkins.war under the appserver rather than standalone becomes feasible. Likewise, we could then pre-package a lot of other J2SE apps like OpenGrok, OpenDJ et al, Magnolia, Alfresco, LifeRay, etc. with default configs or slightly better, so people looking to deploy the stuff could get started with OpenIndiana with a simple `pkg install`. They would still have some homework to actually set these up, but lifecycle of actually installing things would be simplified, making it a bonus point for our distro as a successor to Solaris which was the original platform for those projects (and in many cases, Sun having started them).

pyhalov:
As for mediators I mostly was asking about 'link path=usr/share/payara mediator=payara mediator-implementation=payara-server'. Why do we need this if it's even not versioned? 
As for separate logs, I'm concerned that payara tried to write logs to /usr/share/payara-server/something. I expected to see them at /var/lib/payara-server/*. 
As for user, it was required to run asadmin as root and server was started as root user, using webservd would be much better.

jimklimov:
I think that thanks to symlinks, the domain logs would end up under /var. Maybe some config files like a saved generated master-password for service startup might go elsewhere. My setups were ok with everything under /opt so I did not dig into this separation much.

The appserver JVM can certainly run non-root, but this is IMHO up to end-user setup - to chown files, tweak SMF, maybe custom-name the domain, etc. We can document some suggestions of course to help along.

jimklimov:
Updated per similar nits from #3356 review

PR #3402, opened 2017-08-05
mesa: Remove libGLw dependency

optarg:
The OpenGL widget library (provided by x11/library/glw) is neither required to build nor use Mesa's core drivers/libs. It seems to be a historical relic that it gets included with the package at all. If it is desirable to continue delivering glw along with mesa, I'd like to propose setting its depend type to group so that it, and the legacy packages it depends on, can be avoided. Otherwise, I see no harm in simply removing it from the manifest.


pyhalov (approved):
I'm fine with this. 
@alarcher: What do you think?

alarcher (approved):

PR #3405, opened 2017-08-09
update to vala-0.36.4 last before valadoc merge

ghost:
Signed-off-by: jason <agrellum@gmail.com>

Please read thread https://mail.gnome.org/archives/vala-list/2017-August/msg00000.html

It applies to all vala packages, I have used so far.

It is the way vala is being built here or the C code vala generates or a library problem


components/developer/vala/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2016 <contributor>
+# Copyright 2017 <agrellum@gmail.com>
```
pyhalov: Don't edit sample-manifest

pyhalov:
Can you provide sample code, which dumps core?

pyhalov:
OK, got vala sample. Now would be good to get minimal C sample...

pyhalov:
It's a bug in valac (or in valac example https://valadoc.org/glib-2.0/GLib.Array.prepend_vals.html). It generates the following line of code (with -C):
<pre>  g_array_prepend_val (_tmp1_, "3. entry");</pre>
But glib manual explicitly says: 
<pre>
g_array_prepend_val() is a macro which uses a reference to the value parameter v . This means that you cannot use it with literal values such as "27". You must use variables.
</pre>

components/developer/vala/Makefile
```diff
@@ -20,37 +20,40 @@
 #
 # Copyright (c) 2013 David Hoeppner. All rights reserved.
 # Copyright (c) 2017 Andreas Wacknitz
+# Copyright (c) 2017 agrellum@gmail.com
```
pyhalov: please, use your name

ghost:
Fixed Makefile and sample-manifest; New to Openindiana and Solaris but did run package through 1000 samples. Using vala with webkitgtk2 it is easy to add another browser to desktop for local app development

pyhalov:
Will you report issue in valac to the developers?

components/developer/vala/vala.p5m
```diff
@@ -1,25 +1,16 @@
...
-# information: Portions Copyright [yyyy] [name of copyright owner]
-#
-# CDDL HEADER END
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
 #
 # Copyright (c) 2013 David Hoeppner. All rights reserved.
 # Copyright (c) 2017 Andreas Wacknitz
+# Copyright 2017 <agrellum@gmail.com>
```
pyhalov: your name here

ghost:
Fixed name. Will run all samples again, plus the ones at https://rosettacode.org/wiki/Category:Vala
Will try to isolate the bugs in the 13 vala samples that core on me and report upstream.

PR #3424, opened 2017-08-11
Updated GO to version 1.8.3

Toasterson:
Tested by Compiling https://github.com/Toasterson/opencloud

Works as expected.


components/developer/golang-18/Makefile
```diff
@@ -0,0 +1,159 @@
...
+#
+# - verify 1.5 produces "correct" result for bootstrapping 1.5 (upstream
+#   still uses 1.4 officially)
+# - after 1.5 has been available for at least a few builds in the WOS, switch
+#   to building Go with system version of Go
+# - when SPARC support is integrated into Go upstream, build upstream version
+#   using system Go for x86, then build Go for SPARC using upstream version
+#   just built for x86 (cross-compilation), package both at the same time
+# - after SPARC version has been available for at least a few builds in the
+#   WOS, switch to building Go with system version for x86 and SPARC
```
xen0l, line 63: Is this block still valid?

xen0l (commented):
I'll give it a try later this weekend, overall looks good. This work is very much appreciated! Thanks for looking into this!

components/developer/golang-18/Makefile
Toasterson: Yes I am still using 1.4 to bootstrap 1.8 compilation. I could look into upstream userland to see if that has changed yet but for now a simple version upgrade and some mediator fixes. Best of course would be an internal boostrapper in go 1.8 but I have not checked if that is the case yet. But this works and it's not much of a delay. It still compiles faster than any C or C++ thus far.

components/developer/golang-16/golang-16.p5m
```diff
@@ -25,8 +26,8 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 <transform file path=usr/lib/golang/1.6/pkg/tool/solaris_$(MACH64)/.+ -> \
     default mode 0555>
 
-link path=usr/bin/go target=../lib/golang/1.6/bin/go
-link path=usr/bin/gofmt target=../lib/golang/1.6/bin/gofmt
+link path=usr/bin/go target=../lib/golang/1.6/bin/go mediator=golang mediator-version=1.6
```
xen0l: COMPONENT_REVISION has to be bumped in the Makefile, so the new package is published. Otherwise, golang-18 won't be installable and will fail with conflicting actions error.

components/developer/golang-18/Makefile
```diff
@@ -0,0 +1,159 @@
...
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2014, 2015, Oracle and/or its affiliates. All rights reserved.
+# Modified by Aurelien Larcher, 2016.$
```
xen0l: Use the proper copyright line please

components/developer/golang-18/Makefile
```diff
@@ -0,0 +1,159 @@
...
+# Modifies by Till Wegmüller, 2017
```
xen0l: Use the proper copyright line please

components/developer/golang-18/Makefile
```diff
@@ -0,0 +1,159 @@
...
+#
+
+include ../../../make-rules/shared-macros.mk
+
+PATCH_EACH_ARCHIVE= 1
+
+COMPONENT_NAME=			golang
+COMPONENT_VERSION=		1.8.3
+COMPONENT_REVISION=		0
```
xen0l: COMPONENT_REVISION is zero by default, drop this line

xen0l (changes requested):

xen0l:
I found some problems when trying out the package. These need to be fixed, so the package is installable.

components/developer/golang-18/Makefile
Toasterson: Even if Golang always has had this Copyright?

xen0l: I don't think we are noting "modified by" in copyright, we always used copyright lines. However, I am not sure on this. @pyhalov, what's your take on this?

pyhalov: Just add @alarcher's copyright and be done with this

xen0l:
Man page also needs to be mediated.

Toasterson:
Changes made. Anything else I missed?

components/developer/golang-16/golang-16.p5m
```diff
@@ -25,8 +26,12 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 <transform file path=usr/lib/golang/1.6/pkg/tool/solaris_$(MACH64)/.+ -> \
     default mode 0555>
 
-link path=usr/bin/go target=../lib/golang/1.6/bin/go
-link path=usr/bin/gofmt target=../lib/golang/1.6/bin/gofmt
+<transform file path=usr/share/man/.+ -> default mediator-version 1.6 >
```
pyhalov: What are you trying to do with this transform?

components/developer/golang-18/golang.p5m
```diff
@@ -22,4 +23,4 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-depend type=require fmri=developer/golang-16
+depend type=require fmri=developer/golang-18
```
pyhalov, line 26: do we want to depend on the latest or have require-any dependency?

pyhalov (approved):

xen0l (approved):

PR #3431, opened 2017-08-16
Problem: userland-fetch --keep was lost in refactoring

jimklimov:
Solution: bring it back, with a comment nearby now

Signed-off-by: Jim Klimov <jimklimov@gmail.com>

### DIFF
--- a/tools/userland-fetch
+++ b/tools/userland-fetch
@@ -278,8 +278,8 @@ def download_from_paths(search_list, file_arg, url, link_arg, quiet=False):
 
 def usage():
     print "Usage: %s [-a|--user-agent (user-agent)] [-f|--file (file)] [-l|--link] " \
-        "[-h|--hash (hash)] [-s|--search (search-dir)] [-S|--sigurl (signature-url)] " \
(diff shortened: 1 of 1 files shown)


tools/userland-fetch
```diff
@@ -383,11 +383,22 @@ def main():
...
+                # downloaded twice.
                 if not sig_valid:
-                    try:
-                        os.remove(name)
-                    except OSError:
-                        pass
+
+                    if keep_arg == False:
+                        try:
+                            print "\nWARN: Removing the corrupt downloaded file"
```
pyhalov, line 396: please, keep it silent by default

jimklimov: It already does a lot of diagnostics prints above - like what it is pulling, what checksums were expected and actual etc. And it did print this line until the block was lost to refactoring a couple of commits back. And in fact knowing the verdict and action taken in the end may be as (or more) useful than the 10 or so lines printed anyway before this.

I agree that the tool could use a verbosity setting, but that is out of scope for this change and would impact many other lines of code here.

Also note that the usual use-case is of downloading or using already-cached file which has its expected checksum correctly, so this block does not fire at all.

PR #3441, opened 2017-08-20
support intl API . Some Thunderbird AddOns need this. e.g. Enigmail .

robberer:


pyhalov:

robberer:
updated

AndWac:
Changing the COMPONENT_VERSION is wrong here. You have been asked to set the COMPONENTG_REVISION, eg. COMPONENT_REVISION = 1.
The COMPONENT_VERSION is what we get from the outside (typically sources), while
changing the COMPONENT_REVISION makes sure that changes after the first release (COMPONENT_REVISION unset) will make it to the machines that already have the version installed. When you make changes to an already released version you will have to increase the COMPONENT_REVISION. When you change the COMPONENT_VERSION (eg. when a new upstream version has been released) you need to unset the COMPONENT_REVISION (re-start it).

AndWac:
When you are done with everything the last step is to squash the commits into one.
(Search for "How can I merge commits in git?")

PR #3442, opened 2017-08-20
add fish-2.6.0

robberer:
patch reference: https://github.com/fish-shell/fish-shell/issues/3340#issuecomment-243014980


components/shell/fish/patches/path_max.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/wutil.cpp b/src/wutil.cpp
+index 5b69c47f..9c4bedd2 100644
+--- a/src/wutil.cpp
++++ b/src/wutil.cpp
+@@ -139,7 +139,7 @@ bool wreaddir_for_dirs(DIR *dir, wcstring *out_name) {
+ const wcstring wgetcwd() {
+     wcstring retval;
+ 
+-    char *res = getcwd(NULL, 0);
++    char *res = getcwd(NULL, PATH_MAX);
```
pyhalov, line 10: PATH_MAX + 1 ?

pyhalov:
Where are p5m file and sample-manifest ?

components/shell/fish/Makefile
```diff
@@ -0,0 +1,50 @@
+#
```
pyhalov, line 1: Please, use new template for header (look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/templates/configure.mk)

components/shell/fish/patches/path_max.patch
robberer: Don't know. https://github.com/fish-shell/fish-shell/issues/3340#issuecomment-269692610 says just "PATH_MAX" and it fixed the `getcwd() failed with errno 22/Invalid argument` error for me

components/shell/fish/Makefile
robberer: OK, i will do so. To generate the sample-manifest and p5m: is this the right document to study ? https://wiki.openindiana.org/oi/Building+an+IPS+manifest

This is my first contribution to oi-userland. I have to learn a lot.

robberer: OK, sample-manifest is generated, how to get a p5m file ?

components/shell/fish/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2017 <contributor>
+# Copyright 2017 Longrin Wischnewski
```
pyhalov: Don't do this. sample-manifest shouldn't be changed. It's just a reference for future use. Only p5m file is used  by build system.

components/shell/fish/patches/path_max.patch
pyhalov: I still feel that PATH_MAX is one symbol too short ('\0').

components/shell/fish/Makefile
pyhalov: Copy sample-manifest to component directory and edit as necessary

robberer: done

components/shell/fish/patches/path_max.patch
robberer: I've tested with PATH_MAX + 1 and still no error.  So if you think that PATH_MAX should be + 1 then i will update the patch

xen0l: @pyhalov from http://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html:
<pre>
{PATH_MAX}
Maximum number of bytes in a pathname, including the terminating null character.
Minimum Acceptable Value: {_POSIX_PATH_MAX}
[XSI] [Option Start] Minimum Acceptable Value: {_XOPEN_PATH_MAX} [Option End]
</pre>

So I think the terminating null character is included (assuming our strong posix compliance). I haven't found any direct command/proof though.

xen0l:
I ran a test build and I found some issues in the packaging of the component, but it seems to be working from what I have tried (I tried out fish several times in the past on different systems).

components/shell/fish/fish.p5m
```diff
@@ -0,0 +1,915 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/$(MACH64)/fish
```
xen0l: Please, deliver these in /usr/bin/ instead. Just do something similar:
<pre>
diff --git a/components/shell/fish/Makefile b/components/shell/fish/Makefile
index dadc1fcf1..e1e4bb964 100644
--- a/components/shell/fish/Makefile
+++ b/components/shell/fish/Makefile
@@ -30,6 +30,8 @@ include $(WS_MAKE_RULES)/prep.mk
 
+CONFIGURE_BINDIR.64 = $(USRBINDIR)
+
 COMPONENT_PRE_CONFIGURE_ACTION = ($(CLONEY) $(SOURCE_DIR) $(@D))
 
 build: $(BUILD_64)
</pre>

components/shell/fish/Makefile
```diff
@@ -0,0 +1,42 @@
...
+COMPONENT_PRE_CONFIGURE_ACTION = ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+build: $(BUILD_64)
+
+install: $(INSTALL_64)
+
+test: $(NO_TESTS)
+
+REQUIRED_PACKAGES += system/library
+# Auto-generated contents below.  Please manually verify and remove this comment
```
xen0l: Drop this comment.

components/shell/fish/Makefile
```diff
@@ -0,0 +1,42 @@
...
+
+COMPONENT_PRE_CONFIGURE_ACTION = ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+build: $(BUILD_64)
+
+install: $(INSTALL_64)
+
+test: $(NO_TESTS)
+
+REQUIRED_PACKAGES += system/library
```
xen0l: REQUIRED_PACKAGES is wrong, I am getting these:

<pre>
REQUIRED_PACKAGES += system/library/g++-4-runtime
</pre>

Please double check this, otherwise if somebody will be installing/modifying package in the future, he will receive build errors.

xen0l (changes requested):

components/shell/fish/patches/path_max.patch
pyhalov: @xen0l OK, thanks

components/shell/fish/fish.p5m
```diff
@@ -0,0 +1,915 @@
...
+file path=usr/bin/$(MACH64)/fish_indent
+file path=usr/bin/$(MACH64)/fish_key_reader
+file path=usr/etc/fish/config.fish
```
pyhalov: This is clearly wrong, likely you have to specify --sysconfdir=/etc

pyhalov: Also add preserve=true attribute to this file.

components/shell/fish/Makefile
robberer: OK, last time i did `gmake REQUIRED_PACKAGES`, which has given me just the autogenerated line. Today i got the same required packages as you mentioned.

robberer:
updated

components/shell/fish/Makefile
```diff
@@ -0,0 +1,49 @@
...
+
+CONFIGURE_BINDIR.32 = $(USRBINDIR)
+CONFIGURE_BINDIR.64 = $(USRBINDIR)
+
+COMPONENT_PRE_CONFIGURE_ACTION = ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+CONFIGURE_OPTIONS.32 += --sysconfdir=/etc
+CONFIGURE_OPTIONS.64 += --sysconfdir=/etc/$(MACH64)
+
+build: $(BUILD_32_and_64)
```
pyhalov: Why build both? Build just 64-bit version, deliver files in /usr/bin and use  use /etc sysconfdir.

robberer: I thought it's a good idea to offer this built for 32 bit systems too as OI seems to have some kind of support for it. But clearly I'm wrong :-)

Fixed

pyhalov:
Fails during gmake publish:
```
Couldn't find 'usr/share/doc/fish/ftv2blank.png' in any of the specified search directories:...
```
Do you have doxygen installed? Likely no. Install it, regenerate sample-manifest, fix p5m file and add to REQUIRED_PACKAGES list

robberer:
No I didn't had doxygen installed and it was mentioned nowhere in the documentation i've read. How should i've known.

And how can it be that every `gmake publish` builds fine on my computer while on yours it does not ?

pyhalov:
Because in my case it finds doxygen, regenerates documentation and produces different set of files.

PR #3443, opened 2017-08-20
efl: Fix missing execute bits on some files

optarg:
Plus some cosmetic changes.

### DIFF
--- a/components/desktop/e/efl/Makefile
+++ b/components/desktop/e/efl/Makefile
@@ -24,17 +24,18 @@ include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=           efl
 COMPONENT_VERSION=        1.18.5
-COMPONENT_FMRI=           library/e/efl
+COMPONENT_REVISION=       1
+COMPONENT_FMRI=           library/e/$(COMPONENT_NAME)
 COMPONENT_CLASSIFICATION= System/Libraries
 COMPONENT_SRC=            $(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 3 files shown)


components/desktop/e/efl/Makefile
```diff
@@ -50,16 +51,17 @@ LDFLAGS  += $(JPEG_LDFLAGS)
...
+CONFIGURE_OPTIONS+= --with-profile=release
 CONFIGURE_OPTIONS+= --enable-debug
 CONFIGURE_OPTIONS+= --enable-fribidi
 CONFIGURE_OPTIONS+= --enable-harfbuzz
+CONFIGURE_OPTIONS+= --disable-libeeze
 CONFIGURE_OPTIONS+= --enable-lua-old
 CONFIGURE_OPTIONS+= --enable-pulseaudio
 CONFIGURE_OPTIONS+= --disable-cserve
 CONFIGURE_OPTIONS+= --enable-physics
-CONFIGURE_OPTIONS+= --disable-gstreamer
```
alarcher, line 60: Any particular reason to remove these lines?

optarg: They're already the default in efl moving forward; removing them doesn't change the build. I just didn't find specifying them explicitly useful. I can certainly leave them in if you'd prefer.

alarcher: I usually prefer to explicit important options even if they have defaults because then the Makefile itself is informative and not bound to changes in default behaviour.
But since these options are likely not changing anymore I am fine with this.

alarcher (approved):

PR #3444, opened 2017-08-21
Bump inkscape to 0.92.2

alarcher:


pyhalov:
Have you checked gmake test? Fails here with 
```
uk.po: unclosed <small>:
#: ../src/file-update.cpp:342
...
make[3]: *** [Makefile:1927: check_markup] Error 1
```

PR #3450, opened 2017-08-23
5709 Add binary compatibility with Solaris 10 update 10

pyhalov:


pyhalov:
@andy-js I think I'll add this locally...

xen0l (approved):

PR #3462, opened 2017-08-31
Adding Putty command line tools

Toasterson:
This adds putty command line tools as package. As the Compilation is a little bit complicated i.e its not ported. I thought that a package would be nicer than a wiki entry how to compile putty cmd tools on illumos.


alarcher:
Any reason to deliver 32-bit binaries?

components/putty/Makefile
```diff
@@ -0,0 +1,43 @@
+#
```
pyhalov: please, move it to corresponding category (desktop or network)

components/putty/Makefile
```diff
@@ -0,0 +1,43 @@
...
+# Copyright 2017 Till Wegmueller
+#
+
+include ../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=putty
+COMPONENT_VERSION=0.70
+COMPONENT_SUMMARY=A terminal integrated SSH/Telnet client commandline tools
+COMPONENT_PROJECT_URL=http://www.chiark.greenend.org.uk/~sgtatham/putty/
+COMPONENT_FMRI=network/security/$(COMPONENT_NAME)
```
pyhalov: We have no other fmris in network/security/* , perhaps, use just network/putty or security/putty?

components/putty/Makefile
```diff
@@ -0,0 +1,43 @@
...
+LDFLAGS += -lnsl -lsocket
+
+CONFIGURE_OPTIONS+=--without-gssapi 
+CONFIGURE_OPTIONS+=--without-gtk
+
+build:		$(BUILD_32)
+
+install:	$(INSTALL_32)
+
+test:		$(TEST_32)
```
pyhalov: Why REQUIRED_PACKAGES list is empty?

components/putty/patches/02-cmdgen.patch
```diff
@@ -0,0 +1,69 @@
+--- putty-0.70/cmdgen.c	2017-07-04 21:30:55.000000000 +0000
++++ cmdgen.c	2017-08-30 04:57:16.469173728 +0000
+@@ -236,8 +236,8 @@
```
pyhalov: What are we changing here and why?

components/putty/Makefile
Toasterson: Putty Compiles to static Binaries. Other than system/Library there are no dependencies. (Except compiler ofc) and afaik we did not mention system/Library as it needs to be installed anyway. Has that changed?

components/putty/patches/02-cmdgen.patch
Toasterson: Apparently there is a C Variable called Private which is used in other places somewhere. GCC complained about double Declared Variable. Thus I renamed it to PPRIVATE meaning PuttyPrivate so that putty compiles.

Toasterson:
@alarcher I have only taken the template from templates/configure.mk and adapted the variables listed there. AFAIK we deliver 64bit only binaries in /usr/bin and not /usr/bin/amd64 however when I edit the Makefile to only build 64bit I am unable to make it put those under usr/bin. Do we want 64bit only to be delivered under /usr/bin/amd64 or is there an example from which I could work from?

components/putty/Makefile
pyhalov: OK

components/putty/patches/02-cmdgen.patch
pyhalov: sorry, I haven't spotted double P

pyhalov:
Likely you have to set CONFIGURE_BINDIR.32=$(CONFIGURE_PREFIX)/bin or something similar to make packaging system put 64-bit binaries in /usr/bin

Toasterson:
Changes made as requested.

components/network/putty/Makefile
```diff
@@ -0,0 +1,45 @@
...
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+LDFLAGS += -lnsl -lsocket
+
+CONFIGURE_OPTIONS+=--without-gssapi 
+CONFIGURE_OPTIONS+=--without-gtk
+CONFIGURE_BINDIR.64=$(CONFIGURE_PREFIX)/bin
+CONFIGURE_BINDIR.32=$(CONFIGURE_PREFIX)/bin
```
pyhalov: You can delete this line (setting 32-bit variant)

Toasterson:
Changes made as requested. Via Github. Anything else to fix?

pyhalov:
Can I ask why don't we ship putty itself (perhaps, in separate package)?

Toasterson:
I have not had the time yet to make it compile. This was changes made during $Work as we needed puttygen on the server. Putty Itself needs other changes probably in a similar fashion but it's not as easy to as just changing the Option and Compiling Unfortunately. As there could be more people out there needing a way to make putty keys at all on a solaris box I thought i contribute at least the commandline tools to make that task easier for others.

pyhalov:
This is fine. Perhaps, name package not putty, but network/putty/SOMETHING to preserve putty name for GUI client?

Toasterson:
Hmm. It sounds reasonable but could have negative implications as the package would get a shortname people don't expect. And I don't want to name it something like puttygen as that would confuse even more. If you have an idea for an name that is general enough but not general enough then I am all ears. But I would say network/putty contains all the Putty tools available for our Platform.

pyhalov:
Then you'll have to install gtk on the server when putty itself is added to the set. What about network/putty/putty-utils or network/putty/putty-tools ?

Toasterson:
Good point. I'll go with putty-tools.

Toasterson:
As Putty Will be from the same source I figured that changing the FMRI in the manifest is the best approach. AFAIK we do that with some other packages aswell.

PR #3468, opened 2017-09-04
Bump PNG libraries

alarcher:


pyhalov:
Were other libpng* updated?

jimklimov:
Recipe-wise, LGTM. Is it possible to use macros for bumping the SONAME versions, as I did in a few recipes where SO version is like component version? This could simplify maintentance in the long run.

alarcher:
It seems other libpngs can be bumped to 1.2.58 and 1.4.21.

pyhalov:
@alarcher : can you include their update in this PR?

PR #3478, opened 2017-09-07
Bump cmake to 3.9.1

alarcher:


pyhalov:
cmake bundles libuv. Please, add the following patch https://github.com/libuv/libuv/commit/26daa99e2c93db2692933e8d141e44661a444e68

pyhalov:
Have you checked it, building different mysqls?

alarcher:
Patch added. I had checked a bunch of components but not mysql/percona.
I have built percona-server-57 now.

PR #3480, opened 2017-09-08
Bump mariadb-101 to 10.1.26

alarcher:
Previous mirror has become inactive.
One test failing in 32-bits to be checked.


pyhalov:
Do I remember correctly that it also fails on 10.1.24?

alarcher:
@pyhalov I just tried, the same test failed with 10.1.24.

PR #3495, opened 2017-09-14
dnsmasq: remove unsupported privilege set from SMF manifest

olbohlen:
our SMF version does not support the privilege sets specified in <method_credential/>


pyhalov:
@xen0l what do you think? Is it good as is or do we want to preserve at least privileges='basic,net_privaddr,net_icmpaccess' ?

xen0l:
I definitely prefer preserving those privileges we can.

pyhalov (approved):

xen0l (approved):

PR #3499, opened 2017-09-16
Patch node, node-6 and node-7 so that getifaddrs may return more address families

wiedi:
This is in preparation for illumos #3729 so that getifaddrs may return other address families than AF_INET and AF_INET6.
This is a backport of a libuv fix (https://github.com/libuv/libuv/commit/26daa99e2c93db2692933e8d141e44661a444e68) that is also already included in node-8.


wiedi:
hold off, I missed something. Updated patch comming soon

wiedi:
Ok, should be good now.
Did "gmake publish" for all three and tested `./build/i86/node -e 'console.log(require("os").networkInterfaces())'` which should execute the patched codepath.

PR #3501, opened 2017-09-17
Bump net-snmp to version 5.7.3.

grueni:
The PR contains the patches from oracle-userland and smartos.
All enabled tests pass successfully.


grueni:
Should the folder net-snmp-57 renamed to net-snmp?

components/sysutils/net-snmp-57/Makefile
```diff
@@ -0,0 +1,188 @@
...
+#
+# net-snmp has a "special" build where only 32-bit items are installed into
+# proto area; so prefer 32-bit to preserve existing sourcing.
+#
+
+#
+## Copyright 2016 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
+#
+
+BUILD_BITS= 32_and_64
```
xen0l: Is this needed?

components/sysutils/net-snmp-57/Makefile
```diff
@@ -0,0 +1,188 @@
...
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+  ($(CLONEY) $(SOURCE_DIR) $(@D); $(CLONEY) $(COMPONENT_DIR)/sun $(@D)/sun)
+
+CPPFLAGS += "-DFALSE_SHARING_ALIGN=64"
+
+CONFIGURE_ENV +=	PYTHONPROG=$(PYTHON.$(BITS))
+CONFIGURE_ENV.32     += PERL=$(PERL.5.22)
```
xen0l: Why use different perl versions?

pyhalov:
As for folder name, I don't care much. I care about compatibility.

- [x] Is ABI compatibility broken?
- [ ] Do oi-userland software, dependent on net-snmp, compile happily against new version?
- [ ] Does illumos-gate compile happily against new version?

I see at least following dependent packages:
<pre>
pkg:/application/cluster/cluster-glue
pkg:/application/cluster/corosync
pkg:/application/cluster/pacemaker
pkg:/network/asterisk
pkg:/print/filter/hplip
pkg:/service/fault-management (comes from illumos-gate)
pkg:/service/fault-management/snmp-notify (comes from illumos-gate)
pkg:/service/hal (comes from illumos-gate)
pkg:/system/management/nut/drivers
pkg:/system/management/powerman
</pre>

pyhalov:
Old net-snmp version should be removed from the gate in this PR. Also we likely need updated  https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/sysutils/net-snmp/patches/__FUNCTION__.patch here. And, please, move component to sysutils/

components/sysutils/net-snmp-57/Makefile
pyhalov: because perl 5.22 is 32-bit-only, 5.24 - 64-bit-only

components/sysutils/net-snmp-57/Makefile
pyhalov: Likely, as illumos-gate depends on net-snmp libs.

pyhalov: No

components/sysutils/net-snmp-57/Makefile
```diff
@@ -0,0 +1,188 @@
...
+COMPONENT_PROJECT_URL=	http://www.net-snmp.org/
+COMPONENT_SRC=    $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= \
+    sha256:12ef89613c7707dc96d13335f153c1921efc9d61d3708ef09f3fc4a7014fb4f0
+COMPONENT_ARCHIVE_URL=	http://ftp.ntua.gr/mirror/net-snmp/net-snmp/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_CLASSIFICATION= System/Enterprise Management
+COMPONENT_SUMMARY= Net-SNMP Agent files and libraries
+
+BUILD_32_and_64 += $(BUILD_DIR_32)/.docs
```
pyhalov: We don't use this variable. And I'm not sure where INSTALL_TARGET and TEST_TARGET are used.

grueni: I removed 
- INSTALL_TARGET
- TEST_TARGET 
- BUILD_32_and_64

components/sysutils/net-snmp-57/Makefile
grueni: Ok, removed.

components/sysutils/net-snmp-57/Makefile
grueni: Perl 5.22 is 32-bit and 5.24 is 64-bit.
I thought it would be better to use the same ELF binary type when the modules for perl are built.

grueni:
AFAIK __FUNCTION__.patch  is obsolete. If not compilation if Illumos will fail?
The folder net-snmp-57 is already in component/sysutils.

pyhalov:
No, look at https://www.illumos.org/issues/2245 . Try to remove FUNCTION patch and compile illumos-gate with Sun Studio

pyhalov:
This update breaks ABI, so illumos-gate and dependent oi-userland components should be recompiled.

components/sysutils/net-snmp/net-snmp-base.p5m
```diff
@@ -403,10 +403,11 @@ file build/prototype/$(MACH64)/usr/bin/$(MACH64)/ipf-mod.pl \
 file build/prototype/$(MACH64)/usr/bin/$(MACH64)/mib2c path=usr/bin/mib2c
 file build/prototype/$(MACH64)/usr/bin/$(MACH64)/mib2c-update \
     path=usr/bin/mib2c-update
-link path=usr/bin/net-snmp-config target=./net-snmp-config-64
-file usr/bin/net-snmp-config path=usr/bin/net-snmp-config-32
+
+file usr/bin/net-snmp-config path=usr/bin/net-snmp-config
 file build/prototype/$(MACH64)/usr/bin/$(MACH64)/net-snmp-config \
-    path=usr/bin/net-snmp-config-64
+    path=usr/bin/$(MACH64)/net-snmp-config
```
pyhalov: Please, also preserve old names (as links)

components/sysutils/net-snmp/net-snmp-base.p5m
```diff
@@ -407,6 +407,8 @@ file build/prototype/$(MACH64)/usr/bin/$(MACH64)/mib2c-update \
 file usr/bin/net-snmp-config path=usr/bin/net-snmp-config
 file build/prototype/$(MACH64)/usr/bin/$(MACH64)/net-snmp-config \
     path=usr/bin/$(MACH64)/net-snmp-config
+link path=usr/bin/net-snmp-config-32 target=usr/bin/net-snmp-config pkg.linted=true
```
pyhalov: link is incorrect. It should have been 
link path=usr/bin/net-snmp-config-32 target=net-snmp-config 
link path=usr/bin/net-snmp-config-64 target=./$(MACH64)/net-snmp-config

grueni:
Yes, I realized it during testing.
The other 2 PRs can be changed to used the old files. So they are independent from this PR.

grueni:
Current Illumos compiles without problems.

State of dependent packages:

- [x] pkg:/service/fault-management (comes from illumos-gate)
- [x] pkg:/service/fault-management/snmp-notify (comes from illumos-gate)
- [x] pkg:/service/hal (comes from illumos-gate)
- [x] pkg:/network/asterisk
- [x] pkg:/print/filter/hplip
- [x] pkg:/system/management/nut/drivers
- [x] pkg:/system/management/powerman
- [x] pkg:/application/cluster/cluster-glue
- [x] pkg:/application/cluster/corosync
- [x] pkg:/application/cluster/pacemaker

pyhalov:
Why did you close it?

grueni:
erroneously, was toot tired.

pyhalov:
So, do I understand correctly that you have checked every dependent component?

grueni:
Yes.
We have additional 2 PRs pending.
And I checked in several branches ending with -64 with the cluster modules.
Actually I am testing the deployment of the packages.
Tomorrow I will update crmsh and pcs to the newest versions.
And add 2 PRs for powerman and hplip.

grueni:
powerman: nothing to change, just recompile, creates 32 and 64 bits
hplip: nothing to change, just recompile, creates 32 in 2 variants.

components/sysutils/net-snmp/history
```diff
@@ -1,2 +0,0 @@
-library/python-2/net-snmp-26@5.4.4,5.11-2016.0.1.0
```
pyhalov: please, preserve this file

pyhalov:
We have an issue. When old net-snmp is installed on system, we get 
```
.../oi-userland/components/sysutils/net-snmp/build/manifest-i386-net-snmp-base.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libnetsnmptrapd.so.15 \
        pkg.debug.depend.reason=usr/perl5/vendor_perl/5.22/i86pc-solaris-64int/auto/NetSNMP/TrapReceiver/TrapReceiver.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=lib \
        pkg.debug.depend.path=usr/lib \
        pkg.debug.depend.path=usr/perl5/5.22/lib/i86pc-solaris-64int/CORE'.
/export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/build/manifest-i386-net-snmp-base.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libnetsnmpagent.so.15 \
        pkg.debug.depend.reason=usr/perl5/vendor_perl/5.22/i86pc-solaris-64int/auto/NetSNMP/agent/agent.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=lib \
        pkg.debug.depend.path=usr/lib \
        pkg.debug.depend.path=usr/perl5/5.22/lib/i86pc-solaris-64int/CORE'.
/export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/build/manifest-i386-net-snmp-base.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libnetsnmpagent.so.15 \
        pkg.debug.depend.reason=usr/perl5/vendor_perl/5.22/i86pc-solaris-64int/auto/NetSNMP/TrapReceiver/TrapReceiver.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=lib \
        pkg.debug.depend.path=usr/lib \
        pkg.debug.depend.path=usr/perl5/5.22/lib/i86pc-solaris-64int/CORE'.
```

pyhalov:
It seems there's some issue in build logic. If I rebuild already built net-snmp, it builds as expected. Likely some missing dependency.

pyhalov:
@grueni did you use/test updated snmpd in real applications?

grueni:
I created a zone, relaxed the constraints for snmp and installed the new package. After reboot Ic checked the log of snmpd.log and found no errors.
In this zone I build all packages which are dependent from net-snmp.
And I built Illumos too.

grueni:
Please try it again.
I forgot to push my local already rebased version. The changes of alp are already integrated.
When I compile this version in a zone with net-snmp 5.4.4 installed I get no errors concerning libnetsnmptrapd.so.15.

PR #3506, opened 2017-09-19
libxml2: update to 2.9.5 (upstream change)

pyhalov:


pyhalov:
Built illumos-gate with it. Without https://github.com/illumos/illumos-gate/commit/919f1431f035c123b6b6e1efa2554b6e8729b494 it builds fine. And that commit just has broken lint...

PR #3508, opened 2017-09-20
Asterisk

grueni:
compiled with net-snmp-5.7.3 and amd64 as default for SNMP creates this error:
ERROR userland.action001.2        64-bit object 'usr/lib/asterisk/modules/res_snmp.so' in 32-bit path

This patch is just to avoid the lint error message.

### DIFF
--- a/components/network/asterisk/Makefile
+++ b/components/network/asterisk/Makefile
@@ -28,7 +28,7 @@ include $(WS_TOP)/make-rules/prep.mk
 include $(WS_TOP)/make-rules/configure.mk
 include $(WS_TOP)/make-rules/ips.mk
 
-COMPONENT_PRE_CONFIGURE_ACTION += ($(CLONEY) $(SOURCE_DIR) $(@D));
+COMPONENT_PRE_CONFIGURE_ACTION += ($(CLONEY) $(SOURCE_DIR) $(@D))
 
 PATCH_LEVEL = 0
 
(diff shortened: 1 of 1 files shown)


pyhalov:
This looks like a hack, and ugly one... Can we just have /usr/bin/$(MACH32)/net-snmp-config (and possibly /usr/bin/$(MACH64)/net-snmp-config), as with other tools and just set PATH as needed to find 32/64-bit version?
Also, please, preserve only one commit in this PR and bump COMPONENT_REVISION.

grueni:
net-snmp creates
lrwxrwxrwx 1 root root    20 Sep 17 23:18 /usr/bin/net-snmp-config -> ./net-snmp-config-32
-r-xr-xr-x 1 root bin  26207 Sep 17 23:18 /usr/bin/net-snmp-config-32
-r-xr-xr-x 1 root bin  26070 Sep 17 23:18 /usr/bin/net-snmp-config-64
and checks then in configure a custom path for net-snmp-config.
This was the simplest method to avoid the lint failure.

It worked with net-snmp-5.4.4 but will not with net-snmp-5.7.3. net-snmp-config is a link to net-snmp-config-32.

We can change net-snmp-5.7.3 to use 
/usr/bin/net-snmp-config
/usr/bin/amd64/net-snmp-config
and not the files shown above.
With this this patch would be:
CONFIGURE_OPTIONS +=  $(BINDIR) 
or
CONFIGURE_OPTIONS +=  $(BINDIR)/amd64
depending on binary type of snmpd.

pyhalov:
changing net-snmp to deliver 
/usr/bin/net-snmp-config
/usr/bin/amd64/net-snmp-config

seems to be the sanest way. Perhaps, preserve -32/-64 symlinks for compatibility.

pyhalov:
Not needed

PR #3509, opened 2017-09-20
Updated Puppet to 4.10.6 analog to oracle-userland

Toasterson:
I kept our custom types and providers even if oracle-userland does not, as they have not landed yet upstream.


components/ruby/puppet/patches/0001-no-upstream-Solaris-auditing-support.patch
```diff
@@ -0,0 +1,98 @@
+From d68b018f2d30df2824335b579ee3bff4c10cb1e0 Mon Sep 17 00:00:00 2001
```
pyhalov: This patch relies on PuppetAudit class, which is provided by ext/*  in upstream repository and likely will need some work to port to OI.

pyhalov:
Also solaris-userland has more patches. At least 0004-no-upstream-install-path-changes.patch, 0005-PUP-7745-package-latest-method-does-not-refresh-meta.patch look like they should be imported.

Toasterson:
I SPecifically did not include these patches because for 0004 only changes the enterprise puppet location and does not influence the path where puppet is put. It would not be correct acccording to puppet documentation. For 0005 we don't need that as our pkg still refreshes metadata on every run. At least it say to do that when I want to install Software.

pyhalov:
I was under impression that it changes location where puppet looks for things, and as we put it under /usr, it should be changed.

pyhalov:
As for 0005-PUP-7745-package-latest-method-does-not-refresh-meta.patch, it does 'pkg refresh' before pkg list, and we do not do this automatically, so I consider it should be included.

Toasterson:
For the Path changes as far as I can tell they are all configurable with puppet config utility / configurationfile. From the Upstream Site they say the merged the enterprise and opensource code into one and thus the paths aswell. It is up to us if we want to change them. However I am not sure if this would not break something if somebody decides to install enterprise edition (unlikely but possible).  It is also something that is wildly inconsistent among *Nix oses. I personally always prefer to stick to upstream in such cases. Just so that we are not breaking something. I will be using puppet over the next few weeks so we could include the patch and if something breaks take it out again. What would be your preference?

Toasterson:
As for the pkg changes: I did not think about pkg list -a.... Inegrating that patch now.

pyhalov:
> I will be using puppet over the next few weeks so we could include the patch and if something breaks take it out again. What would be your preference?

Yes. When we are on Solarish system, I believe Solaris team knows better than software developers...

Toasterson:
True that.

Sidenote from my vpn zone this happens when issuing pkg list -a
```
$ pkg list -a openvpn
Refreshing catalog 1/1 openindiana.org 316.81 kB
```

Toasterson:
All changes puhed as requested

pyhalov:
Check manifests, they include *.`~1`~ files!!!
```
Couldn't find 'usr/ruby/2.3/lib/ruby/vendor_ruby/2.3.0/puppet/agent.rb.~1~' in any of the specified search directories:
...
```

Toasterson:
Huh Wat are those even? can't be vi files. Anyway fixed. They are removed.

pyhalov:
```
puppet agent --test --verbose
Info: Creating a new SSL key for buildzone.oi-build.r61.net
/usr/sbin/svcadm: illegal option -- s
```

Where does it come from??? Looks like command is missed...

Toasterson:
i would say either one of our providers or one of patches. Could be the mark as degraded? I'll have more time tomorrow to check.

pyhalov:
And we still miss 0004-no-upstream-install-path-changes.patch . Without it puppet tries to create /opt/puppetlabs/...

Toasterson:
Oh sorry thought I added that one...

Toasterson:
I have tested it on my system the past few days and it works as expected. If you don't have anything else to add I would say ship it.

pyhalov:
I've created https://github.com/Toasterson/oi-userland/pull/1 to fix issues on 'puppet agent --test --verbose'

components/ruby/puppet/puppet-common.p5m
```diff
@@ -34,131 +34,133 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
...
 set name=org.opensolaris.smf.fmri value=svc:/application/puppet:agent \
     value=svc:/application/puppet:master
-dir  path=etc/puppet owner=puppet group=puppet mode=0755
+dir  path=etc/puppetlabs/ owner=puppet group=puppet mode=0755
+dir  path=etc/puppetlabs/code owner=puppet group=puppet mode=0755
+dir  path=etc/puppetlabs/puppet owner=puppet group=puppet mode=0755
 dir  path=var/lib/puppet owner=puppet group=puppet mode=0755
 dir  path=var/log/puppet owner=puppet group=puppet mode=0755
-file path=etc/puppet/auth.conf owner=puppet group=bin mode=0444
+file path=etc/puppetlabs/puppet/auth.conf owner=puppet group=bin mode=0444
```
pyhalov: mode=0644 preseve=true?
The same  - for other etc/puppetlabs/*.conf files....

Toasterson: Ah yep I have added that in the latest revision. I thought we had a general transform for etc.

Toasterson:
Hey All I have Puppet 5.2 now in production use at $Work from this branch and it works very well. Do you want to merge this now or is there still missing stuff?

pyhalov:
@Toasterson I'll take a look in several days. Ping me if I forget :)

pyhalov:
What are we going to do with puppetserver?

Toasterson:
This Version of Puppet master works. If we also want to package puppetserver it would be a seperate component I would say.

pyhalov:
For me "puppet master --no-daemonize" exits with 
Notice: Starting Puppet master version 5.2.0
svcadm: Instance "svc:/application/puppet:master" is not online.
Error: Could not run: getaddrinfo: node name or service name not known

From tcpdump I see that it tries to resolve *.my.domain.name....

Toasterson:
Yes that is intentional. Both *.mydomain.com and  puppet.mydomain.com must exist and point to the puppetmaster. Alternatively you can add a directive to puppet.conf to force a hostname. I don't remember the directive off hand but I think it was hostname in the [master] or [main] section

Toasterson:
Bump. Is there any work still required for this?

Toasterson:
@pyhalov Any more comments on this?

pyhalov:
Yes. We should provide puppet master, which will start and run on svcadm enable puppet:master. We are still not here, server dies with 
```
 Could not run: getaddrinfo: node name or service name not known
```

pyhalov:
There's nothing about '*' resolution in https://puppet.com/docs/puppet/5.2/install_pre.html

Toasterson:
Well Point 4. It was always in the Best Practices. Might be that it changed to only require puppet.mydomain.com which makes sense. But I could check if I can add a section under master to force hostname to be 127.0.0.1

pyhalov:
The issue is not that it searches for 'puppet.my.domain', but that it searches for '*.my.domain', which is really strange.

Toasterson:
As far as i remember *.mydomain.com is a fallback if it does not find a listening Puppet master under puppet.mydomain.com have you tried adding puppet as an alias to 127.0.0.1 to your hosts file? I can check this evening once i am back ad my build computer.

PR #3510, opened 2017-09-20
fix: compile with net-snmp-5.7.3

grueni:
The PR depends on net-snmp 5.7.3 and should be merged after the PR for net-snmp 5.7.3.

### DIFF
--- a/components/sysutils/nut/Makefile
+++ b/components/sysutils/nut/Makefile
@@ -104,9 +104,9 @@ CONFIGURE_OPTIONS +=	--with-macosx_ups=no
 
 # Note: Link with the correct ISA of net-snmp (for drivers) and gdlib (for cgi)
 CONFIGURE_OPTIONS +=	--with-snmp
-CONFIGURE_OPTIONS +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-$(BITS)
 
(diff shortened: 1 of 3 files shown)


components/sysutils/nut/Makefile
```diff
@@ -104,9 +104,9 @@ CONFIGURE_OPTIONS +=	--with-macosx_ups=no
 
 # Note: Link with the correct ISA of net-snmp (for drivers) and gdlib (for cgi)
 CONFIGURE_OPTIONS +=	--with-snmp
-CONFIGURE_OPTIONS +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-$(BITS)
 
-CONFIGURE_OPTIONS.32 +=	--with-gdlib-config=$(USRBINDIR)/gdlib-config
+CONFIGURE_OPTIONS.32 +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-32
```
pyhalov: Why should it be changed?

grueni: I don't like it too but with this change it ssems to the usual way.
I don't mind to use the net-snmp-config in /usr/bin/ or /usr/bin/amd64 respective.
But we should do it the same way when net-snmp-config is used.

pyhalov:
This PR is not needed, as net-snmp was fixed to preserve net-snmp-config-$(BITS)

PR #3519, opened 2017-09-26
arpsend: compile as 64-bit

grueni:


components/cluster/arpsend/Makefile
```diff
@@ -35,11 +35,11 @@ CONFIGURE_OPTIONS += --with-libnet=/usr
 
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D))
```
AndWac, line 38: You need to set COMPONENT_REVISION here, too.

xen0l (changes requested):

xen0l (approved):

PR #3521, opened 2017-09-26
corosync: compile as 64-bit

grueni:


components/cluster/corosync/Makefile
```diff
@@ -15,6 +15,8 @@
 
 include ../../../make-rules/shared-macros.mk
 
+PATH= $(PATH.gnu)
+
 COMPONENT_NAME= corosync
 COMPONENT_VERSION= 2.4.2
```
AndWac, line 21: If you don't change COMPONENT_VERSION you should add (or increment if already there) COMPONENT_REVISION.

xen0l (changes requested):

xen0l (approved):

xen0l (approved):

PR #3525, opened 2017-09-26
libesmtp: compile as 64-bit

grueni:


components/cluster/libesmtp/libesmtp.p5m
```diff
@@ -13,6 +13,8 @@
 # Copyright 2016 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
```
AndWac: You probably want to update your copyright.
You also need to set COMPONENT_REVISION

xen0l (changes requested):

components/cluster/libesmtp/Makefile
```diff
@@ -40,11 +41,13 @@ COMPONENT_PREP_ACTION=(cd $(SOURCE_DIR); mkdir m4; $(AUTORECONF) -i)
 
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D))
 
-build:          $(BUILD_32)
```
pyhalov, line 43: Why do we drop 32-bit library?

grueni: We don't need it for the x64 builds of the other components.
I am not aware of any existing dependency of an 32bit binary.
Of course it could be activated again.
Should I do it?

xen0l: The general policy is that we deliver libraries for both 32 and 64 bit. However, if we know that no 32-bit application is going to use these libraries, I'd say it can stay 64-bit only. I remember that when I was packaging this, it wasn't being used by anything else apart from the cluster suite. If somebody or something will need for these libraries to be shipped 32-bit, I would open the discussion just then. I don't think it is reason to accept this contribution. @pyhalov what do you think?

pyhalov: We delivered libesmtp as 32-bit in several releases. I don't see why we can't deliver both. If we don't, it's a clear abi breakage and at least it should be documented at http://wiki.openindiana.org/oi/oi_hipster. Also think about possibility of updating libesmtp without updating dependent packages on clients' systems. It's clearly a flag day, but we can avoid it. So why not?

xen0l: @grueni @pyhalov is right here. We should ship both 32-bit and 64-bit if we did so in the past. This is also a general rule for libraries. Though, I still think that if we know that consumers are 64-bit only, why should we bother with 32-bit. However, let's ship it, so please update library PRs accordingly, so we can merge this big piece of great work!

grueni: PR updated to deliver 32- and 64-bits

xen0l (approved):

PR #3526, opened 2017-09-26
libqb: compile as 64-bit

grueni:


xen0l (changes requested):

xen0l (approved):
Looks good to me I just wonder the merge order. Libraries first, binaries second? Or is there different order these should be merged in?

grueni:
This should be the sequence during merging:
arpsend
libesmtp
libstatgrab
libqb
hacluster-common
resource-agents
cluster-glue
corosync
pacemaker
crmsh
pcs
service-hacluster
service-hacluster-remote
hacluster-remote

xen0l:
@grueni  I spoke with @pyhalov and we agreed that we should continue shipping both  32-bit even if we know that cluster is the sole consumer. Would you mind editing this PR? Sorry for troubles!

components/cluster/libqb/Makefile
```diff
@@ -45,11 +46,13 @@ COMPONENT_PREP_ACTION=(cd $(@D); $(ENV) $(CONFIGURE_ENV) $(CONFIG_SHELL) autogen
 
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D)) 
 
-build:          $(BUILD_32)
+#CONFIGURE_SBINDIR.64= $(CONFIGURE_SBINDIR.32)
```
pyhalov: Why do we add this comment? Does CONFIGURE_SBINDIR.64=/usr/sbin work?

grueni:
Too fast reviewed:)

components/cluster/libqb/Makefile
```diff
@@ -45,11 +44,11 @@ COMPONENT_PREP_ACTION=(cd $(@D); $(ENV) $(CONFIGURE_ENV) $(CONFIG_SHELL) autogen
...
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D)) 
 
-build:          $(BUILD_32)
+build:          $(BUILD_32_and_64)
 
-install:        $(INSTALL_32)
+install:        $(INSTALL_32_and_64)
 
-test:           $(TESTS_32)
+test:           $(TESTS_32_and_64)
```
pyhalov: $(TEST_32_and_64)

components/cluster/libqb/Makefile
```diff
@@ -39,16 +40,31 @@ CONFIGURE_OPTIONS += CFLAGS="$(CFLAGS)"
 CONFIGURE_OPTIONS += --enable-debug
 CONFIGURE_OPTIONS += --localstatedir=/var
 CONFIGURE_OPTIONS += --with-socket-dir=/var/run/hacluster
+CONFIGURE_OPTIONS += --enable-slow-tests
 
 COMPONENT_PREP_ACTION=(cd $(@D); $(ENV) $(CONFIGURE_ENV) $(CONFIG_SHELL) autogen.sh)
 
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D)) 
 
+# I can't figure out why this don't work if 'make test' is used:
```
pyhalov: how does it fail?

pyhalov: Likely, oi-userland environment can influence this. 
Quick fixes: "unexport SHELLOPTS" or "ENV = /usr/bin/env -i" can help with this.

grueni:
Squash commits now?

components/cluster/libqb/Makefile
```diff
@@ -40,16 +42,21 @@ CONFIGURE_OPTIONS += CFLAGS="$(CFLAGS)"
...
 CONFIGURE_OPTIONS += --with-socket-dir=/var/run/hacluster
+CONFIGURE_OPTIONS += --enable-slow-tests
 
 COMPONENT_PREP_ACTION=(cd $(@D); $(ENV) $(CONFIGURE_ENV) $(CONFIG_SHELL) autogen.sh)
 
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D)) 
 
-build:          $(BUILD_32)
+COMPONENT_PRE_TEST_ACTION += \
+    ($(MKDIR) /var/run/hacluster);
```
pyhalov: This will not work, as non-root can't do this. Can we run tests with another run directory?

grueni: The value of --with-socket-dir defines SOCKETDIR in config.h and this variable is used in the source code.
The folder is created by svc-hacluster when the service will be enabled.
There is another PR https://github.com/grueni/oi-userland/blob/hacluster-common/components/cluster/hacluster-common/hacluster-common.p5m for hacluster-common.
The folder can be created in hacluster-common and libqb will be defined as dependent from hacluster-common.

pyhalov (approved):

xen0l (approved):

PR #3527, opened 2017-09-26
libstatgrab: compile as 64-bit

grueni:


components/cluster/libstatgrab/Makefile
```diff
@@ -36,11 +36,11 @@ PATCH_LEVEL= 1
 COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D))
```
AndWac: I'll stop here with my COMPONENT_REVISION comments.
The general rule is: If you change anything either set the new COMPONENT_VERSION (and delete COMPONENT_REVISION) or set/increment COMPONENT_REVISION (starting at 1).

components/cluster/libstatgrab/libstatgrab.p5m
```diff
@@ -24,15 +25,21 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-file path=usr/bin/saidar
-file path=usr/bin/statgrab
-file path=usr/bin/statgrab-make-mrtg-config
-file path=usr/bin/statgrab-make-mrtg-index
+file path=usr/bin/$(MACH64)/saidar
```
xen0l: Please, set CONFIGURE_BINDIR.64 like [here](https://github.com/OpenIndiana/oi-userland/blob/b6ca80a20fb2f5e4e777be54f0bdf34372fde345/components/shell/dash/Makefile#L41)

xen0l (changes requested):

components/cluster/libstatgrab/libstatgrab.p5m
grueni: This will not work:
echo "CONFIGURE_BINDIR.64=/usr/bin/amd64"
We need a varaiable with the value "usr/bin/amd64".

grueni: Ok, I got it.
I thought all 64bit binaries should go in usr/(s)bin/$MACH64).
But this seems to be wrong and if only 64bit binaries are created we use the the paths without $(MACH64).

grueni:
PR updated to deliver 32- and 64-bits

xen0l (approved):

components/cluster/libstatgrab/Makefile
```diff
@@ -31,16 +32,11 @@ include $(WS_MAKE_RULES)/prep.mk
...
-COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D))
+install:        $(INSTALL_32_and_64)
 
-
-build:          $(BUILD_32)
-
-install:        $(INSTALL_32)
-
-test:           $(TESTS_32)
+test:           $(TESTS_32_and_64)
```
pyhalov: This is wrong target, correct one is TEST_32_and_64 . BTW, do tests pass?

xen0l (approved):

PR #3528, opened 2017-09-26
pacemaker: compile as 64-bit

grueni:


components/cluster/pacemaker/Makefile
```diff
@@ -14,6 +14,8 @@
 
 include ../../../make-rules/shared-macros.mk
 
+PATH= $(PATH.gnu):$(USRBINDIR64)
+
 COMPONENT_NAME= pacemaker
 COMPONENT_VERSION= 1.1.17
```
xen0l: Bump COMPONENT_REVISION

components/cluster/pacemaker/Makefile
```diff
@@ -64,16 +65,18 @@ CONFIGURE_OPTIONS += --with-acl=no
...
 
 COMPONENT_POST_INSTALL_ACTION=( \
-    mkdir -p $(PROTO_DIR)/usr/share/pacemaker/examples; \
-    mv $(PROTO_DIR)/etc/logrotate.d $(PROTO_DIR)/usr/share/pacemaker/examples )
+    $(RM) -r $(PROTO_DIR)/usr/share/pacemaker/examples; \
+    $(MKDIR) $(PROTO_DIR)/usr/share/pacemaker/examples; \
+    $(MV)    $(PROTO_DIR)/etc/logrotate.d $(PROTO_DIR)/usr/share/pacemaker/examples )
 
-build:          $(BUILD_32)
+build:          $(BUILD_64)
```
xen0l: Please, set CONFIGURE_BINDIR.64 like [here](https://github.com/OpenIndiana/oi-userland/blob/b6ca80a20fb2f5e4e777be54f0bdf34372fde345/components/shell/dash/Makefile#L41)

components/cluster/pacemaker/Makefile
```diff
@@ -51,8 +51,9 @@ LD_OPTIONS.32 += -L/usr/lib/gnutls-3 -R/usr/lib/gnutls-3
 LD_OPTIONS.64 += -L/usr/lib/$(MACH64)/gnutls-3 -R/usr/lib/$(MACH64)/gnutls-3
 LD_OPTIONS += $(LD_OPTIONS.$(BITS))
 
+CONFIGURE_OPTIONS.32 += --libexecdir=$(CONFIGURE_PREFIX)/lib
```
xen0l: If we are building 64-bit only, we don't need 32-bit options, do we?

xen0l (changes requested):

components/cluster/pacemaker/manifests/sample-manifest.p5m
```diff
@@ -134,6 +134,29 @@ file path=usr/lib/python2.7/site-packages/cts/logging.py
 file path=usr/lib/python2.7/site-packages/cts/patterns.py
 file path=usr/lib/python2.7/site-packages/cts/remote.py
 file path=usr/lib/python2.7/site-packages/cts/watcher.py
+file path=usr/sbin/$(MACH64)/attrd_updater
```
xen0l: Is the manifest-sample.p5m correct? Binaries shouldn't move out of /usr/bin.

components/cluster/pacemaker/pacemaker.p5m
```diff
@@ -96,23 +97,36 @@ file path=usr/lib/ocf/resource.d/pacemaker/o2cb
...
+file path=usr/lib/$(MACH64)/pacemaker/cibmon
+file path=usr/lib/$(MACH64)/pacemaker/crmd
+file path=usr/lib/$(MACH64)/pacemaker/lrmd
+file path=usr/lib/$(MACH64)/pacemaker/lrmd_internal_ctl
+file path=usr/lib/$(MACH64)/pacemaker/lrmd_test
+file path=usr/lib/$(MACH64)/pacemaker/pengine
+file path=usr/lib/$(MACH64)/pacemaker/stonith-test
+file path=usr/lib/$(MACH64)/pacemaker/stonithd
+
+link path=usr/lib/pacemaker/attrd              target=/usr/lib/$(MACH64)/pacemaker/attrd
```
xen0l: Please, don't do this. Let's rather not ship 32-bit even though it will be ABI breakage. @pyhalov what do you think about such breakage in this case?

xen0l (changes requested):

components/cluster/pacemaker/pacemaker.p5m
pyhalov: Don't ship /usr/lib/pacemaker/* symlinks (especially these ones, which are incorrect - symlinks should be relative.

I'd shipped binaries in /usr/sbin or /usr/bin and libraries in /usr/lib/$(MACH64). But note, that pacemaker uses binaries from $(libdir)/pacemaker and as I understand searches for resources definitions in $(libdir)/ocf/resource.d . You should be careful with that, so that pacemaker could find resource descriptions of resource-agents package.
I'd prefer pacemaker-specific binaries to live where they are now (/usr/lib/pacemaker), it's consistent, for example, with sendmail. As for dlls, they should live in /usr/lib/$(MACH64) .

Was I clear? :)

components/cluster/pacemaker/manifests/sample-manifest.p5m
grueni: corosync and pacemaker use mainly /usr/sbin.
And tools loke crm and pcs search the binaries in this paths.

components/cluster/pacemaker/pacemaker.p5m
grueni: I hope I got it now. Just making clear the rules.

components/cluster/pacemaker/Makefile
```diff
@@ -43,16 +42,14 @@ CONFIGURE_SCRIPT= $(@D)/configure
...
 # Force use of gnutls-3 pkgconfig during 2.x->3.x transition
-GNUTLS_PKG_CONFIG_PATH_32 = /usr/lib/pkgconfig/gnutls-3
 GNUTLS_PKG_CONFIG_PATH_64 = /usr/lib/$(MACH64)/pkgconfig/gnutls-3
 PKG_CONFIG_PATH = $(GNUTLS_PKG_CONFIG_PATH_$(BITS)):$(PKG_CONFIG_PATH.$(BITS))
 
-LD_OPTIONS.32 += -L/usr/lib/gnutls-3 -R/usr/lib/gnutls-3
 LD_OPTIONS.64 += -L/usr/lib/$(MACH64)/gnutls-3 -R/usr/lib/$(MACH64)/gnutls-3
 LD_OPTIONS += $(LD_OPTIONS.$(BITS))
 
+CONFIGURE_OPTIONS.64 += --libexecdir=$(CONFIGURE_PREFIX)/lib/$(MACH64)
```
pyhalov: @What happens if we set it back to $(CONFIGURE_PREFIX)/lib ? What files does it affect?

grueni: I guess this would be replaced (untested!):
file usr/lib/$(MACH64)/pacemaker/attrd             path=usr/lib/pacemaker/attrd
file usr/lib/$(MACH64)/pacemaker/cib               path=usr/lib/pacemaker/cib
file usr/lib/$(MACH64)/pacemaker/cibmon            path=usr/lib/pacemaker/cibmon
file usr/lib/$(MACH64)/pacemaker/crmd              path=usr/lib/pacemaker/crmd
file usr/lib/$(MACH64)/pacemaker/lrmd              path=usr/lib/pacemaker/lrmd
file usr/lib/$(MACH64)/pacemaker/lrmd_test         path=usr/lib/pacemaker/lrmd_test
file usr/lib/$(MACH64)/pacemaker/pengine           path=usr/lib/pacemaker/pengine
file usr/lib/$(MACH64)/pacemaker/stonith-test      path=usr/lib/pacemaker/stonith-test
file usr/lib/$(MACH64)/pacemaker/stonithd          path=usr/lib/pacemaker/stonithd
file usr/lib/$(MACH64)/pacemaker/lrmd_internal_ctl path=usr/lib/pacemaker/lrmd_internal_ctl

by
file usr/lib/pacemaker/attrd
file usr/lib/pacemaker/cib
file usr/lib/pacemaker/cibmon
file usr/lib/pacemaker/crmd
file usr/lib/pacemaker/lrmd
file usr/lib/pacemaker/lrmd_test
file usr/lib/pacemaker/pengine
file usr/lib/pacemaker/stonith-test
file usr/lib/pacemaker/stonithd
file usr/lib/pacemaker/lrmd_internal_ctl

grueni:
This is the patch to avoid the usage of /usr/lib/amd64/pacemaker/...
~~This should be done for corosync too.~~

components/cluster/pacemaker/manifests/sample-manifest.p5m
pyhalov: He speaks about $(MACH64). Regenerate sample-manifest, please.

xen0l (approved):

PR #3531, opened 2017-09-26
service-hacluster: dependencies

grueni:


components/cluster/service-hacluster/files/svc-hacluster
```diff
@@ -16,7 +16,7 @@ export CLUSTER_USER=hacluster
 export CLUSTER_GROUP=haclient
 export PATH=/usr/gnu/bin:${PATH}
 MYNAME=hacluster
-APPPATH=/usr/sbin/
+APPPATH=/usr/sbin/amd64/
```
xen0l: If we ship 64-bit binaries in 32-bit paths like we do for 64-bit only components, this change is not needed. Shipping 64-bit binaries in 32-bit paths is the preferred solution.

components/cluster/service-hacluster/service-hacluster.p5m
```diff
@@ -36,9 +36,7 @@ set name=pkg.description value="$(COMPONENT_SUMMARY)"
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-group groupname=haclient gid=66
```
pyhalov, line 39: Why do we avoid shipping user/group hacluster?

grueni: This goes to hacluster-common as in other packages.
https://github.com/grueni/oi-userland/blob/cf6f24bb43bc3e22ad9a9325ccd1e01936a40a93/components/cluster/hacluster-common/hacluster-common.p5m

xen0l (approved):

PR #3541, opened 2017-10-09
openvpn: set_ipsec_bypass() doesn't work for IPV6 connections

pyhalov:


pyhalov:
@danmcd , @gwr : please, review

xen0l (approved):

components/network/openvpn/patches/02-ipsec-bypass.patch
```diff
@@ -12,22 +12,32 @@
...
-+  if (setsockopt(sock, IPPROTO_IP, IP_SEC_OPT, &ipsr, sizeof (ipsr)) == -1) {
++  if (ai_family == AF_INET) {
++    result = setsockopt(sock, IPPROTO_IP, IP_SEC_OPT, &ipsr, sizeof (ipsr));
++  } else { /* AF_INET6 */
++    result = setsockopt(sock, IPPROTO_IPV6, IPV6_SEC_OPT, &ipsr, sizeof (ipsr));
++  }
++  if (result == -1) {
 +    switch (errno) {
 +    case EPROTONOSUPPORT:
 +      dmsg (D_SOCKET_DEBUG, "Not using IPsec.");
```
danmcd: "IPsec not loaded" is a more accurate description.

pyhalov: Have one more question... If I receive EPROTONOSUPPORT error in zone, but it works in GZ, does this mean that zone misses some privileges?

pyhalov: Also if it fails in zone with EPROTONOSUPPORT, in GZ when run as usual user, it will fail with EPERM (so tests will fail). The question is if we should also handle this fail gracefully (I mean, ignore EPERM as we do with EPROTONOSUPPORT)?

danmcd: A regular user needs the privilege IP_CONFIG to trigger policy bypasses.  EPERM is not ignorable like EPROTONOSUPPORT is.

pyhalov: yes, after ipseckey flush it fails differently - EPERM.

pyhalov:
Now the only question is if we are going to ignore EPERM error, like we do for EPROTONOSUPPORT.

components/network/openvpn/patches/02-ipsec-bypass.patch
danmcd: So clearly we do the is-loaded check BEFORE the privilege check.  A regular user will need the IP_CONFIG privilege to bypass.  (I thought openvpn needed this privilege anyway?!)

pyhalov: well, at least openvpn tests run without root privileges (without this patch). It can do simple client-server pings...

pyhalov:
With this patch applied openvpn needs IP_CONFIG privilege, so test can fail with EPERM

PR #3575, opened 2017-10-18
add spidermonkey 1.8.5

grueni:


components/runtime/spidermonkey/Makefile
```diff
@@ -0,0 +1,73 @@
...
+# Copyright 2017 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved. 
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=	          spidermonkey
+COMPONENT_VERSION=        1.8.5
+COMPONENT_SUMMARY=        Javascript engine
+COMPONENT_PROJECT_URL=    https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey
+COMPONENT_FMRI=	          runtime/spidermonkey
```
pyhalov: library/spidermonkey (as it's not an interpreter), please move the component correspondingly

components/runtime/spidermonkey/Makefile
```diff
@@ -0,0 +1,73 @@
...
+COMPONENT_CLASSIFICATION= Development/Other Languages
```
pyhalov: System/Library

pyhalov:
Do we need Studio and SPARC-specific patches?

grueni:
We don't need it. But we are gently guys and maybe other people (sparc!) are interested.
Where did you find the Studio-specific patches?
I thought I commented them out.

components/runtime/spidermonkey/patches/02-jsfunc.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/js/src/jsfun.h b/js/src/jsfun.h
+index a52092c..9555a03 100644
+--- a/jsfun.h
++++ b/jsfun.h
+@@ -144,6 +144,7 @@ struct JSFunction : public JSObject_Slots2
+         void            *nativeOrScript;
+     } u;
+     JSAtom          *atom;        /* name for diagnostics and decompiling */
++    void            *dummy;       /* for building with Solaris Studio */
```
pyhalov: seems to be relevant for studio only (based on comment). Is it true?

components/runtime/spidermonkey/patches/05-pgo-ss12_2.patch
```diff
@@ -0,0 +1,18 @@
+diff --git a/js/src/configure b/js/src/configure
+index 0e98daf..11bb34f 100755
+--- a/configure
++++ b/configure
+@@ -14560,10 +14560,10 @@ if test -n "$INTEL_CC"; then
+ fi
+ 
+ if test "$SOLARIS_SUNPRO_CC"; then
+-  PROFILE_GEN_CFLAGS="-xprofile=collect:$_objdir/$enable_application"
```
pyhalov: Relevant only for Studio

components/library/spidermonkey/Makefile
```diff
@@ -18,7 +18,7 @@ COMPONENT_NAME=	          spidermonkey
 COMPONENT_VERSION=        1.8.5
 COMPONENT_SUMMARY=        Javascript engine
 COMPONENT_PROJECT_URL=    https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey
-COMPONENT_FMRI=	          runtime/spidermonkey
+COMPONENT_FMRI=	          library/javascript/spidermonkey
```
pyhalov: as I understand, Oracle took library/javascript/* for JS libraries. Let's use just library/spidermonkey, as it's C library for JS processing :)

components/runtime/spidermonkey/patches/05-pgo-ss12_2.patch
grueni: It is protected so we can use the patch or remove it. It doesn't matter.

components/runtime/spidermonkey/patches/02-jsfunc.patch
grueni: I tested with the patch and all tests passed successfully.
So I would use it.

grueni:
Ok, it is now in component/library.

alarcher:
@pyhalov @grueni Unfortunately it does not build with gcc-6, probably because 1.8.5 is an outdated version. Are we able to use at least version 45 from https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Releases ?

grueni:
Unfortunately no.  In the issues section of the CouchDB project the team explains why. They use anonymous functions which are not available with younger versions of couchdb.

PR #3576, opened 2017-10-18
add couchdb-common

grueni:


components/database/couchdb-common/couchdb-common.p5m
```diff
@@ -0,0 +1,27 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2017 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
+#
+
+set name=pkg.fmri value=pkg:/database/$(COMPONENT_NAME)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
```
pyhalov: Use COMPONENT_FMRI

pyhalov:
BTW, do we need separate couchdb-common component? For example, PG and MySQL components just delivers mysql-common package in the component with lowest version.

grueni:
I just did not know how to do it or better I don't know the rules when to use a *-common package.
In my understanding the common objects should be in such a package. This are at least the user and group definitions.

PR #3577, opened 2017-10-18
add CouchDB 2.1.1

grueni:


components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1334 @@
...
+dir  path=etc/couchdb/local.d/ owner=couchdb group=couchdb mode=0755
+
+file path=etc/couchdb/default.d/README
+file path=etc/couchdb/default.d/default.ini
+file path=etc/couchdb/default.ini
+file path=etc/couchdb/local.d/README
+file path=etc/couchdb/local.ini            owner=couchdb group=couchdb mode=0664
+file path=etc/couchdb/vm.args
+
+link path=usr/lib/$(MACH64)/couchdb/etc target=/etc/couchdb
```
pyhalov: targets are relative, so links are incorrect

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,86 @@
...
+COMPONENT_ARCHIVE=        $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=    http://apache.mirror.iphh.net/$(COMPONENT_NAME)/source/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_LICENSE=        Apache 2.0
+COMPONENT_SUMMARY=        Apache CouchDB is a document-oriented database.
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CFLAGS += -std=c99 -D__EXTENSIONS__ -D_POSIX_PTHREAD_SEMANTICS
```
xen0l: There might be macros in https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/shared-macros.mk for some of these. Not a blocker, but consider using them.

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,86 @@
...
+
+COMPONENT_PRE_CONFIGURE_ACTION=($(CLONEY) $(SOURCE_DIR) $(@D))
+
+COMPONENT_PREP_ACTION = ( \
+	$(CP) $(COMPONENT_DIR)/files/Makefile.openindiana $(@D); \
+	$(CP) $(COMPONENT_DIR)/files/default.ini $(@D); )
+
+CONFIGURE_SCRIPT= $(@D)/configure
+
+CONFIGURE_OPTIONS = --user $(COMPONENT_NAME) --with-curl
```
xen0l: Split this into two separate lines, e.g.
```
CONFIGURE_OPTIONS = --user $(COMPONENT_NAME)
CONFIGURE_OPTIONS+= --with-curl
```

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,86 @@
...
+#
+# This is a mixed 64bit and 32bit installation.
+# All binary applications are built as 64bit but not couchjs.
+# With patch 01-rebarconfigscript.patch couchjs is forced to be built as 32bit and
+# link against /usr/lib/libmozjs185.so (Spidermonkey).
+# Spidermonkey is 32bit only.
+# CouchDB uses couchjs as external process over STDOUT/STDIN/STDERR.
+#
+
+# Te project doesn't have an install section but let user copy the folder rel/couchdb.
```
xen0l: The?

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,86 @@
...
+COMPONENT_INSTALL_ARGS += COMPONENT_DIR=$(USRDIR)/$(COMPONENT_NAME)/$(COMPONENT_VERSION) V=1
+
+COMPONENT_INSTALL_TARGETS = release
+
+#
+# This is a mixed 64bit and 32bit installation.
+# All binary applications are built as 64bit but not couchjs.
+# With patch 01-rebarconfigscript.patch couchjs is forced to be built as 32bit and
+# link against /usr/lib/libmozjs185.so (Spidermonkey).
+# Spidermonkey is 32bit only.
```
xen0l, line 62: If I understand correctly, if we make Spidermonkey 64-bit, we can drop 32-bit part of this component? What's the reason for not doing so?

components/database/couchdb-21/couchdb-service.p5m
```diff
@@ -0,0 +1,36 @@
...
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend type=require fmri=database/couchdb-21
+depend type=require fmri=database/couchdb-common
+
+file files/couchdb.xml path=lib/svc/manifest/application/couchdb.xml restart_fmri=svc:/system/manifest-import:default
```
xen0l: What about RBAC integration? Again, not a blocker, but would be nice to have.

components/database/couchdb-21/couchdb-service.p5m
```diff
@@ -0,0 +1,36 @@
...
+
+dir  path=var/couchdb/data  owner=couchdb group=couchdb mode=0755
+dir  path=var/couchdb/index owner=couchdb group=couchdb mode=0755
+dir  path=var/couchdb/home  owner=couchdb group=couchdb mode=0755
+dir  path=var/log/couchdb   owner=couchdb group=couchdb mode=0755
+
```
xen0l: Delete trailing empty lines

components/database/couchdb-21/couchdb-service.p5m
```diff
@@ -0,0 +1,36 @@
...
+depend type=require fmri=database/couchdb-21
+depend type=require fmri=database/couchdb-common
+
+file files/couchdb.xml path=lib/svc/manifest/application/couchdb.xml restart_fmri=svc:/system/manifest-import:default
+
+dir  path=var/couchdb/data  owner=couchdb group=couchdb mode=0755
+dir  path=var/couchdb/index owner=couchdb group=couchdb mode=0755
+dir  path=var/couchdb/home  owner=couchdb group=couchdb mode=0755
+dir  path=var/log/couchdb   owner=couchdb group=couchdb mode=0755
```
xen0l: Have you considered shipping /etc/logadm.d entry for log rotation?

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1335 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend type=require fmri=runtime/erlang@17.5-2017.0.0.0
+depend type=require fmri=runtime/python-27
+depend type=require fmri=library/python/netaddr-27
+depend type=require fmri=database/couchdb-common
+
+dir  path=etc/couchdb/local.d/ owner=couchdb group=couchdb mode=0755
```
xen0l: Remove traling slash

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1335 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend type=require fmri=runtime/erlang@17.5-2017.0.0.0
```
xen0l: Why specify version here? If we rebuild erlang, then the dependency will not apply.

components/database/couchdb-21/files/Makefile.openindiana
```diff
@@ -0,0 +1,21 @@
+   
+install:
+	if [ -d  $(DESTDIR)$(ETCDIR) ]; then    rm -rf $(DESTDIR)$(ETCDIR); fi
+	if [ -d  $(DESTDIR)$(MANDIR) ]; then    rm -rf $(DESTDIR)$(MANDIR); fi
+	if [ -d  $(DESTDIR)$(DOCDIR) ]; then    rm -rf $(DESTDIR)$(DOCDIR); fi
+	rm -rf   $(DESTDIR)$(RELDIR)/var	
+	mkdir -p $(DESTDIR)$(BINDIR)
+	mkdir -p $(DESTDIR)$(MANDIR)/man1/
```
xen0l: Consider using macros from https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/shared-macros.mk#L155, they might be helpful.

xen0l (changes requested):

components/database/couchdb-21/Makefile
grueni: As already mentioned in package Spidermonkey: Spidermonkey 1.8.5 is strictly 32bit and CouchDB officially only supports Spidermonkey 1.8.5.

components/database/couchdb-21/couchdb.p5m
grueni: I added this comment
# CouchDB 2.1.0 must have an erlang runtime < 19
With Erlang 19.3 you get issues.

pyhalov: Explicit version here will  work the following way: we need erlang >= 17.5-2017.0.0.0

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1336 @@
...
+
+# CouchDB 2.1.0 must have an erlang runtime < 19
+depend type=require fmri=runtime/erlang@17.5-2017.0.0.0
+depend type=require fmri=runtime/python-27
+depend type=require fmri=library/python/netaddr-27
+depend type=require fmri=database/couchdb-common
+
+dir  path=etc/couchdb/local.d owner=couchdb group=couchdb mode=0755
+
+file path=etc/couchdb/default.d/README
```
pyhalov: Sorry for raising this late, but can couchdb follow the common layout of other DBMS on OI (MongoDB, PostgreSQL, MariaDB, etc)?
I mean: 
- we deliver binary files to /usr/couchdb/${MAJOR_VERSION},  so that we have /usr/couchdb/2.1/{bin,lib,etc...}
- we deliver data files directories in /var/couchdb/${MAJOR_VERSION}, 
- we create mediated links from /usr/bin, /usr/share/man to delivered files.

pyhalov (changes requested):

components/database/couchdb-21/couchdb.p5m
grueni: No problem.
In fact I have couchdb-16 also so we can test two versions.
I have no experience with mediators but we will see.
But I will need some time.

grueni:
~~Still untested~~.
~~And I will add configuration file for logrotation.~~
changed names, added logrotate
tested with all changes -> OK

~~Needs work, will be updated to 2.1.1~~

grueni:
tested with all changes -> OK

components/database/couchdb-21/couchdb-service.p5m
```diff
@@ -0,0 +1,38 @@
...
+depend type=require fmri=database/couchdb-common
+
+file files/auth_attr path=etc/security/auth_attr.d/couchdb-21
+file files/exec_attr path=etc/security/exec_attr.d/couchdb-21
+file files/prof_attr path=etc/security/prof_attr.d/couchdb-21
+file files/user_attr path=etc/user_attr.d/couchdb-21
+
+file files/couchdb-21.xml path=lib/svc/manifest/application/couchdb-21.xml restart_fmri=svc:/system/manifest-import:default
+
+dir  path=var/couchdb-21/data  owner=couchdb group=couchdb mode=0755
```
pyhalov: should have been /var/couchdb/2.1/data and so on

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1403 @@
...
+
+link path=usr/lib/$(MACH64)/couchdb-21/etc target=../../../../etc/couchdb-21
+
+dir  path=etc/couchdb-21/local.d owner=couchdb group=couchdb mode=0755
+
+# users should create a link to their preferred logging system
+file files/couchdb-21.logrotate   path=etc/couchdb-21/couchdb-21.logrotate   owner=couchdb group=couchdb mode=0664 preserve=yes
+file files/couchdb-21.logadm.conf path=etc/couchdb-21/couchdb-21.logadm.conf owner=couchdb group=couchdb mode=0664 preserve=yes
+
+file path=etc/couchdb-21/vm.args
```
pyhalov: Please, follow layout of other DBMSes. This should have been /etc/couchdb/2.1/...

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1403 @@
...
+     mediator=couchdb mediator-version=2.1
+link path=usr/bin/couchjs target=../lib/$(MACH64)/couchdb-21/bin/couchjs \
+     mediator=couchdb mediator-version=2.1
+link path=usr/bin/couchup target=../lib/$(MACH64)/couchdb-21/bin/couchup \
+     mediator=couchdb mediator-version=2.1
+
+file path=usr/share/man/man1/couchdb.1 \
+     mediator=couchdb mediator-version=2.1
+
+file path=usr/lib/$(MACH64)/couchdb-21/LICENSE
```
pyhalov: Better place it to /usr/couchdb/2.1

components/database/couchdb-21/files/couchdb-21.xml
```diff
@@ -0,0 +1,51 @@
...
+
+    <dependency name="network" grouping="require_all" restart_on="error" type="service">
+      <service_fmri value="svc:/milestone/network:default"/>
+    </dependency>
+
+    <dependency name="filesystem-local" grouping="require_all" restart_on="none" type="service">
+      <service_fmri value="svc:/system/filesystem/local:default"/>
+    </dependency>
+
+		<exec_method type="method" name="start" exec="/usr/lib/amd64/couchdb-21/bin/couchdb &amp;" timeout_seconds="60">
```
pyhalov: Try to avoid using amd64|sparcv9 in manifests explicitly.

components/database/couchdb-21/patches/02-couchup.patch
```diff
@@ -0,0 +1,8 @@
+--- apache-couchdb-2.1.0/rel/overlay/bin/couchup.orig	2017-10-07 13:09:47.591823720 +0000
++++ apache-couchdb-2.1.0/rel/overlay/bin/couchup	2017-10-07 13:11:25.777635346 +0000
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/amd64/python2.7
```
pyhalov: Is it strictly necessary use 64-bit python here?

pyhalov (changes requested):

components/database/couchdb-21/patches/02-couchup.patch
grueni: No. The script communicates exclusively by the REST interface. No libraries from couchdb are used. I removed the patch.

components/database/couchdb-21/files/couchdb-21.xml
grueni: replaced by adding links to usr/bin/

pyhalov:
Please, move database files to /usr/couchdb/2.1 and data files to /var/couchdb/2.1 to match PostgreSQLs/MySQLs/MongoDB

pyhalov:
Do I understand correctly that CouchDB 2.2 will not be compatible to 2.1?

grueni:
Did I tell this? We have now 2.1.1. 
I am actually adding the tests.

grueni:
Tests are OK.
There is no version 2.2.
Versions 2.* are incompatible with version < 2.* but migration is supported with couchup.
http://docs.couchdb.org/en/latest/install/upgrading.html

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,96 @@
...
+  ($(RM) -r $(PROTO_DIR)/*; \
+   $(MKDIR) $(PROTO_DIR)$(USRLIBDIR64)/couchdb/21; \
+   $(CP) $(BUILD_DIR_64)/LICENSE $(COMPONENT_NAME).license; \
+   $(CP) -r $(BUILD_DIR_64)/rel/couchdb/* $(PROTO_DIR)$(USRLIBDIR64)/couchdb/21/ ); \
+   cd $(BUILD_DIR_64); \
+   $(GMAKE) -f Makefile.openindiana DESTDIR=$(PROTO_DIR) MANDIR=$(USRSHAREMANDIR) BINDIR=$(USRBINDIR) \
+      DOCDIR=$(USRSHAREDOCDIR)/couchdb/21 ETCDIR=$(ETCDIR)/couchdb/21 RELDIR=$(USRLIBDIR64)/couchdb/21
+
+COMPONENT_PRE_TEST_ACTION += \
+   pip install hypothesis
```
xen0l: Never ever do this in a component. If we need a package, it shall be shipped in oi-userland.

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,96 @@
...
+install:        $(INSTALL_64)
+
+test:           $(TEST_64)
+
+REQUIRED_PACKAGES += runtime/erlang
+REQUIRED_PACKAGES += developer/icu
+REQUIRED_PACKAGES += system/library/g++-4-runtime
+REQUIRED_PACKAGES += library/spidermonkey
+# only required for target test
+REQUIRED_PACKAGES += library/python/nose-27
```
xen0l: These will be installed every time, you install CouchDB. Where are they coming from?

components/database/couchdb-21/couchdb-common.p5m
```diff
@@ -0,0 +1,27 @@
...
+
+#
+# Copyright 2017 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
+#
+
+set name=pkg.fmri value=pkg:/database/$(COMPONENT_NAME)-common@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+#set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
```
xen0l: If we don't need this, just drop it

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1407 @@
...
+depend type=require fmri=runtime/erlang
+depend type=require fmri=runtime/python-27
+depend type=require fmri=library/python/netaddr-27
+depend type=require fmri=database/couchdb-common
+
+link path=usr/lib/$(MACH64)/couchdb/21/etc target=../../../../../etc/couchdb/21
+
+dir  path=etc/couchdb/21/local.d owner=couchdb group=couchdb mode=0755
+
+# users should create a link to their preferred logging system
```
xen0l: Logging -> log rotation

xen0l (changes requested):

components/database/couchdb-21/Makefile
grueni: I will comment the two packages.
It is a problem with automatic testing when packages are only required for testing and not for runtime or building.

components/database/couchdb-21/Makefile
grueni: Could this be a comment for the tester?

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,96 @@
...
+# All binary applications are built as 64bit but not couchjs.
+# With patch 01-rebarconfigscript.patch couchjs is forced to be built as 32bit and
+# link against /usr/lib/libmozjs185.so (Spidermonkey).
+# Spidermonkey is 32bit only.
+# CouchDB uses couchjs as external process over STDOUT/STDIN/STDERR.
+#
+
+# The project doesn't have an install section but let user copy the folder rel/couchdb.
+# Here we prepare manually the image in the prototype folder and do the rest with a makefile.
+$(INSTALL_64): COMPONENT_POST_INSTALL_ACTION= \
```
pyhalov: Why do we define COMPONENT_POST_INSTALL_ACTION only for INSTALL_64 target?
I suppose, it's not necessary.

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,96 @@
...
+  ($(RM) -r $(PROTO_DIR)/*; \
+   $(MKDIR) $(PROTO_DIR)$(USRLIBDIR64)/couchdb/21; \
+   $(CP) $(BUILD_DIR_64)/LICENSE $(COMPONENT_NAME).license; \
```
pyhalov: Why do we replace $(COMPONENT_NAME).license ???

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,96 @@
...
+   $(CP) -r $(BUILD_DIR_64)/rel/couchdb/* $(PROTO_DIR)$(USRLIBDIR64)/couchdb/21/ ); \
+   cd $(BUILD_DIR_64); \
```
pyhalov: Can we move all installation logic (without rm -r $(PROTO_DIR)) to Makefile.openindiana, so that we could just set something like "COMPONENT_INSTALL_ARGS = -f Makefile.openindiana install"  in components Makefile?

components/database/couchdb-21/Makefile
```diff
@@ -0,0 +1,96 @@
...
+# link against /usr/lib/libmozjs185.so (Spidermonkey).
+# Spidermonkey is 32bit only.
+# CouchDB uses couchjs as external process over STDOUT/STDIN/STDERR.
+#
+
+# The project doesn't have an install section but let user copy the folder rel/couchdb.
+# Here we prepare manually the image in the prototype folder and do the rest with a makefile.
+$(INSTALL_64): COMPONENT_POST_INSTALL_ACTION= \
+  ($(RM) -r $(PROTO_DIR)/*; \
+   $(MKDIR) $(PROTO_DIR)$(USRLIBDIR64)/couchdb/21; \
```
pyhalov: Please, use /usr/couchdb/2.1 instead of /usr/lib/$(MACH64)/couchdb/21 for database program files and /var/couchdb/2.1 for data files to match mysqls/postgresqls/mongodb.

components/database/couchdb-21/files/Makefile.openindiana
```diff
@@ -1,14 +1,14 @@
-   
+ 
 install:
-	if [ -d  $(DESTDIR)$(ETCDIR) ]; then  $(RM) -r $(DESTDIR)$(ETCDIR); fi
-	if [ -d  $(DESTDIR)$(MANDIR) ]; then  $(RM) -r $(DESTDIR)$(MANDIR); fi
-	if [ -d  $(DESTDIR)$(DOCDIR) ]; then  $(RM) -r $(DESTDIR)$(DOCDIR); fi
-	/bin/rm -rf $(DESTDIR)$(RELDIR)/var
+	rm -rf $(DESTDIR)
```
pyhalov: Why are you doing this? If somehow it's called with DESTDIR=/, it would be pretty interesting

components/database/couchdb-21/patches/03-bin-couchdb.patch
```diff
@@ -5,6 +5,6 @@
  # the License.
  
 -COUCHDB_BIN_DIR=$(cd "${0%/*}" && pwd)
-+COUCHDB_BIN_DIR=/usr/lib/amd64/couchdb/21/bin
++COUCHDB_BIN_DIR=/usr/lib/amd64/couchdb/2.1/bin
```
pyhalov: Please, make COUCHDB base directory /usr/couchdb/2.1

components/database/couchdb-21/files/Makefile.openindiana
grueni: During testing I got rm -rf /*. And I installed a new build environment afterwards.

components/database/couchdb-21/files/couchdb-21.xml
```diff
@@ -0,0 +1,51 @@
...
+
+    <dependency name="network" grouping="require_all" restart_on="error" type="service">
+      <service_fmri value="svc:/milestone/network:default"/>
+    </dependency>
+
+    <dependency name="filesystem-local" grouping="require_all" restart_on="none" type="service">
+      <service_fmri value="svc:/system/filesystem/local:default"/>
+    </dependency>
+
+		<exec_method type="method" name="start" exec="/usr/bin/couchdb2.1 &amp;" timeout_seconds="60">
```
pyhalov: 1) Perhaps, it's better to usr /usr/couchdb/2.1/bin/couchdb instead of link?
2) @xen0l can we do something better here than running 'couchdb &' ?

pyhalov: @grueni will exec='/usr/couchdb/2.1/bin/couchdb -b' work?

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1408 @@
...
+
+# mediators
+link path=usr/bin/couchdb target=../couchdb/$(COMPONENT_VERSION_MAJOR)/bin/couchdb \
+     mediator=couchdb mediator-version=2.1
+link path=usr/bin/couchjs target=../couchdb/$(COMPONENT_VERSION_MAJOR)/bin/couchjs \
+     mediator=couchdb mediator-version=2.1
+link path=usr/bin/couchup target=../couchdb/$(COMPONENT_VERSION_MAJOR)/bin/couchup \
+     mediator=couchdb mediator-version=2.1
+
+file path=usr/share/man/man1/couchdb.1 \
```
pyhalov: What is this? File should be delivered to /usr/couchdb/2.1/man/man1 or /usr/couchdb/2.1/share/man/man1 and mediated links should be created.

components/database/couchdb-21/couchdb.p5m
```diff
@@ -0,0 +1,1408 @@
...
+     mediator=couchdb mediator-version=2.1
+
+file path=usr/couchdb/$(COMPONENT_VERSION_MAJOR)/LICENSE
+file path=usr/couchdb/$(COMPONENT_VERSION_MAJOR)/bin/couchdb
+file path=usr/couchdb/$(COMPONENT_VERSION_MAJOR)/bin/couchjs                 pkg.linted=true
+file path=usr/couchdb/$(COMPONENT_VERSION_MAJOR)/bin/couchup                 pkg.depend.bypass-generate=.*
```
pyhalov: remove multiple spaces after file path. Also, why we have pkg.linted in two places - here and on the top?

components/database/couchdb-21/files/couchdb-21.xml
grueni: I use argument -b now.

grueni:
This man page https://linux.die.net/man/1/couchdb is wrong or not for version 2.1.
The command line arguments for erlexec can be studied here
https://github.com/erlang/otp/blob/master/erts/etc/common/erlexec.c

pyhalov:
Looks better. Now the only question I have is where from does couchdb get its configuration files. It seems to ignore /etc/couchdb/2.1  and looks into /usr/couchdb/2.1/etc (which is a symlink to /etc/couchdb/2.1). Can we patch it to use /etc/couchdb/2.1 directly?
What is a common practice? What does Debian ? For example, in PostgreSQL, configs are supposed to live inside $PG_DATA.

grueni:
The folder etc is searched here:  apache-couchdb-2.1.1/src/config/src/config_app.erl

```erl
 50 default() ->
 51     Etc = filename:join(code:root_dir(), "etc"),
 52     Default = [
 53         filename:join(Etc, "default.ini"),
 54         filename:join(Etc, "default.d"),
 55         filename:join(Etc, "local.ini"),
 56         filename:join(Etc, "local.d")
 57     ],
 58     lists:filter(fun filelib:is_file/1, Default).
```
So it is a cheap operation to provide a folder etc in the root_dir of Couchdb.
I replaced the folder etc with a symlink to /etc/..
But it can also be other way around, a link /etc/couchdb/2.1 to folder etc in Couchdb.

Couchdb looks in its folder structure for binaries and configuration files. 
Data and log folder are defined in the configuration files.
The debian package from CouchDB is defined here:
- https://github.com/apache/couchdb-pkg/blob/master/debian/couchdb.dirs
- https://github.com/apache/couchdb-pkg/blob/master/debian/couchdb.links

I found no other "official" packages for 2.* for newer debian versions.

It is convenient to have all folders (real or as links) in one place.
We could add links from /var/couchdb/2.1/log|data to /usr/couchdb/2.1/.
So logs and data live in /var but are visible if you a a look into the couchdb folder.

Personally I would say that files which are potentially changed like configuration data should never be located in the same folder structure as binaries.
They should be located physically in /etc.
But I am biased by our business rules to manage our windows/linux/unix system.
This is the reason why I moved the the configuration folder from the package to the configuration area from the system. The former is convenient for the developper/project the latter is convenient for the administrators. My point of view is this of an administrator.

pyhalov:
I wouldn't put changing configuration files in /usr. As for preserving link, I have weak preference to avoid keeping it and patch couchdb  to look for configuration in /etc. This would decrease confusion. But if you consider that current structure is fine, so be it. What do you think?

grueni:
I see your point you would like to avoid any temptation to have configuration files in /usr.
That is my point too.

The files live now in /etc/couchdb/2.1/ and administrators should be educated to configure the applications in this folder.

The links in /usr/couchdb/2.1 are just for convenience. And in this project the folder etc is a necessity when source code is not changed.

Changing the source code of this project leads to the expectation that it should be changed for other projects too. And I know that some of the applications we use do it the same way. Not only for etc but for data and log too.

So I fear the future costs of changing source code and expect problems in maintainability in the future.

PR #3579, opened 2017-10-19
ncurses: update to 6.0-20171014

pyhalov:
(Mostly) ABI-compatible. We've delivered libpanelw.so.5 , libformw.so.5 and libmenuw.so.5, which we don't longer deliver. However, from all consumers of these additional libraries I've found only /usr/bin/samba-regedit, and it consumes non-w versions of these libraries. So I think it's safe to omit them.
Tested with vim, nano, clamdtop.


pyhalov:
Have to test it on some sun-color (console) terminal...

xen0l (approved):
LGTM, though question regarding GPG signature verification. Should we add it, shouldn't we?

components/library/ncurses/Makefile
```diff
@@ -61,31 +59,26 @@ CFLAGS += -std=c99 -D_POSIX_C_SOURCE=200112L
...
-CONFIGURE_OPTIONS +=	--enable-pc-files
+CONFIGURE_OPTIONS +=	--enable-symlinks
+CONFIGURE_OPTIONS +=	--disable-lib-suffixes
 CONFIGURE_OPTIONS +=	--includedir=$(USRINCDIR)/ncurses
+CONFIGURE_OPTIONS +=	--with-pkg-config-libdir=$(PKG_CONFIG_PATH)
+CONFIGURE_OPTIONS +=	--enable-pc-files
 
 CONFIGURE_ENV += CPPFLAGS="$(CPPFLAGS)"
 
-# XXX: Ugly hack copied directly from the desktop SUNWncurses.spec file:
```
pyhalov, line 71: Also need to restore this hack

pyhalov: With or without it after updating ncurses text-installer draws '?' instead of window boundaries in console.

pyhalov: Likely, we have to wait for console update in illumos-gate to happen before integrating this update.

pyhalov:
Closing this, as updated version is available - https://github.com/OpenIndiana/oi-userland/pull/4745

PR #3595, opened 2017-10-25
Lightdm: use /var/dt/sdtlogin interface to pass xorg server user ids

pyhalov:


components/desktop/lightdm/patches/20-sdtlogin.patch
```diff
@@ -0,0 +1,188 @@
...
++#ifdef __sun
++    {
++        struct stat statbuf;
++        int r;
++
++        r = stat (DT_DIR, &statbuf);
++        if (r < 0) {
++           g_mkdir (DT_DIR, 0755);
++        }
++
```
pyhalov, line 178: Original patch does 
```
g_rmdir(SDTLOGIN_DIR);
g_mkdir(SDTLOGIN_DIR,0700); 
```
But this will not work if this directory contains any pipes. So we just check that directory exists, and if it doesn't, create it.

alarcher: OK , thanks for the explanation.

alarcher:
@pyhalov Looks like lightdm does not publish if yelp is not installed. Also itstool, gobject-introspection, and vala as well.

pyhalov:
@alarcher fixed

alarcher (approved):

PR #3598, opened 2017-10-25
Fix various dependencies

alarcher:


components/x11/libfontenc/libfontenc.p5m
```diff
@@ -24,7 +24,7 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 <transform file path=usr/lib(.*)/pkgconfig/(.+).pc$ -> set action.hash usr/lib/xorg%<\1>/pkgconfig/%<\2>.pc >
 
-depend fmri=x11/library/libfontenc/encodings type=require
+#depend fmri=x11/library/libfontenc/encodings type=require
```
pyhalov: What's wrong in circular dep? It's not a build dep. IIRC, the idea is that if application require libfontenc (which is auto-detected), it automatically gets encodings, which are required for normal libfontenc work.

pyhalov:
Some changes touch component's files, but not bump COMPONENT_REVISION. Please, fix it.

alarcher:
@pyhalov Do we really want to republish packages although only packaging metadata is touched?

pyhalov:
For mkdocs, for example, you touch not only metadata, but p5m file. I'd incremented COMPONENT_REVISION. The same goes for libfontenc, however, for this one I'm not sure that change is required.

alarcher:
@pyhalov The encodings dependencies prevented me from building everything from scratch without installing from the repo first. But since it was a long time ago I do not remember how, so I just removed it.

PR #3613, opened 2017-10-26
ibus: add auto-start management interface to ibus-setup

pyhalov:


pyhalov:
It's work-in-progress attempt to add "enable/disable on startup" management to ibus-setup properties. The question is  - do we need it if imf-selector can do it for us?

components/inputmethod/ibus/patches/01-enable-autostart-management.patch
```diff
@@ -0,0 +1,81 @@
...
+-            os.symlink(ibus_desktop, link_file)
++        return False
++
++    def __checkbutton_auto_start_toggled_cb(self, button):
++        IBUS_SCRIPT="/etc/X11/xinit/imf.d/0010.ibus"
++
++        home = os.getenv("HOME")
++        if home:
++            if self.__checkbutton_auto_start.get_active():
++               os.system("/usr/bin/cat %s > %s/.imf" % (IBUS_SCRIPT, home))
```
xen0l: Why os.system and not subprocess.call or subprocess.check_call?

pyhalov: What's the actual difference?  We need to execute shell command. I'd worried more about path escaping here, perhaps, this exact call could be rewritten in clear python.

pyhalov: Avoided using os.system here. Also new code shouldn't care about file path escaping

kenmays:
LGTM

components/inputmethod/ibus/patches/01-enable-autostart-management.patch
```diff
@@ -0,0 +1,164 @@
...
++        try:
++            home = os.getenv("HOME")
++            if home:
++                config_file = path.join(home, ".imf")
++                if self.__checkbutton_auto_start.get_active():
++                   shutil.copyfile(IBUS_SCRIPT, config_file)
++                elif os.path.exists(config_file):
++                   os.unlink(config_file)
++        except Exception as e:
++            dlg = Gtk.MessageDialog(message_type = Gtk.MessageType.ERROR,
```
xen0l, line 66:
```
dlg = Gtk.MessageDialog(message_type = Gtk.MessageType.ERROR,
```
->
```
dlg = Gtk.MessageDialog(message_type=Gtk.MessageType.ERROR,
```

etc

xen0l (changes requested):

components/inputmethod/ibus/patches/01-enable-autostart-management.patch
pyhalov: It's a common style in that file

PR #3616, opened 2017-10-26
wget: update to 1.19.2

pyhalov:


pyhalov:
Needs libidn2 to be installed on the build server.

PR #3625, opened 2017-10-27
Build dvd+rw-tools as 64-bit

alarcher:


components/sysutils/dvd+rw-tools/patches/05-growisofs-manpage.patch
```diff
@@ -0,0 +1,13 @@
+Debian: https://sources.debian.net/src/dvd%2Brw-tools/7.1-11.1/debian/patches/02-growisofs-manpage.patch/
+
+Author: Daniel Baumann <daniel@debian.org>
+Description: Fix manpage section
+
+--- a/growisofs.1
++++ b/growisofs.1
+@@ -1,4 +1,4 @@
+-.TH GROWISOFS 1m "1 Mar 2008" "growisofs 7.1"
```
pyhalov: this is correct for us, no need to import this patch :)
/usr/share/man/man1m/growfs.1m

alarcher:
@pyhalov Removed.

alarcher:
@pyhalov Burned new snapshot on DVD with this version.

pyhalov (approved):

PR #3677, opened 2017-11-06
sane-fontends: fix build with gcc-6

alarcher:


AndWac (approved):
Why don't you download from alioth.debian.org,
the officially mentioned source? Otherwise ok.

alarcher:
@AndWac I could not download from them yesterday.

PR #3678, opened 2017-11-06
xsane: fix and rebuild with libjpeg-turbo

alarcher:


AndWac (approved):
Looks good and builds fine here.

PR #3696, opened 2017-11-07
nimbus: fix archive url

alarcher:


xen0l (commented):
Bump COMPONENT_REVISION?

alarcher:
@xen0l Even if the archive has not changed?

PR #3702, opened 2017-11-08
Libvisual

alarcher:


components/library/libvisual-plugins/Makefile
```diff
@@ -30,19 +31,29 @@ COMPONENT_LICENSE = GPLv2
...
 
 
 CONFIGURE_OPTIONS += --sysconfdir=/etc
 CONFIGURE_OPTIONS += --enable-shared=yes
 CONFIGURE_OPTIONS += --enable-static=no
 CONFIGURE_OPTIONS += --disable-corona
 CONFIGURE_OPTIONS += --disable-gforce
+CONFIGURE_OPTIONS += --disable-esd
+# Only 0.8 supported
+CONFIGURE_OPTIONS += --disable-gstreamer-plugin
```
pyhalov, line 46: Did p5m file change after disabling some options?

alarcher:
It did not when I regenerated sample-manifest.p5m but I can double check just in case I forgot to clobber.

components/library/libvisual-plugins/Makefile
alarcher: Just re-checked: no change.

PR #3704, opened 2017-11-08
thunderbird: fix runpath resolution

alarcher:


components/mail/thunderbird/Makefile
```diff
@@ -108,7 +109,7 @@ COMPONENT_POST_INSTALL_ACTION += \
 	unzip -u -d $(LIGHTNING_DIR) $(USERLAND_ARCHIVES)$(COMPONENT_ARCHIVE_1); \
 	for file in $(PROTO_DIR)/usr/lib/thunderbird-$(COMPONENT_VERSION)/*.so \
           $(PROTO_DIR)/usr/lib/thunderbird-$(COMPONENT_VERSION)/plugin-container; do \
-	  /usr/bin/elfedit -e 'dyn:value -s  RPATH "/usr/lib:/usr/lib/thunderbird"' $$file ;\
+	  /usr/bin/elfedit -e 'dyn:value -s  RPATH "/usr/lib:/usr/lib/thunderbird:$$ORIGIN:$$ORIGIN/.."' $$file ;\
```
pyhalov: What are you trying to do? /usr/lib/thunderbird == $ORIGIN for these libraries...

pyhalov:
There is an issue with libxul.so - it was linked by GNU ld, and it seems it's broken enough for elfedit to fail to operate on it correctly. Correct solution here would be to try (likely, together with @richlowe) to find out where our ld is wrong and use it to link libxul.so.

alarcher:
I did the same thing as in the firefox component to try to convince pkgdepend that the library should be resolved.

pyhalov:
Firefox is a bit different, it does the same for $(PROTO_DIR)/usr/lib/firefox/sdk/lib/*.so and there $ORIGIN has different meaning.

alarcher:
Unfortunately without doing something about it I cannot publish in my zone. Not sure what changed.

pyhalov:
GCC 6 zone or usual one? How does it fail?

alarcher:
@pyhalov GCC 6. I sent the output to you by email.

pyhalov:
Can't find the message.

pyhalov:
I know the reason. GCC 6 encodes RUNPATH: 
<pre>
 elfdump build/prototype/i386/usr/lib/thunderbird-52.4.0/libprldap60.so |grep PA
      [14]  RUNPATH           0x9aa               /usr/gcc/6/lib
      [15]  RPATH             0x9b9               /usr/lib:/usr/lib/thunderbird
      [45]  SUNW_STRPAD       0x1e2               
</pre>
GCC sets RUNPATH and RPATH, but RPATH is overwritten later by elfedit. And it is considered, if present, and so, pkgdepend resolve does correct thing when doesn't find libraries in /usr/lib/thunderbird, as it seems RUNPATH overwrites RPATH.

As I understand, we have to use RPATH here (or both). 
Fix for thunderbird could look like:
```
@@ -109,7 +108,8 @@ COMPONENT_POST_INSTALL_ACTION += \
        unzip -u -d $(LIGHTNING_DIR) $(USERLAND_ARCHIVES)$(COMPONENT_ARCHIVE_1); \
        for file in $(PROTO_DIR)/usr/lib/thunderbird-$(COMPONENT_VERSION)/*.so \
           $(PROTO_DIR)/usr/lib/thunderbird-$(COMPONENT_VERSION)/plugin-container; do \
-         /usr/bin/elfedit -e 'dyn:value -s  RPATH "/usr/lib:/usr/lib/thunderbird"' $$file ;\
+         /usr/bin/elfedit -e 'dyn:value -s  RUNPATH "/usr/lib:/usr/lib/thunderbird:$(GCC_ROOT)/lib"' $$file ;\
+         /usr/bin/elfedit -e 'dyn:value -s  RPATH "/usr/lib:/usr/lib/thunderbird:$(GCC_ROOT)/lib"' $$file ;\
        done ;
```

alarcher:
Commited your fix.

pyhalov:
The question which I have is if we should to merge this now or with GCC 6 switch?

alarcher:
I do not mind either. As long as it builds in the current gcc-6 zone for now it is fine.

pyhalov:
So, I think you can move this commit to GCC-6 PR.

PR #3717, opened 2017-11-09
xterm: temporarily move back to 32bit as 64bit is broken

alarcher:


pyhalov:
Is it related to 'vim with xterm bug', discussed in illumos-dev?

alarcher:
Yes it is. Just want to provide a working xterm until we find the fix.

PR #3718, opened 2017-11-10
Bump nghttp2 to 1.27.0

alarcher:
Moving to category and updating at the same time, but:

pkg search -r -o pkg.name 'depend:require:library/nghttp2'
PKG.NAME
web/curl
web/server/apache-24

so some testing needed.


pyhalov:
need to check if apache still can speak HTTP2

pyhalov (approved):

PR #3720, opened 2017-11-10
hacluster-common: Add download target

hliebel:
Hi.

I'm trying to do a top level 'gmake download' and apply some simple fixes where I can.

Thanks,
Harry
PS. This is my first one, so apologies if I get it wrong.

### DIFF
--- a/components/cluster/hacluster-common/Makefile
+++ b/components/cluster/hacluster-common/Makefile
@@ -35,6 +35,8 @@ COMPONENT_SUMMARY= Common items for hacluster components.
 
 include $(WS_MAKE_RULES)/ips.mk
 
+download:
+
 build:
 
 install:


pyhalov:
Can you also add 'prep' target, while you are on this?

hliebel:
I could. Could you explain to me why that target it is needed? If I do a 'gmake prep' in the component directory it seems to 'do nothing' as expected, Whereas a 'gmake download' causes the build to fail and stop.

There is a '.PHONY prep' target in the shared_macros.mk. Should there perhaps be a 'download' target too?

pyhalov:
Perhaps. Also I don't understand why 'gmake prep' works now for this component, I don't see where it inherits rule for prep target.

hliebel:
prep and other PHONY targets are declared at the bottom of
make-rules/shared-macros.mk
I think I'm going to add 'download' in here and see how it goes. Is a more general solution if it works.

PR #3727, opened 2017-11-13
Fix xorg module drivers path

alarcher:


jimklimov:
Haven't seen such transforms before, but if I get their purpose correctly - LGTM :)

pyhalov:
I have the only question - how did it work before this change?

alarcher:
Most likely because the auto-config patch skips non-64bit paths.

PR #3728, opened 2017-11-13
shared-macros: Add PHONY download target

hliebel:
Some components do not specify/require a 'download' target. We do not
want the top level 'gmake download' operation to fail in this case.


pyhalov:
This is wrong. This makes 'gmake prep ' always do something - extract and patch archive, even if it's already extracted.

PR #3730, opened 2017-11-13
gnome-connection-manager: Fix download URL

hliebel:
Main project site is gone, use a mirror for now.


pyhalov:
Fixing url is fine thing, supporting dead project is not so fine. @jimklimov is upstream alive? If it's not, should we continue providing this package?

PR #3732, opened 2017-11-13
Bump cmake to 3.9.6

alarcher:
Manifest and test suite unchanged.


pyhalov:
Have you built some mysqls with it?

PR #3737, opened 2017-11-14
Fix dictionaries

alarcher:


pyhalov:
Why new empty lines in end of files are there?

alarcher:
Erf... because I ran gmake REQUIRED_PACKAGES which was empty for these components and forgot to remove them...

alarcher:
@pyhalov New lines are fixed.

PR #3744, opened 2017-11-16
Make pkgdepend ignore GCC runtime libraries in /usr/lib and /usr/lib/…

pyhalov:
…$(MACH64) if  current compiler is not gcc-4

### DIFF
--- a/make-rules/ips.mk
+++ b/make-rules/ips.mk
@@ -62,6 +62,10 @@ GENERATE_TRANSFORMS +=		$(WS_TOP)/transforms/generate-cleanup
 PKGMOGRIFY_TRANSFORMS +=	$(WS_TOP)/transforms/libtool-drop
 PKGMOGRIFY_TRANSFORMS +=	$(WS_TOP)/transforms/ignore-libs
 
+ifneq ($(GCC_ROOT), /usr/gcc/4.9)
+PKGMOGRIFY_TRANSFORMS +=	$(WS_TOP)/transforms/ignore-gcc-usr-lib
+endif
+
 COMPARISON_TRANSFORMS +=	$(WS_TOP)/transforms/comparison-cleanup
 COMPARISON_TRANSFORMS +=	$(PKGMOGRIFY_TRANSFORMS)
 
--- a/transforms/ignore-gcc-usr-lib
+++ b/transforms/ignore-gcc-usr-lib
@@ -1,0 +1,17 @@
+# Don't depend on GCC libraries in /usr/lib or /usr/lib/$(MACH64)
+<transform file -> add pkg.depend.bypass-generate usr/lib/libgcc_s\\.so\\.1 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libgcc_s\\.so\\.1 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/libgfortran\\.so\\.3 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libgfortran\\.so\\.3 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/libgomp\\.so\\.1 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libgomp\\.so\\.1 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/libgobjc\\.so\\.4 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libgobjc\\.so\\.4 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/libgobjc_gc\\.so\\.4 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libgobjc_gc\\.so\\.4 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/libquadmath\\.so\\.0 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libquadmath\\.so\\.0 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/libssp\\.so\\.0 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/$(MACH64)/libssp\\.so\\.0 >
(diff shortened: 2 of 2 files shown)


transforms/ignore-gcc-usr-lib
```diff
@@ -0,0 +1,17 @@
+# Don't depend on GCC libraries in /usr/lib or /usr/lib/$(MACH64)
+<transform file -> add pkg.depend.bypass-generate usr/lib/libgcc_s\\.so\\.1 >
+<transform file -> add pkg.depend.bypass-generate usr/lib/(sparcv9|amd64)/libgcc_s\\.so\\.1 >
```
xen0l: Just a nitpick, maybe change to MACH64?

xen0l (approved):

pyhalov:
@alarcher does it work for you?

alarcher:
It does not seem to affect in good or bad the components that I recompiled. Which specific one do you want me to check? I lost track of the discussion...

make-rules/ips.mk
```diff
@@ -62,6 +62,10 @@ GENERATE_TRANSFORMS +=		$(WS_TOP)/transforms/generate-cleanup
 PKGMOGRIFY_TRANSFORMS +=	$(WS_TOP)/transforms/libtool-drop
 PKGMOGRIFY_TRANSFORMS +=	$(WS_TOP)/transforms/ignore-libs
 
+ifneq ($(GCC_ROOT), /usr/gcc/4.9)
```
alarcher, line 65: What I meant with the definition of a USERLAND_GCC_ROOT is that there are two variables needed:
1) the userland gcc used as system compiler
2) the component gcc with overriden GCC_ROOT
That's what you do by comparing if GCC_ROOT == /usrr/gcc/4.9; in this case USERLAND_GCC_ROOT=/usr/gcc/4.9.
My suggestion was to avoid hardcoding /usr/gcc/4.9 and put it in a variable, so that we avoid changing the hardcoded value in different locations whenever we migrate to a new compiler. @pyhalov @xen0l

alarcher:
@pyhalov I tried with clang and get for example:

    /export/home/userland/oi-userland/components/developer/clang-40/build/manifest-i386-clang-40.depend has unresolved dependency '
        pkg.debug.depend.reason=usr/clang/4.0/bin/llvm-strings \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.fullpath=lib/libgcc_s.so.1 \
        pkg.debug.depend.fullpath=usr/clang/4.0/bin/../lib/libgcc_s.so.1'.

although:

      [11]  NEEDED            0x109bd             libgcc_s.so.1
      [15]  RUNPATH           0x10a46             /usr/gcc/6/lib:$ORIGIN/../lib
      [16]  RPATH             0x10a46             /usr/gcc/6/lib:$ORIGIN/../lib

make-rules/ips.mk
pyhalov: Do I understand correctly that you want to set USERLAND_GCC_ROOT to /usr/gcc/4.9? When and why will we have to change it?

alarcher: It should match the host system compiler. Maybe the name is not correct.

pyhalov: still do not understand :(

pyhalov:
Where from can I get updated clang component? Current one evidently incorrectly uses elfedit...

alarcher:
If you talk about the compilation with GCC6, you need the updated patch to set the paths for gcc-6 in openindiana.patch.

pyhalov:
I'm asking, because without changes to Makefile, I get 
```
ldd build/prototype/i386/usr/clang/4.0/bin/llvm-strings 
        libdl.so.1 =>    /lib/libdl.so.1
        libcurses.so.1 =>        /lib/libcurses.so.1
        libstdc++.so.6 =>        /usr/lib/libstdc++.so.6
        libstdc++.so.6 (GLIBCXX_3.4.21) =>       (version not found)
        librt.so.1 =>    /lib/librt.so.1
        libgcc_s.so.1 =>         /usr/lib/libgcc_s.so.1
elfdump build/prototype/i386/usr/clang/4.0/bin/llvm-strings  |grep PATH
      [15]  RUNPATH           0x10a55             $ORIGIN/../lib
      [16]  RPATH             0x10a55             $ORIGIN/../lib
```

pyhalov:
publishing clang succeeds for me with the following changes to clang's Makefile:
```
--- a/components/developer/clang-40/Makefile
+++ b/components/developer/clang-40/Makefile
@@ -55,11 +57,7 @@ COMPONENT_PRE_CONFIGURE_ACTION =        (cp -a $(SOURCE_DIR)/* $(@D))
 
 COMPONENT_POST_INSTALL_ACTION =  \
 for file in `echo $(PROTO_DIR)/$(CMAKE_PREFIX)/bin/*`; do \
-       elfedit -e 'dyn:runpath $$ORIGIN/../lib' $$file ; \
-done && \
-for file in `echo $(PROTO_DIR)/$(CMAKE_PREFIX)/lib/*.so`; do \
-       elfedit -e 'dyn:delete RUNPATH' $$file ; \
-       elfedit -e 'dyn:delete RPATH' $$file ; \
+       elfedit -e 'dyn:runpath $(GCC_ROOT)/lib:$$ORIGIN/../lib' $$file ; \
 done && \
 for file in \
        `ggrep -rlh '^\#!*/usr/bin/env python' $(PROTO_DIR)/$(CMAKE_PREFIX)`; do \
```

alarcher:
@pyhalov Thanks, I though I had fixed that but apparently not ... possibly committed to the wrong branch... Only hunspell and fltk's fluid seem to need to the same fix.

pyhalov:
I still don't understand the question about USERLAND_GCC_ROOT. I suppose, every gcc > 4.9 will deliver runtimes to /usr/gcc/X/lib/.... . So, the only gcc, which should use runtimes from /usr/lib is GCC 4.9. So if we don't compile programs with GCC 4.9, we should ignore runtimes in /usr/lib.

alarcher:
@pyhalov Then this is the piece of information I was missing :) Some months ago when we discussed with tomww, I was told that the default/system compiler should have runtime libraries in /usr/lib. If the idea is that any compiler > 4.9 has runtime libraries in its own prefix then I take my remark back :)

PR #3749, opened 2017-11-17
Add 'nvme' driver to installation media

Mno-hime:
With `nvme` driver OpenIndiana can be installed to NVMe devices.

On VirtualBox OI installs but won't boot as that is possible only in
UEFI mode: https://www.virtualbox.org/manual/ch05.html#harddiskcontrollers.

QEMU should boot as NVMe support was added to it's BIOS recently
(https://www.seabios.org/Releases#SeaBIOS_1.11.0) but won't install due
to illumos#8808 bug.


pyhalov:

Mno-hime:
Bumped.

PR #3758, opened 2017-11-21
7755 library/gnu-libiconv should not deliver files to /usr/gnu/ prefix

pyhalov:


tomww:
Could gnu-libiconv package get an automatic uninstall/retirement?
That would help to not keep a dead package.
Thank you!

pyhalov:
It will be removed automatically.

pyhalov:
smbclient/smbd seems to work fine, but we also have to remove all '-liconv's from pkgconfig files.

pyhalov:
Added noincorporate attribute.

xen0l (approved):

PR #3762, opened 2017-11-22
Add mupdf 1.11

alarcher:
Libraries only delivered as static.

### DIFF
--- a/components/desktop/mupdf/Makefile
+++ b/components/desktop/mupdf/Makefile
@@ -1,0 +1,67 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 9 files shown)


pyhalov:
Is it a build dependency for something or a separate pdf reader? If it's a separate reader, why do we need one more?

alarcher:
The main interest is the epub reader. I tried Atril's Epub support but it fails to open samples from Adobe while MuPDF has not failed on any.

pyhalov:
Perhaps, we could update fbreader?

alarcher:
OK. I thought the project was dead.

pyhalov:
It's in strange shape. New mobile versions are appearing. New desktop version is promised. Old desktop versions are 0.12.10 and 0.99.6 , but they are rather old...

pyhalov:
@alarcher Do you have interest in updating it?

pyhalov:
@alarcher What are we going to do with this PR? Will you update it or should I close it for now?

alarcher:
Since nowadays I cannot ensure that I can maintain it I'll just close.

PR #3768, opened 2017-11-24
Make rules

pyhalov:


tools/userland-fetch
```diff
@@ -230,6 +230,14 @@ def download_paths(search, filename, url):
     if url is not None and url not in urls:
         urls.append(url)
 
+    # last resort path
+    if filename is not None:
+        tmp = os.getenv('DOWNLOAD_FALLBACK_PATH')
+
+        file = os.path.basename(filename)
+
+        urls += [base + '/' + file for base in tmp.split(' ')]
```
xen0l: This is preferred:
```
urls += ['/'.join(base, file) for base in tmp.split()]
```

make-rules/shared-macros.mk
```diff
@@ -44,11 +44,16 @@ PATH = $(PATH.illumos)
...
 #	http://static.opensolaris.org/action/browse/userland/tarball/userland
 
+DLC_ARCHIVE_MIRROR = http://dlc.openindiana.org/oi-userland/source-archives
+
 # Default to looking for source archives on the internal mirror and the external
 # mirror before we hammer on the community source archive repositories.
 #export DOWNLOAD_SEARCH_PATH +=	$(INTERNAL_ARCHIVE_MIRROR)
 #export DOWNLOAD_SEARCH_PATH +=	$(EXTERNAL_ARCHIVE_MIRROR)
 
+# Look for file at DLC server as last resort
```
xen0l, line 54: as -> as the

xen0l: Is the rsync job from the build machine to DLC enabled again?

pyhalov: yes

PR #3771, opened 2017-11-24
mod_wsgi: update to 4.5.22, fix unsigned int overflow

pyhalov:


grueni (commented):
I gues you should change COMPONENT_REVISION too.

pyhalov:
No, as COMPONENT_VERSION is increased

grueni:
Tested and works again with pkg/depot

PR #3774, opened 2017-11-26
augeas: bump to version 1.9.0

grueni:
IMHO this component should be in system/management/augeas.
The current locations is library/augeas


components/sysutils/augeas/Makefile
```diff
@@ -11,26 +11,28 @@
...
-  http://download.augeas.net/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	sha256:2b463d398cabc9b42747aa61d3e83ed6a93ce03d9074cf8e7a7bd3107a668343
+COMPONENT_ARCHIVE_URL= http://download.augeas.net/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL=	http://augeas.net/index.html
 COMPONENT_FMRI=	library/$(COMPONENT_NAME)
 COMPONENT_CLASSIFICATION=	System/Libraries
 COMPONENT_LICENSE=	LGPLv2.1
 COMPONENT_LICENSE_FILE=	COPYING
 
+PKG_MACROS += LIBAUGEAS_VER=so.0.23.0
```
pyhalov: Where do you use this macros?

pyhalov:
Is it ABI-compatible?

pyhalov:
Does nut build after this update?

components/sysutils/augeas/Makefile
grueni: I don't use the definitions. I guess they will be used by the auto* facility.
Comment out or remove?

grueni:
nut builds.

There are API changes at least for bug fixes. Exiisting applications should be recompiled. I found only nut.

components/sysutils/augeas/Makefile
pyhalov: remove

components/sysutils/augeas/Makefile
```diff
@@ -41,19 +40,48 @@ CFLAGS.32=	-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE
...
+        '-e "/passed/p" ' \
+        '-e "/failed/p" ' \
+        '-e "/Checks:/p" ' \
+        '-e "/Total:/p" ' \
+        '-e "/Passed/p" ' \
+        '-e "/Failed/p" ' \
+        '-e "/Failure expected/p" ' \
+        '-e "/SKIP/p" ' \
+        '-e "/PASS/p" ' \
+        '-e "/FAIL/p" '
```
pyhalov: File with test results seems to be missing

pyhalov:
Differences found in tests:
```
@@ -1,6 +1,6 @@
 # TOTAL: 99
-# PASS:  77
-# SKIP:  22
+# PASS:  87
+# SKIP:  12
 # XFAIL: 0
 # FAIL:  0
 # XPASS: 0
```
http://buildzone.oi-build.r61.net/augeas-test-results.tar

grueni:
I assume the build environments used differ in the installed packages. For some packages augeas has lenses and use the defined tests.
The numbers of test passed or skipped successfully are different but this may be tolerable.
No tests fail and this is the most important information.
How to handle this situation? Must the test of each package called in  a clean (same for all packages) or in a especially for this package defined environment?
How did you handle the testing actually?

PR #3782, opened 2017-11-27
ejabberd: bump to version 17.11

grueni:
ejabberd 2.1.13 does not work with erlang > 17.5


components/web/ejabberd/Makefile
```diff
@@ -21,60 +21,77 @@
...
+COMPONENT_LICENSE_FILE= $(COMPONENT_NAME).license
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CC=gcc -m64 -I/usr/include/gd2
+CPP=gcc -m64 -E -I/usr/include/gd2
+CXX=g++ -m64 -I/usr/include/gd2
+LDFLAGS += -L/usr/lib/libjpeg8-turbo/lib/amd64 -R/usr/lib/libjpeg8-turbo/lib/amd64 -lnsl -lsocket
```
pyhalov: s/amd64/$(MACH64)/

components/web/ejabberd/Makefile
```diff
@@ -21,60 +21,77 @@
...
+COMPONENT_DESCRIPTION= the ejabberd Jabber/XMPP instant messaging server
+COMPONENT_CLASSIFICATION= Web Services/Communications
+COMPONENT_LICENSE= GPL-2.0
+COMPONENT_LICENSE_FILE= $(COMPONENT_NAME).license
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CC=gcc -m64 -I/usr/include/gd2
```
pyhalov: Can we use CFLAGS/CXXFLAGS/CPPFLAGS here?

components/web/ejabberd/Makefile
```diff
@@ -21,60 +21,77 @@
...
-CONFIGURE_ENV += "PATH=/usr/xpg4/bin:/usr/bin"
+COMPONENT_PRE_BUILD_ACTION+= \
+  $(CLONEY) $(SOURCE_DIR) $(@D)
 
 CONFIGURE_OPTIONS +=	--sysconfdir=$(ETCDIR)
 CONFIGURE_OPTIONS +=	--with-erlang=$(USRBINDIR)
 CONFIGURE_OPTIONS +=	--localstatedir=/var
 CONFIGURE_OPTIONS +=	--enable-odbc
-
+CONFIGURE_OPTIONS +=	--enable-mysql
```
pyhalov: You enable mysql, but doesn't force using of mariadb-101. The same goes for pgsql. Also, what about REQUIRED_PACKAGES? Do they need updating?

components/web/ejabberd/Makefile
alarcher: You should use the predefined macros instead, for reference:
    `CFLAGS   += $(JPEG_CPPFLAGS) $(JPEG_CFLAGS)`
    `CXXFLAGS += $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS)`
    `LDFLAGS  += $(JPEG_LDFLAGS)`

components/web/ejabberd/ejabberd.p5m
```diff
@@ -19,283 +19,834 @@
...
+file files/prof_attr         path=etc/security/prof_attr.d/ejabberd
+file files/svc-xmpp-ejabberd path=lib/svc/method/svc-xmpp-ejabberd
+file files/xmpp-ejabberd.xml path=lib/svc/manifest/network/xmpp-ejabberd.xml
+
+file path=etc/ejabberd/ejabberd.yml    mode=0644 preserve=true
+file path=etc/ejabberd/ejabberdctl.cfg mode=0644 preserve=true
+file path=etc/ejabberd/inetrc          mode=0644 preserve=true
+
+file path=usr/sbin/$(MACH64)/ejabberdctl
+link path=usr/sbin/ejabberdctl target=./$(MACH64)/ejabberdctl
```
pyhalov: Can we just set CONFIGURE_SBINDIR.64 =   $(CONFIGURE_PREFIX)/bin ?

pyhalov (changes requested):

components/web/ejabberd/Makefile
grueni: This options compile the adapters for the databases.
It is up to the user to use one of them instead of the builtin mnesia database.

pyhalov: I mean, does it make ejabberd link to mysql libraries? If so, it should use MYSQL_VERSION of libraries. The same goes for pgsql.

components/web/ejabberd/Makefile
```diff
@@ -45,10 +45,14 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 
-CC=gcc -m64 -I/usr/include/gd2
-CPP=gcc -m64 -E -I/usr/include/gd2
-CXX=g++ -m64 -I/usr/include/gd2
-LDFLAGS += -L/usr/lib/libjpeg8-turbo/lib/amd64 -R/usr/lib/libjpeg8-turbo/lib/amd64 -lnsl -lsocket
+CC=gcc
+CPP=gcc -E
+CXX=g++ -m64
```
pyhalov: Why do we need this?

grueni:
I didn't figure out how to avoid this definitions.
CC=gcc
CPP=gcc -E
CXX=g++ -m64
If not used rebar compiles *.cpp as 32 bit.

components/web/ejabberd/Makefile
grueni: I didn't figure out how to avoid this definitions.
CC=gcc
CPP=gcc -E
CXX=g++ -m64
If not used rebar compiles *.cpp as 32 bit.
And uses cc instead of gcc.

alarcher: At least $(CC_BITS) would be preferred.

grueni: The next commit solves the problem.

components/web/ejabberd/Makefile
```diff
@@ -21,60 +21,80 @@
...
 CONFIGURE_OPTIONS +=	--with-erlang=$(USRBINDIR)
 CONFIGURE_OPTIONS +=	--localstatedir=/var
 CONFIGURE_OPTIONS +=	--enable-odbc
+CONFIGURE_OPTIONS +=	--enable-mysql
+CONFIGURE_OPTIONS +=	--enable-pgsql
+CONFIGURE_OPTIONS +=	--enable-sqlite
+CONFIGURE_OPTIONS +=	--enable-pam
+CONFIGURE_OPTIONS +=	--enable-riak
+CONFIGURE_OPTIONS +=	--enable-redis
+CONFIGURE_OPTIONS +=	--enable-user=daemon
```
pyhalov: When this option is set, gmake install fails, because it tries to chown files to daemon.

pyhalov (changes requested):

components/web/ejabberd/Makefile
```diff
@@ -21,60 +21,80 @@
...
+
+CFLAGS   += $(JPEG_CPPFLAGS) $(JPEG_CFLAGS) -I$(USRINCDIR)/gd2
+CXXFLAGS += $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS) -I$(USRINCDIR)/gd2
+CPPFLAGS += $(JPEG_CPPFLAGS) -I$(USRINCDIR)/gd2
+LDFLAGS  += $(JPEG_LDFLAGS) -lnsl -lsocket
 
-COMPONENT_BUILD_ENV += "HOME=/tmp"
 CONFIGURE_ENV += "HOME=/tmp"
+CONFIGURE_ENV += "CPP=$(CC) -E"
+CONFIGURE_ENV += "CXX=$(CXX) $(CC_BITS)"
```
pyhalov: Better avoid using it, as CONFIGURE_ENV already contains "CXX=$(CXX)". 
The easiest workaround is to add 
```
CXX += $(CC_BITS)
```
for example, above line 47.

components/web/ejabberd/Makefile
grueni: I removed the option.
I thought user daemon exists in each system.

pyhalov: The issue is not that it doesn't exist. The issue is that gmake install is run as usual user, but tries to change files owner.

pyhalov (approved):

PR #3787, opened 2017-11-27
Bump git to 2.16.2

alarcher:
Review of the test suite WIP.


pyhalov:
I got different test results. Part of tests requires library/perl-5/sqlite-dbi , "git svn branch tests"  also succeed here (do you have subversion and perl subversion modules installed?), part of test results appears in another order, part of test results is just not reproducible (such as 'test-lib-functions.sh: line 388: 17311 Terminated              git fast-import $options 0<&8 1>&9').

alarcher:
Yes all these components are installed but as usual I do not get the same results ... Maybe it is related to the locale ... just annoying that they do not even enforce an ordering of the tests.

AndWac (approved):
Looks good to me. Albeit I also have different test results (slightly better). I have set locale "C" for my build user. That should give culture neutral results.

PR #3797, opened 2017-11-29
Bump lcms2 to 2.9 and import test transforms from solaris-userland

alarcher:
ABI compatible, maintenance release.
One test failure in 32-bit, same as 2.8.0.

### DIFF
--- a/components/library/lcms2/Makefile
+++ b/components/library/lcms2/Makefile
@@ -23,7 +23,7 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME =        lcms2
-COMPONENT_VERSION=      2.8
+COMPONENT_VERSION=      2.9
 COMPONENT_FMRI=         library/lcms2
 COMPONENT_CLASSIFICATION=System/Libraries
 COMPONENT_SUMMARY=      The Little Color Management System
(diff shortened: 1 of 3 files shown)


components/library/lcms2/Makefile
```diff
@@ -46,12 +46,38 @@ CFLAGS             +=           $(JPEG_CPPFLAGS) $(JPEG_CFLAGS)
 CXXFLAGS           +=           $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS)
 LDFLAGS            +=           $(JPEG_LDFLAGS)
 
+# Seems that we have no other way to designate that we want large file support.
+CFLAGS += $(CPP_LARGEFILES)
```
pyhalov, line 50: Have you checked that this didn't break ABI?

alarcher: I only tested eom, eog and xsane with the new lcms2. What should be done?

pyhalov: Check, first of all, if previous 32-bit version of lcms2 was compiled with _FILE_OFFSET_BITS=64  (for example, in config.h).
This affects size of file-related structures (and their members) and can lead to interesting crashes. If it is not so, the safest thing to do is to build it with _FILE_OFFSET_BITS 64 and rebuild dependent apps.

pyhalov: Given that poppler depends on lcms2 and IS compiled with _FILE_OFFSET_BITS=64, I'd added this option.

pyhalov (approved):

components/library/lcms2/Makefile
alarcher: I compared the headers and cannot find any trace of change after this flag is added, should be safe.

PR #3799, opened 2017-11-30
Bump ngrep to 1.47

alarcher:


jimklimov:
I'd also stuff `$(COMPONENT_NAME)` into the download URL but maybe that's overkill.

Is the project URL at SourceForge still valid?

Beside these small nits, LGTM in general.

alarcher:
@jimklimov You are right the sourceforge page is not active anymore. I just fixed the project url accordingly.

xen0l (approved):

PR #3811, opened 2017-12-02
Avoid linking luajit

alarcher:
Linking to LUAjit is broken, so we choose Lua.

### DIFF
--- a/components/desktop/hexchat/Makefile
+++ b/components/desktop/hexchat/Makefile
@@ -16,6 +16,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		hexchat
 COMPONENT_VERSION=	2.12.4
+COMPONENT_REVISION=	1
 COMPONENT_PROJECT_URL=	https://hexchat.github.io/
 COMPONENT_SUMMARY=	HexChat IRC client
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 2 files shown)


components/library/libpeas/Makefile
```diff
@@ -56,6 +57,7 @@ CONFIGURE_ENV.64 += PYTHON3_CONFIG=/usr/bin/$(MACH64)/python3.4-config
 CONFIGURE_OPTIONS += --disable-static
 CONFIGURE_OPTIONS += --enable-gtk
 CONFIGURE_OPTIONS += --enable-lua5.2
+CONFIGURE_OPTIONS += --disable-luajit
```
pyhalov, line 60: This is not strictly required, luajit is not found and so disabled in libpeas (as I see from configure)

components/desktop/hexchat/Makefile
```diff
@@ -27,33 +28,32 @@ COMPONENT_ARCHIVE_URL=	https://dl.hexchat.net/hexchat/$(COMPONENT_ARCHIVE)
...
 					$(RM) build-aux/ltmain.sh && \
-					$(LN) -s /usr/share/libtool/config/ltmain.sh build-aux/ltmain.sh)
+					$(LN) -s /usr/share/libtool/build-aux/ltmain.sh build-aux/ltmain.sh)
 
+CONFIGURE_ENV += OBJC=$(CC)
 CONFIGURE_ENV += PERL=$(PERL)
 CONFIGURE_ENV += INTLTOOL_PERL=$(PERL)
 
 CONFIGURE_OPTIONS += --enable-python=python2.7
+# Avoid using luajit since static linking is broken
```
pyhalov, line 48: Perhaps, better say 'Avoid using luajit as we don't have it'?

pyhalov: ah, sorry, I just don't have it installed...

components/library/libpeas/Makefile
pyhalov: Ah, OK, I just don't have it installed.

pyhalov:
Perhaps, we should fix luajit..

alarcher:
@pyhalov I tried but it is not just about adding -fPIC so it may take time.

PR #3812, opened 2017-12-02
spidermonkey: add patch for gcc6

grueni:
Compiles with gcc6 and passes the test.

I changed in ../../../make-rules/shared-macros.mk 
GCC_VERSION = 4.9
to
GCC_VERSION = 6
to test the patch.
Is this the correct way?


pyhalov:
You could just do gmake GCC_ROOT=/usr/gcc/6 publish. 
Have you checked that it still builds with gcc 4?

grueni:
Yes, I tested both.

PR #3814, opened 2017-12-03
open-vm-tools: bump to version 10.1.15

grueni:
Tested with
- make publish
- make GCC_VERSION=6 publish


pyhalov:
How have you you tested it?

grueni:
I installed the package in a VM under ESX 6.5.
The log does not show any errors.
The GUI of Vmware recognizes the vmware tools as current version (10287).

~~I have to update the commit. The logs were wriiten in the wrong location.~~
Logs are now written to /var/log/vmtools/.

components/sysutils/open-vm-tools/patches/04-nicInfo.patch
```diff
@@ -0,0 +1,43 @@
...
++#if defined(USERWORLD) || defined(sun)
+ /*
+  ******************************************************************************
+  * CountNetmaskBits --                                                   */ /**
+@@ -259,7 +256,7 @@ GuestInfoGetNicInfo(NicInfoV3 *nicInfo)
+    }
+ 
+    return TRUE;
+-#elif defined(USERWORLD)
++#elif defined(USERWORLD) || defined(sun)
```
pyhalov: This can break in c99 mode. Use ```defined(__sun)``` or ```defined(__sun__)```
I mean that patch is either partially wrong or not really necessary. For example:
```$ elfdump build/i86/lib/nicInfo/.libs/libNicInfo_la-nicInfoPosix.o |grep CountNetmaskBits``` shows , that it was never defined.

File: components/sysutils/open-vm-tools/patches/04-nicInfo.patch [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+    struct ifaddrs *ifaddrs = NULL;
+ 
+    if (getifaddrs(&ifaddrs) == 0 && ifaddrs != NULL) {
+@@ -348,7 +345,7 @@ GuestInfoGetNicInfo(NicInfoV3 *nicInfo)
+  *
+  ******************************************************************************
+  */
+-#if defined(__FreeBSD__) || defined(__APPLE__) || defined(USERWORLD)
++#if 1
```
pyhalov: why not just add defined(__sun) as in other cases?

grueni:
You are right, the patch 04-nicInfo.patch is almost unnecessary if libdnet is used.

PR #3820, opened 2017-12-04
GCC6 fixes

alarcher:
Every time the branch is rebased, Jenkins rebuilds all the touched components in the gcc6 branch.
I would like to merge some of the fixes to avoid rebuilding.


components/library/libdnet/patches/dnet.3.patch
```diff
@@ -1,1489 +0,0 @@
---- libdnet-1.12/man/dnet.3.orig	2012-05-23 08:31:15.785689693 -0700
```
pyhalov, line 1: Why did you remove these patches?

components/text/aspell/patches/01-gcc7.patch
```diff
@@ -0,0 +1,22 @@
...
+ 
+       top.in_what = Parm;
+--- aspell-0.60.6.1/prog/check_funs.cpp.orig	Tue May 16 14:06:49 2017
++++ aspell-0.60.6.1/prog/check_funs.cpp	Tue May 16 14:07:30 2017
+@@ -647,7 +647,7 @@
+     }
+   }
+   if (i == width-1) {
+-    if (word == '\0')
++    if (word[0] == '\0')
```
pyhalov, line 19: Do we really check for non-NULL or current symbol != '\0' here?

components/library/libdnet/patches/dnet.3.patch
alarcher: Because they showed as already applied when I did gmake prep.

alarcher: No I misremembered, the guy actually changed the content of the tarball without changing the version so the patches do not apply anymore.

alarcher: Re-added.

components/text/aspell/patches/01-gcc7.patch
alarcher: I kind of expected that the person would not use \0 as a null pointer but now I see that it might have been the case.

PR #3822, opened 2017-12-04
ogl-select should require diagnostic/constype

Mno-hime:
`ogl-select` script uses `constype` binary, but the package does not require `diagnostic/constype` component:
```
components/x11/ogl-select/files/ogl-select:101:DRIVER="$(/usr/bin/constype)"
```
This is visible when VirtualBox Guest Additions on a system from Minimal medium is installed.

constype

### DIFF
--- a/components/x11/ogl-select/ogl-select.p5m
+++ b/components/x11/ogl-select/ogl-select.p5m
@@ -21,6 +21,8 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+depend fmri=diagnostic/constype type=require
+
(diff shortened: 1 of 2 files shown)


components/x11/ogl-select/ogl-select.p5m
```diff
@@ -21,6 +21,8 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+depend fmri=diagnostic/constype
```
pyhalov: depend fmri=diagnostic/constype type=require

pyhalov (changes requested):

components/x11/ogl-select/ogl-select.p5m
Mno-hime: Updated.

PR #3825, opened 2017-12-05
xscreensaver: drop trusted extensions support

pyhalov:


alarcher:
@pyhalov Peter agreed to revert his commit in illumos, not sure if you want to remove the patch temporarily.

pyhalov:
The patch was dropped upstream. Trusted multi-labeled desktop on OI was never tested and I suppose is not working. So I'd dropped any trusted desktop remnants which we have (as it's mostly dead and non-working code).

alarcher:
@pyhalov So I guess my work on the patches was a waste of time then ... :S

PR #3828, opened 2017-12-06
libfuse: fix wrong options and paths for mount/unmount

grueni:
The paths for mount and umount were wrong.
And some options for the commands are not available in Illumos.

Dependent packages to recompile:
- [x] network/sshfs
- [x] system/file-system/ntfsprogs
- [x] system/virtualization/open-vm-tools
marked packages are recompiled and tested

### DIFF
--- a/components/library/libfuse/patches/mount_util.c.patch
+++ b/components/library/libfuse/patches/mount_util.c.patch
@@ -1,5 +1,9 @@
---- libfuse/mount_util.c.ref	2010-06-15 16:46:54.000000000 +0200
-+++ libfuse/mount_util.c	2016-05-03 21:36:52.782854200 +0200
+
+Options -i, -l do not exist in Illumos umount
+Options -i, -f do not exist or differ in meaning in Illumos mount
+ 
+--- libfuse/mount_util.c.orig	2017-12-05 22:13:00.030067695 +0000
++++ libfuse/mount_util.c	2017-12-05 22:14:22.818357506 +0000
 @@ -58,6 +58,7 @@
          return -1;
      }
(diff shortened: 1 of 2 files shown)


pyhalov:
Why do you need to recompile packages?

grueni:
The packages were changed (newer versions, bug fixes).

components/library/libfuse/patches/mount_util.c.patch
```diff
@@ -1,25 +1,31 @@
...
+ 
+@@ -79,9 +80,9 @@
              exit(1);
          }
          rmdir(tmp);
 -        execl("/bin/mount", "/bin/mount", "-i", "-f", "-t", type, "-o", opts,
 -              fsname, mnt, NULL);
-+        execle("/bin/mount", "/bin/mount", "-i", "-f", "-t", type, "-o", opts,
+-        fprintf(stderr, "%s: failed to execute /bin/mount: %s\n", progname,
++        execle("/sbin/mount", "/sbin/mount", "-t", type, "-o", opts,
```
xen0l, line 18: why change the execl to excele?

components/library/libfuse/patches/mount_util.c.patch
```diff
@@ -28,13 +34,15 @@
          setuid(geteuid());
 -        execl("/bin/umount", "/bin/umount", "-i", mnt, lazy ? "-l" : NULL,
 -              NULL);
+-        fprintf(stderr, "%s: failed to execute /bin/umount: %s\n", progname,
 +        if (lazy) {
-+            execle("/bin/umount", "/bin/umount", "-i", mnt, "-l",
++            execle("/sbin/umount", "/sbin/umount", "-a", mnt,
```
xen0l: If the action is the same in both cases, why do we care about **lazy**?

xen0l (changes requested):

components/library/libfuse/patches/mount_util.c.patch
grueni: execle allows to specify the environment in the invocation of the command

components/library/libfuse/patches/mount_util.c.patch
grueni: I was too lazy to clean up the lazy condition.

xen0l (approved):

PR #3831, opened 2017-12-06
open-vm-tools: fix vmhgfs-fuse

grueni:
Changed to support libfuse 2.7.6.
Recompiled with libfuse

Tested in VMware with a VM running under Workstation 12.5.7 with updated open-vm-tools und driver/fuse 

```bash
# show the name of the share defined in VM Workstation
# mount the share via fuse
# list the content
# unmount the share
root@oi-p3:~# vmware-hgfsclient
shared
root@oi-p3:~# vmhgfs-fuse -o allow_other /mnt/hgfs
root@oi-p3:~# ls -la /mnt/hgfs/shared/
total 10
drwxrwxrwx   1 root     root           0 Dec  6 10:58 .
dr-xr-xr-x   1 root     root        4192 Dec  6 17:28 ..
-rwxrwxrwx   1 root     root          22 Dec  6 10:58 a.txt
root@oi-p3:~# umount /mnt/hgfs/
```

### DIFF
--- a/components/sysutils/open-vm-tools/Makefile
+++ b/components/sysutils/open-vm-tools/Makefile
@@ -17,6 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		open-vm-tools
 COMPONENT_VERSION=	10.1.15
+COMPONENT_REVISION= 1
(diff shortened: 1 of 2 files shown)


xen0l (approved):
Thanks for working on this. Good job!

PR #3837, opened 2017-12-08
flac: enable C++ libs

pyhalov:


alarcher (approved):
Built, published with gcc-6 and all tests passed.

PR #3847, opened 2017-12-12
Add audacity 2.1.1

pyhalov:


pyhalov:
Based on FreeBSD port with some patches borrowed from pkgsrc.

pyhalov:
Depends on https://github.com/OpenIndiana/oi-userland/pull/3846

pyhalov:
Works for me. Could create mp3 file with it.

PR #3852, opened 2017-12-13
pidgin: build with gstreamer 1.0

alarcher:


pyhalov:
Have you tried to remove patches/03-hg-remove-SIGCHLD-handler.patch ?
It solved issue with unhandled signal soon after startup. If it's still necessary, pidgin will die in about one minute after startup.

alarcher:
@pyhalov Still required.

PR #3856, opened 2017-12-20
Totem

pyhalov:


pyhalov:
The main motivation for this update is to avoid totem dependency on libunique. We could try updating totem to more recent version, but last time I've tried, I found the following: they depend on clutter now for displaying video, and this doesn't work well when system doesn't have OpenGL acceleration support (like in virtualbox environment). Perhaps, have to retest.

pyhalov:
totem scrollbar doesn't work...

PR #3861, opened 2017-12-21
Gnome2

pyhalov:


pyhalov:
I'm starting to drop Gnome 2 applications which stand in my way while removing old libraries

PR #3867, opened 2018-01-10
openjdk-8: update to 1.8.152

prakashsurya:
This is my first attempt at contributing to oi-userland, so apologies if
I'm doing anything incorrectly. Further, I haven't tested this yet, but
I wanted to open the PR in hopes of getting some feedback in case I'm
doing something wrong (which could invalidate any testing that I do).

The changes to the `Makefile` are straightforward so I won't spend time
explaining those.

The changes to the `patches` directory were generated by replacing the
existing `patches` directory, with the `patches` directory from Joyent's
pkgsrc repository. For example:

    $ git clone https://github.com/OpenIndiana/oi-userland.git ~/oi-userland
    $ git clone https://github.com/joyent/pkgsrc.git ~/pkgsrc

    $ cd ~/pkgsrc
    $ git checkout 91fc60ec7ee226677915690b7dabe15eb43e235d

    $ cd ~/oi-userland/components/runtime/openjdk-8
    $ rm -r patches
    $ cp -r ~/pkgsrc/lang/openjdk8/patches .
    $ find patches -type f | sed -e "p;s/$/.patch/" | xargs -n2 mv

I don't know if this was the correct thing to do, so please let me know
if I should be updating the `patches` directory some other way.

If what I've done so far is correct, I'll look into testing this by
building and installing the new package, and then performaing a full
nightly build of illumos-gate with the new OpenJDK installed.


pyhalov:
You shouldn't have replaced all patches directory. We and pkgsrc have a bit different openjdk layout and some patches were modified, some added. I'd just replaced the patches which were touched with last openjdk update. If everything is fine, you should see that pkgsrc update commit and our commit are mostly identical.

prakashsurya:
@pyhalov does this look better?

pyhalov:
Yes, seems fine.

pyhalov:
Have you checked that 
- [ ] component builds
- [ ] no new files appeared (do gmake sample-manifest && compare results with git diff )
- [ ] new openjdk works as expected on simple hello worlds
- [ ] illumos-gate compiles with it
?

prakashsurya:
I ran into a build failure when building this package last week. I'll try and pick this up again soon; got pulled off to work on something else.

alarcher:
@prakashsurya Do you know if this version has gcc 6 friendly patches applied?

alarcher:
@prakashsurya @pyhalov Since this version has gcc 6 patches applied I have built it, published it to the gcc-6 repository, rebuilt dependent components like tomcat, and just built illumos-gate in the same zone.

prakashsurya:
Sorry for the delays. Do I still need to do more testing, or is what @alarcher has done sufficient?

pyhalov:
Was merged as part of #3014

PR #3882, opened 2018-01-18
git: depend on GNU gettext at runtime, for shell scripts

richlowe:
I added this to both `REQUIRED_PACKAGES` and the manifest directly, since `REQUIRED_PACKAGES` seems to not actually work.  That may just mean I'm missing something about how oi-userland and other userlands differ.

### DIFF
--- a/components/developer/git/Makefile
+++ b/components/developer/git/Makefile
@@ -24,6 +24,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		git
 COMPONENT_VERSION=	2.14.2
+COMPONENT_REVISION=	1
 COMPONENT_PROJECT_URL=	https://git-scm.com/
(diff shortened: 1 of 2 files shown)


pyhalov:
1) REQUIRED_PACKAGES are compile-time dependencies, no need to update them, if you just add dependency to p5m file. However, you can use something like `depend fmri=__TBD pkg.debug.depend.file=usr/gnu/bin/gettext type=require`, in this case it is necessary to add package to REQUIRED_PACKAGES.
2) Please, bump COMPONENT_REVISION and move dependency right after the license (separated from other sections by empty lines) - look at https://wiki.openindiana.org/oi/oi-userland+-+best+practices (yes, I wish we used pkgfmt, but current pkgfmt is unreadable).

richlowe:
Made the changes you asked for (I hope).  

The REQUIRED_PACKAGES thing was prior to me finding #3883 was also necessary :)

PR #3891, opened 2018-01-22
graphviz: add missing binaries for ocaml 32bit

grueni:
Sorry, erroneously deleted.

### DIFF
--- a/components/image/graphviz/graphviz-ocaml.p5m
+++ b/components/image/graphviz/graphviz-ocaml.p5m
@@ -40,6 +40,11 @@ file path=usr/lib/$(MACH64)/graphviz/ocaml/gv.cma
 file path=usr/lib/$(MACH64)/graphviz/ocaml/gv.cmi
 file path=usr/lib/$(MACH64)/graphviz/ocaml/gv.cmo
 file path=usr/lib/$(MACH64)/graphviz/ocaml/libgv_ocaml.so
+file path=usr/lib/graphviz/ocaml/META.gv
+file path=usr/lib/graphviz/ocaml/gv.cma
+file path=usr/lib/graphviz/ocaml/gv.cmi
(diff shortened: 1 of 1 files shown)


pyhalov:
We don't need 32-bit ocaml module, ocaml will not be able to load it. Even now ocaml graphviz extension is close to useless.
1) it requires findlib to load (if I understand google correctly);
2) it needs better packaging.

So far the following partially worked for me:
```
ln -s /usr/lib/amd64/graphviz/ocaml/libgv_ocaml.so /usr/lib/amd64/ocaml/stublibs/dllgv.so
mkdir /usr/lib/amd64/ocaml/site-lib/gv 
ln -s  /usr/lib/amd64/graphviz/ocaml/META.gv  /usr/lib/amd64/ocaml/site-lib/gv/META
mkdir /usr/lib/amd64/graphviz/ocaml/gv
ln -s /usr/lib/amd64/graphviz/ocaml/gv.cma /usr/lib/amd64/graphviz/ocaml/gv
ln -s /usr/lib/amd64/graphviz/ocaml/gv.cmi /usr/lib/amd64/graphviz/ocaml/gv
ln -s /usr/lib/amd64/graphviz/ocaml/gv.cmo /usr/lib/amd64/graphviz/ocaml/gv
```
After this (and installing findlib) I could load ocaml module with
```
#use "topfind"
#require  "gv"
```
But I still don't know how to use it. 
Perhaps, we should add findlib and above symlinks to graphviz-ocaml or just drop it as something which nobody uses.

pyhalov:
We don't need 32-bit binaries, so I close it. The next time I step on graphviz-ocaml, I'll just drop it as totally useless on OI.

grueni:
Same problems with the 64bit build [here](https://github.com/ellson/MOTHBALLED-graphviz/issues/1262).

pyhalov:
Well, updating fixed this.

PR #3900, opened 2018-01-25
guilt: update to 0.36

pyhalov:


pyhalov:
Used patches from oi-build, most of Debian patches and had to add new patches for tests to pass.

jeffpc (approved):
LGTM

PR #3902, opened 2018-01-25
coreutils: cp -a fails on /tmp

pyhalov:


pyhalov:
"timeout 9223372036854775808d sleep 0 && echo 1" seems to fail sporadically, this is why test result is changed.

PR #3906, opened 2018-01-26
tools: avoid publishing binaries with /usr/lib or /usr/lib/amd64 RUNPATH

pyhalov:


pyhalov:
@alarcher specially for you :)

alarcher:
@pyhalov Thank you :)

pyhalov:
@alarcher need to know how much is broken with about hack to configure.mk

PR #3943, opened 2018-01-31
libtool and pkg-config should skip linker default paths 

normjacobs:
The first changeset is unrelated, but addressed the case where the build overwrites log files from component directories that end in the same name.
The next changeset patches libtool so that it drops linker default paths instead of adding them to R*PATH
Then there is a pkglint check for linker default paths in RUNPATH
A userland-mangler change to strip linker default paths from R*PATH
Finally, two changesets to make sure pkg-config has all of the linker default paths when configured and to patch it to drop -R, -rpath, and -Wl, variants for --libs where the path is in the linker defaults.

A full build with and without the new libtool and pkg-config installed shows that fixing libtool and pkg-config addresses roughly half of the R*PATH issues across the gate.  I didn't go back and autoreconf things that weren't addressed, but I expect that would fix some more.  The userland-mangler change fixes things that weren't addressed in the component builds.  It emits a message to stderr (so in the log) letting you know when it makes R*PATH adjustments.

### DIFF
--- a/components/Makefile
+++ b/components/Makefile
@@ -64,7 +64,7 @@ clobber:		TARGET = clobber
 test:			TARGET = test
 component-hook:		TARGET = component-hook
 prep build install pre-publish publish test:	TEMPLATE_ZONE=$(ZONE)
-prep build install pre-publish publish test:	LOG = >$(WS_LOGS)/$(@F).$(TARGET).log 2>&1
+prep build install pre-publish publish test:	LOG = >$(WS_LOGS)/$(subst /,.,$@).$(TARGET).log 2>&1
 
 # turn off pkglint for the individual component builds.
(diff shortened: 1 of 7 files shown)


pyhalov:
Thanks. One style issue - bump COMPONENT_REVISION in components which you change.

alarcher:
@jacobs Thanks for working on this! I have rebuilt the entire oi-userland with the initial patch to userland-mangler that you shared.
@pyhalov Provided that this is the same as I have tested +1 for me. I guess @xen0l will comment further.

normjacobs:
I added the libtool and pkg-config COMPONENT_REVISION Makefile changes requested and pushed the recommitted changesets up to my github repo.

pyhalov (approved):

xen0l (approved):

PR #3946, opened 2018-01-31
augeas: use gcc 6

pyhalov:
This is not elegant , but works. 
The real issue is that ruby needs fixing. 
It sets LIBRUBYARG_SHARED in ruby*.pc:
LIBRUBYARG_SHARED=-Wl,-rpath,${libdir} -L${libdir} -l${RUBY_SO_NAME}
So we need LD_OPTIONS to overwrite it.
It sets CC=/usr/bin/gcc in /usr/ruby/2.X/lib/ruby/2.2.0/i386-solaris2.11/rbconfig.rb , 
so we need to overwrite it.

### DIFF
--- a/components/ruby/augeas/Makefile
+++ b/components/ruby/augeas/Makefile
@@ -45,6 +45,8 @@ COMPONENT_LICENSE_FILE=	augeas.license
 
 TPNO=                   21871
 
+GCC_VERSION = 6
+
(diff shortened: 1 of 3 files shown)


pyhalov:
@alarcher: perhaps you could pick it up for gcc-6 branch, perhaps we should patch ruby to add rpath to GCC_ROOT/lib and use gcc 6 as default compiler

PR #3950, opened 2018-02-05
Bump gcc-7 to 7.3.0

alarcher:


components/developer/gcc-7/patches/0002-compare_tests-Use-nawk-on-Solaris-since-awk-is-crap.patch
```diff
@@ -0,0 +1,34 @@
+From fec0b212a284965982bb1f2796a1c962cc7d1539 Mon Sep 17 00:00:00 2001
```
pyhalov, line 1: Not that this patch matters much with PATH=$(PATH.gnu), but it supposes REQUIRED_PACKAGES += system/extended-system-utilities

PR #3976, opened 2018-02-13
atril: deliver as 64-bit

alarcher:


components/desktop/mate/atril/Makefile
```diff
@@ -41,17 +42,17 @@ COMPONENT_PREP_ACTION=  cd $(@D)  && NOCONFIGURE=1 ./autogen.sh
 CONFIGURE_BINDIR.64=	$(CONFIGURE_BINDIR.32)
 
 CONFIGURE_OPTIONS+=	--sysconfdir=/etc
-CONFIGURE_OPTIONS+=	--libexecdir=/usr/lib/mate
+CONFIGURE_OPTIONS+=	--libexecdir=$(CONFIGURE_LIBDIR.$(BITS))/mate
```
pyhalov, line 45: Should we change it? I'm not sure. For example, we have 64-bit /usr/lib/git-core/*

alarcher: I would keep the lib paths consistent with the arch but that just me. Anyway I just marked the PRs as pending so we can come back to them when we have a better idea of what we want to do. No hurry :)

PR #3982, opened 2018-02-13
mate-session-manager: deliver as 64-bit only

alarcher:


components/desktop/mate/mate-session-manager/mate-session-manager.p5m
```diff
@@ -27,17 +27,13 @@ depend type=require fmri=desktop/mate/marco
...
-file path=usr/bin/$(MACH64)/mate-session-properties
-file path=usr/bin/$(MACH64)/mate-session-save
-file path=usr/bin/$(MACH64)/mate-wm
 file path=usr/bin/mate-session
 file path=usr/bin/mate-session-inhibit
 file path=usr/bin/mate-session-properties
 file path=usr/bin/mate-session-save
 file path=usr/bin/mate-wm
 file path=usr/share/applications/mate-session-properties.desktop
+#file path=usr/share/doc/mate-session-manager/dbus/mate-session.html
```
pyhalov: Why is it commented?

alarcher: It was not delivered so I thought there was a good reason for that. I wanted to make it visible: if things are deliberately not shipped there should be a sign that it was on purpose and not just an overlook.

pyhalov: I suppose, in this case it was not on purpose.

alarcher: I can add it then :)

PR #3989, opened 2018-02-13
Deliver more mate apps as 64-bit

alarcher:


alarcher:
@pyhalov I use now MATE with all the apps as 64-bit binaries except mozo. Need to check if all the applets like it but so far so good.

components/desktop/mate/mate-menus/Makefile
```diff
@@ -19,6 +19,7 @@ COMPONENT_NAME=		mate-menus
 COMPONENT_MJR_VERSION=	1.18
 COMPONENT_MNR_VERSION=	0
 COMPONENT_VERSION=	$(COMPONENT_MJR_VERSION).$(COMPONENT_MNR_VERSION)
+COMPONENT_REVISION=     1
```
pyhalov: mate-menu is a library, likely we want to deliver both 32 and 64 bit

PR #3990, opened 2018-02-14
GCC7 fixes

alarcher:
Could probably merge these fixes to prepare for gcc-7/gcc-8.
Double checking the suggested modification would be nice.


components/developer/golang-18/Makefile
```diff
@@ -76,6 +77,7 @@ COMPONENT_ARCHIVE_URL_1=	https://storage.googleapis.com/golang/$(COMPONENT_ARCHI
 GOROOT_FINAL=		/usr/lib/golang/$(COMPONENT_MAJOR_VERSION)
 # Enable easier stack unwinding for debugging.
 GOEXPERIMENT=		framepointer
+COMPONENT_BUILD_ENV +=	CFLAGS="-std=gnu89 $(CFLAGS)"
```
Toasterson, line 80: This Verion of Go should only use a previous version of Go to compile itself and does not need gcc bootstrap. Was there something left of golang1.4 bootstrap?

Toasterson: A wait sorry I did not update golang 1.8

1.10 does not use bootstrap anymore this could also be changed in this version of golang if we still want that version when switching to gcc7/8 however I would suggest dropping 1.8 in the near future from userland and then rotate it out of the repos during standard rotation

PR #3994, opened 2018-02-14
enchant: 64-bit version should not look at 32-bit modules

alarcher:
Using pluma as 64-bit I found out that enchant tries to load 32-bit modules.
There is a piece of code that unconditionally appends /usr/lib/enchant to the list of module paths.
Since ENCHANT_GLOBAL_MODULE_DIR is already specified there is no need to do so.

I should mention that in the 32-bit case the module directory was appended twice.


pyhalov:
Seems to be a correct fix. However, why do we need to introduce libexecdir, when we can use libdir for this purpose?

PR #4001, opened 2018-02-17
glib2: update to 2.55.2

pyhalov:

PR #4007, opened 2018-02-19
glib: update to 2.54.3

pyhalov:


pyhalov:
Replaces #4001 , as glib 2.55 is development version

PR #4009, opened 2018-02-21
Add PREFERRED_BITS

alarcher:


alarcher:
@xen0l @pyhalov Any comment?

pyhalov:
Should we also import common.mk and switch components to it one-by one?

alarcher:
That would make sense.
Should we merge this one and do it in a second stage as they are not coupled?
I am away for the week so cannot do anything on it in the coming days.

PR #4018, opened 2018-03-01
update to goffice-0.10.38 and ship 64 bit only

ghost:
The math tests fail for 32 bit builds but not 64 bit builds and since gnumeric, mate and others are shipping 64 bit I just built it that way. Could not find another package using goffice. I could have missed one?


components/library/goffice/Makefile
```diff
@@ -44,19 +44,22 @@ COMPONENT_PREP_ACTION =        ( cd $(@D)  && \
...
 CONFIGURE_ENV+= PYTHON=$(PYTHON)
 
 COMPONENT_INSTALL_ENV+= GCONF_DISABLE_MAKEFILE_SCHEMA_INSTALL=1
 
 # common targets
-build:		$(BUILD_32_and_64)
+build:		$(BUILD_64)
 
-install:	$(INSTALL_32_and_64)
+install:	$(INSTALL_64)
```
pyhalov: It's a library. It was delivered for 32 bits and I think we should continue to deliver it for both architectures, unless there are strong reasons to avoid delivering 32-bit version.

ghost:
Will work on it, no power for days.

components/library/goffice/Makefile
alarcher: I doubt there is any consumer outside gnumeric and gnucash... since 32-bit tests fail not sure it is worth spending time on it.

ghost:
I am working with 32 bit test now. Breaking it down function by function.

ghost:
I think the problem is with return values with func go_cotpil

Some improvements already made by upstream developer.

https://github.com/GNOME/goffice/commit/ca0a1c01e1e890e912ac62d77c513d82d933754a

ghost:
Have not solved this, even with upstream help.

Maybe, my lack of C skills and understanding what I see.

1) Ship 32 and 64 library with No_Test?

2) Leave as is?

components/library/goffice/Makefile
```diff
@@ -56,7 +58,8 @@ install:	$(INSTALL_32_and_64)
 # To prevent "libtool_install_magic: unbound variable"
 unexport SHELLOPTS
 
-test:		$(TEST_32_and_64)
+# TEST_32 fails test-math
+test:		$(TEST_64)
```
pyhalov: preserve $(TEST_32_and_64) and this comment

ghost:
Updating to goffice version 0.10.39 now.

pyhalov:
Why did you close it ?

ghost:
Trying to update both to latest release (yesterdays) but now I am having
problems with latest gnumeric.
Some of the tools core while testing.

PR #4019, opened 2018-03-01
update to gnumeric-1.12.38

ghost:
valgrind tests still fail
1 python test fails but I think it expected .pyc file.
Executable worked for me with python functions and python console.


ghost:
Updating to Gnumeric version 1.12.39

PR #4022, opened 2018-03-04
Add cpuid 1.6.3

alarcher:


components/sysutils/cpuid/cpuid.p5m
```diff
@@ -0,0 +1,25 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/cpuid
```
pyhalov, line 25: it would be good to import some man page

alarcher: There is none, we need to write our own. The Debian utility is another cpuid.

AndWac:
What shall we do with this PR?
There is no man page and nobody seems to care enough to create one.
Shall we merge or close it?

pyhalov:
I'd merged it

PR #4024, opened 2018-03-05
Added: SASL Compile options to postfix

Toasterson:


components/mail/postfix/Makefile
```diff
@@ -38,11 +40,13 @@ CCARGS+=	-DDEF_NEWALIAS_PATH=\\\"/usr/sbin/newaliases\\\"
 CCARGS+=	-DUSE_TLS
 CCARGS+=	-DHAS_DB
 CCARGS+=	-DHAS_SQLITE
+CCARGS+=	-DUSE_SASL_AUTH -DUSE_CYRUS_SASL -I/usr/include/sasl
 
 AUXLIBS=	-lssl
 AUXLIBS+=	-lcrypto
 AUXLIBS+=	-ldb-5.3
 AUXLIBS+=	-lsqlite3
+AUXLIBS+=	-L/usr/lib/sasl -lsasl
```
pyhalov: do we need corresponding -R flag?

Toasterson: Not as far as I know. CentOS/Fedora does not have a -R Flag. AFAIK many linuxes don't

pyhalov: Does it work without -L ?

Toasterson: Yep it works without -L. Don't know why the linuxes include it in that case good catch.

components/mail/postfix/postfix.p5m
```diff
@@ -26,6 +26,10 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+<transform file -> add pkg.depend.bypass-generate libsasl\.so\.1>
+
+depend fmri=system/library/security/libsasl type=require
```
pyhalov: why is it so complex?

Toasterson: No clue why but pkg seems not to like libsasl. It can not identify libsasl as a dependency thus it needs to be bypassed during the depend recognition run and then added manually. This is how I saw other packages solved that particular issue.

pyhalov: it seems pkgdepend finds dependency on libsasl fine here without this.

PR #4025, opened 2018-03-05
9223: bump mate-media to 1.18.2

alarcher:
https://www.illumos.org/issues/9223


alarcher:
@pyhalov This is to fix a bug found in 1.18.1, not icon for the volume applet.

PR #4034, opened 2018-03-08
ffmpeg: bump to 3.2.14

alarcher:


pyhalov:
Why is it marked as "pending"? Does current vlc or audacity strictly require ffmpeg 2?

alarcher:
Yes, vlc 2.x requires ffmpeg 2 although some distros hack around to support more recent versions.
I did not have time to follow up on that given the priority on gcc.
I have used vlc 3.x for some time before moving to the gcc-8 repository, so it would be possible to bump to latest.

alarcher:
@pyhalov This ffmpeg version works with 3.0.4

Mno-hime:
I don't see accelerated H.264 on Ivy Bridge laptop, see https://github.com/OpenIndiana/oi-userland/pull/4035#issuecomment-513556751.

Mno-hime:
Added H.265 support, requires https://github.com/OpenIndiana/oi-userland/pull/5150 to be merged first.

PR #4039, opened 2018-03-11
ncdu: Add new package

hliebel:


AndWac (commented):
Remove the COMPONENT_REVISION because this will be the first release of the package under the version 1.13. Setting (or increasing) the COMPONENT_REVISION is only needed if you provide changes to the installation without changing the COMPONENT_VERSION in order to make it distinquishable from older revsions with the same version.

components/sysutils/ncdu/Makefile
```diff
@@ -0,0 +1,52 @@
...
+COMPONENT_LICENSE_FILE=	COPYING
+COMPONENT_LICENSE=	MIT
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CFLAGS +=	-I/usr/include/ncurses
+
+CONFIGURE_BINDIR.64 = $(CONFIGURE_BINDIR.32)
```
alarcher: You can use PREFERRED_BITS=64 instead of Line 39. Place it just before "include ../../../make-rules/shared-macros.mk"

hliebel:
Hi.
I made the changes, did a 'git amend' and 'force pushed' to my branch again. It that what I'm supposed to do or does that break the flow?

Do you get notified automatically if I push an update or am I supposed to write a comment?

Thanks,
Harry

components/sysutils/ncdu/Makefile
```diff
@@ -0,0 +1,51 @@
...
+COMPONENT_LICENSE=	MIT
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CFLAGS +=	-I/usr/include/ncurses
+
+# Enable Address Space Layout Randomization
+ASLR_MODE = $(ASLR_ENABLE)
```
pyhalov: This directive is no-op on illumos, you can remove it.

PR #4053, opened 2018-03-17
libvncserver: Add new package

hliebel:


components/library/libvncserver/Makefile
```diff
@@ -0,0 +1,60 @@
...
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PREP_ACTION = \
+	( cd $(@D) && autoreconf -fiv )
+
+# Configure does not find the correct JPEG library, use 64bit pkg-config
+PKG_CONFIG +=	/usr/bin/$(MACH64)/pkg-config
```
alarcher: To avoid references to a particular JPEG package we defined JPEG_* macros in shared-macros.mk.See https://wiki.openindiana.org/oi/JPEG+8

alarcher:
@hliebel The package does not publish in my environment due to missing library/gnutls in REQUIRED_PACKAGES, you should add it manually before auto-generated deps.

hliebel:
@alarcher 
Thanks, will do.

How do you normally catch all the required packages? Do you just create a brand new minimal zone and build from there or do you go through all the 'ldd' output libs or? Is there a way for me to setup a local build bot or Jenkins to catch these? I would like to add a few more packages so would like to know best practice to get it right. Is there some documentation I can look at?

Regards,
Harry

alarcher (approved):

PR #4054, opened 2018-03-17
x11vnc: Add new package

hliebel:
This depends on the pull request I just submitted for libvncserver.

### DIFF
--- a/components/x11/x11vnc/Makefile
+++ b/components/x11/x11vnc/Makefile
@@ -1,0 +1,62 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2018 Harry Liebel
+#
+
+PREFERRED_BITS=64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		x11vnc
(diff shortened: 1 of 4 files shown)


alarcher:
@hliebel I am not familiar with this package, how does it differ from TigerVNC? Should we document the alternative somewhere?

3eka:
@alarcher,

this is not another VNC package: this one allows connection to already running GUI session.
It works just fine (i used to use it in the past).

Regards.

hliebel:
Hi.
I've been using it on a headless machines with the X framebuffer server (Xvfb) to run a GUI app. When looking online on how to run a GUI app on a persistent X session Xvfb and X11vnc are often mentioned as a good solution. Xvfb is used to run the headless X session and X11vnc allows you to connect to it as needed from a remote PC using a VNC client. Perhaps this is possible from TigerVNC, but is seems to want to spawn its own X sessions..

Actually, looking around to find a better answer for you I found this on the Arch wiki:
https://wiki.archlinux.org/index.php/x11vnc
-> x11vnc is not the only way to accomplish a remote control of the real X display; x0vncserver is part of tigervnc and allows for the same operation.

So looks like there can be some overlap..

Thanks

hliebel:
I just tried x0vncserver as it is supposed to do the same as x11vnc, but it crashes with a SEGV and dumps core when I try to connect to it from a remote machine. Sorry, I dont yet know enough to debug it, best I can do is try to build another version..

hliebel:
Hi.
Any news on this one or the VNC library it depends on?
Thanks.

alarcher:
@hliebel Once libvncserver is fixed we can merge. Make sure flags are explicitly set to avoid different dependencies requirements depending on the environment.

PR #4056, opened 2018-03-17
xeyes update to 1.1.2

alarcher:


make-rules/x11.mk
```diff
@@ -0,0 +1,90 @@
...
+# Define library install paths
+#
+
+XORG_LIBDIR.32 = $(USRLIBDIR)/xorg	
+XORG_LIBDIR.64 = $(USRLIBDIR)/xorg/$(MACH64)	
+XORG_LIBDIR    = $(XORG_LIBDIR.$(BITS))
+
+#
+# Common macros for font packages
+#
```
pyhalov: Sorry, I don't see their usage in our xorg components. Do we really need them now?

make-rules/x11.mk
```diff
@@ -0,0 +1,90 @@
...
+# CDDL HEADER START
```
pyhalov: Can you use new Makefile header?

make-rules/x11.mk
alarcher: The corresponding component is in another branch. I can split if you like.

make-rules/x11.mk
alarcher: OK

PR #4077, opened 2018-04-10
timgm6mb-soundfont: Add new package

hliebel:
This is an older file, but is small and seems to be kept around by people like Debian. Alternatives are much bigger.

PR #4078, opened 2018-04-10
timidity: Add new package

hliebel:
Midi player for game/openttd

### DIFF
--- a/components/multimedia/timidity++/patches/01-sys-soundcard.patch
+++ b/components/multimedia/timidity++/patches/01-sys-soundcard.patch
@@ -1,25 +1,21 @@
 diff -puNr TiMidity++-2.14.0.orig/timidity/midi_a.c TiMidity++-2.14.0/timidity/midi_a.c
 --- TiMidity++-2.14.0.orig/timidity/midi_a.c	2009-03-14 15:49:59.000000000 +0000
-+++ TiMidity++-2.14.0/timidity/midi_a.c	2018-04-10 07:35:49.743687930 +0000
-@@ -49,14 +49,14 @@
++++ TiMidity++-2.14.0/timidity/midi_a.c	2018-04-17 14:36:21.184520463 +0000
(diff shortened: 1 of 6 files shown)


alarcher:
The package gets an unresolved dependency on libao if it is installed on the system. You should either enable the option and update REQUIRED_PACKAGES, or disable it explicitly.

alarcher:
@hliebel Sorry for the delay, everyone seems pretty busy lately.

components/multimedia/timidity++/patches/01-sys-soundcard.patch
```diff
@@ -0,0 +1,25 @@
+diff -puNr TiMidity++-2.14.0.orig/timidity/midi_a.c TiMidity++-2.14.0/timidity/midi_a.c
+--- TiMidity++-2.14.0.orig/timidity/midi_a.c	2009-03-14 15:49:59.000000000 +0000
++++ TiMidity++-2.14.0/timidity/midi_a.c	2018-04-10 07:35:49.743687930 +0000
+@@ -49,14 +49,14 @@
+ #endif
+ #include <fcntl.h>
+ #include <stdlib.h>
+-#ifdef HAVE_SYS_SOUNDCARD_H
```
alarcher: Maybe guard it with a conditional on __sun to make it cleaner.

alarcher:
Depends on #4077

hliebel:
I have 'library/audio/libao' in the Makefile REQUIRED_PACKAGES already. When I do a 'gmake env-check' the library is listed as required. When I do a ldd on the binary it refers to the library provided by that package. If I loook at the raw manifest for this package it lists the libao library as a depend.
Not sure what to look at otherwise..?

alarcher:
Nevermind... I had library/libao installed from before I renamed it and added the package to the repository...

alarcher:
Did not see that there were 2 commits... sorry...

PR #4082, opened 2018-04-12
gnugo: Move to games section and make 64bit

hliebel:


alarcher (approved):
For future components, you do not need to add COMPONENT_LICENSE_FILE=	gnugo.license if it is set to $(COMPONENT_NAME).license

PR #4084, opened 2018-04-14
Minimal install: Add USB Ethernet driver

hliebel:
Hi.
I'm using a USB dongle with one of my machines as it has no supported network devices (I'm hoping to fix that in the future).
To get it online I have to pull the driver onto a memory stick from a different OI machine and install from a local repo. Can we add the driver to the installer please?
Thanks.


alarcher:
@pyhalov Added you as reviewer.

pyhalov (approved):

PR #4085, opened 2018-04-15
elinks: Move to web section and make 64bit only

hliebel:


alarcher (changes requested):
Need to investigate why the encoding is wrong with the 64-bit version. Accented characters cannot be displayed.

hliebel:
Which page did you use for your test?
I looked up some random site that looked ok:
-> elinks https://ascii.cl/es/
Compared to Firefox it just removes the accents for me, looks ok. I'm running a default US english install of OI, Do you use a different localisation or fonts?

alarcher:
@hliebel My locale is fr_FR.UTF-8

alarcher:
@hliebel i just tried to open the same page on Debian for comparison: elinks is not able to open it, so it seems our elinks is less broken ^ ^

alarcher:
Actually it is now completely broken with a bunch of websites...

pyhalov:
This has been already made.

alarcher:
@pyhalov The encodings on the 64-bit build were completely broken so I assume that the merged version has been tested and works correctly?

pyhalov:
@Mno-hime  ?

Mno-hime:
With `LC_ALL=de_DE.UTF-8` I see "tofu" in `F10` menus characters in both 32- and 64-bit `elinks`. When pointer cursor to a link it's somewhat doubling the text of the link with 32-bit elinks... So I guess it is broken either way?

PR #4101, opened 2018-04-18
freealut: Add new package

hliebel:


components/multimedia/freealut/freealut.p5m
```diff
@@ -0,0 +1,32 @@
...
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/freealut-config
+file path=usr/include/AL/alut.h
+file path=usr/lib/$(MACH64)/libalut.a
```
pyhalov: do we need to deliver static library?

components/multimedia/freealut/Makefile
```diff
@@ -0,0 +1,49 @@
...
+COMPONENT_ARCHIVE_URL= \
+  https://github.com/vancegroup/$(COMPONENT_NAME)/archive/$(COMPONENT_ARCHIVE)
+COMPONENT_LICENSE=	LGPLv2
+COMPONENT_CLASSIFICATION= System/Multimedia Libraries
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/cmake.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+build:		$(BUILD_64)
```
pyhalov: Why only 64-bit version is delivered? If we deliver library, we usually deliver both 32/64-bit versions.

alarcher (approved):

PR #4102, opened 2018-04-18
ftgl: Add new package

hliebel:


components/x11/ftgl/ftgl.p5m
```diff
@@ -0,0 +1,57 @@
...
+file path=usr/include/FTGL/FTGlyph.h
+file path=usr/include/FTGL/FTLayout.h
+file path=usr/include/FTGL/FTOutlineGlyph.h
+file path=usr/include/FTGL/FTPixmapGlyph.h
+file path=usr/include/FTGL/FTPoint.h
+file path=usr/include/FTGL/FTPolyGlyph.h
+file path=usr/include/FTGL/FTSimpleLayout.h
+file path=usr/include/FTGL/FTTextureGlyph.h
+file path=usr/include/FTGL/ftgl.h
+file path=usr/lib/$(MACH64)/libftgl.a
```
pyhalov: Why do we deliver static library?

components/x11/ftgl/Makefile
```diff
@@ -0,0 +1,56 @@
...
+COMPONENT_CLASSIFICATION= System/X11
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/configure.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+CONFIGURE_OPTIONS += --with-x
+CONFIGURE_ENV += LIBS="-lm"
+
+build:		$(BUILD_64)
```
pyhalov: Why do we build only 64-bit version?

hliebel: Hi.
I thought all apps were now preferred as 64bit. If so, then do we expect to build 32bit apps against this library in future?
Thanks

components/x11/ftgl/ftgl.p5m
hliebel: Its just a install product of the default library build. What if someone wants to statically link against this library when building something else? I see in /var/lib that a few other libraries do the same (like boost, sdl2, qt)..

components/x11/ftgl/Makefile
pyhalov: We are mostly providing both 32/64 bit libraries, as 32-bit applications are still supported.

components/x11/ftgl/ftgl.p5m
pyhalov: We shouldn't encourage user to do silly things (like static linking). Think about library updates. We deliver static variants of libraries when there's no dynamic version, but generally static libs shouldn't be shipped until strictly necessary.

PR #4120, opened 2018-04-27
kvm: import fix for OS-6610

pyhalov:


pyhalov:
Tested, fix works for me.

PR #4124, opened 2018-04-28
Updated Protobuf to 3.5.1

Toasterson:
Notes: I have no machine to test SPARC compilation thus removing the blocking Patch for Now. @alarcher If you think it is still needed feel free to add.

Also some Libraries protobuf now depends are still 32bit only thus 64bit compilation fails disabled for now.

I have successfully compiled libArcus with this


alarcher:
@Toasterson Any dependency that we should test rebuild with the new version?

Toasterson:
I have tested a new Package libArcus. I have no clue what else uses protobuf. Any idea how to find out what does?

alarcher:
@Toasterson 

helios> pkg search -r -o pkg.name "depend:require:library/c++/protobuf"
PKG.NAME
desktop/compiz/library/libcompizconfig
service/network/mosh

Toasterson:
All those three compile. Do you have any use case for one of them? Or do you think compiling is enough?

alarcher:
@Toasterson Should at least check that powerdns works as expected. Possibly just check that the testsuite is fine.

Toasterson:
There does not seem to be a testsuite and I have no idea/time to configure a power-dns recursor

PR #4125, opened 2018-04-28
Added: libArcus requires protobuf >= 3.0.0

Toasterson:


components/library/libArcus/Makefile
```diff
@@ -0,0 +1,43 @@
+# This file and its contents are supplied under the terms of the
```
pyhalov, line 2: Header misses upper line

components/library/libArcus/Makefile
```diff
@@ -0,0 +1,43 @@
...
+COMPONENT_LICENSE_FILE=LICENSE
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/cmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+CMAKE_OPTIONS += -DBUILD_PYTHON=OFF
+CMAKE_OPTIONS += -DCMAKE_CXX_STANDARD_LIBRARIES="-lsocket -lnsl" 
+
+build: $(BUILD_32)
```
pyhalov: Should we deliver also 64-bit library?

alarcher:
@Toasterson Next time you should look at dependent packages like the ones I listed for protobuf 3.5.1. Since the update is not ABI compatible, these components need rebuild. A PR like #4142 should be opened together with such incompatible update.

PR #4126, opened 2018-04-30
Introduce recipe for znapzend baseline version 0.18.0

jimklimov:


components/sysutils/znapzend/Makefile
```diff
@@ -0,0 +1,76 @@
...
+# Copyright 2016-2018 Jim Klimov
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=	znapzend
+COMPONENT_VERSION=	0.18.0
+COMPONENT_SUMMARY=	zfs backup with remote capabilities and mbuffer integration
+COMPONENT_DESCRIPTION=	ZnapZend is a ZFS centric backup tool. It relies on snapshot,\
+ send and receive todo its work. It has the built-in ability to manage both local\
```
pyhalov: to do?

components/sysutils/znapzend/Makefile
```diff
@@ -0,0 +1,76 @@
...
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_PREP_ACTION = ( cd $(SOURCE_DIR) && autoreconf --force --install --verbose --make )
+
+COMPONENT_PRE_CONFIGURE_ACTION =	( $(CLONEY) $(SOURCE_DIR) $(@D) )
+
+# NOTE: Code below assumes that BINDIR64 is one layer under BINDIR32
+# Othewise might user GNU ln -rsf ...
+COMPONENT_POST_INSTALL_ACTION =	( $(MKDIR) $(PROTO_DIR)/$(CONFIGURE_BINDIR.$(BITS)) \
```
pyhalov: Can we simplify it, delivering only 64-bit version?

components/sysutils/znapzend/Makefile
```diff
@@ -0,0 +1,76 @@
...
+        && for BINFILE in znapzend znapzendzetup znapzendztats ; do \
+            ln -sf "$$ORIGDIR/$$BINFILE" "$(PROTO_DIR)/$(CONFIGURE_BINDIR.$(BITS))/" || exit ; \
+        done )
+
+# NOTE: znapzend is a script, but it can pull some PERL modules and some may be
+# ARCHed, and/or these may (package-)conflict with dependencies of other pkgs.
+# TODO: package dependency modules as separate IPS requirements, and build only
+# a 32-bit znapzend? For now we work around the mess by having it all in a
+# dedicated directory per ARCH, including the perl modules.
+CONFIGURE_OPTIONS  =	--prefix=$(CONFIGURE_PREFIX)/share/$(COMPONENT_NAME)/$(MACH$(BITS))
```
pyhalov: @alarcher What do you think about this? Perhaps, we have to package perl modules separately?

components/sysutils/znapzend/Makefile
jimklimov: Yes i think. Copypasted intro ;p

components/sysutils/znapzend/Makefile
jimklimov: Probably, though this being a perl script (and perl module libs it pulls) the point may be moot - we do aim to deliver all variants of libs right?

This is related to the other comment: if dependencies are an internal detail of this component, we can build it whichever way. If it would depend on many perl-something packages, it may be a bit of a different story.

components/sysutils/znapzend/Makefile
jimklimov: There is benefit to making many packaged libs, though i'm not sure how znapzend would like them (especially as things would evolve independently, not how authors test it).

It is also complicated IMHO (as work and as an horde of PRs to merge before this one then), so can be refactored later ;p

pyhalov: Then we don't need two versions of libraries, they should be private interfaces, not usable by any external  tools.

jimklimov:
Interesting, now a `gmake publish` gives me this:

````
/usr/bin/pkgdepend resolve -e /export/home/jim/shared/oi-userland/components/sysutils/znapzend/build/resolve.deps -m /export/home/jim/shared/oi-userland/components/sysutils/znapzend/build/manifest-i386-znapzend.depend
/export/home/jim/shared/oi-userland/components/sysutils/znapzend/build/manifest-i386-znapzend.depend has unresolved dependency '
        pkg.debug.depend.reason=usr/share/znapzend/amd64/lib/i86pc-solaris-64int/auto/List/Util/Util.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.fullpath=lib/libssp.so.0 \
        pkg.debug.depend.fullpath=usr/gcc/6/lib/libssp.so.0'.
/export/home/jim/shared/oi-userland/components/sysutils/znapzend/build/manifest-i386-znapzend.depend has unresolved dependency '
        pkg.debug.depend.reason=usr/share/znapzend/i86/lib/i86pc-solaris-64int/auto/List/Util/Util.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.fullpath=lib/libssp.so.0 \
        pkg.debug.depend.fullpath=usr/gcc/6/lib/libssp.so.0'.
````

Adding `GCC_RUNTIME_PKG` to explicit build and packaging dependencies did not help... any ideas? :)

File: components/sysutils/znapzend/Makefile (hunk 1 above)
Comment (jimklimov) [reply components/sysutils/znapzend/Makefile]: fixed (push pending)

File: components/sysutils/znapzend/Makefile (hunk 2 above)
Comment (jimklimov) [reply components/sysutils/znapzend/Makefile]: fixed (push pending)

File: components/sysutils/znapzend/Makefile (hunk 3 above)
Comment (jimklimov) [reply components/sysutils/znapzend/Makefile]: so internal detail of the 32-bit-only build, at least for now (push pending)

File: components/sysutils/znapzend/Makefile (hunk 2 above)
Comment (jimklimov) [reply components/sysutils/znapzend/Makefile]: Actually for PERL in OI we only build 32-bit versions, as I found in `makemaker.mk` ;)

Comment (pyhalov) [reply components/sysutils/znapzend/Makefile]: As I see, BITS is set based on PERL_VERSION, we build 64-bit modules for 64-bit perl (5.24).

**Comment by pyhalov** [comment]
What RPATH and RUNPATH does Util.so have?

**Comment by jimklimov** [comment]
Looking at builddir, it has configs from perl, and ignores any attempts at exported env variables.
The offending feature is likely `-fstack-protector` (several variants) and `-shared`.

So far I overrode with generate-bypass... at least it passes the packaging now :\

So this adventure may be an incentive to (refactor) provide at least that perl lib as a package, built with `makemaker.mk` which probably avoids the hassle (I did not try).

File: components/sysutils/znapzend/Makefile [hunk 4]
```diff
@@ -39,38 +39,55 @@ include $(WS_MAKE_RULES)/prep.mk
...
+            | while read LIBFILE ; do \
+                $(CHMOD) +w "$$LIBFILE" \
+                && $(ELFEDIT) -e "dyn:rpath $(GCCLIBDIR)" "$$LIBFILE" \
+                && $(CHMOD) -w "$$LIBFILE" \
+                || exit ; \
+            done \
+        )
+
+# Enable ASLR for this component
+ASLR_MODE = $(ASLR_ENABLE)
```
pyhalov: ASLR_MODE oi-userland macros is no-op on OI

jimklimov: Fixed

jimklimov:
Bump? Is there any more required to improve?

components/sysutils/znapzend/Makefile
```diff
@@ -0,0 +1,90 @@
...
+CONFIGURE_OPTIONS +=	--with-svcname-smf=system/filesystem/zfs/znapzend
+
+build: $(BUILD_32)
+
+install: $(INSTALL_32)
+
+test: $(TEST_32)
+
+REQUIRED_PACKAGES += runtime/perl-522
+REQUIRED_PACKAGES += runtime/perl-524
```
pyhalov: Do I understand correctly that as we build only 32-bit version, we depend on perl-522, not -524? So second dependency can be removed.

pyhalov:
@jimklimov: Are you going to update it to 0.19.0 ?

PR #4133, opened 2018-05-01
Fixed: Now Compiling uuid extension aswell as delivering the docs

Toasterson:


pyhalov:
Should other PostgreSQL versions also be modified?

pyhalov:
Also, please, bump COMPONENT_REVISION

Toasterson:
Yeah that would be an idea. added to old postgresql versions aswell

PR #4136, opened 2018-05-04
ghostscript: update to 9.23

pyhalov:


pyhalov:
Likely ABI-incompatible - https://abi-laboratory.pro/tracker/timeline/ghostscript/ 
Dependent packages: 
```
pkg:/image/editor/gimp
pkg:/image/graphviz
pkg:/library/libspectre
pkg:/print/cups
pkg:/print/cups-filters 
```
Checked, that all of them build fine.

PR #4148, opened 2018-05-16
9497 Bump to cdrtools 3.02a09

Mno-hime:
Tested that `mkisofs` is able to create ISO.


pyhalov:
please, bump COMPONENT_REVISION, so that we could identify this package (as IPS_COMPONENT_VERSION is fixed for now)

Mno-hime:
`COMPONENT_REVISION` updated.

PR #4153, opened 2018-05-16
sbcl: update to 1.4.7

pyhalov:


pyhalov:
Fixes https://www.illumos.org/issues/9513 . But as real case for the issue seems to be https://smartos.org/bugview/OS-6906 , I'd like to postpone merging this until I check old sbcl and patched illumos-gate.

pyhalov:
It's not https://smartos.org/bugview/OS-6906, have to extract reproducer. sbcl 1.3.6 binaries are available here - http://buildzone.oi-build.r61.net/sbcl-affected/

PR #4157, opened 2018-05-18
Bump GNU coreutils to 8.29

Mno-hime:
Tested couple of updated utilities.


pyhalov:
There are new test failures, they require investigation/fixing
+FAIL: tests/misc/shred-remove.sh
+FAIL: tests/misc/timeout-parameters.sh

Mno-hime:
Fixed the former fail by an upstream patch. I don't see the latter fail in my (very new) zone. Can you share the fail, please?

pyhalov:
don't see second failure now. Perhaps, it's a big sporadic.

PR #4162, opened 2018-05-19
tigervnc: switch to 64-bit

pyhalov:


pyhalov:
Initial attempt to deliver only 64-bit Xvnc. So far have issues with linking wrong libjpeg version.

PR #4163, opened 2018-05-19
9516 Marco is compiled without Xinerama support, 1.18.3 bump

Mno-hime:
MATE marco is build without XFree Xinerama support, so maximized windows are spread
over both screens. The reason seems to be that `configure` is looking for Solarix
Xinerama on `*-*-solaris*` hosts. As we ship only XFree Xinerama, `configure` won't
bother finding it. Removing the check.

Added couple of build deps which I had to install manually.

Bumped marco to 1.18.3.

Tested that maximized windows work as expected when two screens are on.

Should I move `x11/library/libxinerama` to the auto-generated dependencies section? Some script after `gmake publish` warned about this dependency.

### DIFF
--- a/components/desktop/mate/marco/Makefile
+++ b/components/desktop/mate/marco/Makefile
@@ -17,15 +17,14 @@ include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		marco
 COMPONENT_MJR_VERSION=	1.18
-COMPONENT_MNR_VERSION=	1
-COMPONENT_REVISION=	2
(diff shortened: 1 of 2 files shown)


alarcher:
If gmake REQUIRED_PACKAGES lists x11/library/libxinerama as deps then yes, otherwise leave it under a separate block so that it is not removed by mistake. Thanks for working on this!

Mno-hime:
Thanks for the hint @alarcher! `REQUIRED_PACKAGES` target found it.

PR #4179, opened 2018-05-21
binutils 2.30

alarcher:
@pyhalov @richlowe @Mno-hime @tsoome Maybe time to have a closer look at binutils in the coming weeks?

I have used this to build gcc-8 and in the past weeks on my machines.

@normjacobs Thanks for preparing this update, I squashed recent CVEs from userland also. I left out the cross-compilation part (though I played with a gcc-6 arm cross-compiiler :)). Could you tell why  $(CPP_LARGEFILES) $(CPP_POSIX) is not needed anymore?


pyhalov:
As https://www.illumos.org/issues/6653 was fixed, I suppose this update to be more-or-less safe if properly tested. Has somebody used these binutils for some time?
@alarcher

alarcher:
@pyhalov The gcc-8 repository is built with this and andyf has used binutils 2.30 for a while on bloody.

normjacobs:
Sorry I didn't see this sooner.  I have been using this version of binutils several month now in a zone that I periodically build and test bits in.  As for the changes in the Makefile, I vaguely remember looking at the C*FLAGS and verifying that we didn't need to augment them when we build.

pyhalov (approved):

PR #4187, opened 2018-05-22
GCC 8.1

alarcher:


pyhalov:
Do we need something like https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/developer/gcc-6/patches/use-cxa_atexit.patch ?

alarcher:
@pyhalov Apparently not, since Solaris 12 was not released the file was modified as:
```
case ${target} in
    *-*-solaris2.1[1-9]*)
      # __cxa_atexit was introduced in Solaris 11.4.
      default_use_cxa_atexit=yes
      ;;
  esac
```
So I removed the patch.

PR #4188, opened 2018-05-22
GCC runtime macros

alarcher:
For future compiler migration we need to use the compiler runtime macros.
There is no functional change so avoiding mass rebuild would be better.


pyhalov:
Should we just change them or also sort  dependencies list (as gmake REQUIRED_PACKAGES) would sort them?

alarcher:
@pyhalov I think changing them is enough, people can modify them as components are updated. This is more a question of timing since any rebuild with gcc-7 or gcc-8 requires the macros to be in place right now. What do you think?

PR #4209, opened 2018-05-27
zziplib: CVE-2018-7726 CVE-2018-7725

Mno-hime:


alarcher:
These are already fixed in 0.13.69.
EDIT: OK I seee that SUSE developers reworked the patches with additional checks.

alarcher (approved):

PR #4213, opened 2018-05-27
pinentry 1.1.0 & 64-bit only

Mno-hime:


alarcher:
You should check that sysutils/gnupg is not broken with this.

Mno-hime:
With the updated pinentry I am able to sign commits in local zone via curses backend and in global zone via GTK+ backend. In both cases while entering passphrase pinentry took the terminal, or desktop, respectively, for itself.

PR #4224, opened 2018-05-30
pinentry: fix build to find the correct fltk-config

alarcher:


Mno-hime:
I am sorry for this complication. I don't have `library/fltk` in my zone but I guess we do on the server. Since we don't ship `usr/bin/pinentry-fltk` nor require `library/fltk` in Makefile, can we just `--disable-pinentry-fltk`?

PR #4257, opened 2018-06-06
xscreensaver: remove broken hacks

Mno-hime:
Added those hacks in 5.39 bump but they segfault and were not
distributed before.


alarcher:
Please bump COMPONENT_REVISION.

alarcher (approved):

PR #4258, opened 2018-06-06
SQLITE_ENABLE_COLUMN_METADATA

ghost:
Trying to build gnome-db with system sqlite instead of embedded one.

### DIFF
--- a/components/database/sqlite/Makefile
+++ b/components/database/sqlite/Makefile
@@ -25,6 +25,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME			= sqlite
 COMPONENT_VERSION		= 3.19.3
+COMPONENT_REVISION		= 1
 TARBALL_VERSION			= 3190300
 COMPONENT_SRC			= sqlite-autoconf-$(TARBALL_VERSION)
 COMPONENT_ARCHIVE		= $(COMPONENT_SRC).tar.gz
@@ -59,6 +60,7 @@ CFLAGS			+= -DHAVE_STATVFS
 CFLAGS			+= -DSQLITE_ENABLE_UNLOCK_NOTIFY
 CFLAGS			+= -DSQLITE_ENABLE_STAT2
 CFLAGS			+= -DSQLITE_ENABLE_STMT_SCANSTATUS
+CFLAGS			+= -DSQLITE_ENABLE_COLUMN_METADATA
 
 #Fixes firefox crash with new sqlite, don't remove until proper fix
 gcc_OPT                 =  -O2


AndWac (commented):
Please set the COMPONENT_REVISION.

AndWac (commented):
If firefox needs "-O2 -fno-schedule-insns2 -fno-lifetime-dse -fno-delete-null-pointer-checks"
do we need the same here?

alarcher:
That very much depends on the code here... The problematic part in Firefox is mainly the JS code that relies on undefined behaviour. My guess is to just run the test suite... Any segfault at object destruction (in particular at exit) hints that -fno-lifetime-dse may be required.

ghost:
sqlite test suite passed for me

alarcher:
Are we able to add support for the testsuite so that we can compare results are version bumps?

ghost:
That I am unsure about or beyond me right now.

alarcher:
Apparently there is no testsuite. I guess you should just install the new sqlite package on your system and check that Firefox + other dependencies are not broken.

ghost:
OK. Will do.

ghost:
I am happy, actually impressed due to the number of segfaults I get building things.
Open hundreds of sites I had bookmarked, somtimes as many as sixty at a time.
No chatter from firefox run in terminal window so I could see errors.
My sites are not facebook and the like though.

alarcher:
Out of curiosity, do you run the latest Firefox package that I pushed yesterday?

ghost:
Yes, I am normally right with you on updated stuff. If I remember, I update firefox twice in last day.
web/browser/firefox                               52.8.0-2018.0.0.2          i--
web/data/firefox-bookmarks                        52.8.0-2018.0.0.2          i--

alarcher:
Great thanks, sorry about the double update yesterday... -Os is broken on gcc-7 so I changed to -O2 then figured out that -O2 is broken on gcc-8 unless some optimizations are disabled. We have now the same compilation flags as other recipes out there.

So should I gather that the PR can be merged?

ghost:
I do not see problems but I am no expert either. Works for me.

alarcher:
@agrellum I'll let @pyhalov look at it as I do not want to break anything.

ghost:
I will need to add the functions I need to the sqlite mapfile so they are global and not local.
They are there just not useable yet.
Will work on so more before commit.

ghost:
I am deleting this one. I already have libgda built with the embedded one and only have a few tests failing. Anjuta builds against it and passes all tests and seems to run OK. I think I would be missing to many functions and maybe it is better security wise to leave it as it is.

PR #4265, opened 2018-06-10
Bump pcre to 8.42

Mno-hime:


components/library/pcre/Makefile
```diff
@@ -20,17 +20,19 @@
 
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2018, Michal Nowak
 #
 
+PREFERRED_BITS=64
```
pyhalov: I'd be careful with this. Delivering 64-bit pcre-config in /usr/bin will likely confuse consumers and cause build failures

PR #4266, opened 2018-06-11
Hack font 3.003

Mno-hime:
* license changed from custom Hack license to MIT (Vera is kept)
* a bit saner versioning

@timmooney: Just FYI.


timmooney:
Thanks for the FYI and for the updates.  The changes look good to me.

alarcher (approved):

PR #4268, opened 2018-06-13
Fix several desktop icons

Mno-hime:
* XTerm

`Icon=terminal` seems to be a better match, which works for all icon
themes we ship (tested that), otherwise stub icon is used.

Disabled tests as I could not find any test target in Makefile.in.

* XScreensaver

We don't ship `Icon=gnome-ccscreensaver` outside Nimbus theme, replacing
with the same icon under a different name.

* system-config-printer

We don't ship `Icon=print-manager` outside Nimbus, replacing with
similar one.

### DIFF
--- a/components/desktop/xscreensaver/Makefile
+++ b/components/desktop/xscreensaver/Makefile
@@ -21,7 +21,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		xscreensaver
 COMPONENT_VERSION=	5.39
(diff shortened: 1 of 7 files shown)


alarcher:
@Mno-hime I had to revert the ove to 64-bit that I had done ealier. There are issues with terminal definitions and this will not work as expected.

Mno-hime:
@alarcher I see, will stick to the `Icon` fix then.

PR #4277, opened 2018-06-19
Mutt 1.10.0

Mno-hime:


components/mail/mutt/Makefile
```diff
@@ -68,11 +70,12 @@ install:	$(INSTALL_64)
...
 test:		$(NO_TESTS)
 
+# Auto-generated dependencies
+REQUIRED_PACKAGES += SUNWcs
 REQUIRED_PACKAGES += library/database/gdbm
-REQUIRED_PACKAGES += library/libidn
+REQUIRED_PACKAGES += library/libidn2
 REQUIRED_PACKAGES += library/security/openssl
 REQUIRED_PACKAGES += library/slang
-REQUIRED_PACKAGES += runtime/perl-522
```
pyhalov: Also add perl-522 to REQUIRED_PACKAGES, this is necessary for mutt to depend on any perl

Mno-hime: `runtime/perl-522`is kept.

PR #4279, opened 2018-06-20
aria2 1.34.0

Mno-hime:
Version bump, fixed `aria2c` bash completion, 64-bit-only.

Tested: file download, bash completion.

Internal tests fail: https://github.com/aria2/aria2/issues/1222

### DIFF
--- a/components/web/aria2/aria2.p5m
+++ b/components/web/aria2/aria2.p5m
@@ -19,6 +19,7 @@
 # CDDL HEADER END
 #
 # Copyright (c) 2016 Predrag Zečević
+# Copyright (c) 2018 Michal Nowak
 #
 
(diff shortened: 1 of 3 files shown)


components/web/aria2/aria2.p5m
```diff
@@ -30,13 +31,12 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-file path=usr/bin/$(MACH64)/aria2c
+file path=etc/bash_completion.d/aria2c
```
pyhalov: should we move it to /usr/share/bash-completion/completions/ ?

Mno-hime: We do now.

PR #4283, opened 2018-06-25
Move vim.desktop back to gvim package

Mno-hime:
vim's `Icon` depends on gvim package content.


pyhalov:

PR #4296, opened 2018-07-02
dbus-glib 0.110

Mno-hime:
Tested: `lshal`

### DIFF
--- a/components/library/dbus-glib/Makefile
+++ b/components/library/dbus-glib/Makefile
@@ -10,12 +10,13 @@
 
 #
 # Copyright 2015 Alexander Pyhalov
+# Copyright 2018 Michal Nowak
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		dbus-glib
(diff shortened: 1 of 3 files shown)


pyhalov:
Does illumos-gate builds fine with it?

Mno-hime:
I think so, as I had it installed in the zone when rebuilding illumos-gate for the DBus 1.12 PR.

pyhalov:
For test target to be successful dbus-daemon should be present in $PATH. Please, fix this.

Mno-hime:
The test target works for me with the new DBus, the fix in DBus is here: https://github.com/OpenIndiana/oi-userland/pull/4297/files#diff-3996e4ed855c54e429a0ff7908195d21. If I remember correctly, `dbus-run-session` was unable to exec `dbus-daemon` as it can't find it in the `$PATH`. This is addressed by the patch above. In light of this you'd still rather see `dbus-daemon` in `/usr/bin/`?

pyhalov:
No

PR #4298, opened 2018-07-02
Add 64-bit-only lzip 1.20 & tests

Mno-hime:


components/archiver/lzip/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+build:		$(BUILD_64)
+
+include $(WS_MAKE_RULES)/shared-targets.mk
+
+install:	$(INSTALL_64)
+
+test:		$(TEST_64)
+
+BUILD_PKG_DEPENDENCIES =	$(BUILD_TOOLS)
```
pyhalov: Please, clean up shared-targets.mk usage, BUILD_PKG_DEPENDENCIES and depend.mk usage. They seem to be not necessary here.

components/archiver/lzip/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_TEST_TRANSFORMS = "'/testing|tests/'"
+
+build:		$(BUILD_64)
+
+install:	$(INSTALL_64)
+
+test:		$(TEST_64)
+
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/g++-4-runtime
```
pyhalov: Sorry, missed this. Why is it g++-4-runtime? It should be either not present or be $(GXX_RUNTIME_PKG)

Mno-hime: Fixed. It was what `REQUIRED_PACKAGES` target provided...

PR #4312, opened 2018-07-08
sqlite3 3.24.0

Mno-hime:


components/database/sqlite/Makefile
```diff
@@ -72,11 +74,10 @@ build:		$(BUILD_32_and_64)
 
 install:	$(INSTALL_32_and_64)
 
-test:		$(TEST_32_and_64)
```
pyhalov, line 75: At least add test: $(NO_TESTS) target

Mno-hime: Done.

PR #4315, opened 2018-07-11
FreeTDS 1.00.94

Mno-hime:


components/database/freetds/freetds.p5m
```diff
@@ -26,16 +27,6 @@ file etc/freetds.conf path=usr/share/doc/freetds/freetds.conf.sample
 file etc/locales.conf path=usr/share/doc/freetds/locales.conf.sample
 file etc/pool.conf path=usr/share/doc/freetds/pools.conf.sample
 
-hardlink path=usr/bin/bsqldb target=../lib/isaexec pkg.linted.userland.action002.0=true
```
pyhalov, line 29: Restore isaexec symlinks or ship 64-bit binaries only

Mno-hime: 64-bit-only now.

pyhalov: There's a misunderstanding here - I've spoken about /usr/bin/* files. We still need 32-bit libraries.

PR #4329, opened 2018-07-22
GNU parallel 20180722 & 64-bit-only

Mno-hime:
Upstream changelog (https://gitlab.com/GNU/Parallel/blob/master/NEWS)
says: "Better CPU detection on OpenIndiana".

Tested several examples from GNU parallel book.

### DIFF
--- a/components/shell/parallel/Makefile
+++ b/components/shell/parallel/Makefile
@@ -21,21 +21,25 @@
 
 #
 # Copyright (c) 2012, 2016, Oracle and/or its affiliates. All rights reserved.
+# Copyright 2018, Michal Nowak
 #
+
+PREFERRED_BITS=64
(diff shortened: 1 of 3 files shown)


AndWac (commented):
Can you explain why and what you changed tools/.gnupg/pubring.gpg?

Mno-hime:
@AndWac The build system adds GPG signatures (or something of the sort) there when new `$(COMPONENT_SIG_URL)` is retrieved. But as it's binary file, Git is unable to resolve conflicts when other PR updates the file, and, hence, PRs gets in conflict. I pulled out the change.

Btw, you can view content of the file by `gpg2 --no-default-keyring --keyring pubring.gpg --list-keys`.

PR #4332, opened 2018-07-23
sudosh2 64-bit-only

Mno-hime:


AndWac (commented):
Why did you remove the tests here?

Mno-hime:
@AndWac I don't see any actual tests in the source.

PR #4333, opened 2018-07-23
GNU which 64-bit-only

Mno-hime:


AndWac (commented):
Why did you remove the test target?

mnowaksuse:
@AndWac I don't see any actual tests in the source.

Mno-hime:
@AndWac I don't see any actual tests in the source.

PR #4336, opened 2018-07-25
sample-manifest template: We have 2018 now

AndWac:


pyhalov:
Any interest to do this auto-updated? :)

PR #4339, opened 2018-07-25
zsh: Remove .~1~ files from completions directory

Mno-hime:


AndWac (commented):
The sample-manifest is also affected, please fix it, too.

pyhalov:
No need to fix sample-manifest, it's an autogenerated file.

PR #4344, opened 2018-07-26
Remove .~1~ files before installation target

Mno-hime:
Try harder not to install .\~1\~ files so they won't end up in generated
sample manifest.


components/shell/zsh/Makefile
```diff
@@ -87,6 +87,8 @@ license: $(COMPONENT_SRC)/LICENCE
 
 CLEAN_PATHS += license
 
+COMPONENT_PRE_BUILD_ACTION = find $(COMPONENT_SRC) -name '*.~1~' -exec rm {} \;
```
pyhalov: Please, don't remove files from $(COMPONENT_SRC). It complicates creating patches and finding out what files were patched. Better add this to POST_INSTALL_ACTION or patch zsh makefiles to avoid installing *.~*~ files.

pyhalov (changes requested):

components/shell/zsh/Makefile
Mno-hime: Done.

PR #4356, opened 2018-08-03
nicstat: bump to 1.95

grueni:


alarcher:
@grueni Any testing performed?

grueni:
Just use in real life on our servers.

PR #4367, opened 2018-08-16
pcs: Add missing build dependency

hliebel:


alarcher (approved):

alarcher:
@hliebel On a 2nd thought, is it automatically generated or did you add it manually?
Whenever the dependency is added manually it should be listed separately before the generated section e.g:
```
# Build dependencies
REQUIRED_PACKAGES += library/ruby/bundler
```

hliebel:
Hi.
All the PRs I added have been fixed by hand. I'll go through and separate them out. I've done that already for some patches on what the Makefile looked like.

Mno-hime:
@hliebel Just wondering how you found out about those missing build dependencies: do you always use clean zone for building? I remember that when my build zone was clean I found several missing build dependencies, not so much these days :).

hliebel:
@Mno-hime I just freshly installed on a laptop and did the top level build from scratch. It's not perfect as I'm not cleaning my environment after each package, but it caught a few things.

I guess in a ideal world one could make a snapshot in a clean zone and try to roll back after each package built.. or build each in their own clean zone.. It will take a lot of time though I guess. The build as is took multiple days to complete on the Laptop I'm using..

Mno-hime:
@hliebel Maybe you'd find use of https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/build-zone.mk (probably outdated to a degree...), though I did not tried it myself.

PR #4385, opened 2018-08-20
file/magic 5.34

Mno-hime:
Run-tested.


alarcher:
Can you remove the change to the sample-manifest as it is just about the year and not the actual content?

PR #4388, opened 2018-08-20
coreutils: bump to 8.30

alarcher:


components/sysutils/coreutils/patches/04-set-permission.c.patch
```diff
@@ -1,38 +1,34 @@
-GNU Coreutils bug 21062 coreutils-8.24 - cp(1) check failures on tmpfs filesystem (Solaris 10 / Solaris 11)
```
pyhalov: Please, preserve comments

alarcher:
Done.

pyhalov:
Test results differ here

+FAIL: tests/misc/timeout-parameters
+FAIL tests/misc/timeout-parameters.sh (exit status: 1)

pyhalov:
Failed test case:
timeout 1.189731495357231765e+4932 true && echo 1 - should always echo 1
Also fails in current coreutils version.

pyhalov:
Failed test is caused by https://www.illumos.org/issues/9858

PR #4392, opened 2018-08-21
update to libgsf-1.14.44 - add test suite

ghost:


components/library/libgsf/Makefile
```diff
@@ -8,7 +8,6 @@
 # source.  A copy of the CDDL is also available via the Internet at
 # http://www.illumos.org/license/CDDL.
 #
-
```
pyhalov: Avoid removing this line

ghost: Does it trigger something in the build system?

pyhalov: No, just standard header format from https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/templates/configure.mk

ghost:
Updated file from web - Not sure about squash commits from there

PR #4394, opened 2018-08-21
update to gnumeric-1.12.43 - adjust patch offsets - add -k to test

ghost:
Tests
1 fail because help is only 32 bit

3 like this bug; was supposed be fixed
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=903194
diff dies.

Installed gnumeric seems ok to me.


ghost:
Updated file from web - Not sure about squash commits from there

PR #4397, opened 2018-08-23
samba: update to 4.8.4

pyhalov:


3eka:
Not sure if this is right place to ask:
Does this new package really needs to create /usr/local structure (old one does not have it)?

```
$ pkg contents -r pkg:/service/network/samba@4.8.4-2018.0.0.0 | grep usr/local
usr/local
usr/local/samba
usr/local/samba/bind-dns
```
Maybe this should go to /var or similar?

Regards.

pyhalov:
Well, it uses $(prefix)/samba/bind-dns to store BIND configurations and keytabs, and is set as this in several places. ${prefix} is not set to /usr, because this forces specific samba layout. If someone comes with (tested) set of patches, modifying this location, I'll gladly accept them.

PR #4400, opened 2018-08-25
Python 2.7.15

Mno-hime:
Resolves CVE-2018-1060 and CVE-2018-1061.

Changes: https://gist.github.com/Redjumpman/3b49508ff7b86bbee678bfac1ddc2674

32-bit test suite passed, 64-bit one fails with "not enough space" and
then stucks.


components/python/python27/test/results-32.master
```diff
@@ -0,0 +1,29 @@
+373 tests OK.
```
pyhalov, line 1: Why only 32-bit tests results are there?

Mno-hime: The reason is now in the PR descriptions, was "hiding" in on of the commit descriptions. Looked as if all possible file descriptors were taken... Would be cool if someone could provide that 64-bit results file.

PR #4401, opened 2018-08-25
expat 2.2.6

Mno-hime:
Changes: https://github.com/libexpat/libexpat/blob/R_2_2_6/expat/Changes

Notably, fixes UTF-8 bug required for Python 2.7.15 (https://github.com/OpenIndiana/oi-userland/pull/4400) test suite to pass.

Test suite now seems to have a different output.

### DIFF
--- a/components/library/libexpat/Makefile
+++ b/components/library/libexpat/Makefile
@@ -22,24 +22,26 @@
 #
 # Copyright (c) 2011, 2012, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2017, Aurelien Larcher
+# Copyright (c) 2018, Michal Nowak
 #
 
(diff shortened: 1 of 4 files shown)


AndWac (commented):
Have you checked whether all (or at least some) of the dependants are still working as expected?
'pkgtree dependants expat' shows a lot of dependencies...

Mno-hime:
I tested `xmlwf`, "etree" test cases of Python 2.7.15 test suite, and the internal test suite, I also run it in my global & build zones, to see if there are problems. Let me know if you have some particular test case in mind.

AndWac:
I don't have any particular test case in mind but expat has many dependants and I try to avoid new problems.

PR #4406, opened 2018-08-27
9761 Starting VM (KVM/Virtualbox) Causes panic since July

pyhalov:


pyhalov:
We need OS-6998 KVM should use HMA FPU framework for KVM to work correctly.

PR #4419, opened 2018-09-06
Support for Python 3.5 in oi-userland

pyhalov:


make-rules/shared-macros.mk
```diff
@@ -661,6 +666,9 @@ PYTHON.2.7.64 =	/usr/bin/$(MACH64)/python2.7
 PYTHON.3.4.32 =	/usr/bin/python3.4
 PYTHON.3.4.64 =	/usr/bin/$(MACH64)/python3.4
 
+PYTHON.3.5.32 =	/usr/bin/python3.5
```
Mno-hime, line 669: Do we need this for Python 3.5?

pyhalov:
```
PYTHON =	$(PYTHON.$(PYTHON_VERSION).$(BITS))
```
so if you set PYTHON_VERSION to 3.5, 32-bit build environment will find /usr/bin/python, and not the one which was intended. Note, that in most cases when we build not python-related software, we don't care about PYTHON bitness .

PR #4425, opened 2018-09-08
Remove nodejs-7

AndWac:
nodejs-7 is unsupported and has several unpatched security flaws.
I expect that nobody uses it anymore because nodejs-8 and -10 have been out for a while.


components/meta-packages/history/history
```diff
@@ -943,6 +943,7 @@ release/openindiana-welcome@0.5.11,5.11-0.148 release/os-welcome@0.5.11-0.151.1.
 release/registration@0.5.11,5.11-2018.0.0.0
 runtime/java/runtime64@0.5.11,5.11-2017.0.0.0 runtime/java/openjdk8@1.8.112,5.11-2016.1.0.0
 runtime/java@0.5.11,5.11-2017.0.0.0 runtime/java/openjdk8@1.8.112,5.11-2016.1.0.0
+runtime/nodejs-7@7.10.1-2018.0.0.1:20180524T222656Z
```
pyhalov: Please, remove timestamp, increase last number in package version (2018.0.0.2).

pyhalov:
What about node-6 (components/runtime/node6) or node@0.12.18 (runtime/node)?

AndWac:
Nodejs 6 is still supported (until April 2019). I am not sure about 0.12 as earlier we wanted to keep it because Joyent was using it in SmartOS. I am not sure which version SmartOS is using atm.

pyhalov:
Please, add note to openindiana.org/oi/oi_hipster

PR #4428, opened 2018-09-11
gst-libav1: rebuild after ffmpeg update

alarcher:
After #4034


pyhalov:
@alarcher  Do we really need to rebuild gst-libav when it uses internal libav code (included with it)???

pyhalov:
It seems gst-libav doesn't have to be rebuilt.

PR #4449, opened 2018-09-20
cffi: provide Python 3.5 version, fix 64-bit cffi version

pyhalov:


pyhalov:
Needs pycparser-35 to be installed on build server

PR #4459, opened 2018-09-26
Upgrade tornado to 4.5.3

Toasterson:
In Preparation to Package saltstack. Tested and functioning with saltstack 2018.11.0-170-g221bd63


pyhalov:
Additional dependencies are missing, for example, dependency on singledispatch, backports_abc. 
Likely you can find most of them if you remove
```'<transform path=usr/lib/python$(PYVER)/vendor-packages/tornado/.*\.py -> default pkg.depend.bypass-generate .* >'``` line, examine pkgdepend output and look at source code to find out if dependencies are really missing or guarded by try/Except .

PR #4460, opened 2018-09-26
Vim: bump to 8.1.0436

kenmays:


pyhalov:
- As I understand, we don't use build_date.patch, as SOURCE_DATE_EPOCH is not set. It can be removed from Makefile and patch can be dropped completely.
- sample-manifest and vim.p5m need updating

kenmays:
@alarcher @pyhalov - please update manifest/p5m as needed.

alarcher:
@kenmays @pyhalov I pushed an update to the manifests.

PR #4461, opened 2018-09-27
nasm: bump to 2.13.03

kenmays:


pyhalov (changes requested):
01-3392409.patch was integrated and should be removed

PR #4464, opened 2018-09-27
clang-40: fix gcc-8 build and deliver as 64-bit

alarcher:
Requires #4463

### DIFF
--- a/components/developer/clang-40/Makefile
+++ b/components/developer/clang-40/Makefile
@@ -55,7 +55,15 @@ COMPONENT_POST_UNPACK_ACTION += \
   $(UNPACK) $(USERLAND_ARCHIVES)$(COMPONENT_ARCHIVE_1) && \
 	mv $(CLANG_NAME)-$(COMPONENT_VERSION).src $(COMPONENT_SRC)/tools/clang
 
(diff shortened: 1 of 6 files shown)


components/developer/clang-40/Makefile
```diff
@@ -47,15 +49,25 @@ CMAKE_PREFIX= /usr/clang/$(COMPONENT_MAJOR_VERSION)
...
 PATH=$(PATH.illumos)
 
+PKG_MACROS+= GCC_VERSION=$(GCC_VERSION)
+
 COMPONENT_POST_UNPACK_ACTION += \
   $(UNPACK) $(USERLAND_ARCHIVES)$(COMPONENT_ARCHIVE_1) && \
 	mv $(CLANG_NAME)-$(COMPONENT_VERSION).src $(COMPONENT_SRC)/tools/clang
 
-COMPONENT_PRE_CONFIGURE_ACTION =        (cp -a $(SOURCE_DIR)/* $(@D))
+COMPONENT_PREP_ACTION = ( \
```
pyhalov, line 58: Can we instead use compile-time make variables? This would be a bit prettier. 
However, if it's hard, I'm fine with this.

alarcher: That was my original attempt but it would require adding macros to stringify the value. Can we do this later when I add clang-60?

PR #4476, opened 2018-10-06
Nimbus update

pyhalov:


pyhalov:
In this PR we finally switch nimbus to https://github.com/OpenIndiana/nimbus/ repository. However, we fetch specific tag, as after updating theme manifests likely will need updating.

Mno-hime:
Tested, looks good to me.

PR #4492, opened 2018-10-09
arpack: fix build

pyhalov:


pyhalov:
@alarcher - Perhaps, we should enable some options to deliver additional api and files, but so far I'm just going to fix build.

alarcher:
@pyhalov No problem. I am fixing rustc now but it takes a while to test the rebuild. Maybe we can ignore the component for now?

pyhalov:
There's no need for doing this if you can fix it today

PR #4493, opened 2018-10-09
net-snmp: fix CVE-2018-18065

pyhalov:


pyhalov:
Tested according to https://seclists.org/oss-sec/2018/q4/24 . We already have patch for CVE-2018-18066, but miss for CVE-2018-18065.

alarcher:
@pyhalov It seems that net-snmp is broken with gcc-8.

pyhalov:
@alarcher How exactly?

PR #4497, opened 2018-10-10
git: Use gawk for completion

Mno-hime:
`git add ` bash completion fails since git 2.19 with:
```
$ git add awk: syntax error near line 1
awk: bailing out near line 1
```

As for the rest of the bash-completion we should use `gawk` instead of
`awk`.

Tested.


3eka:
We might need text/gawk being listed in REQUIRED_PACKAGES (Makefile)?

Mno-hime:
Thanks @3eka, I added run-time dependency on `gawk`.

PR #4501, opened 2018-10-11
ninja: bump to 1.8.2

alarcher:


Mno-hime:
@alarcher Wondering if we want https://github.com/ninja-build/ninja/commit/1952afa56177d006077447477a237ad4e98270d5? Happened after the 1.8.2 release.

alarcher:
@Mno-hime I'll look into it thanks.

alarcher:
@Mno-hime Does not seem like we need this.

PR #4513, opened 2018-10-13
gnome-themes-standard: don't force dependency on adwaita icon theme

pyhalov:


pyhalov:
Nimbus doesn't need anything from adwaita icon theme. Or at least shouldn't. Removing this dependency, we free about 20 MB on install media, but user will not be able to use standard gnome theme. So I'm asking, if we want  this.

Mno-hime:
How do we free those 20 MB? I don't see `gnome/theme/gnome-themes-standard` nor `gnome/theme/adwaita-icon-theme` on the Live medium. How does it relate to Nimbus?

pyhalov:
Sorry, I was sure that nimbus depends on it, but it doesn't

PR #4514, opened 2018-10-14
fontconfig: fix pkg-config file

alarcher:


Mno-hime:
@alarcher Is this expected to fix this weird font https://openqa.oi.mnowak.cz/tests/3969#step/pkg_update_to_latest_packages/22? I think it started with this https://hipster.openindiana.org/jenkins/job/oi-userland/4253/ batch of updates.

alarcher:
@Mno-hime No it is a build issue, but fontconfig is known for not reloading the configuration properly in some configurations. Logging out and logging in again should be enough.

alarcher:
@Mno-hime If this is a real problem for some configurations we can change the default font configuration. On my systems nothing changed after update.

Mno-hime:
You are right @alarcher, it was just a transient problem, to get rid of it now it will only update to new BE.

PR #4515, opened 2018-10-14
Gstreamer 1.14

alarcher:
To be split later.

### DIFF
--- a/components/multimedia/gstreamer1/Makefile
+++ b/components/multimedia/gstreamer1/Makefile
@@ -10,13 +10,13 @@
 #
 
 #
-# Copyright 2016 Aurelien Larcher.  All rights reserved.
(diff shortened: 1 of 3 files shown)


components/encumbered/ffmpeg/Makefile
```diff
@@ -94,11 +95,9 @@ CONFIGURE_ENV += CPPFLAGS="$(CPPFLAGS)"
...
 
-$(TEST_32):	$(INSTALL_32)
+build: $(BUILD_32_and_64)
 
-build: $(BUILD_32)
-
-install: $(INSTALL_32)
+install: $(INSTALL_32_and_64)
 
 test:	$(TEST_32)
```
Mno-hime: Should the test target also be 64-bit?

alarcher: Probably :)

PR #4524, opened 2018-10-15
mate-system-monitor: do not throw if icon not found

alarcher:
A quick piggy workaround but should be better than dumping if the corresponding icon is not found.
It seems that load_icon() does not handle this.

### DIFF
--- a/components/desktop/mate/mate-system-monitor/patches/03-default-icon.patch
+++ b/components/desktop/mate/mate-system-monitor/patches/03-default-icon.patch
@@ -1,0 +1,14 @@
+--- mate-system-monitor-1.20.1/src/iconthemewrapper.cpp.orig	2018-10-16 01:12:23.890830040 +0000
++++ mate-system-monitor-1.20.1/src/iconthemewrapper.cpp	2018-10-16 01:12:30.260851246 +0000
+@@ -12,7 +12,10 @@
+ {
+     try
+     {
+-      return Gtk::IconTheme::get_default()->load_icon(icon_name, size, flags);
++	  if (Gtk::IconTheme::get_default()->has_icon(icon_name))
++	    return Gtk::IconTheme::get_default()->load_icon(icon_name, size, flags);
++      else
++	    return Gtk::IconTheme::get_default()->load_icon("image-missing", size, flags);
+     }
+     catch (Gtk::IconThemeError &error)
+     {
--- a/components/desktop/mate/mate-system-monitor/Makefile
+++ b/components/desktop/mate/mate-system-monitor/Makefile
@@ -20,6 +20,7 @@ include ../../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		mate-system-monitor
 COMPONENT_MJR_VERSION=	1.20
 COMPONENT_MNR_VERSION=	1
+COMPONENT_REVISION= 1
 COMPONENT_VERSION=	$(COMPONENT_MJR_VERSION).$(COMPONENT_MNR_VERSION)
 COMPONENT_PROJECT_URL=	http://www.mate-desktop.org
 COMPONENT_SUMMARY=	MATE system monitor


pyhalov:
The fix seems to be incorrect. load_icon() result is checked in mate-system-monitor-1.20.1/src/prettytable.cpp . Perhaps, there also should be some try/catches. 
With this fix applied, I see 'icon-unknown' pictures in process name columns, but there should be "application-x-executable" icons.

pyhalov:
Have you checked that Gtk::IconThemeError is catched in procman::IconThemeWrapper::load_icon()? Perhaps, it's a compiler issue?

components/desktop/mate/mate-system-monitor/patches/03-default-icon.patch
```diff
@@ -0,0 +1,14 @@
...
++++ mate-system-monitor-1.20.1/src/iconthemewrapper.cpp	2018-10-16 01:12:30.260851246 +0000
+@@ -12,7 +12,10 @@
+ {
+     try
+     {
+-      return Gtk::IconTheme::get_default()->load_icon(icon_name, size, flags);
++	  if (Gtk::IconTheme::get_default()->has_icon(icon_name))
++	    return Gtk::IconTheme::get_default()->load_icon(icon_name, size, flags);
++      else
++	    return Gtk::IconTheme::get_default()->load_icon("image-missing", size, flags);
```
Mno-hime, line 11: Some lines have tabs & spaces, some do only spaces, after patching the code is not aligned correctly.

alarcher:
@pyhalov It is not a fix but more of a workaround. The code does not select a default icon if bad things happened. I tried different solutions but the code throws anyway. The idea of relying on a exception to select an icon is questionable. In any case I can keep this workaround just for myself for now.

alarcher:
@pyhalov I do not see the catch in the other load_icon, get_icon_from_theme does not check the return.

pyhalov:
We have 
```
Glib::RefPtr<Gdk::Pixbuf>
procman::IconThemeWrapper::load_icon(const Glib::ustring& icon_name,
                                     int size, Gtk::IconLookupFlags flags) const
{
    try
    {
        return Gtk::IconTheme::get_default()->load_icon(icon_name, size, flags)
    }
    catch (Gtk::IconThemeError &error)
    {
        if (error.code() != Gtk::IconThemeError::ICON_THEME_NOT_FOUND)
            g_error("Cannot load icon '%s' from theme: %s", icon_name.c_str(), error.what().c_str());
        return Glib::RefPtr<Gdk::Pixbuf>();
    }
    catch (Gio::Error &error)
    {
        g_debug("Could not load icon '%s' : %s", icon_name.c_str(), error.what().c_str());
        return Glib::RefPtr<Gdk::Pixbuf>();
    }
}
````
If load_icon raises error, it should be caught by load_icon() . And as we can see in mate-system-monitor-1.20.1/src/prettytable.cpp:
```
    f = fopen (icon_name, "r");
    if (f != NULL) {
        fclose (f);
        icon = that->theme->load_icon(icon_name, APP_ICON_SIZE, Gtk::ICON_LOOKUP_USE_BUILTIN);
    }

    if (not icon) {
        icon = Glib::wrap(wnck_application_get_icon(app), /* take_copy */ true);
        icon = icon->scale_simple(APP_ICON_SIZE, APP_ICON_SIZE, Gdk::INTERP_HYPER);
    }
....
 if (it == this->defaults.end()) {
            pix = this->theme->load_icon(name, APP_ICON_SIZE, Gtk::ICON_LOOKUP_USE_BUILTIN | Gtk::ICON_LOOKUP_FORCE_SIZE);
            if (pix)
                this->defaults[name] = pix;
            } else
                pix = it->second;
````
So, it's checked in some cases. Unconditionally returning 'not-found' breaks application appearance. 
Perhaps, check in other place is also necessary. 

Have you tried filing issue against https://github.com/mate-desktop/mate-system-monitor/ ? Perhaps, devs will have some ideas?

pyhalov:
This seems to me a compiler issue - something like thrown exception is not caught or differences in ```Glib::RefPtr<Gdk::Pixbuf>();``` behavior.  I consider this to be not ready for integration as is.

components/desktop/mate/mate-system-monitor/patches/03-default-icon.patch
pyhalov: BTW, does it help if you do ```return Glib::RefPtr<Gdk::Pixbuf>();``` in ```else``` branch?

PR #4528, opened 2018-10-16
Update adwaita-icon-theme & gnome-icon-theme

Mno-hime:
* Update Adwaita to the latest version
* Update gnome-icon-theme to the latest version


components/desktop/gnome2/gnome-icon-theme/Makefile
```diff
@@ -11,20 +11,21 @@
...
 # Copyright (c) 2014 Alexander Pyhalov
 # Copyright (c) 2014 Ken Mays
+# Copyright (c) 2018 Michal Nowak
 #
 
 include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         gnome-icon-theme
-COMPONENT_VERSION=      2.31.0
+COMPONENT_VERSION=      3.12.0
```
pyhalov, line 20: Move component to component/desktop/gnome3 if it's already at 3.x version

components/desktop/nimbus/nimbus.p5m
```diff
@@ -22,6 +23,8 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+depend type=require fmri=gnome/theme/gnome-icon-theme
```
pyhalov: Do we know which icons are missing?

pyhalov:
I'm generally fine with this, but it would be interesting 
a) to find out which icons are missing in nimbus
b) to see how nimbus looks with updated icons, do they feel right (especially when they can change drastically on 2.x => 3.x update).
c) BTW, does mate icon theme contain all necessary icons?
I wan to look at it tomorrow.

components/desktop/nimbus/nimbus.p5m
pyhalov: I've already added to nimbus some missing icons, for example, https://github.com/OpenIndiana/nimbus/commit/da64fb0aecec84ec48ee4f69c05e0836a6e63a61 . Perhaps, we just have to add several more icons.

Mno-hime: I dumped the runtime require, I'll have a look for what icons are missing in Nimbus. If we are not confident that we found every icon missing, we can always add the require.

Mno-hime:
ad a) Close button & icon in Alt-F2 dialog. Some "weathers" in the Weather applet. I'll have a look for more yesterday.
ad b) I tend not to see any differences in this updated gnome theme...
ad c) MATE icon sets tend to be compete apart from OI-specific apps.

Mno-hime:
Or we can have dependency on `gnome/theme/adwaita-icon-theme` instead of `gnome/theme/gnome-icon-theme`. Adwaita is the more recent, actively developed, and bigger project of default GNOME icons.

pyhalov:
We likely want to be sure that our applications never use adwaita (unless explicitly said so). I don't remember, what is the order of looking for icons. Perhaps, we even have to patch gtk 3. I have to find out, what is going on when icon is not present in nimbus...

pyhalov:
Now I know the answer, and I can say that nimbus dependency  is a bit wrong.
We can see in https://github.com/OpenIndiana/nimbus/blob/oi/hipster/icons/index.theme.in
```
Inherits=Tango
```
So, nimbus should depend on 
gnome/theme/tango-icon-theme
And it should depend on gnome/theme/gnome-icon-theme

pyhalov:
The question is if we should switch tango-icon-theme to mate-icon-theme?

pyhalov:
Can you split it in two different commits?

PR #4529, opened 2018-10-17
Correct nimbus theme to work without gnome-icon-themes

pyhalov:


Mno-hime:
I noticed that these apps are missing icon in mate-menu or on in title bar once started:
- Sound Preferences
- Take Screenshot
- Input Method Framework Selector
- Font Viewer missing title icon
- Recent Documents

Can you check they are fixed with this rebase?

pyhalov:
Font viewer is missing title icon. Everything else seems to be fine.

PR #4534, opened 2018-10-18
Generate a component mapping file

xen0l:
This PR adds a new tool, which create a mapping of component FMRI to component paths. This is a first stage for "gmake rebuild-dependent". This PR is also dependent on @alarcher's change for fixing `gmake print-package-names`.


tools/userland-mapping
```diff
@@ -0,0 +1,117 @@
...
+                            'component_fmri': component_fmri,
+                            'component_dir': component_relative_path})
+
+    component_mapping_file = os.path.join(workspace_path, subdir, COMONENT_MAPPING_FILENAME)
+    with open(component_mapping_file, 'w') as f:
+        f.write(json.dumps(mapping, sort_keys=True, indent=4))
+
+
+def main():
+    try:
```
pyhalov: Will it fail also when -w switch is provided?

tools/userland-mapping
```diff
@@ -0,0 +1,117 @@
...
+        os.environ['WS_TOP']
+    except KeyError:
+        sys.stderr.write('WS_TOP is not set and workspace was not provide!')
```
Mno-hime: Perhaps better: `WS_TOP variable not set but workspace was not provided`.

tools/userland-mapping
xen0l: Good catch. Maybe I can remove the check completely given that the main consumer is the build system and there is no point it running it outside of it.

xen0l:
@pyhalov @Mno-hime PR updated

tools/userland-mapping
```diff
@@ -0,0 +1,111 @@
...
+COMONENT_MAPPING_FILENAME = 'mapping.json'
+
+
+def find_component_paths(path, subdir='components', debug=False):
+    expression = re.compile(r'.+\.p5m$', re.IGNORECASE)
+
+    paths = []
+    workspace_path = os.path.join(path, subdir)
+
+    for dirpath, dirnames, filenames in walk(workspace_path):
```
pyhalov, line 50: Should we really walk the directory or better use git ls-files? What will happen when we walk non-clean oi-userland (with extracted archives and build directories)?

xen0l: Yes, we should as we are doing it every time when we are doing builds. That code is also in `bass-o-matic` and it isn't causing any issues :)

pyhalov:
hm... Before @alarcher's work?

PR #4535, opened 2018-10-18
Print package names

alarcher:
After these changes component mapping and dependencies are generated correctly.

### DIFF
--- a/components/database/postgresql-95/postgresql95-contrib.p5m
+++ b/components/database/postgresql-95/postgresql95-contrib.p5m
@@ -11,7 +11,7 @@
 
 #
 # Copyright 2016 Alexander Pyhalov
-# Copyright 2018 Till Wegm√ller
+# Copyright 2018 Till Wegmueller
 #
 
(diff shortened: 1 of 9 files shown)


components/mapping.sorted
```diff
@@ -0,0 +1,1886 @@
+(origin:  meta-packages/history
```
pyhalov: 1) How long is it generated?
2) Obviously, there should be some hard-coded values or exception list.

alarcher: 1. A few minutes with the Makefile target but this is only for testing and comparing with the userland-mapping tool by @xen0l. I'll remove it before integration.
2. This mapping would be used to map the packages so probably need to add a list of provided packages for non-ips.mk components and overload the target. Otherwise the graph will be missing chunks. How it should be handled is still a question. Hopefully we'll move forward on this today.

alarcher:
Can we merge this?

xen0l:
I think so.

PR #4546, opened 2018-10-20
openssh: update to 7.8p1

pyhalov:


pyhalov:
Need to fix login from OI ssh client to Sun SSH server 1.2
Now it fails with 
no matching key exchange method found. Their offer: gss-group1-sha1-toWM5Slw5Ew8Mqkay+al2g==,diffie-hellman-group-exchange-sha1,diffie-hellman-group1-sha1

PR #4547, opened 2018-10-20
[5/7] Update libcompizconfig to 0.8.14

Mno-hime:


components/desktop/libcompizconfig/patches/libcompizconfig-01-solaris-port.patch
```diff
@@ -1,35 +1,22 @@
...
-+++ libcompizconfig-0.8.4/config/config	2009-10-15 08:06:29.308065755 +0200
-@@ -14,7 +14,8 @@ integration = true
- plugin_list_autosort = true
- 
+--- libcompizconfig-0.8.14/config/config	2017-04-18 02:51:42.000000000 +0000
++++ libcompizconfig-0.8.14/config/config.new	2018-10-20 17:46:20.433742905 +0000
+@@ -1,3 +1,4 @@
  [general]
 -backend = ini
 +backend = gconf
```
pyhalov: I wonder how it works if gconf backend is disabled?
@Mno-hime

Mno-hime: I kept the upstream `ini` format and verified that `ccsm` saves activated plugins.

PR #4551, opened 2018-10-22
Added: Saltstack

Toasterson:
Please note that master and minion require an entry inside /etc/hosts to function. Or the IP of the master inside /etc/salt/minion

https://docs.saltstack.com/en/latest/ref/configuration/minion.html

The Services will run without it but will not work properly i.e. Connect to one another.


components/python/salt/Makefile
```diff
@@ -0,0 +1,48 @@
...
+include $(WS_MAKE_RULES)/setup.py.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+build:          $(BUILD_NO_ARCH)
+
+install:        $(INSTALL_NO_ARCH)
+
+test:           $(NO_TESTS)
+
+REQUIRED_PACKAGES+=library/python/futures-$(PYV)
```
pyhalov: What have you tried to achieve here?

components/python/salt/salt-PYVER.p5m
```diff
@@ -0,0 +1,1637 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)-$(PYV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value='Portable, distributed, remote execution and configuration management system'
+set name=info.classification value='org.opensolaris.category.2008:System/Administration and Configuration'
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='Apache License 2.0'
+
+# Tornado does a lot of try/import/except, making pkgdepend unhapyy
```
pyhalov: 1) Please, fix comment
2) What happens if you disable this hack?

components/python/salt/salt-PYVER.p5m
```diff
@@ -0,0 +1,1637 @@
...
+depend type=require fmri=library/python/futures-$(PYV)
+
+# SMF Manifest and distribution directories
+set name=org.opensolaris.smf.fmri value=svc:/application/salt:minion value=svc:/application/salt:master
+dir  path=etc/salt owner=root group=root mode=0755
+dir  path=var/log/salt owner=root group=root mode=0755
+file files/salt.xml path=lib/svc/manifest/application/salt.xml
+
+
+file path=usr/bin/salt
```
pyhalov: If manifest is versioned, better move usr/bin/* to usr/bin/*-2.7/3.4/3.5 and make mediated symlinks to them

components/python/salt/Makefile
pyhalov: Likely, you meant ```REQUIRED_PACKAGES += library/python/futures-27 ```
and so on

Toasterson: The Line is copied from other Python packages. It forces a dependency on the correct Python version no matter the Python version. Should I hardcode 2.7?

pyhalov: It doesn't work in such way in Makefile. PYV is defined only in manifests, IIRC.

components/python/salt/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+COMPONENT_NAME=salt
+COMPONENT_VERSION=2018.3.2
+COMPONENT_SUMMARY=	Portable, distributed, remote execution and configuration management system
+COMPONENT_PROJECT_URL=http://saltstack.com
+COMPONENT_FMRI=system/management/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Administration and Configuration
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=https://files.pythonhosted.org/packages/9b/84/48b46bad5fa13b47c10a71b4f58cf0f3a30fcb32b6a6599fca454b8c6256/$(COMPONENT_ARCHIVE)
```
Mno-hime: This looks odd to me. What's the reason for this. Salt Stacke releases are here: https://github.com/saltstack/salt/releases.

components/python/salt/Makefile
pyhalov: This is a Makefile REQUIRED_PACKAGES, list of packages, against which pkgdepend resolve dependencies + packages, requied to build component. $(PYV) is not defined in Makefile, you should use explicit versions here.

components/python/salt/salt-PYVER.p5m
Toasterson: mediated as implementation or version?

components/python/salt/Makefile
Toasterson: Ah good to know changed.

components/python/salt/salt-PYVER.p5m
Toasterson: Done and done.

components/python/salt/salt-PYVER.p5m
Toasterson: Using version mediation. and setting version to 2.7

components/python/salt/Makefile
Toasterson: Ah thanks i did only find the pipy URL via the documentation. even the checksum is the same. changed to the github url.

components/python/salt/salt-PYVER.p5m
```diff
@@ -22,7 +22,10 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='Apache License 2.0'
 
-# Tornado does a lot of try/import/except, making pkgdepend unhapyy
+# Salt does a lot of try/import/except, making pkgdepend unhapyy
+# If one disables this line pkg will try to add dependencies to a giant list of dependencies that
```
pyhalov: 'will try to add dependencies to a giant list of dependencies' seems to be something non-english even for me
Perhaps, just remove second line and later comment?

Toasterson: Yep that was me getting distracted.

components/python/salt/Makefile
```diff
@@ -0,0 +1,48 @@
...
+COMPONENT_SUMMARY=	Portable, distributed, remote execution and configuration management system
+COMPONENT_PROJECT_URL=http://saltstack.com
+COMPONENT_FMRI=system/management/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Administration and Configuration
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=https://github.com/saltstack/salt/releases/download/v$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=sha256:d86eeea2e5387f4a64bbf0a11d103bfc8aac1122e19d39cc0945d33efdc797bd
+COMPONENT_LICENSE=	Apache License 2.0
+COMPONENT_LICENSE_FILE=LICENSE
```
Mno-hime: In various places above, add tab after `=` so the text aligns better, please.

Toasterson: Sure. Done

components/python/salt/Makefile
```diff
@@ -0,0 +1,48 @@
+# This file and its contents are supplied under the terms of the
```
pyhalov, line 2: Missing first license line '#\n'

Toasterson: It' s missing in python.mk template aswell

Toasterson: Fixed

components/python/salt/salt-PYVER.p5m
```diff
@@ -0,0 +1,1653 @@
...
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/django_orm.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/ec2_pillar.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/etcd_pillar.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/extra_minion_data_in_pillar.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/file_tree.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/foreman.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/git_pillar.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/gpg.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/hg_pillar.py
+file path=usr/lib/python$(PYVER)/vendor-packages/salt/pillar/
```
pyhalov, line 1654: Please, create mediated symlinks fot these man pages (look at barman for example)

components/python/salt/Makefile
```diff
@@ -0,0 +1,49 @@
...
+include $(WS_MAKE_RULES)/setup.py.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+build:          $(BUILD_NO_ARCH)
+
+install:        $(INSTALL_NO_ARCH)
+
+test:           $(NO_TESTS)
+
+REQUIRED_PACKAGES+=library/python/futures-27
```
pyhalov: Only runtime/python-27 should be in REQUIRED_PACKAGES (use gmake REQUIRED_PACKAGES)

components/python/salt/salt-PYVER.p5m
```diff
@@ -0,0 +1,1653 @@
...
+depend type=require fmri=library/python/markupsafe-$(PYV)
+depend type=require fmri=library/python/requests-$(PYV)
+depend type=require fmri=library/python/tornado-$(PYV)
+depend type=require fmri=library/python/pyzmq-$(PYV)
+depend type=require fmri=library/python/cryptography-$(PYV)
+depend type=require fmri=library/python/futures-$(PYV)
+
+# SMF Manifest and distribution directories
+set name=org.opensolaris.smf.fmri value=svc:/application/salt:minion value=svc:/application/salt:master
+dir  path=etc/salt owner=root group=root mode=0755
```
pyhalov, line 47: Do we have some example configuration files?

pyhalov:
```
[ Oct 23 10:54:22 Executing start method ("/usr/bin/salt-master -d --log-file=/var/log/salt/master.log -c /etc/salt"). ]
[ Oct 23 10:54:28 Method "start" exited with status 0. ]
[ Oct 23 10:54:31 Stopping because process dumped core. ]
[ Oct 23 10:54:31 Executing stop method (:kill). ]
```

components/python/salt/salt-PYVER.p5m
Toasterson: Not needed. But the Directory is needed to force salt to use it. We force that in the manifest with the -d switch. If we do not force the configuration directory salt does not use the same paths as documented leading to people being confused.

Toasterson:
Any more details to the crash? I can not make it happen on any of my systems.

pyhalov:
On crash? My host missed https://github.com/illumos/illumos-gate/commit/8c3347c4989237d6ebeeb03a67b21a523fc105d6

Toasterson:
:) Yep saltstack knows of events

pyhalov:
Will you be able to fix it (manpages + REQUIRED_PACKAGES) in several hours?

Toasterson:
Yes that is possible. I just need to setup my buildzone again. Should be done in about two or so.

pyhalov:
OK. Once this is fixed and integrated, I'll prepare snapshot.

components/python/salt/salt-PYVER.p5m
Toasterson: Done

components/python/salt/Makefile
Toasterson: Done

pyhalov:
Couldn't add some fixes to your branch,so merged https://github.com/OpenIndiana/oi-userland/pull/4554 . Sorry, author was rewritten by github.

PR #4554, opened 2018-10-23
Salt

pyhalov:


Toasterson:
Erm this mediation will break salt when switching python versions

pyhalov:
Does it work with python 3?

pyhalov:
Perhaps, we can build with PYTHON_VERSIONS= 2.7 3.4 3.5?

Toasterson:
Nope not yet. they are still in the process of. WHich is why i mediated it to mediator salt

pyhalov:
OK, let's do links non-mediated for now.

Toasterson:
works aswell

PR #4559, opened 2018-10-24
Remove gtkhtml

AndWac:


pyhalov:
Should we update it?

AndWac:
Or better delete it. It used to be the HTML rendering engine for evolution. But we don't provide evolution anymore and I can't find any other dependant.

pyhalov:
Is it   still developed? If no, drop it.

AndWac:
The latest release seems to be 4.10.0 from September, 2015.
This version needs gail (libgail?) which we dropped.

pyhalov (approved):
Don't forget to mention this on http://wiki.openindiana.org/oi/oi_hipster

PR #4566, opened 2018-10-29
Fonts

alarcher:


components/fonts/arphic-uming/arphic-uming.p5m
```diff
@@ -0,0 +1,39 @@
...
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2018 <contributor>
```
pyhalov: Fix this

make-rules/common.mk
```diff
@@ -0,0 +1,198 @@
...
+# Always process LDFLAGS.
+COMPONENT_BUILD_ENV += LDFLAGS="$(strip $(LDFLAGS))"
+endif
+
+# Assume components built with ant do not need ASLR.
+ifeq ($(strip $(BUILD_STYLE)),ant)
+ASLR_MODE= $(ASLR_NOT_APPLICABLE)
+else
+# Enable ASLR by default unless target build is NO_ARCH.
+ifneq ($(strip $(BUILD_BITS)),NO_ARCH)
```
pyhalov, line 96: We can comment out this... Perhaps, we should patch ld to respect environment variables as Oracle one does. But we likely don't want ASLR to be suddenly activated.

make-rules/common.mk
```diff
@@ -0,0 +1,198 @@
...
+
+# For the system-test target, assume that none have been implemented.
+SYSTEM_TEST_TARGET ?= $(SYSTEM_TESTS_NOT_IMPLEMENTED)
+ifneq ($(strip $(SYSTEM_TEST_TARGET)),)
+system-test:    $(SYSTEM_TEST_TARGET)
+endif
+endif # COMMON_TARGETS
+
+# Always needed; every component builds packages.
+include $(WS_MAKE_RULES)/ips.mk
```
pyhalov, line 177: Do we want to use this Makefile for IPS or other gates? If so, this   this should be conditional.

alarcher: I imagined that gates are a special case and would not rely on this file.
For now introducing this files helps factoring for fonts and allow testing the logic before we extend the use to other components.

make-rules/common.mk
alarcher: Yes I was not sure how this should be handled. I am using the file for fonts only but of course this should be a discussion.

components/fonts/arphic-uming/arphic-uming.p5m
alarcher: Thanks for catching it.

PR #4569, opened 2018-10-30
Virtualbox 5.2.22

alarcher:
Need to write install and patch USB possibly.


timmooney:
Just tried a build using the commit from @pyhalov  and @alarcher 's pull with the 5.2.20 diffs, and I get a build failure on my build system because of missing sys/soundcard.h:

```
/export/home/mooney/oi-userland/components/sysutils/virtualbox/build/amd64/src/VBox/Devices/Audio/DrvHostOSSAudio.cpp:22:27: fatal error: sys/soundcard.h: No such file or directory
 #include <sys/soundcard.h>
                           ^
compilation terminated.
kmk: *** [/export/home/mooney/oi-userland/components/sysutils/virtualbox/build/amd64/out/solaris.amd64/release/obj/VBoxDD/Audio/DrvHostOSSAudio.o] Error 1
```

I see that `--enable-pulse` is specified as one of the configure options.  Should that preclude the OSS stuff being built?  I'm just trying to understand whether the intent is that `pkg:/system/header/header-audio` should be required to build or not.

alarcher:
@timmooney Since I have not been all the way to the REQUIRED_PACKAGES stage, build dependencies are not specified yet. You are correct that headers should be installed.

alarcher:
Works but some issue with the interface, only runs as root...

components/sysutils/virtualbox/files/Makefile
```diff
@@ -1,34 +1,161 @@
-BINDIR = out/solaris.amd64/release/bin
-DRVS = vboxnet vboxdrv vboxflt vboxusb vboxusbmon
+PREFIX = /opt/VirtualBox
```
pyhalov: is this file used?

components/sysutils/virtualbox/virtualbox.p5m
```diff
@@ -22,6 +22,111 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+set name=variant.opensolaris.zone value="global"
```
pyhalov: Can we mark only drivers with it (if necessary)? I suppose, one can run VirtualBox in zone, if corresponding drivers are passed to zone.

components/sysutils/virtualbox/virtualbox.p5m
```diff
@@ -22,6 +22,111 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
...
+set name=pkg.depend.runpath value=$PKGDEPEND_RUNPATH:opt/VirtualBox/amd64
```
pyhalov: s/amd64/$(MACH64)/

components/sysutils/virtualbox/virtualbox.p5m
```diff
@@ -32,3 +137,108 @@ file path=platform/i86pc/kernel/drv/vboxflt.conf
...
+file path=usr/share/icons/hicolor/96x96/mimetypes/virtualbox-vbox.png
+file path=usr/share/icons/hicolor/96x96/mimetypes/virtualbox-vdi.png
+file path=usr/share/icons/hicolor/96x96/mimetypes/virtualbox-vhd.png
+file path=usr/share/icons/hicolor/96x96/mimetypes/virtualbox-vmdk.png
+file path=usr/share/mime-info/virtualbox.keys
+file path=usr/share/mime-info/virtualbox.mime
+file path=usr/share/mime/packages/virtualbox.xml
+file path=usr/share/pixmaps/VBox.png
+file path=usr/share/pixmaps/virtualbox.png
+file path=var/svc/manifest/application/virtualbox/virtualbox-autostart.xml
```
pyhalov, line 237: Should we place it in /lib/svc/manifest ?

components/sysutils/virtualbox/virtualbox.p5m
```diff
@@ -32,3 +137,108 @@ file path=platform/i86pc/kernel/drv/vboxflt.conf
...
+file path=var/svc/manifest/application/virtualbox/virtualbox-balloonctrl.xml
+file path=var/svc/manifest/application/virtualbox/virtualbox-zoneaccess.xml
+#file path=vbox.pkginfo
+#file path=vbox.space
+
+driver name=vboxdrv
```
pyhalov, line 243: Do we want to have separate driver package or mark drivers as GZ-only content?

components/sysutils/virtualbox/Makefile
```diff
@@ -61,29 +88,58 @@ LOCAL_CONFIG+='\nVBOX_WITH_DTRACE_R0DRV ='
...
 CONFIGURE_OPTIONS+= --disable-dbus
 CONFIGURE_OPTIONS+= --disable-docs
-CONFIGURE_OPTIONS+= --disable-pulse
+CONFIGURE_OPTIONS+= --enable-pulse
 CONFIGURE_OPTIONS+= --disable-python
 CONFIGURE_OPTIONS+= --disable-sdl-ttf
 CONFIGURE_OPTIONS+= --disable-libvpx
 CONFIGURE_OPTIONS+= --nofatal
 
+VBOX_PREFIX   = /opt/VirtualBox
```
pyhalov, line 106: Do we want to stick to /opt/VirtualBox layout or move it to /usr/virtualbox or /usr/lib/virtualbox to be more consistent with other packages?

components/sysutils/virtualbox/virtualbox.p5m
alarcher: OK

components/sysutils/virtualbox/virtualbox.p5m
alarcher: Thanks

components/sysutils/virtualbox/virtualbox.p5m
alarcher: Problem is that the script for post/pre-install steps assumes that all files are delivered. Would need some time to look at splitting the logic if a different package are provided for drivers.
Should I use a transform to mark drivers are GZ-only content? I have never done this.

components/sysutils/virtualbox/Makefile
alarcher: If we want another prefix we need to patch configuration and install scripts.
That would be better probably but I will not have time to look at this in the coming two weeks.

components/sysutils/virtualbox/files/Makefile
alarcher: I do not think so but I left it since you added it, just in case.
Should be cleaned up just before merge.

alarcher:
@pyhalov Given that I will be busy with relocation, feel free to push whichever change you want to the PR and merge it if you like. I cannot guarantee that I'll have time to review this in the coming two weeks.

components/sysutils/virtualbox/virtualbox.p5m
pyhalov: Just add variant.opensolaris.zone=global to driver action

alarcher: It does not work, it chokes on the actual driver files and if the same is added to them pkg returns an internal error.

pyhalov: Sorry, looking at illumos-gate build receipts, it seems variant.opensolaris.zone=global should be added only to actual files.

pyhalov:
```
kmk: *** No rule to make target `/libQt5Core.so', needed by `/export/home/alp/srcs/oi-userland/components/sysutils/virtualbox/build/amd64/out/solaris.amd64/release/obj/VBoxDbg/VBoxDbg.so'.  Stop.
kmk: *** Waiting for unfinished jobs....
```

components/sysutils/virtualbox/Makefile
```diff
@@ -0,0 +1,145 @@
...
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+PATH=$(PATH.gnu)
+
+X11_SERVERMODS_DIR=/usr/lib/xorg/modules
+X11_SERVERLIBS_DIR=/usr/lib/xorg
+
+QT5_ROOT=/usr/lib/qt/5
```
pyhalov: currently we don't have /usr/lib/qt/5, we have /usr/lib/qt/5.8

components/sysutils/virtualbox/Makefile
```diff
@@ -0,0 +1,145 @@
...
+PATH=$(PATH.gnu):$(QT5_ROOT)/bin/amd64
```
pyhalov: s/amd64/$(MACH64)/g

pyhalov:
I can install virtualbox in zone, but I haven't tried to actually run it there...

components/sysutils/virtualbox/Makefile
```diff
@@ -0,0 +1,144 @@
...
+
+#
+# Copyright 2017 Aurelien Larcher
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=         VirtualBox
+COMPONENT_VERSION=      5.2.22
+COMPONENT_SUMMARY=      VirtualBox - Host drivers
```
pyhalov: COMPONENT_SUMMARY should be fixed

pyhalov:
Symbol relocation issue with qt5 is an issue with qt5/suid application. Briefly - for suid application to respect RUNPATH, it shouldn't contain $ORIGIN.

PR #4571, opened 2018-11-01
slim_source: use local nightly copy

pyhalov:


pyhalov:
Depends on https://github.com/OpenIndiana/slim_source/pull/47

PR #4573, opened 2018-11-01
gfx-drm: use local bldenv copy

pyhalov:


pyhalov:
Depends on https://github.com/illumos/gfx-drm/pull/14

PR #4574, opened 2018-11-01
build-essential: add python runtime packages

alarcher:
Required by illumos-gate as suggest by @jlevon

### DIFF
--- a/components/meta-packages/build-essential/Makefile
+++ b/components/meta-packages/build-essential/Makefile
@@ -16,7 +16,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		build-essential
 COMPONENT_VERSION=	1.0
-COMPONENT_REVISION=	12
+COMPONENT_REVISION=	13
 
 include ../../../make-rules/ips.mk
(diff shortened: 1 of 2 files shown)


pyhalov:
They are not strictly required and make build essential much less essential. Also note, that illumos-gate can be built without one of the pythons. Likely it's better to be added to illumos-gate component REQUIRED_PACKAGES.

xen0l (approved):

pyhalov:
Should we also add runtime/gcc-7 there? 
Or better split out illumos-gate build dependencies from build-essential?

Mno-hime:
Maybe these Python runtimes and gcc-7 can at least/also be added as `REQUIRED_PACKAGES` to illumos-gate component so it's obvious what's required.

pyhalov:
@alarcher  Please, rebase :)

pyhalov:
@alarcher Also, what do you think about splitting illumos build dependencies in separate package?

pyhalov:
Merged as https://github.com/OpenIndiana/oi-userland/pull/5139

PR #4575, opened 2018-11-02
golang-110: bump to 1.10.4

alarcher:
@Toasterson if you have time to build and test :)


Toasterson (changes requested):
You'll need to include the manpage from files and not from proto directory. Build failes because there is no manpage in proto area acording to pkg

Toasterson:
you can take the smf manifest section of salt as an example

alarcher:
@Toasterson I guess there is an other issue behind this. I have published from a clean component with this recipe.

Toasterson:
Could be. Nethertheless there is a manpage in files so why not just use that?

alarcher:
@Toasterson OK it was easy, I forgot to commit changes to the manifest. Anyhow sample-manifest.p5m lists this file so it is in the proto area. The file you mention was not used until now in your previous commit and probably requires update, I'd rather trust upstream to produce up-to-date manpages. I'll check if the file they deliver requires modifications.

Toasterson:
OK well iam getting:

Couldn't find 'usr/share/golang/1.10/man/man1/go.1' in any of the specified search directories:
        /export/home/toast/workspace/openindiana/userland/components/developer/golang-110
        /export/home/toast/workspace/openindiana/userland/components/developer/golang-110/build
        /export/home/toast/workspace/openindiana/userland/components/developer/golang-110/build/prototype/i386
        /export/home/toast/workspace/openindiana/userland/components/developer/golang-110/build/prototype/i386/mangled
        /export/home/toast/workspace/openindiana/userland/components/developer/golang-110/golang-1.10.4

alarcher:
@Toasterson The problem was that the paths referred to COMPONENT_VERSION instead of COMPONENT_MAJOR_VERSION. I addressed your comments in the current changeset, thank you!

Toasterson:
Brilliant now it works :)

Toasterson (approved):

PR #4579, opened 2018-11-05
gc: bump to 8.0.0

alarcher:
ABI compatible. Tested with inkscape.


Toasterson (approved):
Compiles and Tests work.

LGTM

PR #4598, opened 2018-11-19
itstool: don't crash on non-utf8 msgid

pyhalov:


pyhalov:
Fixes https://github.com/itstool/itstool/issues/22

PR #4605, opened 2018-11-29
Update dash to 0.5.10.2

Mno-hime:


components/shell/dash/Makefile
```diff
@@ -44,6 +44,5 @@ build:		$(BUILD_64)
 
 install:	$(INSTALL_64)
 
-test:		$(TEST_64)
```
pyhalov, line 47: test: $(NO_TESTS) ?

PR #4607, opened 2018-11-30
Port bash-completion & git completion from GNU awk to nawk

Mno-hime:
I tested that `git add` completion, which was broken with illumos `awk` and worked with `gawk`, works with `nawk`.

Casually tested bash-completion, but there are too many of them to test them all.


pyhalov:
nawk is in system/extended-system-utilities . It is installed in zone or in GZ by default, but better explicitly to depend on it.

PR #4619, opened 2018-12-01
9968 Update open-vm-tools to 10.3.5, 64-bit-only, GTK+3

Mno-hime:
Release notes: https://github.com/vmware/open-vm-tools/blob/stable-10.3.5/ReleaseNotes.md

### DIFF
--- a/components/sysutils/open-vm-tools/open-vm-tools.p5m
+++ b/components/sysutils/open-vm-tools/open-vm-tools.p5m
@@ -12,6 +12,7 @@
 #
 # Copyright 2016 Adam Stevko
 # Copyright 2017 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
+# Copyright 2018 Michal Nowak
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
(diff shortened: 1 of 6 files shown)


Mno-hime:
Tested with VMware ESXi 6.7U1: General information from `vmtoolsd` service, screen resolution change, managed restart/shutdown.

image

~~I added workaroung for `vmtoolsd` to find plugins as it now looks for them (incorrectly) in `usr/lib/$(MACH64)/open-vm-tools/plugins/amd64/`.~~

components/sysutils/open-vm-tools/open-vm-tools.p5m
```diff
@@ -65,6 +65,7 @@ file path=usr/lib/$(MACH64)/libhgfs.so.0.0.0
 link path=usr/lib/$(MACH64)/libvmtools.so target=libvmtools.so.0.0.0
 link path=usr/lib/$(MACH64)/libvmtools.so.0 target=libvmtools.so.0.0.0
 file path=usr/lib/$(MACH64)/libvmtools.so.0.0.0
+link path=usr/lib/$(MACH64)/open-vm-tools/plugins/$(MACH64) target=.
```
pyhalov: Should we do it? This looks ugly.
Better patch open-vm-tools-stable-10.3.0/services/vmtoolsd/pluginMgr.c:
It has 
```
#if defined(sun) && defined(__x86_64__)
   const char *subdir = "/amd64";
#else
   const char *subdir = "";
#endif
```
Just set it to ""

Mno-hime: Thanks for finding the culprit. I removed the uglyness and re-tested.

pyhalov: Sorry, I don't see the patch in PR.

Mno-hime: Thanks for catching it, forgot to add it.

PR #4620, opened 2018-12-01
Update glibmm to 2.58.0

Mno-hime:
Needs testing.

ABI Factory looks odd: https://abi-laboratory.pro/?view=timeline&l=glibmm


Mno-hime:
Haven't spotted any issues with this package in my GZ.

PR #4631, opened 2018-12-05
Jenkins weekly 2.154 and lts 2.138.4

jimklimov:
Security updates, per https://jenkins.io/security/advisory/2018-12-05/

Note: Current "latest" stable is 2.150.1, I am clarifying with Jenkins IRC whether it is as safe as 2.138.4 which is recommended in the doc.


jimklimov:
Update: per https://jenkins.io/blog/2018/12/05/security-updates/ 2.150.1 also includes the fix, as well as a new core baseline.

jimklimov:
#4632 follows up on this PR to update the LTS from earlier stable to the newest and bloodier stable :)

Just in case, I think it makes sense to have a package of both revisions in our history, hence the proposed separate PR which should get merged after the 2.138.4 package is in the OI repo.

PR #4632, opened 2018-12-05
Jenkins LTS update from 2.138.4 to 2.150.1

jimklimov:
Follows up on #4631 to update the LTS from earlier stable to the newest and bloodier stable :)

Just in case, I think it makes sense to have a package of both revisions in our history, hence the proposed separate PR which should get merged after the 2.138.4 package is in the OI repo.


pyhalov:
please, rebase branch on master

jimklimov:
Done

PR #4633, opened 2018-12-05
qt5: avoid using relative runpaths

pyhalov:


pyhalov:
@normjacobs: What do you think about it?

alarcher:
@pyhalov You can probably remove the GCC_LIBDIR macro definition as it is now in shared-macros.mk Otherwise it should be fine. Thanks.

pyhalov:
Have to check that runpaths set in COMPONENT_POST_INSTALL_ACTION are still correct.

PR #4637, opened 2018-12-07
virtualbox: build guest additions

pyhalov:
@xen0l @alarcher - well, building them was easy. Trickier part is to correctly install binaries from build/amd64/out/solaris.amd64/release/bin/additions .


pyhalov:
What does work:
- shared clipboard
- drag and drop for files
- mouse integration

What doesn't work:
- vboxvideo: kernel module doesn't load with a lot of symbols undefined, part of them coming from dtrace parts, which are epxlicitly disabled, part of them - are drm functions, which we miss...

What wasn't tested - shared folders.

pyhalov:
vboxvideo works when xorg driver is installed

pyhalov:
Last commit needs #4660

PR #4641, opened 2018-12-12
Update sudo to 1.8.26

Mno-hime:
* Support patch-level zero in Makefile
* Add test result file
* `HAVE_AU_CLOSE_SOLARIS11` added upstream means that `#ifndef` from
  patches/01-bsm_audit.c.patch is no longer needed
  (see https://bitbucket.org/millert/sudo/commits/4a4f91e28bbc5766ab7c7cc8c196acc59a99cb96).
* Enabled `AUE_sudo` as illumos#8309 is now fixed

Is there a way to test `AUE_sudo` or at least some way to check it's not causing problems?


pyhalov:
https://www.oracle.com/technetwork/articles/servers-storage-admin/sol-audit-quick-start-1942928.html

PR #4647, opened 2018-12-13
 Update tmux to 2.8

benf33:
Version bump, no changes.


pyhalov:
1) Remove COMPONENT_REVISION
2) Have you checked if 01-fix-tty.patch is still needed after https://github.com/illumos/illumos-gate/commit/6cfa0a7079eb1baa0a742f1e4def7cdabc8dcdab ?

benf33:
Compiles and runs well without the patch.
I cannot say more, strconf does say the same with or without.

PR #4652, opened 2018-12-15
Update crypto/ca-certificates (NSS) to 3.41

Mno-hime:
* Updated `mk-ca-bundle.pl` to the latest code on the curl master
* Regenerated set of certificates (StartCom, WoSign, Symantec are gone)

Only `package/pkg` seems to depend on it?


pyhalov:
Almost everything, relying on HTTPS or SSL depends on it. You can check if curl or wget is happy with new bundle.

Mno-hime:
Both `curl` and `wget` say about https://www.howsmyssl.com/: "Your SSL client is Probably Okay".

3eka:
What about aria2 ?

Mno-hime:
`aria2` works for me as well.

pyhalov:
Should we merge this?

Mno-hime:
I run it for some 5 days on my desktop, I think we can merge it.

PR #4655, opened 2018-12-17
Update SQLite to 3.26.0, add test suite, enable FTS4/5

Mno-hime:
Test suite passes on amd64, on i386 fails for all date-2.2c-XXX tests:

```
! date-2.2c-1 expected: [06:28:00.001]
! date-2.2c-1 got:      [06:28:00.000]
```

Real-world testing pending.


Mno-hime:
Supposedly there's a vulnerability in SQLite (https://blade.tencent.com/magellan/index_en.html), fixed in 3.26.0. Now I have it on my desktop, so far no problems.

pyhalov (approved):

components/database/sqlite/manifests/sample-manifest.p5m
```diff
@@ -30,6 +30,8 @@ link path=usr/lib/$(MACH64)/libsqlite3.so target=libsqlite3.so.0.8.6
 link path=usr/lib/$(MACH64)/libsqlite3.so.0 target=libsqlite3.so.0.8.6
 file path=usr/lib/$(MACH64)/libsqlite3.so.0.8.6
 file path=usr/lib/$(MACH64)/pkgconfig/sqlite3.pc
+file path=usr/lib/$(MACH64)/sqlite3/libtclsqlite3.so
```
pyhalov, line 33: Should we ship it? How is it related to components/tcl/tcl-sqlite ?

Mno-hime: We can ship it, but I never managed to install 32-bit `libtclsqlite3.so`, it gets overwritten by 64-bit one. I can include the 64-bit one and the associated `pkgIndex.tcl` file, but I would not bother fixing the 32-bit one. I don't know how it relates to `runtime/tcl-8/tcl-sqlite-3`.

PR #4658, opened 2018-12-18
Update OpenLDAP to 2.4.47

Mno-hime:
Originally this was more ambitious but better reverting to version bum, cleanup, and mdb backend enablement.

Move hdb backend to module (needs to be enabled explicitly, but for a long time no one should be using it as it's are deprecated). Enabled mdb as embedded backend. Synced with solaris-userland.

ABI Laboratory: https://abi-laboratory.pro/index.php?view=timeline&l=openldap
Release Changes: https://www.openldap.org/software/release/changes.html

Testsuite passed for all backends.


pyhalov:
As for 07-fopen.patch: 
- First part is no-op (it's guarded by ifdef HAVE_WINSOCK)
- Another parts can be applied to illumos.
As for ldap-olslapd update, some parts can be applied here, but I don't think it's worth doing.
Disabling deprecated backends can break current users.

Mno-hime:
Thanks @pyhalov. I added the `07-fopen.patch` as-is. Modified database backends bdb, hdb, and mdb from compiled-in to modules. Tested the service, manually run `slapd`, and run the internal test suite (of which only `test058-syncrepl-asymmetric` partially failed for all backends).

Mno-hime:
As reported on the mailing list, the service currently starts 32-bit `slapd`, any reason not to run 64-bit one? Otherwise I'll change that.

pyhalov:
If we switch to 64-bit slapd by default, will it work with database, created by 32-bit slapd?

pyhalov:
I see two ways here:
1) We make it tunable, like it had earlier been for mysql - see https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/database/mariadb-101/files/mariadb_101.xml#L81 and https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/database/mariadb-101/files/mariadb_101#L60 . 
2) We make it 64-bit only.

In both cases we should check what happens on 32->64  bit switch (and back). For example, is data format compatible or should we use separate data directory for first case? If data format is not compatible, what options do users have to keep their data?

components/network/openldap/files/ldap-olslapd
```diff
@@ -31,7 +33,7 @@ typeset -r VARRUNDIR=/var/openldap/run
 typeset -r VARDATADIR=/var/openldap/openldap-data
 typeset -r PIDFILE=${VARRUNDIR}/slapd.pid
 typeset -r CONF_FILE=/etc/openldap/slapd.conf
-typeset -r SLAPD="/usr/lib/slapd -u ${LDAPUSR} -g ${LDAPGRP} -f ${CONF_FILE}"
+typeset -r SLAPD="/usr/lib/amd64/slapd -u ${LDAPUSR} -g ${LDAPGRP} -f ${CONF_FILE}"
```
pyhalov: 1) Don't use amd64 in scripts (we have /usr/lib/64)
2) If we always use 64-bit slapd, what's the purpose of delivering 32-bit one? (PREFERRED_BITS= 64 ?)

pyhalov:
1) Transitioning to 64-bit server will break compatibility for old installation.  Should we do it right now? I don't know. Perhaps, we should deliver 2 versions of slapd as earlier and run one based on service setting (32-bit being default for now).
2) To avoid breaking users on update we have to patch configuration file to default to bdb.

components/network/openldap/files/ldap-olslapd
Mno-hime: Not going with this this time.

Mno-hime:
Updated PR summary, now it's about version bump, and clean up. 64-bit and other more complicated changes for the later.

PR #4660, opened 2018-12-20
mesa: explicitly link dri drivers to libglapi

pyhalov:


pyhalov:
@alanc - Does it look sane?

The issue is that swrast and other dri modules use _glapi_Dispatch, which is defined in libglapi.so. But libGL.so is linked to libglapi, so we don't see this. But when we use another libGL.so, not libGL.so (for example, one from VirtualBox), we have undefined symbols in swrast_dri.so

pyhalov:
Needs more testing.

PR #4670, opened 2018-12-21
MATE 1.20 update December 2018

Mno-hime:
~~Upstream keeps adding new releases of the 1.20 stream, let's wait till they are done with it and merge afterwards.~~

I think they are done with dot-updates for now.


components/desktop/mate/mate-applets/Makefile
```diff
@@ -76,8 +76,6 @@ REQUIRED_PACKAGES += library/glib2
 REQUIRED_PACKAGES += library/libgtop
 REQUIRED_PACKAGES += library/libnotify
 REQUIRED_PACKAGES += library/libxml2
-REQUIRED_PACKAGES += library/python/pygobject-3-27
```
pyhalov, line 79: Are you sure that this is not build deps?

Mno-hime: It actually is required, I brought it back. Thanks.

components/desktop/mate/mate-applets/Makefile
```diff
@@ -62,7 +62,8 @@ install:	$(INSTALL_64)
 
 test:		$(NO_TESTS)
 
-REQUIRED_PACKAGES += desktop/character-map/gucharmap
```
pyhalov: gucharmap is still a build dependency.

Mno-hime: Now it's back.

PR #4675, opened 2018-12-23
Update D-Bus to 1.12.12

Mno-hime:
illumos-gate rebuild went fine.


pyhalov:
Does illumos-gate still builds fine with this?

Mno-hime:
Yes.

PR #4677, opened 2018-12-24
Update libarchive to 3.3.3

Mno-hime:


components/library/libarchive/libarchive.p5m
```diff
@@ -54,10 +54,12 @@ link path=usr/lib/$(MACH64)/libarchive.so \
 file path=usr/lib/$(MACH64)/libarchive.so.1$(COMPONENT_VERSION)
 link path=usr/lib/$(MACH64)/libarchive.so.13 \
     target=libarchive.so.1$(COMPONENT_VERSION)
+file path=usr/lib/$(MACH64)/libarchive.so.13.3.2
```
pyhalov: is it really a product of previous build?

Mno-hime: Right, sorry, that was from the previous version.

PR #4679, opened 2018-12-25
rdesktop: update to latest head, enable CredSSP support

pyhalov:


pyhalov:
Updated rdesktop works if I have preinitialized kerberos ticket. Sometimes it fails to work even in this case - for example, when I connect to HV cluster shared address. But anyway this is better than old rdesktop.

Mno-hime:
FreeRDP might be worth a try.

PR #4681, opened 2018-12-28
Update OpenSSH to 7.9p1

Mno-hime:
* Update OpenSSH to 7.9p1

Tested with key-based authentication against v7.9p1 (itself), v7.2p2 (openSUSE Leap 42.3), and v6.7p1 (Debian).

* 10156 Remove deprecated options from sshd_config

Deprecation warnings on `sshd` start are gone.

* Add OmniOs patch for VMware guests with NAT networking

### DIFF
--- a/components/network/openssh/patches/0032-revert-dscp.patch
+++ b/components/network/openssh/patches/0032-revert-dscp.patch
@@ -1,0 +1,68 @@
+https://github.com/openssh/openssh-portable/commit/5ee8448ad7c306f05a9f56769f95
+is reverted in OmniOS as it causes problems for VMware guests where NAT mode
+is in use - see https://communities.vmware.com/thread/590825
+
+diff -wpruN '--exclude=*.orig' a~/readconf.c a/readconf.c
+--- a~/readconf.c	1970-01-01 00:00:00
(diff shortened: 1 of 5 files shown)


pyhalov (approved):

Mno-hime:
I run it for two weeks, removing the "testing needed" label.

components/network/openssh/patches/0032-revert-dscp.patch
```diff
@@ -0,0 +1,85 @@
+From cbbada124c90a23b78fd4c83f4c4ab9b1a2add23 Mon Sep 17 00:00:00 2001
```
pyhalov: Ooops... You don't patch OpenSSH, your patch creates patch to patch openssh.

pyhalov (changes requested):
Patch from OmniOS CE is not applied properly.

components/network/openssh/patches/0032-revert-dscp.patch
Mno-hime: Thanks for spotting that, now it applies as it should.

pyhalov (approved):

PR #4684, opened 2018-12-30
librsvg 2.40.20 dependecies rebuild

Mno-hime:
Dependencies rebuild after https://github.com/OpenIndiana/oi-userland/pull/4683.


pyhalov:
@Mno-hime Why do we need this? 
https://abi-laboratory.pro/?view=timeline&l=librsvg

Mno-hime:
I guess not needed after all.

PR #4688, opened 2018-12-31
Update glib to 2.58.2

Mno-hime:
Stream 2.58 is the last one with autotools configuration, further versions are Meson-only.

Documentation is not being built by default, it seems to save 30 MB, likely worth it.

ABI changes are likely compatible: https://abi-laboratory.pro/index.php?view=timeline&l=glib.


components/library/glib/patches/12-pfexec.patch
```diff
@@ -1,3 +1,7 @@
+(1) Although we have sys/auxv.h, we don't have getauxval(3), Linux does.
```
pyhalov: please, don't mix build patch and patch for fixing behavior.

components/library/glib/Makefile
```diff
@@ -37,13 +38,13 @@ PATH=$(PATH.gnu)
...
-COMPONENT_PREP_ACTION =	( cd $(@D) && autoreconf -f )
+COMPONENT_PREP_ACTION =	( cd $(@D) && gtkdocize && autoreconf -fi )
 
-COMPONENT_PRE_CONFIGURE_ACTION = ( cp -a $(SOURCE_DIR)/* $(@D))
+COMPONENT_PRE_CONFIGURE_ACTION = ( cp -a $(SOURCE_DIR)/* $(@D) )
 
 CONFIGURE_SCRIPT = $(@D)/configure
 
-# GCC 6
+# Fails with GCC 6+ in gio/gunixconnection.c
```
pyhalov: What does this comment mean? :))

PR #4690, opened 2019-01-01
Update fish shell to 3.0.0, build against system PCRE2, move to CMake

Mno-hime:
There are non-backward compatible changes: https://fishshell.com/release_notes.html.

Requires PCRE2 with 32-bit character support enabled for build against system PCRE2: https://github.com/OpenIndiana/oi-userland/pull/4689.

Build rewritten to CMake.

Filed following upstream tickets wrt v3.0.0:
- https://github.com/fish-shell/fish-shell/issues/5458
- https://github.com/fish-shell/fish-shell/issues/5461
- https://github.com/fish-shell/fish-shell/issues/5462
- https://github.com/fish-shell/fish-shell/issues/5463
- https://github.com/fish-shell/fish-shell/issues/5472


mqudsi:
Note that fish ships with its own copy of PCRE2 and can use that instead of a system distribution (assuming stock PCRE2 builds on OpenIndiana).

Mno-hime:
The change in our PCRE2 component is to enable it for `fish` (which is what I prefer to using the old fish-bundled one).

pyhalov (approved):

PR #4698, opened 2019-01-08
Update synergy to synergy core v1.10.1

benf33:
- build 64bits only
  - only core components (no GUI)
  - added changes from Michal Nowak comments

### DIFF
--- a/components/library/pcre2/Makefile
+++ b/components/library/pcre2/Makefile
@@ -28,11 +28,11 @@ PREFERRED_BITS=64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		pcre2
-COMPONENT_VERSION=	10.31
+COMPONENT_VERSION=	10.32
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.bz2
 COMPONENT_ARCHIVE_HASH=	\
-    sha256:e07d538704aa65e477b6a392b32ff9fc5edf75ab9a40ddfc876186c4ff4d68ac
+	sha256:f29e89cc5de813f45786580101aaee3984a65818631d4ddbda7b32f699b87c2e
 COMPONENT_ARCHIVE_URL=	ftp://ftp.csx.cam.ac.uk/pub/software/programming/pcre/$(COMPONENT_ARCHIVE)
 COMPONENT_SIG_URL=	$(COMPONENT_ARCHIVE_URL).sig
 COMPONENT_PROJECT_URL=  http://pcre.org/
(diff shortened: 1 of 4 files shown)


components/desktop/synergy/Makefile
```diff
@@ -34,28 +33,34 @@ include $(WS_MAKE_RULES)/cmake.mk
...
-CMAKE_OPTIONS += CMAKE_BUILD_TYPE=Release
+CMAKE_OPTIONS += -DCMAKE_BUILD_TYPE=Release
+CMAKE_OPTIONS += -DSYNERGY_BUILD_LEGACY_GUI:BOOL=OFF
+CMAKE_OPTIONS += -DSYNERGY_BUILD_LEGACY_SERVICE:BOOL=OFF
+CMAKE_OPTIONS += -DSYNERGY_BUILD_LEGACY_INSTALLER:BOOL=OFF
 
 COMPONENT_PREP_ACTION += ($(RM) $(COMPONENT_SRC)/src/test/CMakeLists.txt)
 
-MY_INSTALL_32 =            $(BUILD_DIR_32)/.my_installed
+MY_INSTALL_64 =            $(BUILD_DIR_64)/.my_installed
```
Mno-hime: You can as easily use `INSTALL_64` instead of `MY_INSTALL_64`. In such case you just re-define the default `INSTALL_64` one.

Mno-hime:
Apart from `MY_INSTALL_64`, LGTM.

Note that it's better to use local branch for changes, keeping `oi/hipster` (our "`master`") clean.

components/desktop/synergy/Makefile
benf33: If I replace the install target by
> install:     $(INSTALL_64)

make install fails miserably with
> make[1]: *** No rule to make target 'install'.  Stop.

Mno-hime: Did you replace all instances of `MY_INSTALL_64` with `INSTALL_64`? Not only the reference but also the definition. It worked for me.

benf33: In that case, it works.
I was trying to use a hypothetical pre-existing INSTALL_64 rather than having to redefine the INSTALL_64 target.

Mno-hime: I understand that. Not a blocker for me.

Mno-hime (approved):

components/desktop/synergy/Makefile
```diff
@@ -9,20 +9,19 @@
 #
 
 #
-# Copyright 2015 Alexander Pyhalov
+# Copyright 2019 benf33
```
pyhalov: Please, keep my Copyright and add your below. Also, copyright with nickname seems to be no-op to me (please, use full name).

PR #4701, opened 2019-01-08
10181 Update vte-291 to 0.54.3 

Mno-hime:


pyhalov:
Why do we want to rebuild mate-terminal?  vte 0.54.1 should be ABI-compatible to 0.54.3.

Mno-hime:
I couldn't find vte on ABI tracker, so just to be on safe side.

pyhalov:
Well, it's minor update to GNOME-supplied library. It doesn't change SONAME. It should be ABI-safe.

Mno-hime:
MATE Terminal rebuild is gone.

PR #4702, opened 2019-01-09
Bump RELEASE_MAJOR to 2019

Mno-hime:


pyhalov:
Do we want to bump RELEASE_MAJOR now or wait for really good reason to do this (by good reason I mean rebuild of large portion of components)?

Mno-hime:
Sure, np, lets bump it on a more important occasion.

PR #4703, opened 2019-01-09
Bring back OpenIndiana-specific colors to the loader

Mno-hime:
Recent update of loader with framebuffer sets colors to black on white, though traditional OpenIndiana colors are vice versa. This change brings `boot/loader.rc.oi` with OI-custom/former colors.

Tested in QEMU VM.

### DIFF
--- a/components/openindiana/illumos-gate/Makefile
+++ b/components/openindiana/illumos-gate/Makefile
@@ -583,7 +583,6 @@ $(BUILD_DIR)/$(MACH)/.published: $(BUILD_DIR)/$(MACH)/publish.transforms
 	$(RM) -r $(@D)/pkgrecv.dir
 	$(MKDIR) $(@D)/pkgrecv.dir
 
-
 ifeq ($(DEBUG),yes)
 	$(RM) -r $(@D)/pkgrepo-merged.dir
 	$(MKDIR) $(@D)/pkgrepo-merged.dir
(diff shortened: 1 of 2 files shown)


tsoome:
while this is certainly also one way, setting variables can be done via /boot/loader.conf or /boot/defaults/loader.conf, there you use syntax name=value.

You also can add them into /boot/solaris/bootenv.rc - this one is read before anything else.

The thing you want to consider is, /boot/loader.conf is searched for anyhow, loader.rc.oi will add extra disk access, also the /boot/loader.rc is meant for distribution and /boot/loader.rc.local for users.

pyhalov:
@Mno-hime  Why avoid delivering just custom /boot/loader.rc.local? 
Is it the issue of distribution-default vs user default? It seems to be consistent with what we do to /boot/loader.conf.

pyhalov:
/boot/loader.rc seems fine.

tsoome:
/boot/loader.rc.local should be left for users. For example, if the system does not provide EDID, it will be excellent place for framebuffer set  mode

Mno-hime:
Actually it wasn't clear to me which files belong to illumos and which to the distribution. I'll go with `/boot/loader.rc` then.

Mno-hime:
Here we go, tested with `0.5.11-2018.0.0.17771` (`illumos-9b40c3052b`).

tsoome (approved):

PR #4704, opened 2019-01-09
Rebuild gfile after build server is clean

Mno-hime:
I get following with official gfile builds:
```
$ gfile Xorg.0.log.old
gfile: compiled magic version [524] does not match with shared library magic version [535]
Xorg.0.log.old: JSON data
```
I think the build environment on the build server is polluted as my
build are OK. Clean the build environment on the build server before
merge, please.


pyhalov:
Performed clean and publish manually.

PR #4711, opened 2019-01-11
percona-server-57: fix build after https://www.illumos.org/issues/8548

pyhalov:
Needs fix for https://www.illumos.org/issues/10224 (likely we should install just updated headers on the build server to avoid dependency on newer libc).


Mno-hime (approved):
Verified with percona-server 5.7.24-27 build. Feel free to monkey-patch the build server's headers.

PR #4715, opened 2019-01-18
Percona 5.6.42-84.2 & 5.7.24-27

Mno-hime:
Tested service start, ping, status, etc. Internal test suite of v5.7 is unusable for me doe to core dumps (not a regression), v5.6 test suite looked good, though my setup brought some flip-flopping tests.


components/database/percona-server-56/Makefile
```diff
@@ -114,15 +112,16 @@ install:	$(INSTALL_32_and_64)
 
 test:		$(TEST_32_and_64)
 
+REQUIRED_PACKAGES += runtime/perl-522
 # Auto-generated dependencies
+REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
+REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
+REQUIRED_PACKAGES += database/percona-server-56/client
```
pyhalov: don't include this, we don't need mysql-client to build mysql.

components/database/percona-server-57/Makefile
```diff
@@ -35,6 +34,8 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/cmake.mk
 include $(WS_MAKE_RULES)/ips.mk
 
+GCC_VERSION = 7
```
pyhalov: Does it really require GCC 7 and cannot be patched to use GCC 6?

PR #4717, opened 2019-01-18
[WIP] cmake: set CMAKE_HOST_SYSTEM_PROCESSOR to `isainfo -k`

Mno-hime:
Previously `CMAKE_HOST_SYSTEM_PROCESSOR` was always set to `uname -p`, that is "i386". Some components like Percona Server 8.0's rockdb treat is as a 32-bit CPU and fail to compile.

This change sets `CMAKE_HOST_SYSTEM_PROCESSOR` to the more sane `isainfo -k`, i.e. "amd64".

Components including `cmake.mk` should be rebuild and tested if they are fine with this.


components/developer/cmake/patches/03-use-isainfo-for-CMAKE_SYSTEM_PROCESSOR.patch
```diff
@@ -0,0 +1,12 @@
+--- cmake-3.12.4/Modules/CMakeDetermineSystem.cmake	2019-01-13 08:55:16.128456978 +0000
++++ cmake-3.12.4/Modules/CMakeDetermineSystem.cmake.new	2019-01-13 09:01:53.015398514 +0000
+@@ -55,6 +55,9 @@ if(CMAKE_HOST_UNIX)
+     elseif(CMAKE_HOST_SYSTEM_NAME MATCHES "OpenBSD")
+       exec_program(arch ARGS -s OUTPUT_VARIABLE CMAKE_HOST_SYSTEM_PROCESSOR
+         RETURN_VALUE val)
++    elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "SunOS")
++      exec_program(isainfo ARGS -k OUTPUT_VARIABLE CMAKE_HOST_SYSTEM_PROCESSOR
```
pyhalov, line 8: Well, it behaves as described in manual: https://cmake.org/cmake/help/v3.12/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html

on Debian Linux it `uname -p` outputs "unknown". What decision can be based on this?

PR #4719, opened 2019-01-18
virutalbox: update to 5.2.24

pyhalov:


pyhalov:
Video recording was disabled earlier due to bug in configuration script. Needs patching to work correctly, so disable it explicitly for now.

PR #4734, opened 2019-01-25
samba: update to 4.9.4

chincheta0815:
Hope this new clean branch will work.
Sorry for the mess.


pyhalov:
```
ERROR userland.action002.0        link usr/bin/eventlogadm has unresolvable target '../lib/samba/bin/eventlogadm'
```
It seems eventlogadm has bin moved to /sbin/, please, fix link:  /usr/bin/eventlogadm should point to usr/lib/samba/sbin/eventlogadm. We can move link to /usr/sbin, but this can break compatibility for some scripts, I'm not sure if it's worth doing.

chincheta0815:
> 
> 
> ```
> ERROR userland.action002.0        link usr/bin/eventlogadm has unresolvable target '../lib/samba/bin/eventlogadm'
> ```
> 
> It seems eventlogadm has bin moved to /sbin/, please, fix link: /usr/bin/eventlogadm should point to usr/lib/samba/sbin/eventlogadm. We can move link to /usr/sbin, but this can break compatibility for some scripts, I'm not sure if it's worth doing.

You refer to the samba.p5m right?
Done.

Maybe the link change can be done some versions later. I left a comment.

components/network/samba/Makefile
```diff
@@ -248,7 +247,9 @@ clean::
 
 # Auto-generated dependencies
 REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
+REQUIRED_PACKAGES += database/lmdb
 REQUIRED_PACKAGES += library/file-monitor/gamin
+REQUIRED_PACKAGES += library/glib2
```
pyhalov: Don't see glib2 in required packages... Have you rechecked it with updated avahi?

chincheta0815: I am looking into that right now.

@pyhalov Concerning tracker support I checked other distro how these are managing the issue. They do package a whole tracker or and BETTER they packge tracker in parts, e.g. the libraries, binaries etc.. Is there a way to build an OI package and split it into parts: e.g. tracker-gui and tracker-libs? That would help me.

pyhalov: I suppose, we could split tracker in parts, so that samba depended only library parts and one could install tracker daemon separately if needed.

pyhalov:
@chincheta0815 Thanks and sorry that it took so long.

components/network/samba/Makefile
chincheta0815: > 
> 
> I suppose, we could split tracker in parts, so that samba depended only library parts and one could install tracker daemon separately if needed.

That's also my idea, but how do I do that? Is there a best practice? I saw there is a bunch of different emacs pkgs which seem to be be concurrent and built by one routine...

chincheta0815: @pyhalov I looked into the emacs packages. There are different ones made as I think it would be useful for tracker. As long as I do not have to split samba I think it is worth a try... I will look into tracker, splitting it into with and without the GUI stuff.

PR #4739, opened 2019-02-02
Update texinfo to 6.5

Mno-hime:


pyhalov:
IIRC, gcc depend on texinfo. Have you checked that they build fine (especially old ones, including openindiana/illumos-gcc)?

Mno-hime:
illumos-gcc needs a patch, will provide it later today here or to separate PR. GCC 4.9 and GCC 6 are fine. Can't test GCC 3.4 as dl.openindiana.org is down.

pyhalov:
dlc is up again

Mno-hime:
GCC 3.4 need the same kind of patch illumos-gcc does.

Mno-hime:
Fixes for GCCs in #4749.

PR #4740, opened 2019-02-02
Update bash to 5.0

Mno-hime:
At least `varenv` test looks odd in test suite output. `set -o posix` is set, so arguments like `-n` to `echo` are not treated as program options, but printed as such (see note on `xpg_echo`: https://www.gnu.org/software/bash/manual/html_node/Bash-POSIX-Mode.html), but is expected otherwise?


components/shell/bash/bash.p5m
```diff
@@ -28,21 +28,25 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
...
-file etc.bash.bashrc path=etc/bash/bashrc.example
-file etc.bash.inputrc path=etc/bash/inputrc.example
-file etc.skel.bashrc path=etc/skel/.bashrc
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-file path=usr/bin/bash
-file path=usr/bin/bashbug
-#
+legacy pkg=SUNWbash desc="GNU Bourne-Again shell (bash) version 4.2" \
```
pyhalov: Should we just globally drop legacy metadata?

PR #4743, opened 2019-02-04
Update GNU binutils to 2.32

Mno-hime:
Was able to build binutils with the new binutils and to build GCC 6 with updated binutils and then GCC 6 with the rebuild GCC.


pyhalov:
Does illumos-gate builds fine with new binutils?

Mno-hime:
Yes, it builds fine. So does illumos-gcc, GCC 4.9.

PR #4752, opened 2019-02-08
rdesktop: Fix 21 vulnerabilities

Mno-hime:
This commit includes fixes for a set of 21 vulnerabilities in rdesktop when a malicious RDP server is used.

Testing with Windows needed (I don't have any).


Mno-hime:
Tested with WS 2012 R2 and 2016.

PR #4760, opened 2019-02-14
Update fbreader to 0.99.4

Mno-hime:
FBreader now uses QT5 as GTK+ is not supported anymore.

New patches are from openSUSE.

Tested with couple of epubs.


pyhalov:
```
QPushButton::minimumSizeHint() const /jenkins/jobs/oi-userland/workspace/components/desktop/fbreader/build/i86/fbreader/../zlibrary/ui/libzlui.so
```

PR #4763, opened 2019-02-15
PostgreSQL 11.2, 10.7, 9.6.12, 9.5.16, and 9.4.21

Mno-hime:
https://www.postgresql.org/about/news/1920/


pyhalov:
I see that only manifest for PostgreSQL 9.4 is updated. Do other files need updating?

Mno-hime:
Forgot to push them, they are here now.

PR #4765, opened 2019-02-16
Remove obsoleted lint-libraries from build system

AndWac:
I had to re-generate sample-manifests for library/readline, library/trousers and text/tidy as in my build env there were some files missing.


components/library/openssl/openssl-1.0.2/Makefile
```diff
@@ -199,14 +199,8 @@ install:	$(INSTALL_32_and_64) $(PROTO_DIR)/.openssl-0.9.8
...
-LFLAGS_32 := -I$(BUILD_DIR_32)/include $(LINT_FLAGS) -lsoftcrypto
-LFLAGS_64 := -I$(BUILD_DIR_64)/include $(LINT_FLAGS) -lsoftcrypto
-
-# Set modified lint flags for our lint library targets.
-$(BUILD_DIR_32)/llib-lcrypto.ln: LINT_FLAGS=$(LFLAGS_32)
-$(BUILD_DIR_32)/llib-lssl.ln: LINT_FLAGS=$(LFLAGS_32)
-$(BUILD_DIR_64)/llib-lcrypto.ln: LINT_FLAGS=$(LFLAGS_64)
-$(BUILD_DIR_64)/llib-lssl.ln: LINT_FLAGS=$(LFLAGS_64)
+LFLAGS_32 := -I$(BUILD_DIR_32)/include -lsoftcrypto
+LFLAGS_64 := -I$(BUILD_DIR_64)/include -lsoftcrypto
```
Mno-hime: These variables does not seem useful anymore (and the comment above as well).

Mno-hime (commented):
Following files seems to me like a candidates for removal as well:
```
library/fastcgi/llib-lfcgi
library/openssl/openssl-1.0.2/llib-lcrypto
library/openssl/openssl-1.0.2/llib-lssl
library/gd2/llib-lgd
library/libxml2/llib-lxml2
library/lcms/llib-llcms
library/pcre/llib-lpcre
library/zlib/llib-lz
library/mpfr/llib-lmpfr
library/readline/llib-lreadline
library/gmp/llib-lgmp
library/trousers/llib-ltspi
library/libmng/llib-lmng
text/tidy/llib-ltidy
web/curl/llib-lcurl
```

components/library/openssl/openssl-1.0.2/Makefile
AndWac: Yes, I already did it but wasn't sure about it.
So you know the purpose of those .ln files?

AndWac: Maybe it would be better to create another PR with these changes because they don't seem to be related to the removal of the lint-libraries. What do you think?

AndWac:
Ok, I added the additional changes. I am not sure about what I have done for openssl, though.

components/library/openssl/openssl-1.0.2/Makefile
Mno-hime: Remove either way you like.

I don't know the purpose of those .ln files, but they are clearly related to lint, hence up for removat.

pyhalov: It seems, LDFLAGS_32 and LDFLAGS_64 here are only used to pacify lint and should be removed to avoid confusion.

pyhalov:
Why haven't you removed  LDFLAGS_32 and LDFLAGS_64 completely from openssl? They used only for lint rules.

components/library/openssl/openssl-1.0.2/Makefile
AndWac: I am not sure whether the comment above should completely removed or only the lint part.
The comment says that we use the install code only for man installation which is 32 bit only.
64 bit is only for lint. So in my understanding we could also omit 64 bit install.
What do you think?

AndWac:
I removed both lines in my latest commit but I am not sure about the comment: what is still relevant and what can be removed from it?

pyhalov:
what comment? And I still see lines with "LFLAGS_32 := -I$(BUILD_DIR_32)/include -lsoftcrypto" in PR

AndWac:
Sorry, forgot to git add Makefile before last push.

I meant these comments:
# We use OpenSSL install code for installing only manual pages and we do that
# for 32-bit version only.

and later:

# We could run OpenSSL install code for 32 bits only to process header files and
# manual pages. However, lint libraries depend on install stamps so we run
# install for 64 bit as well. Note that we must take built binary files from
# build directories, not from the proto area which contains whatever was
# installed first.
install:	$(INSTALL_32_and_64) $(PROTO_DIR)/.openssl-0.9.8

pyhalov:
I think we can transform
```
# We could run OpenSSL install code for 32 bits only to process header files and
# manual pages. However, lint libraries depend on install stamps so we run
# install for 64 bit as well. Note that we must take built binary files from
# build directories, not from the proto area which contains whatever was
# installed first.
```
to
```
# Note that we take built binary files from
# build directories, not from the proto area which contains whatever was
# installed first.
```
and 
```
# We use OpenSSL install code for installing only manual pages and we do that
# for 32-bit version only.
```
to 
```
# We use OpenSSL install code for installing only manual pages and headers
```
We could likely change
```
install:        $(INSTALL_32_and_64) $(PROTO_DIR)/.openssl-0.9.8
```
to 
```
install:        $(INSTALL_32) $(PROTO_DIR)/.openssl-0.9.8
```
but it's not related change and should likely be checked separately.

pyhalov (approved):

PR #4766, opened 2019-02-16
Add rename utility

Mno-hime:
This brings the `rename` utility I missed from Linux. I am happy to add others, but most obviously won't build as are Linux-specific.

Not packaging locales as they are 6 MB, which I find too much for a 20 KB utility.

### DIFF
--- a/components/meta-packages/install-types/Makefile
+++ b/components/meta-packages/install-types/Makefile
@@ -16,7 +16,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		install-types
 COMPONENT_VERSION=	0.1
-COMPONENT_REVISION=	31
+COMPONENT_REVISION=	32
 COMPONENT_SUMMARY=	A meta-packages that install common applications for ISOs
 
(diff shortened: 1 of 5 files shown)


components/sysutils/util-linux/Makefile
```diff
@@ -0,0 +1,51 @@
...
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		util-linux
+COMPONENT_MAJOR_VERSION=2.33
+COMPONENT_MINOR_VERSION=1
+COMPONENT_VERSION=	$(COMPONENT_MAJOR_VERSION).$(COMPONENT_MINOR_VERSION)
+COMPONENT_SUMMARY=	Various Linux utilities
+COMPONENT_PROJECT_URL=	https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git
+COMPONENT_FMRI=		file/util-linux
```
pyhalov: Why not call package file/rename?

pyhalov:
Please, can you change commit messages to 'Add rename utility'?

PR #4767, opened 2019-02-16
Replace watch with new implementation, gain UTF-8

Mno-hime:


components/sysutils/procps/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+PREFERRED_BITS=		64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		procps
+COMPONENT_VERSION=	3.3.15
+COMPONENT_SUMMARY=	Linux ps utilities
+COMPONENT_PROJECT_URL=	https://gitlab.com/procps-ng/procps
+COMPONENT_FMRI=		utilities/procps
```
pyhalov: at least utility/procps to be consistent, perhaps, better keep existing name as it just the same watch?

pyhalov:
You forgot to remove actual watch component. Perhaps, preserve the same fmri and component location, just switch source?

Mno-hime:
OK. Just replaced the component with new source.

pyhalov:
Hey, it doesn't build. Likely you forgot patches.

Mno-hime:
Yup, they are here now.

PR #4774, opened 2019-02-18
gcc: don't alter mpc/mpfr timestamp to avoid reconfiguring with wrong…

pyhalov:
… automake version


Mno-hime:
Looks like https://www.illumos.org/issues/9980.

pyhalov:
yes

PR #4777, opened 2019-02-19
gcc49: Remove elfedit-ing of libgfortran.so.3.0.0

Mno-hime:
`COMPONENT_POST_INSTALL_ACTION` actually silently fails when `elfedit`-ing of `libgfortran.so.3.0.0` (see the end of https://hipster.openindiana.org/logs/oi-userland/latest/developer.gcc49.publish.log). It's because GNU elfedit is picked.

We can easily "fix" it by providing full path to `elfedit`, but it's not clear to me why we need it at all. What it does, when fixed, is that `libgfortran.so.3.0.0` can't find `/usr/gcc/4.9/lib/libquadmath.so.0` anymore.

### DIFF
--- a/components/developer/gcc49/Makefile
+++ b/components/developer/gcc49/Makefile
@@ -18,7 +18,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= gcc
 COMPONENT_VERSION= 4.9.4
-COMPONENT_REVISION=5
+COMPONENT_REVISION=6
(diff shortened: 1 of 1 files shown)


pyhalov:
The issue is the following. GCC libs have embedded runpaths, but GCC 4.9 libs are delivered to /usr/lib/. So this runpath was unnecessary.

pyhalov (approved):

PR #4781, opened 2019-02-20
Update OpenVPN to 2.4.7, 64-bit-only

Mno-hime:
**Testing**
- Test suite passed
- Tested three examples in the `openvpn(1m)` man page against OpenVPN 2.4.6 on Linux


pyhalov:
Please, rebase and merge it.

PR #4786, opened 2019-02-22
Update bison to 3.3.2

Mno-hime:
Any idea how to test this?

@citrus-it Did you have any problems with bison 3.3?


citrus-it:
> Any idea how to test this?
> 
> @citrus-it Did you have any problems with bison 3.3?

The first release of 3.3 broke the gate build (see https://www.illumos.org/issues/10312) but 3.3.1 reverted the behaviour that caused that. I haven't seen any other issues with 3.3.x

Mno-hime:
Thanks @citrus-it!

I was able to build illumos-gate and old gcc3.

pyhalov:
bison is tricky - it's sometimes a hidden build-time dependency. Don't know a good way to test if it breaks something besides either rebuilding entire oi-userland or at least looking at components which were earlier broken by bison update - look at git log - these are webkit, gstreamer, ptlib.

PR #4792, opened 2019-02-26
Update git to 2.21.0

Mno-hime:
Tests now run with `PATH.gnu` as they increasingly fail with illumos tools.


pyhalov:
Is it test tooling failure or git is trying to use some illumos tools?

Mno-hime:
git test suite tries to use several illumos tools in a failing test case, which was introduced in 2.21.0.

PR #4794, opened 2019-02-27
Build Python 3 with -std=c99

pyhalov:


pyhalov:
This fixes inconsistency between Python 3 and Python 2 which appears when we try to use boolean constants with cffi. The inconsistency makes it hard to build IPS on OI with both python 2 and 3.

PR #4800, opened 2019-02-28
Deliver 64-bit-only Apache 2.4

pyhalov:
Second part of this PR breaks all our Apache modules, so it depends on module update...


pyhalov:
#4801 has to be merged immediately after this.
Also php modules should be rebuilt (TBD)

pyhalov:
Brief user update instructions:
1) Update apache 2.4 and modules
2) Inspect your conf and ini files - mod_event is used by default, 64bit is not set, so ensure to examine all IfDefine 64bit, IfModule prefork.c - these conditions are likely false now. Carefully compare /etc/apache2/2.4/httpd.conf and /etc/apache2/2.4/httpd.conf.new - modules*.load files have gone, modules load is happening in httpd.conf and /etc/apache2/2.4/conf.d/*.conf files. Be careful - on update most apache modules will not touch /etc/apache2/2.4/conf.d/*.conf files if they have been changed by hands.
3) Read apache24(1M), it describes apache24 service properties which can be used to manage service.

Mno-hime (approved):

jimklimov:
Do we generate post-upgrade notes files made prominent for people to review after `pkg update`? Alexander's post above should end up in those too, then.

FWIW, in discussions I advocated for not replacing the existing apache config with a new empty one - having the service broken right after upgrade is the least surprise here, compared to it seemingly loaded with an empty config and then users complaining that the site is wrong.

pyhalov:
I don't know how to do it in IPS in a sane way. The only thing which comes to mind is to consider such a noitce 'a license' and force user to accept it.

PR #4806, opened 2019-03-02
mdata-client: add mdata-client release-20190228

jclulow:
This adds a binary package for the metadata client utilities that SmartOS HVM guests can use to get information from the hypervisor (or the cloud orchestration suite managing the hypervisor).

I built and installed this package in one of my OpenIndiana VMs running under KVM on a SmartOS host:

```
root@oi0:~# pkg set-publisher -g file:///ws/oi-userland/i386/repo userland
root@oi0:~# pkg install mdata-client
           Packages to install:  1
       Create boot environment: No

Completed                                1/1           9/9      0.1/0.1    0B/s

Installing new actions                         22/22
Reading search index                            Done 
Updating search index                            1/1 

root@oi0:~# man mdata-list | head -20
MDATA-LIST(1m)                 Metadata Commands                MDATA-LIST(1m)






NAME
       mdata-list -- Fetch the list of available metadata key-value pairs.


SYNOPSIS
       /usr/sbin/mdata-list


DESCRIPTION
       The mdata-list command allows the user (or a script) to query the
       metadata for a guest instance running in a SmartDataCenter (SDC) cloud.
       Metadata values are generally set programmatically via CloudAPI, or
       during interactive provisioning via a Web Portal.  They may also be set

root@oi0:~# mdata-list

root@oi0:~# mdata-put date <<< $(date)

root@oi0:~# mdata-list
date

root@oi0:~# mdata-get date
March  2, 2019 at 02:05:10 PM PST

root@oi0:~# mdata-get sdc:alias
oi0

root@oi0:~# mdata-get sdc:server_uuid
36343330-3836-5553-4533-3238384d5053
```

I tried to make this look as much like the other Makefiles nearby as I could.  Please let me know if I need to fix anything!  Thanks.

### DIFF
--- a/components/sysutils/mdata-client/Makefile
+++ b/components/sysutils/mdata-client/Makefile
@@ -1,0 +1,44 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2019 Joyent, Inc.
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		mdata-client
+COMPONENT_VERSION=	20190228
+COMPONENT_SUMMARY=	Cross-platform metadata client tools for use in Joyent Triton guests (both Zones and KVM)
+COMPONENT_PROJECT_URL=	https://github.com/joyent/mdata-client
+COMPONENT_FMRI=		system/virtualization/mdata-client
+COMPONENT_CLASSIFICATION= Applications/System Utilities
+COMPONENT_SRC=		$(COMPONENT_NAME)-release-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=	https://github.com/joyent/mdata-client/archive/release-$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:1165ddb2c81139320e184577889ce3548416aef4835e26ad39d8e0f2d9f28135
+COMPONENT_LICENSE=	MIT
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_BUILD_ARGS +=	CC='$(CC)'
+
+build:		$(BUILD_32)
+
+install:	$(INSTALL_32)
+
+test:		$(NO_TESTS)
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += system/library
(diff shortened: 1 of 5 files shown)


richlowe:
Probably should refer to it as Triton, not SDC.  I hear the cool kids do.

jclulow:
> Probably should refer to it as Triton, not SDC. I hear the cool kids do.

Fixed.  Thanks.

pyhalov:
Can you set/overwrite $(CC) so that userland compiler is certainly used, not just gcc? 
Also, please do "gmake sample-manifest" and add resulting file.

components/sysutils/mdata-client/Makefile
```diff
@@ -0,0 +1,41 @@
...
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+build:		$(BUILD_32)
+
+install:	$(INSTALL_32)
+
+test:		$(NO_TESTS)
+
+REQUIRED_PACKAGES += system/library
```
Mno-hime, line 44: Also run `gmake REQUIRED_PACKAGES`, so it's clear in the Makefile, what dependencied the build system is able to find on it's own. You'll get something like:
```Makefile
# Auto-generated dependencies
REQUIRED_PACKAGES += ...
```
Whatever is needed, but not found automatically, move above the comment.

jclulow: > Can you set/overwrite $(CC)

Done!  The software appears to build correctly, now using a fully qualified `/usr/gcc/6/bin/gcc`.

> please do "gmake sample-manifest" and add resulting file

I will do this, but it's not immediately clear to me what this file is for!  Can you point me to a document that explains it?

> run `gmake REQUIRED_PACKAGES`, so it's clear in the Makefile, what dependencied the build system is able to find on it's own

Done!  It's still just `system/library` but I now have the `# Auto-generated dependencies` banner.

Mno-hime: > > please do "gmake sample-manifest" and add resulting file
> 
> I will do this, but it's not immediately clear to me what this file is for! Can you point me to a document that explains it?

It's a good starting point for the actual mdata-client.p5m component file and on version update one can see via `git diff`, what new files were installed (or were removed) and decide if they should be put into the package.

E.g. see https://github.com/OpenIndiana/oi-userland/pull/4818/files. On version update three new files emerged in the sample manifest (reflects the install area) and then I decided to put them to the right component p5m file.

> Done! It's still just `system/library` but I now have the `# Auto-generated dependencies` banner.

That banner is a good sign that someone actually run the `REQUIRED_PACKAGES` target and that the list of required packages is more or less complete.

components/sysutils/mdata-client/Makefile
```diff
@@ -0,0 +1,44 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2019 Joyent, Inc.
+#
+
+include ../../../make-rules/shared-macros.mk
```
Mno-hime, line 16: Adding `PREFERRED_BITS=64` before the include and changing common targets `s/32/64/` makes sure 64-bit binary is build and placed to the (usually 32-bit) area of `/usr/bin/`. E.g. see `components/archiver/gtar/` component.

Any reason to go with 32-bit binaries? We prefer 64-bit where possible, certainly for new stuff.

jclulow: At the moment, the software is only built and shipped 32-bit on any platform.  The `Makefile` doesn't currently offer a way to add different `CFLAGS` in a useful way.  I can try and fix that upstream, and then come back to improve the package here, but I'd like to get these binaries shipped as 32-bit first.

Mno-hime (approved):

PR #4807, opened 2019-03-03
Added Freetype extention support to php gd extention

Toasterson:
Additionally fix dependencies to compile
Note the pkg.depend.bypass-generate=.* may be a problem with my build system. It was not finding libpq

### DIFF
--- a/components/web/php/php-7_0/Makefile
+++ b/components/web/php/php-7_0/Makefile
@@ -17,6 +17,7 @@ COMPONENT_NAME=		php
 COMPONENT_VERSION=	7.0.33
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.xz
+COMPONENT_REVISION=	1
 COMPONENT_ARCHIVE_HASH=	sha256:ab8c5be6e32b1f8d032909dedaaaa4bbb1a209e519abb01a52ce3914f9a13d96
 COMPONENT_ARCHIVE_URL=	http://uk1.php.net/get/$(COMPONENT_ARCHIVE)/from/this/mirror
 COMPONENT_SIG_URL=	http://uk1.php.net/get/$(COMPONENT_ARCHIVE).asc/from/this/mirror
@@ -109,14 +110,14 @@ CONFIGURE_OPTIONS     +=	--with-pdo-mysql=shared,$(MYSQL_HOME)
 CONFIGURE_OPTIONS     +=	--with-pgsql=shared,$(PG_BINDIR)
 
 # Enabled Extensions
-#CONFIGURE_OPTIONS     +=	--with-libxml-dir=$(ECPREFIX)
-CONFIGURE_OPTIONS     +=	--with-freetype-dir=$(ECPREFIX)
+#CONFIGURE_OPTIONS     +=	--with-libxml-dir=/usr
+CONFIGURE_OPTIONS     +=	--with-freetype-dir=/usr
 CONFIGURE_OPTIONS     +=	--with-jpeg-dir=/usr
-#CONFIGURE_OPTIONS     +=	--with-png-dir=$(ECPREFIX)
-#CONFIGURE_OPTIONS     +=	--with-zlib-dir=$(ECPREFIX)
-#CONFIGURE_OPTIONS     +=	--with-xpm-dir=$(ECPREFIX)
-#CONFIGURE_OPTIONS     +=	--with-libexpat-dir=$(ECPREFIX)
-#CONFIGURE_OPTIONS     +=	--with-openssl-dir=$(ECPREFIX)
+#CONFIGURE_OPTIONS     +=	--with-png-dir=/usr
+#CONFIGURE_OPTIONS     +=	--with-zlib-dir=/usr
+#CONFIGURE_OPTIONS     +=	--with-xpm-dir=/usr
+#CONFIGURE_OPTIONS     +=	--with-libexpat-dir=/usr
+#CONFIGURE_OPTIONS     +=	--with-openssl-dir=/usr
 
 # Disabled extensions & Options
 CONFIGURE_OPTIONS     +=	--disable-static
--- a/components/web/php/php-7_0/Makefile
+++ b/components/web/php/php-7_0/Makefile
@@ -202,12 +202,13 @@ install:	$(INSTALL_32_and_64)
 
 # Build dependencies
 REQUIRED_PACKAGES += web/server/apache-24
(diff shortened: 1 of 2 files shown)


components/web/php/php-7_0/Makefile
```diff
@@ -110,7 +110,7 @@ CONFIGURE_OPTIONS     +=	--with-pgsql=shared,$(PG_BINDIR)
 
 # Enabled Extensions
 #CONFIGURE_OPTIONS     +=	--with-libxml-dir=$(ECPREFIX)
-#CONFIGURE_OPTIONS     +=	--with-freetype-dir=$(ECPREFIX)
+CONFIGURE_OPTIONS     +=	--with-freetype-dir=$(ECPREFIX)
```
pyhalov: 1) ECPREFIX is not defined, it's a legacy derived from ec-userland, please, fix it with correct value. 
2) Please, bump COMPONENT_REVISION

components/web/php/php-7_0/php70-ext-pgsql.p5m
```diff
@@ -23,5 +23,5 @@ set name=info.classification \
 license LICENSE license="PHP56 License"
 
 file pgsql.ini path=/etc/php/7.0/conf.d/pgsql.ini preserve=true
-file path=/usr/php/7.0/extensions/$(MACH64)/pgsql.so
-file path=/usr/php/7.0/extensions/pgsql.so
+file path=/usr/php/7.0/extensions/$(MACH64)/pgsql.so pkg.depend.bypass-generate=.*
```
pyhalov: Do you have postgresql94 installed?

Toasterson: Yes is that the problem? I know that in some of my zones  python ffi is broken so I thought it might be that problem (to which I have not found a solution)

components/web/php/php-7_0/Makefile
Toasterson: ? I noticed ECPREFIX as an odd variable but interesting that it worked. I replaced it with /usr directly as with the jpeg dir done by the previous contributor. Is that Correct? Or would you like another prefix Variable

components/web/php/php-7_0/php70-ext-pgsql.p5m
pyhalov: No. Component is looking for $(PG_BINDIR)/pg_config, and PG_BINDIR is /usr/postgres/9.4/bin or /usr/postgres/9.4/bin/amd64

Toasterson: Ah yes Zabbix is looking for pg_config and that works. pkgsend dependency resolv however complains. And I guess that is Python FFI.

pyhalov: How exactly does it complain?

pyhalov: @Toasterson works here without pkg.depend.bypass-generate=.* .

Toasterson: @pyhalov Look below the conversation :)

components/web/php/php-7_0/Makefile
```diff
@@ -206,6 +207,7 @@ REQUIRED_PACKAGES += web/server/apache-24
 REQUIRED_PACKAGES += SUNWcs
 REQUIRED_PACKAGES += compress/bzip2
 REQUIRED_PACKAGES += database/mariadb-101/library
+REQUIRED_PACKAGES += database/mariadb-101/client
```
pyhalov: Why mysql client is needed?

components/web/php/php-7_0/php70-ext-pgsql.p5m
Toasterson: This is what I get. As said I believe it is my build zone. However I have not found a way to fix it and it probably will happen again if I recreate the zone.
```
/root/userland/components/web/php/php-7_0/build/manifest-i386-php70-ext-pgsql.depend has unresolved dependency '
        pkg.debug.depend.reason=usr/php/7.0/extensions/pgsql.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.fullpath=lib/libpq.so.5 \
        pkg.debug.depend.fullpath=usr/gcc/6/lib/libpq.so.5 \
        pkg.debug.depend.fullpath=usr/mariadb/10.1/lib/libpq.so.5 \
        pkg.debug.depend.fullpath=usr/postgres/11/lib/amd64/libpq.so.5'.
/root/userland/components/web/php/php-7_0/build/manifest-i386-php70-ext-pgsql.depend has unresolved dependency '
        pkg.debug.depend.reason=usr/php/7.0/extensions/amd64/pgsql.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.fullpath=lib/64/libpq.so.5 \
        pkg.debug.depend.fullpath=usr/gcc/6/lib/amd64/libpq.so.5 \
        pkg.debug.depend.fullpath=usr/mariadb/10.1/lib/amd64/libpq.so.5 \
        pkg.debug.depend.fullpath=usr/postgres/11/lib/amd64/libpq.so.5'.
gmake: *** [/root/userland/make-rules/ips.mk:439: /root/userland/components/web/php/php-7_0/build/.resolved-i386] Error 1
```

components/web/php/php-7_0/Makefile
```diff
@@ -226,3 +228,5 @@ REQUIRED_PACKAGES += text/aspell
 REQUIRED_PACKAGES += text/tidy
 REQUIRED_PACKAGES += web/curl
 REQUIRED_PACKAGES += web/php-70/php-cli
+REQUIRED_PACKAGES += system/library/freetype-2
```
pyhalov, line 232: Please, add mariadb-101/client and developer/icu above "# Auto-generated dependencies" line, as they are not found by pkgdepend

components/web/php/php-7_0/php70-ext-pgsql.p5m
pyhalov: This means it somehow finds and uses pg_config from PostgreSQL 11, not one from PG_BINDIR. What does gmake print-PG_BINDIR say? Do you have /usr/postgres/9.4/bin/pg_config ?

Toasterson: Ah ok I get the following. ANd Yes I have postgres11 installed. Should I try with removed postgres11?
```
gmake print-PG_BINDIR
PG_BINDIR=/usr/postgres/9.4/bin (origin: file, flavor: recursive)
```

pyhalov: No

PR #4813, opened 2019-03-08
Add xbacklight

Mno-hime:
Tested on Lenovo X230 (although it seems that MATE resets brightness back to 100 % in couple of seconds).


pyhalov:
Need to check if it's work of mate-power-manager

PR #4825, opened 2019-03-11
1/2: Update libidn2 to 2.1.1

Mno-hime:
ABI Tracker: https://abi-laboratory.pro/index.php?view=timeline&l=libidn2.

gnutls-3 needed a fix to link as some API calls wen't private in libidn2.

Testing:
* curl builds and IDN works,
* wget builds, test suite (with IDN test cases) passed, IDN works,
* gnutls-3 builds, test suite passed


pyhalov:
Should we split this PR in 2?

Mno-hime:
Yes, will split it.

Mno-hime:
Here's the second part: https://github.com/OpenIndiana/oi-userland/pull/4832.

PR #4826, opened 2019-03-11
Update coreutils to 8.31

Mno-hime:
Release notes: https://savannah.gnu.org/forum/forum.php?forum_id=9394.


Mno-hime:
Added upstream patch so tests don't regress.

pyhalov (approved):

PR #4828, opened 2019-03-11
2/2: libidn rebuild (VLC, pidgin, libgsasl, Lynx)

Mno-hime:
Merge after https://github.com/OpenIndiana/oi-userland/pull/4827.

Updated Lynx to latest release, made it 64-bit. Test suite passed. https://www.howsmyssl.com/ report "Probably Good", IDN works.

### DIFF
--- a/components/meta-packages/history/history
+++ b/components/meta-packages/history/history
@@ -1109,6 +1109,7 @@ system/management/visual-panels/panel-time@0.5.11,5.11-2015.0.2.0
 system/management/visual-panels/panel-usermgr@0.5.11,5.11-2015.0.2.0
 system/management/visual-panels@0.5.11,5.11-2015.0.2.0
 system/management/wbem/resource-management@0.5.11,5.11-2015.0.2.0
+system/management/web/openwsman@2.1.0-2018.0.0.0
 system/management/webmin@1.510,5.11-2014.0.1.0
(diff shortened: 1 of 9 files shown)


pyhalov:
libgsasl also needs rebuilding.
system/management/web/openwsman should be either rebuilt or dropped (likely, dropped)

Mno-hime:
> system/management/web/openwsman should be either rebuilt or dropped (likely, dropped)

I don't see this one on oi-userland. ?

pyhalov:
It is legacy package, inherited from OI /dev, it has been never added to oi-userland.

components/meta-packages/history/history
```diff
@@ -1109,6 +1109,7 @@ system/management/visual-panels/panel-time@0.5.11,5.11-2015.0.2.0
 system/management/visual-panels/panel-usermgr@0.5.11,5.11-2015.0.2.0
 system/management/visual-panels@0.5.11,5.11-2015.0.2.0
 system/management/wbem/resource-management@0.5.11,5.11-2015.0.2.0
+system/management/web/openwsman@2.1.0-2013.0.0.0
```
pyhalov: Mark it with  2018.0.0.0, as 2013 is reserved for 'not-rebuilt legacy'

PR #4839, opened 2019-03-13
geany:bump to 1.34.1

hoewweken:
As noted, update geany to Version 1.34.1
No special reasons, compiles and works as expected.


Mno-hime (changes requested):
Thanks for your contribution @hoewweken!

This branch is polluted with Firefox 60est commits. Clean-up your `oi/hipster` branch first to match `oi/hipster` from `OpenIndiana/oi-userland` repo, then create a new branch of it and add your changes and push. That will require a new PR.

Did you run `gmake sample-manifest`? What did you test?

Also, provide PR for geany plugins, they got updated upstream as well.

PR #4893, opened 2019-03-18
Update open-vm-tools to 10.3.10

Mno-hime:
Release notes: https://github.com/vmware/open-vm-tools/blob/master/ReleaseNotes.md


Mno-hime:
I don't have VMware around anymore. I suggest we merge this.

PR #4900, opened 2019-03-20
10228 GTK3 Color Selection cores application

pyhalov:


pyhalov:
@iigs - Does it fix the issue with mate-color-select?

iigs:
The gtk3 color picker dropper now works correctly with Mate Color Selection and Firefox 60.6.  Software does not crash.

PR #4901, opened 2019-03-20
samba: bugfix avahi_register.c

chincheta0815:
There was an issue with the discovery of the fruit/apple info using avahi.


Mno-hime:
Can you provide more information on what the actual problem is? Commented reproducer will do. I don't see the change upstream. Is there some discussion about it?

Those two commits should be squashed.

chincheta0815:
On OpenIndiana a propagation of device-info and adisk are propagated over avahi/mdns via port 0. This does not work (maybe on OpenIndiana only?) and, more important, even MacOS uses port 9. 

device-info and adisk only serve TXT info. No real service. There seems to be no communication on the propagated ports.

Doing the publish via dns-sd does not work too. So it seems to be a kind of mdns specification thing. Having in mind that even Appel does use port 9 which is "network /dev/null" makes this patch reasonable. Also, the feature can be switched off, if there are worse problems.

I started the discussion in the samba-technical mailing list.

I will squash them, of course.

Language understanding thing: What do you mean by commented reproducer will do?

pyhalov:
@tsoome: Toomas, do you have some comments on it? 
The discussion thread on Samba ML has no comments from Samba devs yet - https://lists.samba.org/archive/samba-technical/2019-March/132972.html . 
@chincheta0815 Please, file a bug at https://www.illumos.org/projects/openindiana/issues, describing the issue and testing so that we could find out later why this patch is added. Also make sure that samba devs aware of issue (file a bug?).

After some reading the change seems sane to me.

chincheta0815:
> @tsoome: Toomas, do you have some comments on it?
> The discussion thread on Samba ML has no comments from Samba devs yet - https://lists.samba.org/archive/samba-technical/2019-March/132972.html .

I will try too keep in touch with the samba devs as I like this avahi thing be unpatched here and officially checked. 

Only thing I did not find out if its some special or buggy thing to Solaris/OI and its avahi/mdns implementation as all I found about Linux seems to work with "device-info" on port "0" and "adisk" on port "9" (e.g. https://www.bootc.net/archives/2010/11/07/apple-time-machine-and-netatalk/). But there are also ones without any ports at all and my Macbook uses e.g. port 311, my time capsule port 9. So there should be something special in OI.

Nevertheless, as IANA states that for device-info is no service and that only the TXT is needed I think that there won't be any communication at all there on the port mentioned and for adisk it seems to be the same stuff.  

> @chincheta0815 Please, file a bug at https://www.illumos.org/projects/openindiana/issues, describing the issue and testing so that we could find out later why this patch is added. Also make sure that samba devs aware of issue (file a bug?).
> 
> After some reading the change seems sane to me.

I will file a bug. I already wrote to samba-technical waiting for their statement.

tsoome:
> @tsoome: Toomas, do you have some comments on it?
> The discussion thread on Samba ML has no comments from Samba devs yet - https://lists.samba.org/archive/samba-technical/2019-March/132972.html .
> @chincheta0815 Please, file a bug at https://www.illumos.org/projects/openindiana/issues, describing the issue and testing so that we could find out later why this patch is added. Also make sure that samba devs aware of issue (file a bug?).
> 
> After some reading the change seems sane to me.

I do not use samba myself but in kernel smb, and I have this config:

$ more /etc/avahi/services/smb.service 
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE service-group SYSTEM "avahi-service.dtd">
<service-group>
    <name replace-wildcards="yes">%h.local</name>
    <service>
        <type>_smb._tcp</type>
        <port>445</port>
    </service>
    <service>
        <type>_device-info._tcp</type>
        <port>0</port>
        <txt-record>model=Xserve</txt-record>
    </service>
</service-group>
```
$

Apparently it is working… at least Finder is suggesting this server is Mac:D (the model there is set for icon only)

tsoome:
btw, I was starting to wonder why you have _adisk there, and indeed, _adisk is not generic smb, but time machine: https://jonathanmumm.com/tech-it/mdns-bonjour-bible-common-service-strings-for-various-vendors/

Of course this means the _adisk should only be announced for time machine share.

chincheta0815:
@tsoome Thanks  for your information on the avahi service. I tried the same with using dns-sd (see the samba-technical thread: https://lists.samba.org/archive/samba-technical/2019-March/132972.html) This did actually not work, there was no logo in my Finder. I just used the entries you did.

The avahi_register part in samba does not have any effect. So there seems to be a problem somewhere. Maybe with avahi itself(?) as there was an update in January in OI.

The thing with adisk is the following: It propagates a defined share for Time Machine backups to MacOS making your server a Time Capsule (just choose the right device-info model). Such an entry might look like: "_adisk._tcp 9 sys=waMA=b4:99:ba:bf:46:4e,adVF=0x100 dk0=adVF=0x83,adVN=Time Machine". For in-kernel smb it won't work properly as this needs FULLSYNC which is an Apple extension. Using samba from samba.org has all that implemented under the "fruit" vfs. Just add it in the config and samba propagates everything right. But as of now there is an issue with the avahi-part I am trying to fix.

Could you help to find the issue by doing dns-sd etc. and send that to samba-technical into the mailing list? Or put it here? I already tried different OI installtions and it really seems to be something like a bug. I do not know where.

chincheta0815:
@tsoome Bug is filed: https://www.illumos.org/issues/10577

chincheta0815:
I close this due to the WIP in samba-4.10.0

PR #4902, opened 2019-03-20
gcc-6: update to 6.5.0

alarcher:


components/developer/gcc-6/patches/use-cxa_atexit.patch
```diff
@@ -1,11 +0,0 @@
---- gcc-6.4.0/gcc/config.gcc.~1~	2017-01-10 00:01:26.582656000 +0000
```
pyhalov, line 1: Is this patch no longer necessary?

alarcher: It is not necessary since https://github.com/OpenIndiana/gcc/blob/il-6_5_0/gcc/config.gcc#L853

PR #4909, opened 2019-03-21
Mate 1.22 part3

Mno-hime:
Faenza icon theme wasn't released - so removing it.


pyhalov:
Where is part2? :)

Mno-hime:
https://github.com/OpenIndiana/oi-userland/pull/4910 :).

PR #4916, opened 2019-03-26
Update nano to 4.0

Mno-hime:
Update GNU nano to 4.0.

Add global `/etc/nanorc` with syntax highlighting on.

Highlighting according to file content (via libmagic) does not work, reported here: http://lists.gnu.org/archive/html/nano-devel/2019-03/msg00047.html.

### DIFF
--- a/components/editor/nano/nano.p5m
+++ b/components/editor/nano/nano.p5m
@@ -11,6 +11,7 @@
 
 #
 # Copyright 2017 Adam Stevko 
+# Copyright 2019 Michal Nowak
 #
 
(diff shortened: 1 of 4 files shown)


components/editor/nano/nano.p5m
```diff
@@ -22,6 +23,7 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+file path=etc/nanorc
```
pyhalov: preserve=true?

PR #4921, opened 2019-03-29
Build with pulse audio support on hipster.

timmooney:
Add a couple missing build dependencies to the Makefile.


Mno-hime (commented):
You should also add `COMPONENT_REVISION=1`.

Mno-hime:
Does it also work when you add `VBOX_WITH_AUDIO_PULSE = 1` to `LOCAL_CONFIG.64` instead of the patch?

timmooney:
Hi Michal!

I didn't test adding that define to LOCAL_CONFIG.64, though I could test it if you think that's better for some reason.

I *did* test using the check_pulse() configure check, but like --enable-pulse, check_pulse() doesn't actually add the VBOX_WITH_AUDIO_PULSE define to any config, so it's ineffective.  Both of those things seem like bugs (configuration code that was never finished).  When building packages for OI and you find what looks like a bug in upstream code, do you typically follow up with upstream, or just patch for OI?

I can submit a revised pull request with COMPONENT_REVISION included, once I have your response on the other questions.

timmooney:
Here's an updated pull with

- Michal's suggestion of not patching AutoConfig.kmk and instead just writing VBOX_WITH_AUDIO_PULSE=1 to the local config.

- added COMPONENT_REVISION = 1 (for just system/virtualbox, not for virtualbox-additions)

Since I didn't hear back about whether I should also pursue the general issue with upstream, I went ahead and opened a ticket in the main VirtualBox bugtracker with some suggestions for fixes:

    https://www.virtualbox.org/ticket/18544

Please review and let me know if additional changes are required.

Thanks

Mno-hime:
@timmooney I like the aproach now, thanks!

Could you add the upstream ticket link somewhere close to the reasoning in the Makefile on why `--enable-pulse` is not needed, please? And feel free to add any patches to the upstream ticket as sometimes it's easier to understand what is needed from patch than from plain words.

pyhalov (approved):

timmooney:
Sure thing.  I'll get the oi-userland patch updated with your suggestion and see about submitting a patch to upstream, to go with my bug report.  You should see an updated pull request within 24 hours.

PR #4922, opened 2019-04-01
9732 Pulseaudio always uses the first audio device

pyhalov:
9655 Both HAVE_OSS_OUTPUT and HAVE_SOLARIS are defined in pulseaudio


pyhalov:
For now I open this PR for discussion only, not for integration.

iigs:
I applied this PR on a system which did not have working Pulse audio.  After reboot, I get audio, but MATE sound control has no effect on audio volume.

Any insight what would happen when  this bug is fixed: https://www.illumos.org/issues/10619 .  The expected behavior is that the user can independently adjust volume for both the "onboard" audio and HDMI audio (the use case I have seen is that the user manually mutes one device and sets the volume for the other).

AndWac:
superceded by #6093

PR #4923, opened 2019-04-02
samba: fix avahi propagation

chincheta0815:
As I am struggling with samba-4.10.0 it would be great to have that patch in.

So here is the original PR I closed since I thought samba-4.10 would be easy...: https://github.com/OpenIndiana/oi-userland/pull/4901


components/network/samba/manifests/sample-manifest.p5m
```diff
@@ -1,27 +1,3 @@
-#
```
pyhalov: Don't touch this file, it didn't change.

components/network/samba/samba.proto-objects
```diff
@@ -1,10 +1,4 @@
-dir  path=etc
```
pyhalov: The same goes for this one.

components/network/samba/patches/11-avahi-register.patch
```diff
@@ -0,0 +1,20 @@
+--- samba-4.9.5/source3/smbd/avahi_register.c	2019-03-12 09:19:04.000000000 +0000
```
pyhalov, line 6: Please, add reference to https://www.illumos.org/issues/10577 on the top of this patch.

pyhalov:
Commit message should be "samba: fix avahi propagation"

components/network/samba/manifests/sample-manifest.p5m
chincheta0815: Thanks. I wondered why it should have changed, but this time I at least did not forget to create that manifest... ;o)

components/network/samba/samba.proto-objects
chincheta0815: Same as above.

components/network/samba/patches/11-avahi-register.patch
chincheta0815: Will do that.

chincheta0815:
Will do the changes and then quash and push...

chincheta0815:
@pyhalov I hope I did everything well. Concerning samba-4.10: Could you provide me some help or "infopoint"? Samba-devs seem to have different topics on their agenda....

PR #4908, opened 2019-03-21
Mate 1.22 part1

Mno-hime:
I started every GUI application visible from the MATE menu and basic functionality is ensured.

However, there seems to be some problems with icons in the default Nimbus theme. Some icons may be missing (Rhythmbox, gThumb, gtkam), "Help" icon is too big so that buttons in the row are too big ("Add to panel"), odd theming on checkbox in context menu ("Lock icon"). See the screenshots.

rhythmbox
gthumb
gtkam

help-icon-nimbus

lock-to-panel

PR #4926, opened 2019-04-10
mate-panel: update to 1.22.0

pyhalov:
We have an issue - when mate applications menu is redrawn, our modification to it ("run application") is lost (for example, when desktop file is add to /removed from "/usr/share/applications"). So, earlier we set up
- a monitor which on mate-applications.menu change sets corresponding flag - needs-appending
- show signal callback to append necessary widgets.

Now after https://github.com/mate-desktop/mate-menus/commit/524d3e4fc324c2cbc2454e8b1fcddc6622dc4714 we've lost ability to set such monitor (or at least I don't see an easy way to do it), but we don't need to - we just set this flag for every modification of every menu. As others menus don't have corresponding show handler, everything works as expected.


Mno-hime:
Thanks, I cherry picked it to https://github.com/OpenIndiana/oi-userland/pull/4908 (MATE 1.22 1/5) as there's dependency on mate-panel in https://github.com/OpenIndiana/oi-userland/pull/4909/files (MATE 1.22 3/5).

Mno-hime (approved):
The "About OpenIndiana" item works. Integrated to MATE 1.22 PRs. Will close this one.

PR #4930, opened 2019-04-12
Update gawk to 5.0.0

Mno-hime:
Release notes: http://lists.gnu.org/archive/html/info-gnu/2019-04/msg00002.html

Studio problem was resolved so patch was dropped.

Gawk extemsion modules are installed to `/usr/share/man/man3/` with `.3am` suffix.


components/text/gawk/gawk.p5m
```diff
@@ -84,6 +88,17 @@ file path=usr/share/locale/sv/LC_MESSAGES/gawk.mo
 file path=usr/share/locale/vi/LC_MESSAGES/gawk.mo
 file path=usr/share/locale/zh_CN/LC_MESSAGES/gawk.mo
 file usr/share/man/man1/gawk.1 path=usr/gnu/share/man/man1/awk.1
+file path=usr/share/man/man3/filefuncs.3am
```
pyhalov, line 91: Our man unlikely to find these pages

Mno-hime: It actually does for me?

```
$ man filefuncs | head
FILEFUNCS(3am)             GNU Awk Extension Modules            FILEFUNCS(3am)

NAME
       filefuncs - provide some file related functionality to gawk

SYNOPSIS
       @load "filefuncs"
```

Mno-hime: @pyhalov ping?

PR #4944, opened 2019-04-18
clang-40: clang should be rebuild after GCC 6 update to see header files

pyhalov:


pyhalov:
@alarcher We should note this dependency somewhere or rework current patches  (e.g. use /usr/gcc/6/include/c++/6 -> 6.5.0 symlink).

PR #4951, opened 2019-04-19
logilab-common should not deliver pytest

Mno-hime:
I guess the pytest package should deliver `usr/bin/pytest-2.7`.


components/python/logilab-common/logilab-common-27.p5m
```diff
@@ -51,7 +51,6 @@ file \
 file path=usr/lib/python2.7/vendor-packages/logilab_common-$(COMPONENT_VERSION)-py2.7.egg-info/requires.txt
 file \
     path=usr/lib/python2.7/vendor-packages/logilab_common-$(COMPONENT_VERSION)-py2.7.egg-info/top_level.txt
-file path=usr/bin/pytest-2.7
```
pyhalov, line 54: Rename it to ligilab-pytest-27?

pyhalov (approved):

PR #4952, opened 2019-04-19
Update pytest to 4.6

Mno-hime:
Changelog: https://docs.pytest.org/en/latest/changelog.html

4.6 is the last version with support of Python 2.7 and 3.4: https://docs.pytest.org/en/latest/py27-py34-deprecation.html.

~It's not delivering versioned `pytest` and `py.test` executables, created them in post install.~ It does now.

This PR is required to run pyopenssl 19.0.0 tests (https://github.com/OpenIndiana/oi-userland/pull/4955).


pyhalov:
We use them in oi-userland test targets, for example, in components/python/pynacl/Makefile.

components/python/pytest/Makefile
```diff
@@ -42,18 +42,30 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/setup.py.mk
 include $(WS_MAKE_RULES)/ips.mk
 
+ASLR_MODE = $(ASLR_NOT_APPLICABLE)
+
+COMPONENT_POST_INSTALL_ACTION = \
+	for pv in $(PYTHON_VERSIONS); do \
```
pyhalov: You have to do it just once for PYTHON_VERSION - COMPONENT_POST_INSTALL_ACTION will run for each of them, look for example, on components/python/fonttools/Makefile.

alarcher:
@Mno-hime This is also blocking the update to latest numpy. What can be done to move forward on this PR?

Mno-hime:
> @Mno-hime This is also blocking the update to latest numpy. What can be done to move forward on this PR?

https://github.com/OpenIndiana/oi-userland/pull/4951 needs to be pulled-in first, so `library/python/logilab-common-27` won't deliver `usr/bin/pytest-2.7`. Then merge this.

Updated to v4.6.4.

alarcher:
@pyhalov @Mno-hime Can we merge this one?

Mno-hime:
> @pyhalov @Mno-hime Can we merge this one?

The only package which depends on `pytest` is `py`. I rebuilt it successfully as well as `paramico`, `bcrypt`, `pynacl`, `pyasn1`, and `cryptography`. For the distribution it's probably safe not sure about `gmake test` (I find them not very stable and requiring deps we don't have).

pyhalov:
What about components using pytest for tests, do tests at least run?

Mno-hime:
I don't feel like working on this one anymore, it's a too big chunk to swallow for me.

PR #4954, opened 2019-04-19
1/2: Update cryptography to 2.6.1

Mno-hime:
Changelog: https://cryptography.io/en/latest/changelog/

Thanks @balrog-nona for identifying the Python 2 Unicode patch.

Tested with `pkg` and Ansible 2.7.10 test suite (no regressions).

This PR is requirement for https://github.com/OpenIndiana/oi-userland/pull/4955.

PR #4955, opened 2019-04-19
Update pyopenssl to 19.0.0

Mno-hime:
Changelog: https://pyopenssl.org/en/stable/changelog.html

`pkg` is the sole consumer of `library/python/pyopenssl`.

Requires cryptography: https://github.com/OpenIndiana/oi-userland/pull/4954


pyhalov:
@Mno-hime How did you test pkg after this change?

Mno-hime:
By daily running it. I would have run `pkg` test suite, but it's results unreliable for me in any case.

PR #4958, opened 2019-04-20
nodejs-10: Update to 10.15.3

AndWac:
I had to change the preparation because $(CLONEY) doesn't work anymore.


Mno-hime:
> I had to change the preparation because $(CLONEY) doesn't work anymore.

What's wrong with it?

AndWac:
Nodejs changed its configure after 10.10.0. When using $(CLONEY) with newer nodejs versions some files will be created in the original source folder instead in the build folder. I haven't found a better solution than to change to what some other package do - make use of cp -a.

PR #4959, opened 2019-04-21
Gnome mount issues

pyhalov:


pyhalov:
There are two issues here:
- gnome-unmount dies when receives error from dbus
- recent dbus update efficiently removed our modifications to detect console user as logic was hidden behind undefined DBUS_CONSOLE_AUTH_DIR .

PR #4482, opened 2018-10-08
automake: add version 1.16.1

alarcher:
@pyhalov This new version generates scripts that assume GNU make... Components using PATH.illumos break.

PR #4967, opened 2019-04-28
Add automake 1.16.1, install with build-essential

Mno-hime:
Release notes: https://lists.gnu.org/archive/html/info-gnu/2018-02/msg00008.html

Component branched off automake-115.

**Testing**

* test suite has three fails

### DIFF
--- a/components/developer/automake-1.16/Makefile
+++ b/components/developer/automake-1.16/Makefile
@@ -1,0 +1,76 @@
+#
+# CDDL HEADER START
+#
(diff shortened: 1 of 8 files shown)


alarcher:
@Mno-hime Actually broken see https://github.com/OpenIndiana/oi-userland/pull/4482

Mno-hime:
Oh, thanks @alarcher for pointing that out, sorry for duplicating work.

PR #4974, opened 2019-05-03
Totem

pyhalov:


pyhalov:
@Mno-hime 
The PR contains several totem-related fixes. The most important is totem tools issue, which causes totem-video-thumbnailer and other tools to crash on startup (we can see this as video files without previews in gthumb. 

Don't ask me when the issue originally appeared...

Mno-hime:
While rebuilding I got:
```
configure: error: 
			Cannot find required GStreamer-0.10 plugin 'gconfaudiosink'.
			It should be part of gst-plugins-good. Please install it.
```
Had to install `library/audio/gstreamer/plugin/good` to build it.

Mno-hime (approved):
Thumbnails now work for video, thanks for fixing them (admittedly, I even forgot we had them).

The only fail I have is MP3 files from one album: `Unable to create loader for mime type audio/mpeg: Unrecognized image file format`.

PR #4975, opened 2019-05-03
net-snmp: bug 828449 (Debian), fix from solaris (Oracle)

grueni:


pyhalov:
Hi. 
Can you reorganize this PR in two commits, so that one contains fix for snmpd crash, and another - OpenSSL 1.1.0 fixes (preferably, as one patch)?

grueni:
Better now?

Mno-hime:
@grueni Did you have a chance to run the test suite (there appears to be some) or test it in some other way?

grueni:
I added test/results-* and the transform in the Makefile. Tests are passed with OK.

PR #4981, opened 2019-05-05
Update bash 5.0 to patchset 7

Mno-hime:
Do differences in the test suite to what is recorded.

Will run it locally for some time.

### DIFF
--- a/components/shell/bash/Makefile
+++ b/components/shell/bash/Makefile
@@ -30,6 +30,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		bash
 COMPONENT_VERSION=	5.0
+COMPONENT_REVISION=	1
 COMPONENT_FMRI=		shell/bash
 COMPONENT_SUMMARY=	GNU Bourne-Again shell (bash)
 COMPONENT_CLASSIFICATION=	System/Shells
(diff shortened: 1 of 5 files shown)


components/shell/bash/Makefile
```diff
@@ -37,7 +37,7 @@ COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_PROJECT_URL=	http://www.gnu.org/software/bash/
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_HASH = \
-    sha256:b4a80f2ac66170b2913efbfb9f2594f1f76c7b1afd11f799e22035d63077fb4d
+	sha256:b4a80f2ac66170b2913efbfb9f2594f1f76c7b1afd11f799e22035d63077fb4d
```
pyhalov: Has something gone wrong????

Mno-hime: It's really easy to miss out new patches. Thanks for pointing that out.

I did not bump versioning as it seems to be doing in the Makefile by counting `00X` patches.

pyhalov: At least, bump COMPONENT_REVISION

PR #4984, opened 2019-05-05
10915 Firefox: Limit default context processes to 1

Mno-hime:
VM with 4 GB of physical memory is unable to run Firefox 60 ESR with 4 context processess, be it in Live environment or installed (where 2 GB swap is present).

With one context process Firefox works on installed system, with 4 it does not.

For Firefox to work in the Live environment, 6 GB of physical memory are needed.


pyhalov:
Looks good, just change commit message to match bug report title - '10915 Firefox fails to run with 4GB of memory in Live environment'

PR #4988, opened 2019-05-07
Update Firefox to 60.6.2

Mno-hime:
Release note: https://www.mozilla.org/en-US/firefox/60.6.2/releasenotes/?utm_source=firefox-browser&utm_medium=firefox-browser&utm_campaign=whatsnew

How do I test that lang packs work?

I did `pfexec pkg change-facet -v facet.locale.cs=true` but it's not clear to me, that any langpack actually installed/enabled in Firefox.

Otherwise firefox works.

### DIFF
--- a/components/web/firefox/Makefile
+++ b/components/web/firefox/Makefile
@@ -21,14 +21,13 @@ PREFERRED_BITS=		64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		firefox
-IPS_COMPONENT_VERSION=	60.6.1
-COMPONENT_REVISION=	1
+IPS_COMPONENT_VERSION=	60.6.2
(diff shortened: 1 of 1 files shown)


pyhalov:
firefox -UILocale 'locale'

Mno-hime:
> firefox -UILocale 'locale'

The cs XPI is there, but `firefox -UILocale cs` produces the usual US English UI. But it does not work in 60.6.1 as well.

pyhalov:
Look at original Martin's announce https://openindiana.org/pipermail/openindiana-discuss/2015-November/018234.html , part "langpacks". We've never had langpacks working for FF43+. I'd merged this patch as is and later tried to look at @MartinBochnig's patch to XPIProviderUtils.js

Mno-hime:
OK, lets move on with this and fix langpacks later.

pyhalov (approved):

PR #4989, opened 2019-05-07
10918 Enable NTP service by default

Mno-hime:
GZ and non-GZ manifests differ only in:

`<instance name="default" enabled="true">` × `<instance name="default" enabled="false">`

Tested both installation and update in GZ and non-GZ.

### DIFF
--- a/components/network/ntp/ntp.p5m
+++ b/components/network/ntp/ntp.p5m
@@ -19,6 +19,7 @@
 # CDDL HEADER END
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2019, Michal Nowak
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
@@ -28,6 +29,7 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+set name=variant.opensolaris.zone value=global value=nonglobal
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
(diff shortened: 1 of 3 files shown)


pyhalov:
Can we use just one xml file (keep existing one) and generate another using sed? 
Something like 
```
$(BUILD_DIR)/ntp-ngz.xml: $(COMPONENT_DIR)/Solaris/ntp.xml $(BUILD_DIR)
       $(GSED) -e '/instance name="default"/s/enabled="true"/enabled="false"/g' $< > $@

$(INSTALL_64): $(BUILD_DIR)/ntp-ngz.xml
```
to avoid duplicating code?

Mno-hime:
Good idea.

Zone:
```
{build-userland} newman:~ $ pkg contents ntp | grep xml
lib/svc/manifest/network/ntp-ngz.xml
```

Global Zone:
```
{global} newman@lenovo:~ $ pkg contents ntp | grep xml
lib/svc/manifest/network/ntp.xml
```

components/network/ntp/ntp.p5m
```diff
@@ -41,7 +43,8 @@ file Solaris/ntp.client path=etc/inet/ntp.conf group=sys preserve=true
 file Solaris/ntp.server path=etc/inet/ntp.server group=sys
 file Solaris/auth_attr path=etc/security/auth_attr.d/ntp
 file Solaris/prof_attr path=etc/security/prof_attr.d/ntp
-file Solaris/ntp.xml path=lib/svc/manifest/network/ntp.xml
+file Solaris/ntp.xml path=lib/svc/manifest/network/ntp.xml variant.opensolaris.zone=global
+file build/ntp-ngz.xml path=lib/svc/manifest/network/ntp-ngz.xml variant.opensolaris.zone=nonglobal
```
pyhalov: Can you put it into the same place?

Mno-hime: What you mean?

pyhalov:
```
file Solaris/ntp.xml path=lib/svc/manifest/network/ntp.xml variant.opensolaris.zone=global
file build/ntp-ngz.xml path=lib/svc/manifest/network/ntp.xml variant.opensolaris.zone=nonglobal
```

tsoome:
This is not the way to do such feature at all.

You have 2 possible better ways:

1. apply proper SMF profile (NTP enabled in global, disabled for local zone)
2. enable NTP via profile (because services are defined disabled by default), but update the start method to disable itself if we are running in local zone. This does work for us because we do not have kernel zones as Oracle Solaris does.

components/network/ntp/ntp.p5m
Mno-hime: Right. Thanks.

pyhalov:
@tsoome Do I understand correctly that profile applies only on installation? Well, generic_limited_net.xml has ntp disabled by default.
Should we fix it? I'm not sure. 
We can fix https://github.com/OpenIndiana/slim_source/blob/oi/hipster/usr/src/cmd/distro_const/utils/generic_live.xml , but this will have effect only on new installations (given that even minimal installation includes ntp).

PR #5003, opened 2019-05-11
2/2: Rebuild libarchive & borg because of zstd 1.4.0 update

Mno-hime:
`zstd` dependents need rebuild because of [`zstd` update to 1.4.0](https://github.com/OpenIndiana/oi-userland/pull/5002). I rebuild `libarchive` and choose to update `borg` as there's a [security fix](https://github.com/borgbackup/borg/blob/1.1.9/docs/changes.rst#version-119-2019-02-10) in it (though, perhaps not relevant for us), also added Python 3.5 variant, as suggested, and made the whole package 64-bit-only.

**Testing**

* `borg` can be started in both Python 3.4 and Python 3.5 variants


components/sysutils/borgbackup/borg-34.p5m
```diff
@@ -98,3 +98,4 @@ file path=usr/lib/python3.4/vendor-packages/borgbackup-$(COMPONENT_VERSION)-py3.
 file path=usr/lib/python3.4/vendor-packages/borgbackup-$(COMPONENT_VERSION)-py3.4.egg-info/not-zip-safe
 file path=usr/lib/python3.4/vendor-packages/borgbackup-$(COMPONENT_VERSION)-py3.4.egg-info/requires.txt
 file path=usr/lib/python3.4/vendor-packages/borgbackup-$(COMPONENT_VERSION)-py3.4.egg-info/top_level.txt
+file path=usr/share/bash-completion/completions/borg
```
pyhalov: This file should be in backup/borg, not in backup/borg-34, as it's likely to be common between borg-34 and borg-35 (which we should add). Also, please add dependency: borg-34 should require borg.

Mno-hime: I added 3.5 version of `borgbackup`, but dumped the Bash completion as I was unable to add sub-packages.

pyhalov: You can add necessary content (only content, no set actions) to borg-GENFRAG.p5m, look for example on gst-python/gst-python-GENFRAG.p5m

pyhalov:
Fixed borg to deliver 32-bit version for Python 3.4 as earlier (to avoid surprises).

PR #5010, opened 2019-05-14
Update GNU gettext to 0.20.1

Mno-hime:
GNU gettext 0.20: http://savannah.gnu.org/forum/forum.php?forum_id=9430
GNU gettext 0.20.1: http://lists.gnu.org/archive/html/info-gnu/2019-05/msg00011.html

**Changes**
* packaged more man pages from section 3
* many hundreds files of documentation and examples were removed
* intl is gone. From the release notes:

  * The --intl option of the gettextize program (deprecated since 2010) is no longer available. Instead of including the intl sources in your package, we suggest making the libintl library an optional prerequisite of your package. This will simplify the build system of your package.
  * Accordingly, the Autoconf macro AM_GNU_GETTEXT_INTL_SUBDIR is gone as well. 
* build and ship 64-bit stuff as GNU Bison 3.4 (https://github.com/OpenIndiana/oi-userland/pull/5024) can leverage `libtextstyle`

**Testing**
* test suite passed
* real-world testing is needed, especially with old stuff


pyhalov:
Double-check that illumos libc gettext() is not broken with new mo files.

alarcher:
@Mno-hime What is left to do before the PR is ready?

Mno-hime:
> @Mno-hime What is left to do before the PR is ready?

@alarcher To do what @pyhalov says: "Double-check that illumos libc gettext() is not broken with new mo files." I could not figure out, how to design such a test. Ideas?

pyhalov:
I suggest to do at least some brief sanity check
Create simple *.po file (or use existing one, for example, from libpeas ).
Compare binary *.mo files generated mo files (with new and old msgfmt).
Check that illumos gettext still can get information from it.

pyhalov:
I've tested gettext a bit. gettext's gmo.h file is not changed, *.mo file format seems to be not affected.  
Tested that files, generated with gmsgfmt work with illumos gettext.
Also removed man(3) pages from this PR as they shadow illumos gettext(3c).

Mno-hime:
Thanks for testing this for me @pyhalov.

PR #5015, opened 2019-05-17
Replace 'rename' with 'text-utilities'

Mno-hime:
`text-utilities` come from util-linux (originally random set of BSD utilities) the same source as the former `rename` component and thus replace it.

**Testing**
- internal test suite
- manual runs with examples from respective man pages


components/meta-packages/history/history
```diff
@@ -578,6 +578,7 @@ editor/gedit@2.30.4,5.11-2018.0.0.3
 editor/gedit/gedit-plugins@2.32.0,5.11-2017.0.0.2
 editor/ghex@2.30.0,5.11-2014.0.0.0
 editor/gobby@0.4.94,5.11-2017.0.0.3
+file/rename@2.33.1-2018.0.0.0 file/text-utilities
```
pyhalov: To be sure that this package version is always greater than older one, bump branch component  (2018.0.0.1)

components/meta-packages/install-types/includes/server_desktop
```diff
@@ -16,7 +16,7 @@ depend type=require fmri=editor/vim
 depend type=require fmri=file/gnu-coreutils
 depend type=require fmri=file/gnu-findutils
 depend type=require fmri=file/slocate
-depend type=require fmri=file/rename
+depend type=require fmri=file/text-utilities
```
pyhalov: Bump install-types COMPONENT_REVISION

components/text/text-utilities/Makefile
```diff
@@ -0,0 +1,118 @@
...
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		text-utilities
+COMPONENT_MAJOR_VERSION=2.33
+COMPONENT_MINOR_VERSION=2
+COMPONENT_VERSION=	$(COMPONENT_MAJOR_VERSION).$(COMPONENT_MINOR_VERSION)
+COMPONENT_SUMMARY=	Various Linux utilities
+COMPONENT_FMRI=		file/$(COMPONENT_NAME)
```
pyhalov: Rename it to text/text-utilities to match Oracle FMRI

components/meta-packages/install-types/includes/server_desktop
pyhalov: Rename it to text/text-utilities to match Oracle

PR #5021, opened 2019-05-18
Update LightDM to 1.30.0

Mno-hime:
Changelog: https://github.com/CanonicalLtd/lightdm/blob/master/NEWS

I added `reboot-needed=true` to the `lightdm` executable (is there a better place where to put it, how to do it?) because during LightDM update some processes tracked by the service are killed, including `Xorg` thus killing the desktop, service ends-up in maintenance.

No visual change this time.

**Testing**
- login
- autologin
- change of: environment, language, accessibility, power down.

### DIFF
--- a/components/desktop/lightdm/Makefile
+++ b/components/desktop/lightdm/Makefile
@@ -11,18 +11,18 @@
 
 #
 # Copyright 2016 Alexander Pyhalov
-# Copyright 2018 Michal Nowak
+# Copyright 2019 Michal Nowak
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		lightdm
-COMPONENT_VERSION=	1.28.0
+COMPONENT_VERSION=	1.30.0
 COMPONENT_SUMMARY=	Light display manager
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.xz
 COMPONENT_ARCHIVE_HASH= \
-	sha256:f20f599d8d7cf90b77a2df33c6e9e615abc443123b1b228de1c20a636aecfa07
+	sha256:a118779650a3a5b20f05d53c4f9e377b484fbc1810a6b0914a60840d9a581f95
 COMPONENT_ARCHIVE_URL= \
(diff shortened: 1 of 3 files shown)


components/desktop/lightdm/lightdm.p5m
```diff
@@ -57,7 +58,8 @@ link path=usr/lib/$(MACH64)/liblightdm-gobject-1.so.0 \
 file path=usr/lib/$(MACH64)/liblightdm-gobject-1.so.0.0.0
 file path=usr/lib/$(MACH64)/lightdm-guest-session mode=0555
 file path=usr/lib/$(MACH64)/pkgconfig/liblightdm-gobject-1.pc
-file path=usr/sbin/lightdm
+# LightDM needs to be updated in new BE, otherwise X crashes
```
pyhalov: crashes or is killed?

Mno-hime: First three warnings is "the usual stuff". I assume killed:

```
** (process:20972): WARNING **: 11:06:21.856: Error getting user list from org.freedesktop.Accounts: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.Accounts was not provided by any .service files
** (process:20972): WARNING **: 11:06:21.988: Error getting XDG_RUNTIME_DIR from ConsoleKit: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Method "GetXDGRuntimeDir" with signature "" on interface "org.freedesktop.ConsoleKit.Session" doesn't exist
** (lightdm:20928): WARNING **: 11:06:21.989: Error activating ConsoleKit session: GDBus.Error:org.freedesktop.DBus.GLib.UnmappedError.CkSeatError.Code0: Activation is not supported for this kind of seat
end from FAM server connection
end from FAM server connection
Failed to get D-Bus connection
** (process:20972): WARNING **: 11:06:54.608: Error ending ConsoleKit session: The connection is closed
```

pyhalov: please, update comment.

pyhalov: Anyway, I'm not sure about reboot-needed for lightdm, as efficiently it is not needed.
What are we trying to achieve? Avoid restarting service on pkg update? But given hipster's nature, pkg update will likely also update kernel module and will create new BE anyway. But if we put reboot-needed on lightdm, it will make developing/testing it a bit harder - you'll have to reboot after each update of lightdm.

PR #3974, opened 2018-02-12
Category

alarcher:
Semi-scripted thing, need to double check.

PR #5022, opened 2019-05-19
Components cleanup #1

Mno-hime:
Moving components from `components/` dir to their respective categories.

In the process I obsolete those which are too old and we have a better replacement, or the upstream is dead for many years and it's not worth having it. Also I will obsolete components which are too broken to be fixed easily (assuming there are crappy things in there anyway).

If the component looks at least minimally viable, I update it to the latest version, run `sample-manifest` and `pre-publish` targets, do minimal `Makefile` and `.p5m` clean-up, to ensure the component meets minimum standards and builds correctly.

Started moving things alphabetically. Will probably separate things into more PR, so review and building is easier.

Let me know, if I go the right way with this, before I am too far :). Also speak-up should you want to preserve something I am obsoleting.

@balrog-nona FYI.


components/meta-packages/history/history
```diff
@@ -529,6 +529,7 @@ desktop/administration/gnome-system-tools@2.30.0,5.11-2017.0.0.1
 desktop/administration/system-tools-backends@1.4.2,5.11-2017.0.0.3
 desktop/archive-manager/file-roller@2.32.2,5.11-2017.0.0.3
 desktop/avant-window-navigator@0.3.2.1,5.11-2017.0.0.3
+desktop/calculator/gcalctool@5.30.2-2018.0.0.0
```
AndWac: From my understanding you will have to increase the build number for entries here.
Thus, it should be
desktop/calculator/gcalctool@5.30.2-2018.0.0.1

components/meta-packages/history/history
```diff
@@ -600,6 +601,7 @@ gnome/applet/gnome-utility-applets@2.32.1.1,5.11-2017.0.0.5
 gnome/applet/deskbar-applet@2.32.0,5.11-2017.0.0.2
 gnome/applet/quick-lounge-applet@2.14.1,5.11-2017.0.0.1
 gnome/cmdassist@0.5.11,5.11-2017.0.0.1
+gnome/config/gconf/gconf-editor@2.32.0-2018.0.0.0
```
AndWac: Same here, it should be
gnome/config/gconf/gconf-editor@2.32.0-2018.0.0.1

AndWac (changes requested):

components/meta-packages/history/history
Mno-hime: I think you are right, changed accordingly.

AndWac (approved):

alarcher:
For subsequent PRs you could save some time cherry-picking some commits from https://github.com/OpenIndiana/oi-userland/pull/3974

pyhalov:
Are we ready to obsolete gconf-editor? There are apps, which still use GConf, and gconf-editor provides a nice way to look at gconf settings.

Mno-hime:
Brought `gconf-editor` back.

pyhalov (approved):

PR #5024, opened 2019-05-19
Update GNU Bison to 3.4.1

Mno-hime:
Changes: http://savannah.gnu.org/forum/forum.php?forum_id=9433

Requires updated GNU gettext (https://github.com/OpenIndiana/oi-userland/pull/5010), which ships `libtextstyle` (both 32-bit and 64-bit) for colored diagnostics

**Testing**
- [x] keep for some time in my development environment
- [x] test suite passed
- [x] illumos-gate rebuild (due to likes of [#10312](https://www.illumos.org/issues/10312))
- [x] illumos-gate debug rebuild


Mno-hime:
Eh, not ready yet as I need to have a look at the gettext PR.

PR #5025, opened 2019-05-20
10691 Want rdesktop with GSS-API and TLS v1.3

Mno-hime:
Really, this is not the stated version as releases are based on the 1.8.x branch but require libgssglue, which we don't want to skip as it's obsolete in the master branch anyway. For Kerberos we use native GSS-API. So, version 1.8.99 is to be read as development version before 1.9.0 is out (also wired in sources).

Has some security fixes as well.

**Changes** (all inherited by the version change)
- uses PulseAudio instead of OSS
- GNU TLS instead of OpenSSL

**Testing**
- Successfully connected to WS 2016 and WS 2012 R2 where NLA is disabled, though some warnings are present:
```
Autoselecting keyboard map 'en-us' from locale
Core(error): tcp_tls_connect(), Could not load system trust database: An unimplemented or disabled feature has been requested.
Core(warning): Certificate received from server is NOT trusted by this system, an exception has been added by the user to trust this specific certificate.
Failed to initialize NLA, do you have correct Kerberos TGT initialized ?
Core(error): tcp_tls_connect(), Could not load system trust database: An unimplemented or disabled feature has been requested.
Core(warning): Certificate received from server is NOT trusted by this system, an exception has been added by the user to trust this specific certificate.
Connection established using SSL.
Protocol(warning): process_pdu_logon(), Unhandled login infotype 1
Clipboard(error): xclip_handle_SelectionNotify(), unable to find a textual target to satisfy RDP clipboard text request
```

Some testing around kerberos and sound are still required.


esmnemon:
Hi, I finally managed to build this :-)
    It looks fine, the credSSP things are working against servers I've tried of different kind (win2008r2, win2016), but on the one server which was the reason I initiated this, it is hardened to only accept TLS1.3, which is not supported in gnutls, openindiana-version (3.5.14) or in openssl: /usr/include/openssl/tls1.h
`# define TLS_MAX_VERSION                 TLS1_2_VERSION`
 

This is what happens: 
$ ./rdesktop1899 -u myuser -d MYDOMAIN.EXAMPLE.COM -k no 10.0.0.26
Password: 
Core(error): tcp_tls_connect(), Could not load system trust database: An unimplemented or disabled feature has been requested.

ATTENTION! The server uses and invalid security certificate which can not be trusted for
the following identified reasons(s);

 1. Certificate issuer is not trusted by this system.

     Issuer: DC=com, DC=example, DC=mydomain, CN=mydomain-enterprise-ca

 2. Certificate was signed using an insecure algorithm.


Review the following certificate info before you trust it to be added as an exception.
If you do not trust the certificate the connection attempt will be aborted:

    Subject: CN=wts.mydomain.example.com, CN=wts, CN=wts.example.com, CN=wts
     Issuer: DC=com, DC=example, DC=mydomain, CN=mydomain-enterprise-ca
 Valid From: Thu May  2 11:46:50 2019
         To: Sun Nov 29 07:49:01 2020

  Certificate fingerprints:

       sha1: 79c5be225143xxxxxxxxxx65d6db6e252258a283
     sha256: 37a4e96af60662b889bcxxxxxxxxxxcb705ddf255cbc2fa831a372e2121f5039


Do you trust this certificate (yes/no)? yes
Failed to initialize NLA, do you have correct Kerberos TGT initialized ?
Failed to connect, CredSSP required by server (check if server has disabled old TLS versions, if yes use -V option).

---------
And yes, I have a krb-ticket, 
$ klist
Ticket cache: FILE:/tmp/krb5cc_1234
Default principal: myuser@MYDOMAIN.EXAMPLE.COM

Valid starting               Expires               Service principal
19/06/2019 11:19  19/06/2019 21:19  krbtgt/MYDOMAIN.EXAMPLE.COM@MYDOMAIN.EXAMPLE.COM
        renew until 21/06/2019 11:19
$

I might do something wrong...
I've asked the person responsible if I can get a test-window wher the server accepts tls1.2 as well as 1.3, just to test if it is OK.

Mno-hime:
Thanks for giving it a shot @esmnemon! Any news on this with TLS 1.2? TLS 1.3 in OpenSSL will only come when we move to OpenSSL 1.1.1, and in GNU TLS with 3.6. In both cases, no plans for the transition yet.

pyhalov:
@Mno-hime Should we merge it now or should we wait for GNU TLS update?

Mno-hime:
From my testing this is no worse than the old version. I am fine with merging.

PR #5034, opened 2019-05-22
Update Go 1.10 to 1.10.8

Mno-hime:
Release notes: https://golang.org/doc/devel/release.html#go1.10.minor

Go 1.10 is not supported anymore but before we remove it, worth a security update.

I dumped `GO_MAJOR` as this brings me problems on minor update (I don't see much in `vimdiff`).

Tests have some fails, but nothing new.


Toasterson:
Was the error about an unused variable? Because GO_MAJOR was before used inside the manifest.

Otherwise LGTM.

Toasterson (approved):

Mno-hime:
Nope, well... this time it even passed: [go110tests.txt](https://github.com/OpenIndiana/oi-userland/files/3210280/go110tests.txt).

PR #5035, opened 2019-05-22
Update Go 1.11 to 1.11.10

Mno-hime:
Release notes: https://golang.org/doc/devel/release.html#go1.11.minor

There's the same security fix as in Go 1.10.8.

I dumped `GO_MAJOR` as this brings me problems on minor update (I don't see much in `vimdiff`).

Tests have some fails, but nothing new.


Toasterson (approved):
LGTM

Mno-hime:
There are few fails but only those we [discussed](https://github.com/OpenIndiana/oi-userland/pull/4612) before: [go111tests.txt](https://github.com/OpenIndiana/oi-userland/files/3210288/go111tests.txt).

PR #5047, opened 2019-05-28
Update hplip to 3.19.5

timmooney:
Update component hplip from 3.15.9 to 3.19.5.

Review needed, @pyhalov and @Mno-hime 

* Add new portability patches 10_* through 17_*
  * the 16_libImageProcessor.arch.patch will need work if/when this package switches to multi-arch or x86_64
* Update a couple of the existing patches to work with the new release
* drop COMPONENT_BUGDB from Makefile
* rerun 'gmake REQUIRED_PACKAGES'
* rerun 'gmake sample-manifest' and replace the existing sample manifest with the new one
* update hplip.p5m with hundreds of new file paths for new and replaced files

Testing is definitely needed, especially from people with multiple models of HP printers and with multifunction printer/scanners.  I've only been able to test with one HP printer so far.

There probably needs to be a discussion about the 'hpijs' build too.  hpijs has been deprecated for apparently years.  Including hpijs in the same package adds hundreds of files to the package.  If we need to continue to provide hpijs, perhaps its time for a package split?

Tim


timmooney:
I'm still not very familiar with github workflow.   I thought I would be able to add the "testing needed" label after opening the pull request, but I cannot.  Can one of you apply that label to this pull, please?

Mno-hime:
> I'm still not very familiar with github workflow. I thought I would be able to add the "testing needed" label after opening the pull request, but I cannot. Can one of you apply that label to this pull, please?

I think only project members can toggle labels.

Mno-hime:
Ad https://github.com/OpenIndiana/oi-userland/pull/5047/commits/2e6935397889536ed25a9139c8332af18275d0cc
```
 Tim Mooney authored and Tim Mooney committed 2 hours ago 
```
Are you using different name/email in your cloned repo than on GitHub? (Not a problem, it just looks weird.)

timmooney:
Hmmm, I might be using a different email address, thanks for pointing it out.  I recently rebuilt my hipster dev environment and the git setup might not be quiet right.

I'll take a look and see if I can fix it.

Thanks for adding the "testing needed" label.

components/print/hplip/patches/16_libImageProcessor.arch.patch
```diff
@@ -0,0 +1,101 @@
+
+Make createPPD.sh choose the architecture for libImageProcessor.so based
+on a compiled binary, not based on the kernel architecture.
+
+Fix Makefile.am and Makefile.in so that it installs the i386 version of
```
pyhalov: The fix seems to be overcomplicated. Can we just pass architecture as compile-time variable and do  something like COMPONENT_BUILD_ENV+="ARCH=$(BITS)"?

timmooney:
Thanks @pyhalov , I've adjusted patch 16 and updated Makefile to use `CONFIGURE_ENV += ARCH=$(BITS) `    (because createPPD.sh is run as part of configure) and `COMPONENT_INSTALL_ENV += ARCH=$(BITS)` (because which link to install is determined as part of `install-data`).

Patch 16 is still somewhat complicated because the `install-data` target is wrong (it misses $(DESTDIR) ) and I have to fix that, in addition to the architecture fix, but your suggestion improved the logic.

pyhalov:
For now I'm going to merge it as-is. 
I'm not sure about splitting package in two. Perhaps, it could be done in separate PR.
How much space does it save?

timmooney:
Thanks for the merge.

The hpijs components add about 10-12 MiB to the package size, and include more than 1000 files.  By modern standards, it's minor.  I wouldn't even have suggested it, except the IJS components are long-deprecated.

PR #5050, opened 2019-05-29
lighttpd: update to 1.4.54

pyhalov:


components/web/lighttpd/patches/05-build.patch
```diff
@@ -0,0 +1,14 @@
+https://github.com/lighttpd/lighttpd1.4/commit/0a46f7ec23a924f4e53874ce32471209b2b70686
+has been never tested on illumos/Solaris
```
Mno-hime, line 2: Let upstream know so we don't have to carry it next time?

pyhalov: Yes, I've notified developer.
Surprisingly, event port polling method seems to work correctly.

PR #5053, opened 2019-06-01
gdb fixes (#11087, #11088)

AndWac:


components/developer/gdb/Makefile
```diff
@@ -20,24 +20,25 @@
...
+# Copyright 2019 Andreas Wacknitz
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		gdb
 COMPONENT_VERSION=	7.10.1
-COMPONENT_REVISION=	4
+COMPONENT_REVISION=	5
+COMPONENT_SUMMARY=  GDB: The GNU Project Debugger
```
Mno-hime: Tab might do better.

Mno-hime (approved):

components/developer/gdb/Makefile
AndWac: I have tried to fix the tabs, does it look better now?

pyhalov:
Regenerated syscall patches with something like 
```
grep 'define     SYS' /usr/include/sys/syscall.h  |sed -e 's:SYS_::' | awk ' { print "  <syscall name=\"" $2 "\" number=\"" $3"\"/>" ; }'
``` 
and manual fixing.

richlowe (approved):
That definitely looks like our syscall table now!

richlowe:
This still leaves a (largely?) non-functional gdb.  Perhaps everything in the patch that introduced this needs regenerating?

I notice that Oracle's gdb has fewer patches (though obviously they'd need regenerating for illumos too).

I also notice that pkgsrc doesn't patch gdb really at all.

PR #5055, opened 2019-06-03
convert gegl to libjpeg8

timmooney:
Convert the gegl image library (used by gimp) to use libjpeg8.

I looked at updating to latest gegl too, but that is a much larger project and will need some decisions about what to do about the "workshop" component.

### DIFF
--- a/components/library/gegl/Makefile
+++ b/components/library/gegl/Makefile
@@ -11,13 +11,14 @@
 
 #
 # Copyright 2017 Alexander Pyhalov
+# Copyright 2019 Tim Mooney
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= gegl
 COMPONENT_VERSION= 0.2.0
-COMPONENT_REVISION= 1
+COMPONENT_REVISION= 2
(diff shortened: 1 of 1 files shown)


Mno-hime (approved):
LGTM.

Btw, since you are touching it, feel free to do cleanup along the way, e.g.:

- `PATH=/usr/gnu/bin:/usr/bin` -> `PATH=$(PATH.gnu)`
- update the component's version (where appropriate)
- rebuild `sample-manifest` to see if new files are not available (may happen from various reasons)

timmooney:
Thanks for the review!

As I said in my comment, I did look at updating the component's version, but that is a much larger project and will require input from you and other core developers, since an entire subpackage will probably need to go away (the gegl/workshop stuff).  I didn't want to delay the libjpeg8 conversion because of that.

I'll keep the other comments in mind.  I do try do cleanups that I've already been made aware of
(things like ($WS_MAKE_RULES), removing COMPONENT_BUGDB, etc.), but until your comment I wasn't aware of the desire to switch to $(PATH.gnu).

As I gain more experience with the types of updates, I'll likely return to updating babl/gegl/gimp.

Mno-hime:
@timmooney I somehow missed the description, my fault. Agreed.

alarcher (approved):

PR #5060, opened 2019-06-05
Update mutt to 1.12.2

Mno-hime:
Release notes: http://www.mutt.org/relnotes/1.12/ & http://www.mutt.org/relnotes/1.11/ & https://gitlab.com/muttmua/mutt/raw/stable/UPDATING

**Testing**
- it starts


components/mail/mutt/mutt.p5m
```diff
@@ -19,18 +19,9 @@
...
 
-<transform file path=usr.*/man/.+/flea\.1$ -> default mangler.man.stability uncommitted>
-<transform file path=usr.*/man/.+/mutt\.1$ -> default mangler.man.stability uncommitted>
-<transform file path=usr.*/man/.+/muttbug\.1$ -> default mangler.man.stability uncommitted>
-<transform file path=usr.*/man/.+/pgpewrap\.1$ -> default mangler.man.stability volatile>
-<transform file path=usr.*/man/.+/pgpring\.1$ -> default mangler.man.stability volatile>
-<transform file path=usr.*/man/.+/smime_keys\.1$ -> default mangler.man.stability volatile>
-<transform file path=usr.*/man/.+/mbox\.5$ -> default mangler.man.stability committed>
-<transform file path=usr.*/man/.+/mmdf\.5$ -> default mangler.man.stability committed>
-<transform file path=usr.*/man/.+/muttrc\.5$ -> default mangler.man.stability volatile>
```
mtelka, line 33: Why is this no longer needed?

pyhalov: I think, because we usually can't guarantee the stability of interface provided by  upstream project. What if it changes at some point? Will we patch it? I doubt we even notice it. So, I think stability of third-party-provided interfaces is mostly a legacy artefact. Of course we can preserve this, but I don't think this makes much sense.

components/mail/mutt/Makefile
```diff
@@ -21,28 +21,33 @@
...
 
 include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 
+CFLAGS +=			-I$(USRINCDIR)/idn2
+
 # Deliver 64-bin binaries in standard paths
-CONFIGURE_BINDIR.64 = $(CONFIGURE_PREFIX)/bin
+#CONFIGURE_BINDIR.64 = $(CONFIGURE_PREFIX)/bin
```
pyhalov: Remove this line and comment above to avoid confusion.

components/mail/mutt/mutt.p5m
mtelka: Right, we cannot guarantee the stability, so it is set to uncommitted or volatile.  The only committed stability is for something (mbox and mmdf) that is really stable.  The documented stability is very helpful in many cases and sets user's (developer's) expectations properly.  Please keep it, if possible.  Thanks.

Mno-hime: > I think, because we usually can't guarantee the stability of interface provided by upstream project. What if it changes at some point? Will we patch it? I doubt we even notice it. So, I think stability of third-party-provided interfaces is mostly a legacy artefact. Of course we can preserve this, but I don't think this makes much sense.

Yes, that's exactly my reasoning behind this change.

> Right, we cannot guarantee the stability, so it is set to uncommitted or volatile. The only committed stability is for something (mbox and mmdf) that is really stable. The documented stability is very helpful in many cases and sets user's (developer's) expectations properly. Please keep it, if possible. Thanks.

For that I need to see upstream's explicit commitment to stability of those component's interfaces, otherwise we are misleading developers in believing this information is to be trusted. Even in case of interfaces being stable in every practical regard, without upstream guaranteeing it, they can change without notice. Personally, I remove those "guarantees" wherever I see them.

pyhalov:
@Mno-hime  Should we merge it?

Mno-hime:
> @Mno-hime Should we merge it?

I updated this to 1.12.1 but still didn't have the time to test it...

PR #5062, opened 2019-06-06
Update vim to 8.1.1526, GTK3, 64-bit, and /etc/vimrc

Mno-hime:
Due to security stuff reported by @AndWac: https://github.com/numirias/security/blob/master/doc/2019-06-04_ace-vim-neovim.md.

Ported to GTK 3, 64-bit, added global `/etc/vimrc` with changes requested at 10342 which make console experience much better.

As Ruby support requires 64-bit Ruby, which we don't have, I disabled it.


pyhalov (approved):

Mno-hime:
Will update it later to https://github.com/vim/vim/releases/tag/v8.1.1473 where https://github.com/OpenIndiana/oi-userland/pull/5062/files#diff-410f1f87aafda62ff93a766e7117f8b5 was fixed.

alarcher:
Should I merge?

Mno-hime:
It's slightly controversial as it disables Ruby support (or something). I'll rather ask on discuss ml.

pyhalov:
You should be able to enable ruby support now in 64-bit version.

Mno-hime:
Thanks, it builds ruby-26. One thing I should add to the default `vimrc` is `set mouse=`, so mouse click does not enter to insert mode, but didn't have luck with it so far, though it works fine on `vim`'s `:` "command line".

components/editor/vim/patches/remove-zh_CN.patch
```diff
@@ -0,0 +1,34 @@
+ko.po: warning: Charset "EUC-KR" is not supported. gmsgfmt relies on iconv(),
```
pyhalov, line 1: Should we open illumos-gate issue?

pyhalov:
test target is broken:
$ gmake clean test pre-publish
make[3]: *** No rule to make target 'zh_CN.ck', needed by 'check'.  Stop

@Mno-hime

PR #5070, opened 2019-06-09
Various components clean ups

balrog-nona:


components/network/iftop/Makefile
```diff
@@ -57,9 +59,9 @@ CONFIGURE_OPTIONS  +=	CPPFLAGS="$(CPPFLAGS)"
 
 
 # common targets
-build:		$(BUILD_32)
```
pyhalov, line 60: Please, remove IFTOP_LIBDIR  + LDFLAGS  (lines 55, 56), they are confusing (and not needed, as libncurses lives in /usr/lib/64)

pyhalov: Besides this small issue, it looks good.

balrog-nona: Done. Thanks.

alarcher (approved):

PR #5075, opened 2019-06-10
new component itstool-2.0.6

timmooney:
This is a new component, [itstool](http://itstool.org) . It's used by some recent GNOME components at build time (not runtime) as part of the internationalization process.

Please review, @pyhalov and @Mno-hime .  In particular
  * COMPONENT_CLASSIFICATION . I used the same classification that glade is using, since this tool is mainly used when building GNOME applications, but perhaps Developer/GNU or something else would have been better?
  * It's a python script, but despite what the comments say in `shared-macros.mk`, there doesn't seem to be any way to do a `$(BUILD_NO_ARCH)/$(INSTALL_NO_ARCH)` for a general component, like Alp had me do when I packaged `python/virtualenv`.
  * since `PYTHON_VERSION` still defaults to 2.7, I built this component to use that version of python.  It does claim to support python 3, so I could rebuild with e.g. 3.4 or 3.5, if you prefer.
  * I added the man page transform in `itstool.p5m`, but I don't know if there are any other transforms that are appropriate?

Feedback welcome.

### DIFF
--- a/components/developer/itstool/COPYING
+++ b/components/developer/itstool/COPYING
@@ -1,0 +1,19 @@
+ITS Tool - XML to PO and back again using ITS definitions
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+As a special exception, the copyright holders give you permission to
+copy, modify, and distribute the ITS definitions bundled with this
+program under the terms of your choosing, without restriction.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


pyhalov:
Just update https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/text/itstool

timmooney:
Thanks.  I did search for itstool before starting this, I don't know how I missed it, but I'll just update that one.  Sorry for the noise.

Mno-hime:
> Thanks. I did search for itstool before starting this, I don't know how I missed it, but I'll just update that one. Sorry for the noise.

Fairly convenient way to search in the oi-userland project is to go to https://github.com/OpenIndiana/oi-userland, press `t`, and go from there.

PR #5076, opened 2019-06-10
update itstool to 2.0.6

timmooney:
Update itstool from 2.0.5 to 2.0.6

I regenerated the sample manifest and verified no file changes.

### DIFF
--- a/components/text/itstool/Makefile
+++ b/components/text/itstool/Makefile
@@ -19,19 +19,19 @@
 # CDDL HEADER END
 #
 # Copyright 2016 Jim Klimov
+# Copyright 2019 Tim Mooney
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		itstool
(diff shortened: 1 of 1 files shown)


components/text/itstool/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2016 <contributor>
+# Copyright 2018 <contributor>
```
Mno-hime: Usually we don't ship this date-only changes, stick with `2016`, please.

Mno-hime (approved):

timmooney:
Thanks Michal.  I've updated the pull request, removing that change.

In the future, to make it clear that I've done a `gmake sample-manifest` to verify no file content changes, I'll just note that in the pull request, rather than including the change to the year in the sample.

alarcher (approved):

PR #5082, opened 2019-06-15
mc: Enable test target

Mno-hime:
Ticket for the fix https://midnight-commander.org/ticket/3971.

Test target now compiles. Though, there several fails.

### DIFF
--- a/components/sysutils/mc/patches/04-clarify_sighandler_t_usage_in_non-GNU_OSes.patch
+++ b/components/sysutils/mc/patches/04-clarify_sighandler_t_usage_in_non-GNU_OSes.patch
@@ -1,0 +1,46 @@
+https://github.com/MidnightCommander/mc/commit/4f06ba93460751bd979ca5cc87a49219b1f78a9f
+
+From 4f06ba93460751bd979ca5cc87a49219b1f78a9f Mon Sep 17 00:00:00 2001
+From: Andrew Borodin <aborodin@vmail.ru>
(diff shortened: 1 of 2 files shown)


alarcher (approved):

components/sysutils/mc/patches/04-clarify_sighandler_t_usage_in_non-GNU_OSes.patch
```diff
@@ -0,0 +1,46 @@
...
+
+Signed-off-by: Andrew Borodin <aborodin@vmail.ru>
+---
+ m4.include/mc-tests.m4                 | 6 ++++++
+ tests/lib/utilunix__my_system-common.c | 5 +++++
+ 2 files changed, 11 insertions(+)
+
+diff --git a/m4.include/mc-tests.m4 b/m4.include/mc-tests.m4
+index 95e948bd7f..257218ecbe 100644
+--- a/m4.include/mc-tests.m4
```
pyhalov, line 16: Do I miss something? It seems you patch m4 file, but do not regenerate configure.

pyhalov: @Mno-hime

Mno-hime: Right... this and that the `test` target is actually starting any tests even without the patch. Ah. Lets do it with the next `mc` release.

PR #5084, opened 2019-06-15
Update tcsh to 6.21.00

Mno-hime:
Announcement: https://mailman.astron.com/pipermail/tcsh/2019-May/000018.html

- ~Currently builds only with GCC 8: https://bugs.astron.com/view.php?id=80~ Actually a GCC optimizer bug.
- ~Test fail with `nice`: https://bugs.astron.com/view.php?id=95~


alarcher:
@Mno-hime Can we merge?

components/shell/tcsh/Makefile
```diff
@@ -67,12 +60,11 @@ COMPONENT_TEST_TRANSFORMS += \
...
-# common targets
 build:		$(BUILD_64)
 
 install:	$(INSTALL_64)
 
 test: 		$(TEST_64)
 
+# Auto-generated dependencies
 REQUIRED_PACKAGES += system/library
-REQUIRED_PACKAGES += text/locale
```
pyhalov: Please, check if it is a build dependency

pyhalov: @Mno-hime

Mno-hime:
> @Mno-hime Can we merge?

I think we can merge now.

PR #5087, opened 2019-06-18
tracker: add 64-bit build

timmooney:
This pull request requires that pull request #5086 be applied first, to add 64-bit support to `totem-pl-parser`.

These changes to `library/desktop/search/tracker` add 64-bit to the build.

I investigated updating tracker, but there are multiple issues:
  * tracker 2.x now uses [meson](http://mesonbuild.com) to build instead of autotools.
  * tracker 2.x also appears to split components into multiple separate tarballs, so there would need to be more extensive packaging changes.
  * I then looked at updating tracker from 1.8.3 to 1.12.x, but that version of tracker in that series have a build problem with sqlite3.  The problem impacts Linux too, as I found lots of problems reported related to `libtracker-data.so` and `sqlite3_bind_pointer`.

The one change that's included **that may not be appropriate** is that my build box had `taglib` installed, and tracker's configure noticed that and added new files to the install because of it.  If you want me to revert that change and the addition of `taglib` to `REQUIRED_PACKAGES`, let me know.

Tracker will also build additional components if the build box has thunderbird installed.  It's possible that we should be more explicit about `--disable` options to `configure` if we don't want things like `trackerbird`.

Overview of changes:
  * bump existing `COMPONENT_REVISION`
  * switch `build` and `install` targets to the 32_and_64 variant
  * move `developer/icu` from the "autogenerated" dependency section to the hand-curated build requirements.  It's not automatically included by `gmake REQUIRED_PACKAGES`
  * re-run `gmake sample-manifest`
  * import changes from the updated `manifests/sample-manifest.p5m` to `tracker.p5m`.  These are mainly the 64-bit related files, but there's also the **inclusion of taglib**
  * re-run `gmake REQUIRED_PACKAGES` and keep the **new** `library/audio/taglib` requirement.  Let me know if this should be reverted.


pyhalov:
Tracker fails to build gobject-introspection. 
Likely needs  CC and CFLAGS set in COMPONENT_BUILD_ENVIRONMENT

timmooney:
I'm sorry, I think I need more information.  I'm not seeing any compile or link failures in my build of tracker on my dev box, and configure is saying that it detected gobject-introspection.

Can you paste in the output from the compile or link failure?

pyhalov:
Hm... Now can't reproduce the issue

PR #5088, opened 2019-06-18
Update brasero to 3.12.2 and include 64-bit

timmooney:
This change requires that pull request #5086 and pull request #5087 be merged first.

This updates `brasero` from 3.12.1 to 3.12.2 and adds 64-bit to the build.

The only interesting change is that the patch for `pfexec` needed to change just slightly because of wording changes in the `brasero.desktop.in.in` file.

The other changes were pretty straightforward:
  * change the `-D_FILE_OFFSET_BITS=64` to only be for `CFLAGS.32`
  * differentiate between the `libexec` location for 32 and 64 bit builds
  * change `build` and `install` to use the 32_and_64 variant
  * **add** `gtk-doc` as a build dependency, since `configure` cannot be rebuilt if it's not installed
  * move the `library/gnome/yelp-tools` from the auto-generated section to the hand-curated build requirements.  It is not part of the output from `gmake REQUIRED_PACKAGES`
  * re-run `gmake sample-manifest`
  * import all the file changes from the `manifest/sample-manifest.p5m' into the main `brasero.p5m`.  Most are 64-bit related, some are new translation-related files with the minor version updated.
  * **add** the `transform` for locale-related files to `brasero.p5m`.
  * re-run `gmake REQUIRED_PACKAGES`

Since the version change is minor, I'm assuming I don't need to worry about triggering a dependency rebuild.

Feedback welcome!

### DIFF
--- a/components/desktop/gnome3/brasero/Makefile
+++ b/components/desktop/gnome3/brasero/Makefile
@@ -15,13 +15,13 @@
 include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=	brasero
-COMPONENT_VERSION=	3.12.1
+COMPONENT_VERSION=	3.12.2
 COMPONENT_PROJECT_URL=	https://projects.gnome.org/brasero/
 COMPONENT_SUMMARY=	Gnome CD/DVD burner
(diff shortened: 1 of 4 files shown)


pyhalov (approved):

pyhalov:
@Mno-hime Should we merge it?

PR #5089, opened 2019-06-18
Install gcc-7-runtime by default after 9996

Mno-hime:
After [9996](https://www.illumos.org/issues/9996) GCC 7 is the default compiler hence we need `gcc-7-runtime` installed by default. It's not installed with `0.5.11-2018.0.0.18599` but GCC 7 `RPATH` and `RUNPATH` are there:

```
newman@minimal:~$ elfdump /usr/sbin/beadm | grep -e RPATH -e RUNPATH
       [9]  RUNPATH           0x6b2               /usr/gcc/7/lib
      [10]  RPATH             0x6b2               /usr/gcc/7/lib
```

To prevent breakage we should install `gcc-7-runtime` never the less.

### DIFF
--- a/components/meta-packages/install-types/Makefile
+++ b/components/meta-packages/install-types/Makefile
@@ -16,7 +16,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		install-types
 COMPONENT_VERSION=	0.1
-COMPONENT_REVISION=	34
+COMPONENT_REVISION=	35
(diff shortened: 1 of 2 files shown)


alarcher (approved):

Mno-hime:
I think @pyhalov will come with a proper solution, so it's installed as illumos dependency. See illumos-developer and #oi-dev.

pyhalov:
I've filed https://www.illumos.org/issues/11255 and working on a fix. This is not a proper fix as it will not help existing installations.

PR #5098, opened 2019-06-21
Update mate-themes 3.22.20

Mno-hime:
There are several visual regressions, filed https://github.com/mate-desktop/mate-themes/issues/255.

But nimbus is fine.


Mno-hime:
How much do we care about themes other than nimbus?

pyhalov:
Well, we need at least one sane not-heavily-customized theme as reference to fix nimbus when it's broken again...

AndWac:
A newer version has been merged

PR #5125, opened 2019-07-16
sbcl 1.5.4

richlowe:
This has two real changes.

The most important one fixes the version number generation to only operate if in an _SBCL_ git checkout, such that we stop smashing the version string
(it current reads `SBCL 10919.-68f8bb0fe`).  I have submitted this upstream as https://bugs.launchpad.net/bugs/1836663

Also, and perhaps less importantly, this updates to SBCL 1.5.4 and adds a target for 'gmake test' to run the test suite.

/cc @jeffpc

### DIFF
--- a/components/runtime/sbcl/Makefile
+++ b/components/runtime/sbcl/Makefile
@@ -15,13 +15,13 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		sbcl
-COMPONENT_VERSION=	1.4.10
+COMPONENT_VERSION=	1.5.4
 COMPONENT_PROJECT_URL=	http://www.sbcl.org/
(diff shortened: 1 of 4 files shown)


richlowe:
There are two unexpected test failures with 1.5.4 which seem to be weirdly related to threading being disabled (with threading enabled they pass -- though others fail).

richlowe:
Actually, I don't see test failures with `:sb-thread` that matter.  One is an overly-short timeout in the `frlock` test of `:sb-concurrency`, and one is due to SBCL finding `cc` and assuming it handles `alloca()` well.

@jeffpc what was wrong with the threading on x86? (beyond it not existing on x86-64, which would prove a problem down the line)

pyhalov (approved):

Mno-hime (approved):

PR #5130, opened 2019-07-19
illumos-gate: avoid running smatch for OI illumos-gate builds

pyhalov:


pyhalov:
We build almost pristine illumos-gate, smatch is useless for regular builds.

Mno-hime (approved):

PR #5131, opened 2019-07-19
Fix mate_install and obsolete mate-icon-theme-faenza

AndWac:


Mno-hime (approved):
Sorry about that, I keep forgetting about having FMRI higher than the shipped one.

pyhalov (approved):

PR #5133, opened 2019-07-19
[2/2] GnuPG 2.2: Update GnuPG to 2.2.17

Mno-hime:
Merge after libassuan & npth (https://github.com/OpenIndiana/oi-userland/pull/5132).


Mno-hime:
GPG signing in `git` works in build zone and keys are unlocked in GZ via `pinentry`.

PR #5145, opened 2019-07-22
sbcl: re-enable threading, ship the manual

richlowe:
This re-enables threading, and ships the manual in HTML and GNU info format. 

It actually re-enables threading by enabling the 'fancy' features, and documenting what they are at the minute, letting upstream pick and choose rather than us picking and choosing.

There are two bugs the comment about the tests failing could refer to.  One was a bug in illumos KPTI mitigation (which is fixed), the other is the bug the patch deals with.  The timeout of the frlock test is not nearly long enough on some build machines (this may indicate a bug, but it's not clear).  I chose a timeout that happens to work for me 100% of the time, it is not particularly scientific.

I skipped the PDF manual because it requires TeX and seems unnecessary, and the internals manual as it requires graphviz and seems like a much more limited audience.

There's one more intermittent build failure in sbcl (with or without the recent changes), that I plan to get reviewed and integrated upstream first.  If sb-posix fails to build because of mktemp, you have hit that failure, a rebuild will almost certainly work (the failure depends on the string mktemp generates)


components/runtime/sbcl/Makefile
```diff
@@ -36,14 +36,15 @@ COMPONENT_ENV +=	GNUMAKE=$(GMAKE)
...
-COMPONENT_BUILD_ARGS += --with-sb-core-compression
+# This is currently --with-sb-thread --with-sb-core-compression
+# --with-sb-xref-internals and :sb-after-xc-core as a build feature.
+COMPONENT_BUILD_ARGS +=	--fancy
 
 $(BUILD_DIR)/%/.built:	$(SOURCE_DIR)/.prep
 	$(RM) -r $(@D) ; $(MKDIR) $(BUILD_DIR)
 	$(CP) -a $(SOURCE_DIR) $(@D)
 	(cd $(@D) ; $(ENV) $(COMPONENT_ENV) /usr/gnu/bin/sh make.sh $(COMPONENT_BUILD_ARGS))
+	(cd $(@D)/doc/manual ; $(ENV) $(COMPONENT_ENV) BUILD_ROOT=$(PROTO_DIR) gmake html info)
```
Mno-hime:
```suggestion
	(cd $(@D)/doc/manual ; $(ENV) $(COMPONENT_ENV) BUILD_ROOT=$(PROTO_DIR) $(GMAKE) html info)
```

Mno-hime (commented):
`COMPONENT_REVISION` needs to be set to `1`.

richlowe:
I have one more update of this incoming to fix the intermittent build failure I referred to with a patch from upstream

richlowe:
I didn't bump `COMPONENT_REVISION` for that second change, should it be done per-commit or per-push?

pyhalov:
Honestly, don't have strong preference here. Two main points of COMPONENT_REVISION  - be certain to trigger rebuild (for example, when patch is added) and to be able to differentiate between component versions. Perhaps, COMPONENT_REVISION per commit would be more consistent.

PR #5149, opened 2019-07-23
x264: update to 157.20190721

pyhalov:


pyhalov:
ABI-incompatible update.

Should be merged prior to #4034 
The following packages should be rebuilt:
```
pkg:/library/audio/gstreamer/plugin/ugly
pkg:/library/audio/gstreamer1/plugin/ugly
pkg:/media/vlc
```

PR #5158, opened 2019-07-26
Update desktop-file-utils to 0.24

Mno-hime:
**Testing**
- will keep it in my build env and desktop for a while


alarcher:
@Mno-hime Probably fine and good to go?

Mno-hime:
> @Mno-hime Probably fine and good to go?

Yes I think so.

PR #5160, opened 2019-07-26
2/2: libass 0.14.0 rebuild (VLC, FFmpeg, Gstreamer plugins bad)

Mno-hime:
Merge after [libass 0.14.0](https://github.com/OpenIndiana/oi-userland/pull/5159).


pyhalov:
@Mno-hime: Do we have to rebuild ffmpeg prior to vlc?

Mno-hime:
I think VLC uses libass via it's plugin not directly via FFmpeg:
```
$ ldd /usr/lib/amd64/vlc/plugins/codec/liblibass_plugin.so | grep ass
	libass.so.9 =>	 /usr/lib/64/libass.so.9
```

PR #5161, opened 2019-07-28
update to samba-4.11.0

chincheta0815:
This is the first shot with samba-4.11. Up to know it is using the samba-4.11.0rc1 and the first patches are updated or deleted where possible due to their incorporation upstream.

Once samba-4.11.0 is no more a release candidate I will inform you.


chincheta0815:
@pyhalov I am getting the following message when typing "gmake publish".
ERROR userland.action002.0        link usr/bin/pidl has unresolvable target '../lib/samba/bin/amd64/pidl'

Do you have a clue what os going wrong there? pdil is in /usr/bin anyway...

Another thing doing the gmake publish: I have a lot of lin warning concerning python3.5 telling me that the files of samba's pythonlibs are in a 32bit path. Can you give me a hint there?

Thanks in advance.

Mno-hime:
In my case `pidl` is _not_ installed to (`build/prototype/i386`):
```
[samba-4.11] {build-userland} newman:~/ws/oi-userland/components/network/samba $ find . -name pidl
./samba-4.11.0rc4/pidl
./samba-4.11.0rc4/pidl/pidl
./build/amd64/bin/default/pidl
./build/amd64/pidl
./build/amd64/pidl/pidl
```
Could be a bug or perhaps some additional Make target needs to be executed. In any case you can reference `build/amd64/pidl/pidl` in p5m file, but better install it somehow, even if "manually".

Note that 32-bit Perl was detected in the 64-bit build. As `pidl` is a Perl script this might be relevant:
```
Checking for program 'perl'                                                       : /usr/bin/perl 
Checking for program 'perl'                                                       : /usr/bin/perl 
Checking for minimum perl version 5.0.0                                           : 5.22.4 
Checking for perl $Config{vendorprefix}:                                          : '/usr/perl5/5.22' 
PERL_ARCH_INSTALL_DIR:                                                            : '${LIBDIR}/perl5' 
PERL_LIB_INSTALL_DIR:                                                             : '${DATADIR}/perl5' 
PERL_INC:                                                                         : ['/usr/perl5/site_perl/5.22/i86pc-solaris-64int', '/usr/perl5/site_perl/5.22', '/usr/perl5/vendor_perl/5.22/i86pc-solaris-64int', '/usr/perl5/vendor_perl/5.22', '/usr/perl5/5.22/lib/i86pc-solaris-64int', '/usr/perl5/5.22/lib'] 
```

Perhaps you need something like `CONFIGURE_ENV.64 += PERL=$(PERL.5.24)`?

> Another thing doing the gmake publish: I have a lot of lin warning concerning python3.5 telling me that the files of samba's pythonlibs are in a 32bit path. Can you give me a hint there?

All Pythons we delivered before Python 3.5 (the one used here) were 32- and 64-bit. Python 3.5 is 64-bit-only and is not using arch specific path for modules, i.e. no `/amd64/` directory in the module path (contrary to 64-bit Python 2.7 and 3.4). I think you can safely ignore this `INFO`.

components/network/samba/Makefile
```diff
@@ -28,12 +28,13 @@ include ../../../make-rules/shared-macros.mk
...
-COMPONENT_SRC =			$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_VERSION =		4.11.0
+COMPONENT_SRC =			$(COMPONENT_NAME)-$(COMPONENT_VERSION)rc4
 COMPONENT_PROJECT_URL =		http://www.samba.org/
-COMPONENT_ARCHIVE =		$(COMPONENT_SRC).tar.gz
-COMPONENT_ARCHIVE_HASH = 	sha256:9efbeb52db1203dc779b118f1c48c161e569f7a6af5101e745497ee6296eef42
-COMPONENT_ARCHIVE_URL =		http://us1.samba.org/samba/ftp/stable/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE =             $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH = 	sha256:29663121eb4ed259bb3bd6c3038b27255f4414a1cca120acd7f29493c22df37f
+COMPONENT_ARCHIVE_URL =		https://download.samba.org/pub/samba/rc/$(COMPONENT_ARCHIVE).tar.gz
```
Mno-hime:
```suggestion
COMPONENT_ARCHIVE_URL =		https://download.samba.org/pub/samba/rc/$(COMPONENT_ARCHIVE)
```

chincheta0815: I know, but until the release is out I will keep it this way. Then I switch to the lines which are there from the beginning...

chincheta0815:
> Note that 32-bit Perl was detected in the 64-bit build. As `pidl` is a Perl script this might be relevant:
> 
> ```
> Checking for program 'perl'                                                       : /usr/bin/perl 
> Checking for program 'perl'                                                       : /usr/bin/perl 
> Checking for minimum perl version 5.0.0                                           : 5.22.4 
> Checking for perl $Config{vendorprefix}:                                          : '/usr/perl5/5.22' 
> PERL_ARCH_INSTALL_DIR:                                                            : '${LIBDIR}/perl5' 
> PERL_LIB_INSTALL_DIR:                                                             : '${DATADIR}/perl5' 
> PERL_INC:                                                                         : ['/usr/perl5/site_perl/5.22/i86pc-solaris-64int', '/usr/perl5/site_perl/5.22', '/usr/perl5/vendor_perl/5.22/i86pc-solaris-64int', '/usr/perl5/vendor_perl/5.22', '/usr/perl5/5.22/lib/i86pc-solaris-64int', '/usr/perl5/5.22/lib'] 
> ```
> 
> Perhaps you need something like `CONFIGURE_ENV.64 += PERL=$(PERL.5.24)`?
> 

Is the 5.24 Perl 64-bit an 5.22 not? Then It would maybe also make sense to delete the PREREQUISITE for Perl 5.22? Right now the system default perl 5.22 is used...

Mno-hime:
> Is the 5.24 Perl 64-bit an 5.22 not? Then It would maybe also make sense to delete the PREREQUISITE for Perl 5.22? Right now the system default perl 5.22 is used...

Yes, check `/usr/perl5/5.22/bin/perl -V` and `/usr/perl5/5.24/bin/perl -V` output.

Yes, that prereq is what was generated when the 32-bit Perl was detected.

chincheta0815:
@pyhalov @Mno-hime Okay. Think I got it for samba-4.11.

chincheta0815:
Forgot to change title of this pr.

components/network/samba/patches/04-int-syscall-sysno.patch.inactive
```diff
@@ -0,0 +1,11 @@
...
++++ third_party/uid_wrapper/uid_wrapper.c	2019-07-27 10:30:30.145913588 +0000
+@@ -235,7 +235,7 @@
+ typedef int (*__libc_setgroups)(size_t size, const gid_t *list);
+ 
+ #ifdef HAVE_SYSCALL
+-typedef long int (*__libc_syscall)(int sysno, ...);
++typedef long int (*__libc_syscall)(long int sysno, ...);
+ #endif
+ 
+ #define UWRAP_SYMBOL_ENTRY(i) \
```
Mno-hime: Do we want this "inactive" patch?

components/network/samba/samba.p5m
```diff
@@ -89,7 +89,7 @@ link path=usr/bin/nmblookup target=../lib/samba/bin/$(MACH64)/nmblookup
 link path=usr/bin/ntlm_auth target=../lib/samba/bin/$(MACH64)/ntlm_auth
 link path=usr/bin/oLschema2ldif target=../lib/samba/bin/$(MACH64)/oLschema2ldif
 link path=usr/bin/pdbedit target=../lib/samba/bin/$(MACH64)/pdbedit
-link path=usr/bin/pidl target=../lib/samba/bin/$(MACH64)/pidl
+#link path=usr/bin/pidl target=../lib/samba/bin/$(MACH64)/pidl
```
Mno-hime, line 98: We don't provide `Parse:Yapp` Perl module, so no `pidl`. I think you can remove the commented link.

chincheta0815: I just commented that out, because there is a patch pending at upstream which should re-add pidl as an install target in the near future. If it is fine let is just commented out. I will add a comment on that in the line above.

components/network/samba/patches/04-int-syscall-sysno.patch.inactive
chincheta0815: removed that inactive patch.

Mno-hime:
It looks good to me.

What scenarios did you test with Samba 4.11?

Thought about enabling test target? From a quick look there's a self test.

chincheta0815:
I actually tested it in my home environment, as I did with the former versions. That is not as deep as it would be for a AD system of a large enterprise, to be honest.

I thought about enabling the self tests. All I found out so far is that these self tests are tagged as optional by the samba how to for compiling the source. Another point I found was that the selftests are quite a lot: approx. 3000+ with some that will fail and throw errors although the code is okay... There seems to be also a quicktest, that should be not that hard. In the end they are often referred to be for the developers.

So far it seems to be a kind of own "project" to incorporate these tests properly into the build process. Therefore, it might make sense to add the package as it is without the tests to get the new samba version and then start working on the test incorporation.

Mno-hime (approved):

chincheta0815:
Thanks.
I am already working on the possibility to implement the tests.

chincheta0815:
Implementing the tests seems a little harder than expected due to some issues with the upstream code.
So, when you are holding back this merge because you are waiting: Please merge. It can take longer implementing the tests...

components/network/samba/patches/10-avahi.patch
```diff
@@ -1,20 +0,0 @@
---- samba-4.9.5/source3/smbd/avahi_register.c	2019-03-12 09:19:04.000000000 +0000
-+++ samba-4.9.5/source3/smbd/avahi_register.c	2019-03-18 17:58:47.199335371 +0000
-@@ -186,7 +186,7 @@
```
pyhalov, line 3: What has happened to this patch?

chincheta0815:
Merged upstream. Should now be unnecassary.

pyhalov (approved):

components/network/samba/patches/10-avahi.patch
Mno-hime: https://github.com/OpenIndiana/oi-userland/pull/5377

PR #5163, opened 2019-07-28
add faad2-2.8.8

chincheta0815:


alarcher:
This component is already in encumbered and since the update breaks ABI multiple dependent components need be checked.

chincheta0815:
@alarcher Thanks for the information. Unfortunately, I do not understand it well. What does "encumbered" in that context mean and what do you mean by "since the update breaks ABI multiple dependent components need be checked"?

Sorry for that, I am not familiar with that wording and want to understand things. Maybe it is not worth to add faad2 then?

chincheta0815:
@alarcher Okay. There popped up something in my mind... Do you mean by "encumbered" that this belongs into the encumbered repository due to I think it was patent restictions? If yes then the "breaks" components part means: Put it into encumbered/update the package in encumbered and all is okay?
(I found the encumbered repo. I have to admit by accident, but I found it.)

pyhalov:
Due to patent issues, some packages are published to separate /hipster-encumbered repository.
No package from /hipster repo should depend on them. If you want to bump encumbered/faad2 component, be sure that dependent components - gst-plugins-bad, gst-plugins-bad1, vlc - build fine with new version. After PR for updating faad2 you'll have to submit PR to bump their COMPONENT_REVISION (if they build/work fine with new faad2 version, otherwise second PR should also include fixes for mentioned components to work correctly with updated faad2) so that they are rebuilt after faad2 bump.

pyhalov (changes requested):
Don't introduce new component, update existing one.

chincheta0815:
@pyhalov Okay. Thanks. Then everything is okay. I thought there was no faad2. So I will stick to the one already there.

PR #5164, opened 2019-07-28
mongodb: bump to 4.0.11

grueni:


Mno-hime (dismissed):
As discussed in https://github.com/OpenIndiana/oi-userland/pull/4630 anything newer than v4.0.3 is under the [Server Side Public License v1](https://github.com/mongodb/mongo/blob/master/LICENSE-Community.txt), which is not open source and, potentially, even dangerous. Linux distros generally does not even ship it.

@pyhalov I will block this PR before we find a way forward license-wise.

~One solution might be to move mongodb 4.0.11 to encumbered/, obsolete it in the global history file, fix `COMPONENT_LICENSE` to include the SSP license v1, and potentially blocking the mongodb service start unless administrator sets special SMF variable acknowledging, that they are aware of the risks.~ Update: As @pyhalov suggested, I think license action asking user to acknowledge the SSPv1 license is enough.

@grueni @pyhalov What do you think?

pyhalov:
I still don't understand the following:
1) are we authorized to redistribute it? 
2) Do I understand correctly, that if our users use it to provide cloud services, it's their problem, not ours?

We and Debian are different in one important thing. 
There are official repositories for Debian, shipping mongodb. And Debian move, perhaps, correct and spectacular, have not affected end users, they just switched repositories, and that's all. For our users this means compiling own patched version, as non-patched version has its own issues. 

Why should we move it to /hipster-encumbered? It's not evident for me. The main reason for /hipster-encumbered to exist is that US users can't legally use it. 

Yes, it's not free, but we ship license. We can mark license action with must-accept, must-display attributes. We already ship closed-source drivers.

pyhalov:
@grueni Tests fail here with `ImportError: No module named urllib3.util.retry`.
Should we switch component to Python3.5 or do we need some additional Python modules?

components/database/mongodb-40/Makefile
```diff
@@ -27,7 +27,7 @@ COMPONENT_CLASSIFICATION= System/Databases
 COMPONENT_SRC=		$(COMPONENT_NAME)-src-r$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_URL=	https://fastdl.mongodb.org/src/$(COMPONENT_ARCHIVE)
-COMPONENT_ARCHIVE_HASH=	sha256:fbbe840e62376fe850775e98eb10fdf40594a023ecf308abec6dcec44d2bce0c
+COMPONENT_ARCHIVE_HASH=	sha256:c2672443f01ea4523a356edc8804cb6af31e831f63a88ee4a100aafbff443e4f
 COMPONENT_LICENSE= AGPLv3, zlib, Apache v2.0
```
pyhalov: COMPONENT_LICENSE has changed

Mno-hime:
>     1. are we authorized to redistribute it?

Yes, I believe we are able to ship it.

>     2. Do I understand correctly, that if our users use it to provide cloud services, it's their problem, not ours?

Yes, I believe this is the case as well.

> We and Debian are different in one important thing.
> There are official repositories for Debian, shipping mongodb. And Debian move, perhaps, correct and spectacular, have not affected end users, they just switched repositories, and that's all. For our users this means compiling own patched version, as non-patched version has its own issues.

Understood. And I believe MongoDB is something we should provide to our users.

> Why should we move it to /hipster-encumbered? It's not evident for me. The main reason for /hipster-encumbered to exist is that US users can't legally use it.

I think I misrepresented hipster-encumbered with "non-free stuff".

> Yes, it's not free, but we ship license. We can mark license action with must-accept, must-display attributes. We already ship closed-source drivers.

If we are able to make user acknowledge the license before install/update, I think we are all set and I don't have any other "extra-mile-safety" requirements (I wasn't aware about the license action possibility).

pyhalov:
Tried to use Python 3 scons... MongoDB is still not ready to this. So likely just need to add urllib3 module.

components/database/mongodb-40/Makefile
grueni: I have installed this packages and AFAIS python-27 is used.
library/python/pip-27
library/python/requests-27
library/python/pip-35
Do all tests fail?

pyhalov:
Tests pass fine with urllib3.
Add it to  REQUIRED_PACKAGES under Build dependencies section and fix license (name of license - it's not AGPL) and mark license action with must-accept, must-display attributes.

components/database/mongodb-40/mongodb40.p5m
```diff
@@ -21,7 +21,7 @@ set name=info.source-url value="$(COMPONENT_ARCHIVE_URL)"
 set name=info.upstream-url value="$(COMPONENT_PROJECT_URL)"
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
-license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)"
+license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)" must-accept=true must-display=true
```
pyhalov: Please, do the same for mongodb40-service.p5m.

pyhalov: @Mno-hime I'm just interested, should we add both or must-display is enough?

pyhalov:
Overall now it looks good. The same license action should be present in both packages. After looking at the way pkg handles this, I suggest change license in three actions.  
One license action for zlib license (with corresponding text), one license action for Apache license, one - for SSPL (with must-display=true and perhaps must-accept=true).
I'm not sure about must-accept, as it breaks `pkg update`. Even when I download Mongo from site it doesn't require manual license acception.

grueni:
If you use `pkg update --accept` it will not break.
IMHO the user is solely responsible in the usage of the software and to respect the license terms.
Our job is to deliver the correct license file.
If I search for files with names containing LICENSE I find a lot of licenses 
`(cd mongodb-src-r4.0.11/; find .|grep -v mongo|grep  LICENSE|xargs head -5) > l.txt`

components/database/mongodb-40/Makefile
```diff
@@ -102,12 +104,11 @@ COMPONENT_TEST_TRANSFORMS +=    '-e "s/\[resmoke\][ -+0-9T:\.\-]*//"' \
                                 '-n ' \
                                 '-e "/Summary of all suites/,\\$$p"'
 
-ENV = /usr/bin/env -i
 
 # Not strictly necessary
 $(TEST_64):     $(INSTALL_64) $(INSTALL_TOOLS)
 
-build:		$(BUILD_64) $(BUILD_TOOLS)
+build:		$(BUILD_64) $(BUILD_TOOLS)
```
Mno-hime: Stray space at the end of the line.

components/database/mongodb-40/mongodb40.p5m
Mno-hime: I am fine with sole `must-display=true`. My point is to go an extra mile with this license, so users are safe. Hopefully with the license displayed, they figure out this time it's different and "Server Side" license on their screen rings some bells.

Mno-hime (commented):
Lifting the dis-approval.

pyhalov:
@grueni: We state that therer's three licenses here. in COMPONENT_LICENSE. It would be fine to show all of them if user asks and we must show only SSPL. So I think, it's worth adding not one, but three license action here (one for zlib, one for Apache, one for SSPL).

grueni:
Like this?
license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)" must-accept=true must-display=true
license src/third_party/s2/LICENSE-2.0.txt license="Apache v2.0"
license src/third_party/zlib-1.2.11/README license="zlib"

pyhalov:
yes, you can just copy Apache 2.0 and zlib license text to apache2.0.license and zlib.license files in components top dir and use that files to avoid issues if their location in source tarball is changed on update
.

pyhalov:
And it seems, must-display=true is fine enough.

components/database/mongodb-40/zlib.license
```diff
@@ -0,0 +1,115 @@
...
+- For PalmOs, see http://palmzlib.sourceforge.net/
+
+
+Acknowledgments:
+
+  The deflate format used by zlib was defined by Phil Katz.  The deflate and
+  zlib specifications were written by L.  Peter Deutsch.  Thanks to all the
+  people who reported problems and suggested various improvements in zlib; they
+  are too numerous to cite here.
+
```
pyhalov: Perhaps, strip everything above 'Copyright notice'?

grueni: OK

pyhalov:
Looks fine. I'll retest and merge after testing.

components/database/mongodb-40/Makefile
```diff
@@ -27,10 +27,9 @@ COMPONENT_CLASSIFICATION= System/Databases
 COMPONENT_SRC=		$(COMPONENT_NAME)-src-r$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_URL=	https://fastdl.mongodb.org/src/$(COMPONENT_ARCHIVE)
-COMPONENT_ARCHIVE_HASH=	sha256:fbbe840e62376fe850775e98eb10fdf40594a023ecf308abec6dcec44d2bce0c
-COMPONENT_LICENSE= AGPLv3, zlib, Apache v2.0
+COMPONENT_ARCHIVE_HASH=	sha256:c2672443f01ea4523a356edc8804cb6af31e831f63a88ee4a100aafbff443e4f
+COMPONENT_LICENSE= SSPL, zlib, Apache v2.0
```
pyhalov: COMPONENT_LICENSE should be only SSPL (other licenses have separate description).

components/database/mongodb-40/mongodb40-service.p5m
```diff
@@ -21,7 +21,7 @@ set name=info.source-url value="$(COMPONENT_ARCHIVE_URL)"
 set name=info.upstream-url value="$(COMPONENT_PROJECT_URL)"
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
-license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)"
+license $(COMPONENT_LICENSE_FILE) license="$(COMPONENT_LICENSE)" must-accept=true must-display=true
```
pyhalov: Fix license action in the same way as for mongodb40.p5m (add 3 licenses and set only must-display=true on SSPL).

pyhalov:
Sorry, have spotted only when testing.

grueni:
No problem

pyhalov:
Thanks

PR #5165, opened 2019-07-28
haproxy: bump to 2.0.3

grueni:


pyhalov:
Build fails here with
```
ld: fatal: unrecognized option '--export-dynamic'
ld: fatal: use the -z help option for usage information
```

grueni:
I missed to attach the patch for Makefile

PR #5168, opened 2019-07-28
net-snmp: bump to 5.8

grueni:


components/sysutils/net-snmp/Makefile
```diff
@@ -20,51 +20,50 @@
...
 COMPONENT_SRC=    $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_HASH= \
-    sha256:12ef89613c7707dc96d13335f153c1921efc9d61d3708ef09f3fc4a7014fb4f0
-COMPONENT_ARCHIVE_URL=	http://ftp.ntua.gr/mirror/net-snmp/net-snmp/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+    sha256:b2fc3500840ebe532734c4786b0da4ef0a5f67e51ef4c86b3345d697e4976adf
+COMPONENT_ARCHIVE_URL=	https://sourceforge.net/projects/net-snmp/files/net-snmp/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
 COMPONENT_CLASSIFICATION= System/Enterprise Management
 COMPONENT_SUMMARY= Net-SNMP Agent files and libraries
+COMPONENT_BUGDB=	service/net-snmp utility/net-snmp library/net-snmp
```
pyhalov: We don't need these IDs (BUGDB, ANITYA_ID, TPNO)

components/sysutils/net-snmp/Makefile
```diff
@@ -20,51 +20,50 @@
...
+COMPONENT_ANITYA_ID=	2062
+
+TPNO=			47655
 
+# obsolete - reminder for common targets build/install/test
 BUILD_32_and_64 += $(BUILD_DIR_32)/.docs
+INSTALL_TARGET= build $(INSTALL_32_and_64)
+TEST_TARGET= $(TEST_32_and_64)
 
-COMPONENT_BUILD_ARGS=
```
pyhalov, line 53: Does it build fine in parallel now? (COMPONENT_BUILD_ARGS=-j4 in build environment)

components/sysutils/net-snmp/Makefile
```diff
@@ -73,15 +72,15 @@ CONFIGURE_OPTIONS  +=	--with-persistent-directory=$(VARDIR)/net-snmp
 
 CONFIGURE_OPTIONS  +=	--with-mibdirs=$(ETCDIR)/net-snmp/snmp/mibs
 CONFIGURE_OPTIONS  +=	--datadir=$(ETCDIR)/net-snmp
+CONFIGURE_OPTIONS  +=	--enable-agentx-dom-sock-only
 CONFIGURE_OPTIONS  +=	--enable-ucd-snmp-compatibility
 CONFIGURE_OPTIONS  +=	--enable-ipv6
 CONFIGURE_OPTIONS  +=	--enable-mfd-rewrites
 CONFIGURE_OPTIONS  +=	--with-openssl=/lib/openssl
-CONFIGURE_OPTIONS  += --sysconfdir=$(ETCDIR)/net-snmp
```
pyhalov, line 80: Why configuration files are moved out of /etc?

components/sysutils/net-snmp/Makefile
```diff
@@ -91,16 +90,19 @@ PYTHON_ARGS =		--basedir=/usr/bin
 CONFIGURE_OPTIONS  +=	--with-python-modules="$(PYTHON_ARGS)"
 
 # Only build 64 bit Perl support
-PERL_ARGS =	DESTDIR=$(PROTO_DIR) INSTALLDIRS=vendor
-CONFIGURE_OPTIONS.32  +=  PERLPROG="$(PERL.5.22)"
+PERL_ARGS =			DESTDIR=$(PROTO_DIR) INSTALLDIRS=vendor
 CONFIGURE_OPTIONS.32  +=	--disable-embedded-perl
-CONFIGURE_OPTIONS.32  +=  --with-perl-modules="$(PERL_ARGS)"
-CONFIGURE_OPTIONS.64  +=	PERLPROG="$(PERL.5.24)"
+CONFIGURE_OPTIONS.32  +=	--without-perl-modules
```
pyhalov, line 89: Why do we avoid building perl modules?

components/sysutils/net-snmp/history
```diff
@@ -1,2 +1 @@
-library/python-2/net-snmp-26@5.4.4,5.11-2016.0.1.0
-library/python-2/net-snmp-27@5.4.4,5.11-2016.0.1.0 library/python/net-snmp-27
+system/management/snmp/net-snmp/documentation@5.7.3-2018.0.0.4 system/management/snmp/net-snmp/demo
```
pyhalov: Why have you removed old content?

components/sysutils/net-snmp/patches/076.21442153.sysconf.patch
```diff
@@ -0,0 +1,23 @@
+Solaris specific patch, not suitable for upstream.
+--- net-snmp-5.8.old/man/snmptrapd.conf.5.def	2018-07-16 07:33:40.000000000 +0000
++++ net-snmp-5.8/man/snmptrapd.conf.5.def	2019-03-08 05:22:34.480553145 +0000
+@@ -298,7 +298,7 @@
+ The default in each case is for the feature to be turned off, so these
+ directives are typically only used to enable the appropriate behaviour.
+ .SH FILES
+-SYSCONFDIR/snmp/snmptrapd.conf
++DATADIR/snmp/snmptrapd.conf
```
pyhalov, line 9: Is snmptrapd.conf moved out of /etc?

pyhalov:
Have you checked that dependent components build fine (including illumos-gate)?
These are 
```
pkg:/application/cluster/cluster-glue
pkg:/application/cluster/corosync
pkg:/application/cluster/pacemaker
pkg:/image/scanner/xsane/sane-backends
pkg:/network/asterisk
pkg:/print/filter/hplip
pkg:/system/management/nut/drivers
pkg:/system/management/powerman
```

components/sysutils/net-snmp/Makefile
grueni: will check it

components/sysutils/net-snmp/Makefile
grueni: This is the same location as the default. 
For the sake of clarification we can add this again:
CONFIGURE_OPTIONS  += --sysconfdir=$(ETCDIR)/net-snmp

components/sysutils/net-snmp/patches/076.21442153.sysconf.patch
grueni: No. If used it should be in /etc/net-snmp/snmp
sysconfdir = datadir

CONFIGURE_OPTIONS  += --sysconfdir=$(ETCDIR)/net-snmp
CONFIGURE_OPTIONS  += --with-mibdirs=$(ETCDIR)/net-snmp/snmp/mibs
CONFIGURE_OPTIONS  += --datadir=$(ETCDIR)/net-snmp

components/sysutils/net-snmp/Makefile
grueni: No chance. The build fails with multiple includes of some libraries.

components/sysutils/net-snmp/Makefile
grueni: It did not compile with --with-perl-modules. But I found the failure in a Makefile.PL and created a patch for it. It compiles now with the perl modules.

components/sysutils/net-snmp/history
grueni: I have to admit that I don't understand this file. 
I guess the added line is unnecessary too.
Drop the changes in history?

Please have a look at the end of net-snmp-demo.p5m.
Actually it is:
depend type=optional \
    fmri=system/management/snmp/net-snmp/documentation@5.7.3,5.11-11.5.0.0.0.50.0
Should this be:
depend type=optional \
    fmri=system/management/snmp/net-snmp/documentation@5.7.3

components/sysutils/net-snmp/Makefile
pyhalov: This means `COMPONENT_BUILD_ARGS=` should be preserved.

components/sysutils/net-snmp/net-snmp-demo.p5m
```diff
@@ -0,0 +1,201 @@
...
+    path=usr/demo/net-snmp/demo_module_9/demo_module_9.h
+file build/$(MACH32)/sun/sdk/demo/demo_module_9/get_demo_module_9 \
+    path=usr/demo/net-snmp/demo_module_9/get_demo_module_9
+file build/$(MACH32)/sun/sdk/demo/demo_module_9/set_demo_module_9 \
+    path=usr/demo/net-snmp/demo_module_9/set_demo_module_9
+file build/$(MACH32)/sun/sdk/demo/demo_module_9/walk_demo_module_9 \
+    path=usr/demo/net-snmp/demo_module_9/walk_demo_module_9
+license COPYING license="BSD, BSD-like"
+depend type=optional \
+    fmri=system/management/snmp/net-snmp/documentation@5.7.3
```
pyhalov, line 199: The idea is the following. This should be an optional dependency on package version from history file. And history file causes renamed package to be published.
Now we have system/management/snmp/net-snmp/documentation package with obsolete and rename (to  system/management/snmp/net-snmp/demo) options. And system/management/snmp/net-snmp/demo package depending on the exact system/management/snmp/net-snmp/documentation version which contains this rename option.
So, when user installs net-snmp/demo package, it will trigger update of net-snmp/documentation package (if it's installed) to renamed version, which leads to its deinstallation.
I hope my explanation was sensible :) If not, look at https://docs.openindiana.org/dev/pdf/ips-dev-guide.pdf , chapter 6, dependency types.

components/sysutils/net-snmp/history
pyhalov: It should be the version of package with rename attribute (which is generated based on history file, look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/ips.mk#L329 )

components/sysutils/net-snmp/Makefile
```diff
@@ -94,17 +93,18 @@ PERL_ARGS =			DESTDIR=$(PROTO_DIR) INSTALLDIRS=vendor
...
 CONFIGURE_OPTIONS.32  +=	--without-perl-modules
 CONFIGURE_OPTIONS.64  +=	PERLPROG="$(PERL)"
-#CONFIGURE_OPTIONS.64  +=	--with-perl-modules="$(PERL_ARGS)"
-CONFIGURE_OPTIONS.64  +=	--without-perl-modules
-#CONFIGURE_OPTIONS.64  +=	--disable-perl-cc-checks
 CONFIGURE_OPTIONS.64  +=  --disable-embedded-perl
+CONFIGURE_OPTIONS.64  +=  --with-perl-modules="$(PERL_ARGS)"
+
+# parallel build not supported
+COMPONENT_BUILD_ARGS=-j1
```
pyhalov: please, change just to `COMPONENT_BUILD_ARGS=` (common idiom)

components/sysutils/net-snmp/history
```diff
@@ -1 +1,2 @@
-system/management/snmp/net-snmp/documentation@5.7.3-2018.0.0.4 system/management/snmp/net-snmp/demo
+library/python-2/net-snmp-26@5.4.4,5.11-2016.0.1.0
```
pyhalov, line 1: appropriate net-snmp/documentation => demo rename SHOULD also be present, otherwise automatic updates will not be possible. Appropriate here means that system/management/snmp/net-snmp/documentation version should be > current one (for example, system/management/snmp/net-snmp/documentation@5.7.3-2018.0.0.5)

components/sysutils/net-snmp/net-snmp-demo.p5m
grueni: Great explanation, got it.

components/sysutils/net-snmp/history
grueni: The man files are now included in net-snmp-base.p5m.
So history should contain
`system/management/snmp/net-snmp/documentation@5.7.3-2018.0.0.5 system/management/snmp/net-snmp/demo`
which means uninstall older packages containing a mix of man and demo files and use the new demo package containing only demo files.

grueni:
Checked with 
illumos-gate 
pkg:/application/cluster/cluster-glue
pkg:/application/cluster/corosync
pkg:/application/cluster/pacemaker
pkg:/image/scanner/xsane/sane-backends
pkg:/network/asterisk
pkg:/print/filter/hplip
pkg:/system/management/nut/drivers
pkg:/system/management/powerman

components/sysutils/net-snmp/Makefile
```diff
@@ -20,68 +20,66 @@
...
-# proto area; so prefer 32-bit to preserve existing sourcing.
+# Copyright (c) 2011, 2019, Oracle and/or its affiliates. All rights reserved.
+# Copyright 2019 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved. 
 #
+BUILD_BITS= 64_and_32
 
-#
-# Copyright 2016 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
-#
+# The Makefile of this component does not support multiple build jobs.
```
pyhalov: COMPONENT_MAKE_JOBS is solaris-userland variable, unused by oi-userland

grueni:
What is the problem with this PR?

alarcher:
@grueni The bot flags PRs as inactive automatically, nothing wrong. If it is ready and/or needs more testing then it is just a matter of asking for review :)

grueni:
Ok, then please review.

pyhalov:
It seems PR has some comments, which are not addressed.

components/sysutils/net-snmp/Makefile
pyhalov: @grueni  This could be fixed.

components/sysutils/net-snmp/Makefile
pyhalov: Can you readd ti?

components/sysutils/net-snmp/Makefile
```diff
@@ -91,65 +82,59 @@ PYTHON_ARGS =		--basedir=/usr/bin
...
+CONFIGURE_OPTIONS.64  +=  --with-perl-modules="$(PERL_ARGS)"
+
+# parallel build not supported
+COMPONENT_BUILD_ARGS=
 
+# Install the net-snmp-57 libraries
 COMPONENT_PRE_CONFIGURE_ACTION = \
-	($(CLONEY) $(SOURCE_DIR) $(@D); $(CLONEY) $(COMPONENT_DIR)/sun $(@D)/sun)
+	($(CLONEY) $(SOURCE_DIR) $(@D); $(CLONEY) $(COMPONENT_DIR)/sun $(@D)/sun; \
+	cd ../net-snmp-57 && $(GMAKE) install);
```
pyhalov: Have to check, why do we need install on pre-configure target

grueni: I don't know how to let disappear the first 6 commits.
The last one is without TPNO et al.
CONFIGURE_OPTIONS += --sysconfdir=$(ETCDIR)/net-snmp was already added.

grueni:
I don't know how to let disappear the first 6 commits.
The last one is without TPNO et al.
CONFIGURE_OPTIONS += --sysconfdir=$(ETCDIR)/net-snmp was already added.

components/sysutils/net-snmp/Makefile
```diff
@@ -20,68 +20,56 @@
...
 CONFIGURE_OPTIONS  +=	--enable-ucd-snmp-compatibility
 CONFIGURE_OPTIONS  +=	--enable-ipv6
 CONFIGURE_OPTIONS  +=	--enable-mfd-rewrites
 CONFIGURE_OPTIONS  +=	--with-openssl=/lib/openssl
-CONFIGURE_OPTIONS  += --sysconfdir=$(ETCDIR)/net-snmp
+CONFIGURE_OPTIONS  +=	--with-transports="UDP TCP UDPIPv6 TCPIPv6"
 
 # Which MIB modules do we want to build
 MIB_MOBULES.sparc = ucd-snmp/lmSensors
-MIB_MOBULES.i386 = ucd-snmp/lmSensors
```
pyhalov, line 84: What's wrong with them?

pyhalov:
On `gmake clean test publish` I get
```/bin/bash: line 0: cd: ../net-snmp-57: No such file or directory```

File: components/sysutils/net-snmp/Makefile [hunk 13]
```diff
@@ -91,65 +79,59 @@ PYTHON_ARGS =		--basedir=/usr/bin
...
+CONFIGURE_OPTIONS.32  +=	--without-perl-modules
 CONFIGURE_OPTIONS.64  +=	PERLPROG="$(PERL.5.24)"
 CONFIGURE_OPTIONS.64  +=  --disable-embedded-perl
-CONFIGURE_OPTIONS.64  +=	--with-perl-modules="$(PERL_ARGS)"
+CONFIGURE_OPTIONS.64  +=  --with-perl-modules="$(PERL_ARGS)"
 
+# parallel build not supported
+COMPONENT_BUILD_ARGS=
+
+# Install the net-snmp-57 libraries
```
pyhalov: We don't need this for sure.

components/sysutils/net-snmp/Makefile
```diff
@@ -91,65 +79,59 @@ PYTHON_ARGS =		--basedir=/usr/bin
...
+# Get correct libraries from the proto area.
+TEST_LIBPATH.32 =       $(PROTOUSRLIBDIR)
+TEST_LIBPATH.64 =       $(BUILD_DIR)/prototype/$(MACH64)$(USRLIBDIR64)
+COMPONENT_TEST_ENV +=   LD_LIBRARY_PATH=$(TEST_LIBPATH.$(BITS))
 
 # perl man pages hiding here in 64-bit PROTO_DIR
 PKG_PROTO_DIRS += $(BUILD_DIR)/prototype/$(MACH64)
 
-COMPONENT_INSTALL_ARGS += PYTHON_VENDOR_PACKAGES=$(PYTHON_VENDOR_PACKAGES.32)
+# The 5.7 version libraries need to be added from the net-snmp-57 proto path.
```
pyhalov: Remove this block, otherwise publish fails.

pyhalov:
Publish fails with
```
Couldn't find 'usr/lib/python2.7/vendor-packages/netsnmp/64/client_intf.so' in any of the specified search directories:
        /export/home/alp/srcs/oi-userland/components/sysutils/net-snmp
        /export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/build
        /export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/build/prototype/amd64
        /export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/build/prototype/i386
        /export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/build/prototype/i386/mangled
        /export/home/alp/srcs/oi-userland/components/sysutils/net-snmp/net-snmp-5.8
```

grueni:
But be aware of the fact that the table for lmSensors is registered but not filled with values.
At least in my test systems (zones and VMware).

grueni:
Usage of net-snmp-57 is dropped now.
The tests succeed.
The dependent packages could be compiled and published but not asterisk.
Asterisk lacks now postgres as 32-bit binary.

pyhalov:
Do you have https://github.com/OpenIndiana/oi-userland/commit/98d5e9e6b486446b87826873b995e990344a80cf ?

grueni:
Yes. 
I fetched the latest commits in a clean build zone.
Asterisk was never build in this zone.
I repeated the build with the same result.

3 files are missing
-file path=usr/lib/asterisk/modules/cdr_pgsql.so
-file path=usr/lib/asterisk/modules/cel_pgsql.so
-file path=usr/lib/asterisk/modules/res_config_pgsql.so

grueni:
PR for asterisk: https://github.com/OpenIndiana/oi-userland/pull/5413

components/sysutils/net-snmp/Makefile
pyhalov: Please, remove misleading comment.

pyhalov:
Please, create PR to rebuild dependent components (just bump COMPONENT_REVISION and ensure that they build fine)

pyhalov:
Running snmpwalk I get
```
CP-MIB::tcpListenerProcess.ipv4."0.0.0.0".111 = Gauge32: 0
TCP-MIB::tcpListenerProcess.ipv4."0.0.0.0".111 = Gauge32: 0
Error: OID not increasing: TCP-MIB::tcpListenerProcess.ipv4."0.0.0.0".111
 >= TCP-MIB::tcpListenerProcess.ipv4."0.0.0.0".111
```
Seems to be not a regression, but anyway should be fixed.

grueni:
Maybe the problem is not in net-snmp but in the underlying libraries which creates results like this with netstat -an.
IMHO should be just on line for each listening protocol but we get 2. (ipv4, ipv6).

TCP: IPv4
   Local Address        Remote Address    Swind  Send-Q Rwind  Recv-Q    State
-------------------- -------------------- ------ ------ ------ ------ -----------
      *.111                *.*                 0      0 128000      0 LISTEN
      *.*                  *.*                 0      0 128000      0 IDLE
      *.111                *.*                 0      0 128000      0 LISTEN
      *.*                  *.*                 0      0 128000      0 IDLE

pyhalov:
Retested once again, issue appears just with new net-snmp.

pyhalov:
@grueni This ugly hack fixes snmpwalk issue. Please, retest if it doesn't have regressions. 
It's awful, but I don't know how to avoid it , as /dev/arp request  can return non-unique entries, for example, when both tcp6 and tcp socket are listened on.

grueni:
I have a patch which avoids the creation of multiple items.
This should be better than changing snmpwalk.

pyhalov:
Should we merge the last two patches and rework in a single style?

pyhalov:
I mean, https://github.com/OpenIndiana/oi-userland/pull/5168/commits/cdd6960f4e2bb22c61e2049bef938e1febdbd683 and https://github.com/OpenIndiana/oi-userland/pull/5168/commits/85195c9213e1a54db86d846027836e6bafa917df ? 

You can avoid touching COMPONENT_REVISION, as component is still not published.

grueni:
It works without  the patches for snmpwalk. The merge would not be necessary. But I used git fetch and and pull and in fact the two patches are now included.

pyhalov:
Why do you think that this is snmpwalk patches? We patch libnetsnmpmibs, which is used by snmpd.

grueni:
Yes, you are right. Don't know where I got this misinformation from.

pyhalov:
I'm just wondering if complex loops can be transformed to CONTAINER_FIND

grueni:
My patch prevents to add an entry if it is already contained. But just for tcpListenerProcess. The entries in a container are indexed and for the construction of the index the entry is used. the usage is the same with a hash and key/value. See here http://www.net-snmp.org/wiki/index.php/Containers#CONTAINER_FIND

pyhalov:
OK, if it works this way, please, add description to patch, format it and also patch udp table.

grueni:
Patching for udp is unneccessary. It works without patches. The indices are differently constructed compared with tcp.

PR #5177, opened 2019-07-30
11507 clang-40 can't compile a simple C or C++ program: clang: error:…

pyhalov:
… linker command failed with exit code 1


pyhalov:
I'm not sure if we should merge it now due to the fact that this add gcc-4 build dependency back. In fact, I suppose we could avoid it and get only runtime gcc-4 dependency, but I'm still not fan of it. However, if we are not be able to fix Clang++ issue, we'll have to merge it.
Should we do it now and later fix it properly or should we wait for a proper fix?

Note, this does more than original bug report, this fixes Clang++-compiled programs (they somehow are broken with newer G++ runtimes).
@richlowe

Mno-hime:
I always thought of `clang-40` as a means to build Firefox. I'd rather have it replaced by `clang-80` if v8 can live with GCC6+ or somehow disabled, so it's not useful on it's own, then having the GCC4 baggage back. Apparently there wasn't much interest in `clang-40` since it's introduction.

alarcher:
@Mno-hime @pyhalov I will have a look at it. I have the recipe for clang 6 and 8 as well but need more testing. I have used clang-40 for my research project until last year so I am surprised about this breakage.

pyhalov:
Superseded by #5181  and #5184

PR #5178, opened 2019-07-31
desktop-cache: switch to Python 3.5

pyhalov:


pyhalov:
Seems to be fine, but it would be good if someone else also tested/reviewed it.

PR #5183, opened 2019-08-02
Move testdisk into sysutils category

AndWac:
This package seems to be a leftover. Alas I haven't been able to update to the newest 7.1 release easily so I left this out for a future step. It seems as if qphotorec needs newer qt5.


components/sysutils/testdisk/Makefile
```diff
@@ -21,16 +21,15 @@
 # Copyright (c) 2016 by Jim Klimov. All rights reserved.
 # Copyright (c) 2019 Tim Mooney <Timothy.V.Mooney@gmail.com>
 #
-include ../../make-rules/shared-macros.mk
+include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		testdisk
 COMPONENT_VERSION=	7.0
-COMPONENT_REVISION=	1
```
Mno-hime, line 28: Shouldn't it be bumped to `2` instead or removed?

AndWac: You are right. I found this problem by myself but you have been faster :)
I removed it when trying to update to 7.1.
Regards

pyhalov (approved):

Mno-hime (approved):

pyhalov:
What is left? docbook && fomatic?

AndWac:
I wasn't sure about those because they themselves contain packages.
What about inputmethod?

pyhalov:
This category potentially contains a lot of components - look at `pkg info -r '*input*method*'`. Nobody has cared enough to update them :)

pyhalov:
Look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/categories-layout.md .
I would add tcl there, as it is the same as python and perl, but I personally would also dismiss scientific. However @alarcher asked to preserve it :)

AndWac:
And I would like to have lang like FreeBSD and others have. The runtime category doesn't fit for all.

pyhalov:
But foomatic clearly goes to print, docbook - even don't know, text or developer

AndWac:
I'd say text because we have already things there mainly interesting for developers.

PR #5192, opened 2019-08-04
Py35

pyhalov:


pyhalov:
@normjacobs Hi. Can you look it through? 
I suppose, we'll move to Python3-only pkg soon, this is required oi-userland part.

pyhalov:
@alarcher Do you have some system to test complete oi-userland rebuild? (with this + https://github.com/OpenIndiana/pkg5/pull/65 installed on the system)

alarcher:
@pyhalov I have installed the new pkg version and I am running now a full rebuild with -j 4 and with build args -j 10. It will take some time.

PR #5193, opened 2019-08-04
illumos-gate: tools should use Python 3 modules

pyhalov:


pyhalov:
One more PR to merge after https://github.com/OpenIndiana/pkg5/pull/65

PR #5194, opened 2019-08-05
setuptools_scm: avoid misdetecting git repository

pyhalov:


pyhalov:
@kulikjak Hi. Have you seen similar issues during solaris-userland gate build?

kulikjak:
> @kulikjak Hi. Have you seen similar issues during solaris-userland gate build?

Hi. We haven't seen similar problems, the reason might be that we are internally using mercurial rather than git. But I am not sure how should this potential problem manifest itself. We are using it just to install other packages (such as portend, tempora and several others) and there is no archive creation happening right?

pyhalov:
@kulikjak The issue which I see here is the following (not sure about hg, but it's definitely an issue with git).
When any python component (it seems even not using setuptools_scm) is installed, for example, python/coverage,  distutils runs file_finder_git from setuptools_scm, then on install stage file_finder_git tries to find files in current archive by processing git archive... of oi-userland.

kulikjak:
@pyhalov Oh, I see.

I tried installing python package on our gate (with some setuptools_scm debug prints), and the installation itself seems to work correctly, I can also see that distutils is listing our whole gate (both git and mercurial) which it shouldn't.

Mno-hime (approved):

PR #5195, opened 2019-08-05
slim_source: set LC_ALL so that pkg could process utf-8 manifests

pyhalov:
Python 3 tries to find out file encoding based on current layout.
We have manifests in /var/pkg, which contain utf-8 data.
To avoid pkgdepend breakage, we set correct encoding in component's Makefile.


pyhalov:
@alarcher This PR and https://github.com/OpenIndiana/slim_source/commit/e300b98de5fbb88511537a1668f2f1b808cb689d fixes slim_source publish.

PR #5197, opened 2019-08-06
php-7_3: Add pdo_pgsql to extension/php-pgsql, update to 7.3.8, fix REQUIRED_PACKAGES

bergerspencer:
I recently noticed that the php-pgsql package was missing pdo_pgsql. Considering the php-mysql packages contains pdo_mysql I think it's a good thing to include. tt-rss with postgres needs this, for example.

It seemed silly to bump the COMPONENT_REVISION just for this, so I also went ahead and updated to 7.3.8.

There were some build errors relating to missing icu-config and libpq-fe.h, so I added the appropriate build dependencies as well. These were not new build errors but present in 7.3.7 as well.


components/web/php/php-7_3/Makefile
```diff
@@ -113,6 +113,7 @@ CONFIGURE_OPTIONS   +=	--with-imap-ssl
 CONFIGURE_OPTIONS  +=	--with-mysqli=shared,$(MYSQL_BINDIR)/mysql_config
 CONFIGURE_OPTIONS  +=	--with-pdo-mysql=shared,$(MYSQL_BINDIR)/mysql_config
 CONFIGURE_OPTIONS  +=	--with-pgsql=shared,$(PG_BINDIR)
+CONFIGURE_OPTIONS  +=	--with-pdo-pgsql=shared
```
pyhalov: pgsql pdo module is compiled using arbitrary postgresql version installed on the system, not Postgresql 9.4

pyhalov (changes requested):

components/web/php/php-7_3/Makefile
bergerspencer: Whoops, fixed.
I'm not sure if it's necessary or if php's configure script can use the PG_CONFIG in the CONFIGURE_ENV. It definitely found and compiled against the correct version on my machine, but better safe than sorry.

PR #5203, opened 2019-08-07
Update cryptography to 2.7

Mno-hime:
Changelog: https://cryptography.io/en/latest/changelog/

**Testing**
- `pkg` works
- `ansible`, [`salt`](https://www.illumos.org/issues/11526) and `pkg` rebuild


pyhalov:
what about pkg test suite?

PR #5204, opened 2019-08-07
Generate more useful sample manifests:

richlowe:
This fixes a handful of bugs and annoyances in sample-manifest output, and at least one real bug.
I don't have the build resources to fully test it, though.  I'm hoping one of you does (@alarcher? @pyhalov?)

- Generate rules for GNU autoconf triplets that actually work
  Previously we'd include SOLARIS_RELEASE and PLAT, which wouldn't expand.

  Instead output $(GNU_TRIPLET), and arrange for its expansion.

  We perhaps would want a 64bit version of this (x86_64-pc-solaris2.11 for
  eg), but the places that seem like they'd use it currently (firefox, rust),
  skip the version number and I'm not really sure what's going on there.

- The IPS_VERSION production never worked, because IPS_VERSION doesn't exist
  replace it with IPS_COMPONENT_VERSION, which does.

- for IPS_COMPONENT_VERSION and COMPONENT_VERSION match against a copy that
  has been regex escaped, so the replacement isn't over-broad.

- use GNU_TRIPLET in place of GNU_ARCH and similar throughout.

- as part of introducing $(GNU_TRIPLET) fix binutils, which in part used the
  old $(GNU_ARCH) but never defined it, and thus delivered
  /usr/gnu/'$(GNU_ARCH)' and contents.


richlowe:
There's also

```
set name=pkg.summary value="Pythonic $(PYVER) binding for the libxml2 and libxslt libraries"
set name=com.oracle.info.description value="Python $(PYVER) bindings for the libxml2 and libxslt libraries"
set name=pkg.summary value="Zope interfaces package for Python $(PYVER)"
set name=com.oracle.info.description value="the Zope interfaces package for Python $(PYVER)"
set name=com.oracle.info.description value="Python $(MAYBE_PYVER_SPACE)setup tools"
set name=com.oracle.info.description value="Python $(MAYBE_PYVER_SPACE)setup tools"
set name=com.oracle.info.description value="Python $(MAYBE_PYVER_SPACE)setup tools"
```

Which need the python version stuff merged from Oracle more completely, but that's beyond me at the moment.

alarcher:
@richlowe I have rebuilt gcc-6, gcc-7, gcc-8 so far, no issue.

alarcher:
@pyhalov @richlowe Rebuilt firefox and rust without any issue.

alarcher (approved):

PR #5209, opened 2019-08-08
Update bind to 9.14.4, fix license, CVE-2019-6471

Mno-hime:
Changelog: https://downloads.isc.org/isc/bind9/9.14.4/RELEASE-NOTES-bind-9.14.4.html

ABI is stable: https://abi-laboratory.pro/index.php?view=timeline&l=bind

License changed in 9.14 from ISC to MPL v2.0.

**Testing**
- `dig` and `host` work


components/network/bind/patches/04-remove-zrelax-transtls.patch
```diff
@@ -0,0 +1,37 @@
+Revert this Solaris-specific patch.
```
pyhalov, line 1: Do I understand correctly that this reverts some gcc-related   tambourine dances which are not needed on illumos?

Mno-hime: Yes, I think so. Our `ld` does not know about `-zrelax=transtls` (and thus openssl link test fails) and our `ld` does not suffer from the problem described in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90912.

I compiled and run the test case from there with our GCC 6..9 and all worked fine, see the output: [zrelax_transtls.test_case.txt](https://github.com/OpenIndiana/oi-userland/files/3480991/zrelax_transtls.test_case.txt).

pyhalov (approved):

PR #5211, opened 2019-08-08
Update ISC DHCP to 4.4.1

Mno-hime:
Sync with solaris-userland component, but revised content of `files/` so the changes are sane.

Also added Solaris patches. I find them same, but someone should have a look as well.

**Testing**

I was able to start the `svc:/network/dhcp/server:ipv4` service in Global Zone after I added `/etc/inet/dhcpd4.conf` with definition of my local subnet:
```
{global} newman@lenovo:~ $ svcs -vp svc:/network/dhcp/server:ipv4
STATE          NSTATE        STIME    CTID   FMRI
online         -             12:29:14    340 svc:/network/dhcp/server:ipv4
               12:29:14     4467 dhcpd
```


components/network/isc-dhcp/files/isc-dhcp-relay.xml
```diff
@@ -69,7 +69,7 @@
 	<exec_method type='method' name='stop' exec=':kill' 
 		timeout_seconds='60'/>
 
-	<exec_method type='method' name='refresh' exec=':kill -HUP' 
+	<exec_method type='method' name='refresh' exec=':true'
```
pyhalov, line 72: Why do we remove refresh method?

components/network/isc-dhcp/files/isc-dhcp-server.xml
```diff
@@ -70,7 +70,7 @@
 	<exec_method type='method' name='stop' exec=':kill' 
 		timeout_seconds='60'/>
 
-	<exec_method type='method' name='refresh' exec=':kill -HUP' 
+	<exec_method type='method' name='refresh' exec=':true'
```
pyhalov, line 73: The same here

components/network/isc-dhcp/patches/002-dhcp_broadcast_offer_bug.patch
```diff
@@ -0,0 +1,179 @@
+This patch file was developed in-house to fix a Solaris specific bug. We are
+working with upstream, but it is unclear at this point whether or not they are
+going to accept it. This is a highly visible bug which many customers have
+encountered.
+
+diff -r 1680283c6725 common/socket.c
```
pyhalov, line 6: The patch makes sense. From what I see, this patch inserts arp entry in OS table before sending unicast BOOTP/DHCP response. Of course, I'd preferred someone else to look at it :)

components/network/isc-dhcp/patches/002-dhcp_broadcast_offer_bug.patch
```diff
@@ -0,0 +1,179 @@
...
+ void if_reinitialize_send (struct interface_info *);
+ void if_register_send (struct interface_info *);
+diff -r 1680283c6725 relay/dhcrelay.c
+--- a/relay/dhcrelay.c	Tue Jan 16 13:35:26 2018 -0800
++++ b/relay/dhcrelay.c	Tue Jan 16 17:55:37 2018 -0800
+@@ -759,6 +759,13 @@
+ 			to.sin_addr = packet->yiaddr;
+ 			to.sin_port = remote_port;
+ 
++#if defined (sun)
```
pyhalov, line 116: @danmcd 
Do I understand this correctly? Before sending response to client we insert (mac, ip) entry in arp table, so that we could do unicast reply...
Full code looks like this: https://source.isc.org/cgi-bin/gitweb.cgi?p=dhcp.git;a=blob;f=relay/dhcrelay.c;h=d8caaaf6d70d013f587416a862fb5ce06fe30789;hb=HEAD#l867

components/network/isc-dhcp/files/isc-dhcp-relay.xml
Mno-hime: Although upstream says that `dhcpd` is [unable to reload itself](https://kb.isc.org/docs/aa-00335), I guess what they mean is that `dhcpd` can't catch the signal and handle it on it's own. But Oracle [got bitten by it](https://github.com/oracle/solaris-userland/commit/22d8d271599a6500d912ac98c1060aa771be874a), it seems.

Trying it manually seems to work fine:
```
{global} newman@lenovo:~ $ svcs -vp svc:/network/dhcp/server:ipv4
STATE          NSTATE        STIME    CTID   FMRI
online         -             14:31:31    351 svc:/network/dhcp/server:ipv4
               14:31:31     4807 dhcpd
{global} newman@lenovo:~ $ pfexec truss -p $(pgrep dhcpd)
pollsys(0xFFFFFD7FFFDFF7E0, 2, 0x00000000, 0x00000000) (sleeping...)
    Received signal #1, SIGHUP, in pollsys() [default]
      siginfo: SIGHUP pid=4812 uid=0
pollsys(0xFFFFFD7FFFDFF7E0, 2, 0x00000000, 0x00000000) Err#4 EINTR
{global} newman@lenovo:~ $ svcs -vp svc:/network/dhcp/server:ipv4
STATE          NSTATE        STIME    CTID   FMRI
online         -             14:31:54    352 svc:/network/dhcp/server:ipv4
               14:31:54     4821 dhcpd
```
SMF handled PID transition correctly.

Maybe if refresh timeout is too short to wait for a bigger setup to load, then it won't do the handover?

pyhalov: OK, let's fetch this change.

components/network/isc-dhcp/patches/002-dhcp_broadcast_offer_bug.patch
danmcd: Seems like an optimization it's making.  Given dhcpd has net_config privs anyway, this should work.

pyhalov (approved):

pyhalov:
@citrus-it Note the patches, perhaps they should be added to omnios-build

PR #5214, opened 2019-08-09
Update xorgproto

alarcher:
@pyhalov If you have time to help me figure out how to do this correctly...


components/x11/x11-protocols/x11-protocols.p5m
```diff
@@ -0,0 +1,327 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.description value="Developer documentation and C language definitions of X11, X extension, and related protocols"
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+# Provided by xorgproto
+#depend fmri=x11/header/bigreqsproto type=require
```
pyhalov: As I understand, a dozen *proto* was merged into x11-protocols. 
To do this update correctly you need to add to history file (in x11-protocol component or meta-packages/history, doesn't matter much), a pair old_FMRI@(old_version+1) NEW_FMRI.
This will lead to generation of dozen rename packages.

To be pedantic you could make new package depend on this old_FMRI@(old_version+1) package, so that installation of new package lead to deinstallation of old packages.

alarcher: OK so my mistake is that I did not bump the version in the history file...

pyhalov: Your mistake is package names - x11/header/xf86bigfontproto, not xf86bigfontproto. 
Also you've sorted history file, this makes reviewing harder.

components/x11/x11-protocols/history
```diff
@@ -0,0 +1,29 @@
+bigreqsproto@1.1.2-2018.0.0.0 x11/header/x11-protocols
```
pyhalov: Don't change it in two places - either use meta-packages/history component or history file in x11-protocols.
Also, use full fmris.

alarcher:
@pyhalov Should be fine now, I tested that it updates to the new version of x11-protocols while deinstalling the old proto packages.

components/x11/x11-protocols/x11-protocols.p5m
```diff
@@ -0,0 +1,327 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.description value="Developer documentation and C language definitions of X11, X extension, and related protocols"
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+# Provided by xorgproto
+depend fmri=x11/header/bigreqsproto@1.1.2-2018.0.0.1 type=require
```
pyhalov: type=optional, 
```
optional
The optional dependency is similar to the  require
 dependency, but the specified package need not be installed.
However, if it is present, it must be at the specified version or greater.
This type of dependency is typically used to handle cases where packages transfer content.
In this case, each version of the package post-transfer would contain an optional dependency on the other package's
post-transfer version, so it would be impossible to install incompatible versions of two packages...
```

alarcher: Thanks for the explanation :) Done.

pyhalov:
Does it affect other xorg components?

alarcher:
I have checked that xorg builds.

pyhalov (approved):

PR #5217, opened 2019-08-10
Add xorg core fonts component

alarcher:
+ minor fixes to make-rules


alarcher:
Need to check update.

Mno-hime:
Does it add weight (as in megabytes) to the default MATE installation? If it updates: LGTM.

alarcher:
It is just a set of packages from the X11 gate that we never migrated to oi-userland.
There is only one component left and we'll have the full gate in oi-userland (finally).

make-rules/shared-macros.mk
```diff
@@ -378,18 +378,23 @@ MACH32 =	$(MACH32_1:i386=i86)
...
 
+BUILD_NO_ARCH =		$(BUILD_DIR_NO_ARCH)/.built
 BUILD_32 =		$(BUILD_DIR_32)/.built
 BUILD_64 =		$(BUILD_DIR_64)/.built
 BUILD_32_and_64 =	$(BUILD_32) $(BUILD_64)
+$(BUILD_DIR_NO_ARCH)/.built:	BITS=32
 $(BUILD_DIR_32)/.built:		BITS=32
 $(BUILD_DIR_64)/.built:		BITS=64
 
+INSTALL_NO_ARCH =	$(BUILD_DIR_NO_ARCH)/.installed
```
pyhalov, line 397: Please, check that these *_NO_ARCH variables don't influence setup.py NOARCH components.

alarcher: I have sucessfully rebuilt all the components in python/

PR #5218, opened 2019-08-10
audacity: update to 2.3.2

pyhalov:
New audacity version depends on portmidi library (which was never ported to illumos/Solaris). Disabling it without disabling midi processing needs significant patches. 
Have several ways here:
- disable portmidi (current approach), needs more work to avoid crashing now and then;
- port portmidi to illumos (likely is doable, but I know nothing about midi processing);
- port libalsa to illumos (at least this is what FreeBSD does, I expect that porting portmidi is much easier)
- disable midi processing completely (if patch will not be very intrusive).


components/encumbered/audacity/patches/16-fix-unknown-method.patch
```diff
@@ -0,0 +1,19 @@
+Rude fix for the following compilation error
```
pyhalov: All over the code NoteTrack is often treated as PlayableTrack, however earlier it could be based either on PlayableTrack or AudioTrack depending on EXPERIMENTAL_MIDI_OUT.

pyhalov:
With portmidi audacity is much more stable, but  midi playback is broken.

pyhalov:
Can look at something else - patch included portmidi to be complete no-op.

pyhalov:
Has somebody tested that it really works? ;) I mean that it can record and play audio.

AndWac:
Am 12/24/21 um 16:58 schrieb Alexander Pyhalov:
>
> Has somebody tested that it really works? ;) I mean that it can record
> and play audio.
>
I have tested it on my system. Albeit I didn't test the recording
capabilities due to my system lacking appropriate equipment. But this PR
was stuck for so long and nobody else seemed to care that I decided to
merge it and let users test it. Alas we neither have testers nor the
infrastructure to do real testing.

pyhalov:
It was stuck for so long, because fresh audacity required alsa to record audio (or some patches). I tried to experiment, but IIRC, it didn't work. So, likely, it's broken now.

PR #5219, opened 2019-08-11
Add portmidi 217

pyhalov:
Portmidi port based on initial OpenBSD one https://github.com/openbsd/ports/commit/c030f0cdc9700530f096171af69497336a2582e8#diff-762884b617e909a974332ff412b73757 
So far, only midi output is supported. Should check if it's enough for audacity.


pyhalov:
With this portmidi library version audacity plays garbage...

pyhalov:
This one is completely wrong.BSD port uses sndio interface to https://man.openbsd.org/midi.4 driver. illumos lacks midi support and needs software synthesizer, so library must be more complex.

PR #5221, opened 2019-08-13
xorg-cyrillic: add using modified solaris-userland recipe

alarcher:


components/meta-packages/X-incorporation/Makefile
```diff
@@ -36,38 +36,8 @@ clean:
 	$(RM) -r $(BUILD_DIR) $(PROTO_DIR)
 
 # Auto-generated dependencies
-REQUIRED_PACKAGES += compatibility/packages/SUNWxwinc
```
pyhalov, line 39: Is it really auto-generated dependencies? I thought  there would be no REQUIRED_PACKAGES for this component.

pyhalov: Yes, you can safely remove them all.
Otherwise looks good.

alarcher: OK I will vacuum with the next PR.

PR #5222, opened 2019-08-13
The walking dead 4

alarcher:
@alanc Who would have thought?


alanc:
> @alanc Who would have thought?

I wouldn't have thought anyone was still using binaries old enough to need this ancient version still.

alarcher:
@alanc I am pretty sure you cannot wait for the sequel... the fifth episode is coming soon...

PR #5224, opened 2019-08-15
Building illumos-gate with UTF-8 locale breaks libcurses

pyhalov:


pyhalov:
This reverts 9d9d4036a13615cc8daccf29ed23f1800a77937c , but luckily we don't need this workaround after https://github.com/OpenIndiana/pkg5/commit/d3187bbf6614769114482c8823a6b1adae05ea3a

PR #5233, opened 2019-08-18
top: use termios instead of termio

tsoome:
termio is using 8 control chars and we end up reading past array end.
Since termios is preferred interface, use it instead.


alarcher:
@tsoome Bumpy bump COMPONENT_REVISION?

tsoome:
> @tsoome Bumpy bump COMPONENT_REVISION?

I thought I did add it but apparently I was too sleepy:)

alarcher (approved):

PR #5235, opened 2019-08-19
Add splix (SPL printing language for Samsung et al. printers)

Mno-hime:
**Testing**

Enabled `svc:/application/cups/scheduler:default`, added Samsung SCX-3205 mono-color later printer via `system-config-printer`, printed test page, which confirms that the `scx3200.ppd` driver is used, printed a PDF page from Atril.


components/print/splix/Makefile
```diff
@@ -0,0 +1,64 @@
...
+SVN_REPO=		https://svn.code.sf.net/p/splix/code/
+SVN_REV=		315
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+PATH=$(PATH.gnu)
+
+export CXXFLAGS			+= -fno-strict-aliasing
```
alarcher: Why not pass the flags the usual way in *_ENV variables?

alarcher (changes requested):

components/print/splix/Makefile
Mno-hime: @alarcher Should be resolved now.

alarcher (approved):

PR #5239, opened 2019-08-19
pacemaker: bump to 1.1.21

grueni:


grueni:
This PR needs some investigation.
Pacemaker could not communicate with its own services.
Wait with the other PRs too.

grueni:
Problem solved with 004-common-ipc.patch.

Mno-hime (approved):

PR #5240, opened 2019-08-19
pcs: bump to 0.9.168

grueni:


components/cluster/pcs/Makefile
```diff
@@ -9,19 +9,19 @@
 #
 
 #
-# Copyright 2016 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved. 
+# Copyright 201999999999 Grueninger, Grueninger GmbH, (grueni). All rights reserved. 
 #
```
AndWac, line 13: Another sed victim?

grueni: I fear my keyboard makes too much from time to time

Mno-hime (approved):

PR #5246, opened 2019-08-20
Pacemaker 2.0.2

grueni:
Both versions of pacemaker can be used.
With version 2.* a rolling upgrade of the cluster from 1.* may not be possible.
Nevertheless I suggest to prefer version 2.0.2.


components/cluster/pacemaker/Makefile
```diff
@@ -53,18 +52,18 @@ CONFIGURE_OPTIONS += --sysconfdir=/etc
...
 CONFIGURE_OPTIONS += --enable-systemd=no
 CONFIGURE_OPTIONS += --enable-fatal-warnings=no
 CONFIGURE_OPTIONS += --disable-rpath
+CONFIGURE_OPTIONS += --with-logdir=/var/log/hacluster
+CONFIGURE_OPTIONS += --with-bundledir=/var/log/hacluster/bundles
+CONFIGURE_OPTIONS += --with-openssl=yes
 CONFIGURE_OPTIONS += $(CONFIGURE_OPTIONS.$(BITS))
 
 CONFIGURE_ENV.64 += LT_SYS_LIBRARY_PATH="/lib/$(MACH64):/usr/lib/$(MACH64)"
+CONFIGURE_ENV.64 += PYTHON=/usr/bin/amd64/python2.7
```
Mno-hime: Will it work with Python 3.5? If so then set `PYTHON_VERSION=3.5` and then:
```suggestion
CONFIGURE_ENV.64 += PYTHON=$(PYTHON)
```

Mno-hime:
> Both versions of pacemaker can be used.
> With version 2.* a rolling upgrade of the cluster from 1.* may not be possible.
> Nevertheless I suggest to prefer version 2.0.2.

Breaking changes are fine if they are [documented](http://docs.openindiana.org/release-notes/latest-changes/) ([source](https://github.com/OpenIndiana/oi-docs/blob/master/docs/release-notes/latest-changes.md)) with information on how to migrate from v1 to v2, or information that it's not possible and requires manual intervention.

So, I think I leave it with you on whether to proceed with updated v1 or move to v2 :).

components/cluster/pacemaker/Makefile
grueni: Following https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/1.1/html/Pacemaker_Development/_python_compatibility.html
it should work with Python 3.5.
I will change this PR and pcs and crmsh (if possible).

grueni:
How to solve this conflicts?
I tried to checkout the PR locally but I get only the last version without conflict markers.
Sorry for the inconvenience.

Mno-hime:
`git fetch` the upstream remote and then rebase to oi/hipster branch while on this branch. Then `git push -f`. Makes sense?

Mno-hime:
```
$ git remote -v
origin	git@github.com:Mno-hime/oi-userland.git (fetch)
origin	git@github.com:Mno-hime/oi-userland.git (push)
upstream	https://github.com/OpenIndiana/oi-userland.git (fetch)
upstream	https://github.com/OpenIndiana/oi-userland.git (push)
```

`git fetch upstream && git rebase -v upstream/oi/hipster`

And then fix conflicts, `git add` changed files and `git push -f`.

grueni:
Thanks a lot.

components/cluster/pacemaker/Makefile
```diff
@@ -32,13 +32,18 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/configure.mk
 include $(WS_MAKE_RULES)/ips.mk
 
+PREFERRED_BITS = 64
+PYTHON_VERSION = 3.5
+PYTHON_VENDOR_PACKAGES=$(PYTHON.$(PYTHON_VERSION).VENDOR_PACKAGES.64)
+PYTHON =  $(PYTHON.$(PYTHON_VERSION).$(BITS))
```
Mno-hime: I think `PREFERRED_BITS` are required only when `32_and_64` bits are being build and one wants 64-bit binaries in `/usr/bin/` instead of `/usr/bin/64/` (and even then the variable should be before `include ../../../make-rules/shared-macros.mk` because it alters things there).

Isn't `PYTHON_VERSION=3.5` enough? Latter `PYTHON` lines look verbatim from `make-rules/shared-macros.mk`.

grueni: Indeed. Just use PYTHON_VERSION = 3.5 is enough.

pyhalov:
@grueni @Mno-hime Is this pr ready? 
Do we need, for example, updated cmrsh to cope with this update?

grueni:
This version of pacemaker is compatible with all other components used for hacluster.
The changes described [in](https://wiki.clusterlabs.org/wiki/Pacemaker_2.0_Configuration_Changes) concern mainly outdated older versions never used in hacluster.

pyhalov:
@grueni Hi. I've rebased your PR against current oi/hipster and updated Pacemaker to the latest release. Does it work for you?

pyhalov:
Works for me.

grueni:
I can check it tomorrow. But it should work. Go for it.

pyhalov:
Why have you closed it?

grueni:
Reopened.
I thought version 2.0.3 is already in the repo.

AndWac:
I there still interest in this? Otherwise I will close it soon.

PR #5250, opened 2019-08-21
XScreenSaver: Import security fix from solaris-userland

Mno-hime:
@alanc hinted following patch on #oi-dev today: https://github.com/oracle/solaris-userland/commit/cae02421cfad00f57fc11f99ec0be121828fe2ec (also https://blogs.oracle.com/solaris/announcing-oracle-solaris-114-sru12).

It's unclear to me what the reason is, I assume security.

**Testing**
- Restarted the daemon on the fly, locks & unlocks fine


alanc:
The reason is: https://techblog.mediaservice.net/2019/10/local-privilege-escalation-on-solaris-11-x-via-xscreensaver/

PR #5257, opened 2019-08-22
chezscheme: Add 9.5.2

richlowe:
This adds Chez Scheme 9.5.2.

The patches are from a pull request I currently have open upstream (cisco/ChezScheme#455)
The files/ directory are the files used to bootstrap on solaris x86 (I've included all of them, though we only use ta6s2, threaded solaris on amd64).  

The files will need to be generated for each future update, on a platform for which Chez ships bootstrap files (how to do so is documented in their build documentation).  The files are somewhat large, and ill-suited to version control, so it'd probably be better to put them in a separate tarball we could host and fetch, but I don't have the means to do that in the pull request.

### DIFF
--- a/components/runtime/chezscheme/Makefile
+++ b/components/runtime/chezscheme/Makefile
@@ -1,0 +1,85 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2019 Richard Lowe
+#
+
+BUILD_BITS= 64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		chezscheme
+COMPONENT_VERSION=	9.5.2
+COMPONENT_SUMMARY=	Chez Scheme
+COMPONENT_PROJECT_URL=	https://github.com/cisco/ChezScheme
+COMPONENT_FMRI=		runtime/chezscheme
+COMPONENT_CLASSIFICATION=Development/Other Languages
+COMPONENT_SRC=		ChezScheme-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+  sha256:3a370fdf2ffd67d6a0ccbb993dfab1cbaf4a0a97983c869cfaab40528c33c48b
+COMPONENT_ARCHIVE_URL= \
+  https://github.com/cisco/ChezScheme/archive/v$(COMPONENT_VERSION).tar.gz
+
+COMPONENT_LICENSE_FILE=	LICENSE
+COMPONENT_LICENSE=	Apache-2.0
+
+# SunOS-ish bootstrap files not included in the release tarballs These must be
+# generated according to the instructions provided by ChezScheme on a system
+# which _is_ supported for bootstrap, and then provided by us
+COMPONENT_NAME_1= chezscheme-bootstrap
+COMPONENT_VERSION_1= $(COMPONENT_VERSION)
+COMPONENT_SRC_1= $(COMPONENT_NAME_1)-$(COMPONENT_VERSION_1)
+COMPONENT_ARCHIVE_1= $(COMPONENT_SRC_1).tar.gz
+COMPONENT_ARCHIVE_URL_1= https://github.com/OpenIndiana/$(COMPONENT_NAME_1)/archive/$(COMPONENT_VERSION_1).tar.gz
+COMPONENT_ARCHIVE_HASH_1= \
+  sha256:1f00d017e5ebf8eb0ee2c6ae6b101917c0e10bf5bcbcebe2827882eb1025015e
+COMPONENT_SRC_BOOTSTRAP= $(COMPONENT_SRC_1)
+
+# Nanopass
+COMPONENT_NAME_2= nanopass-framework-scheme
+COMPONENT_VERSION_2= 1.9
+COMPONENT_SRC_2= $(COMPONENT_NAME_2)-$(COMPONENT_VERSION_2)
+COMPONENT_ARCHIVE_2= $(COMPONENT_SRC_2).tar.gz
+COMPONENT_ARCHIVE_URL_2= https://github.com/nanopass/$(COMPONENT_NAME_2)/archive/v$(COMPONENT_VERSION_2).tar.gz
+COMPONENT_ARCHIVE_HASH_2=	\
+  sha256:625b239f9030d0b1e86b1fffd8b69f7249a63e8b8ca85195a00cf22889f7fc86
+COMPONENT_SRC_NANOPASS= $(COMPONENT_SRC_2)
+
+# stex
+COMPONENT_NAME_3= stex
+COMPONENT_VERSION_3= 1.2.1
+COMPONENT_SRC_3= $(COMPONENT_NAME_3)-$(COMPONENT_VERSION_3)
+COMPONENT_ARCHIVE_3= $(COMPONENT_SRC_3).tar.gz
+COMPONENT_ARCHIVE_URL_3= https://github.com/dybvig/stex/archive/v$(COMPONENT_VERSION_3).tar.gz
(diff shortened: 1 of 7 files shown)


components/runtime/chezscheme/Makefile
```diff
@@ -0,0 +1,87 @@
...
+COMPONENT_NAME_1= nanopass-framework-scheme
+COMPONENT_VERSION_1= 1.9
+COMPONENT_SRC_1= $(COMPONENT_NAME_1)-$(COMPONENT_VERSION_1)
+COMPONENT_ARCHIVE_1= $(COMPONENT_SRC_1).tar.gz
+COMPONENT_ARCHIVE_URL_1= https://github.com/nanopass/$(COMPONENT_NAME_1)/archive/v$(COMPONENT_VERSION_1).tar.gz
+COMPONENT_ARCHIVE_HASH_1=	\
+  sha256:625b239f9030d0b1e86b1fffd8b69f7249a63e8b8ca85195a00cf22889f7fc86
+
+# zlib (configure would get this from github.com/madler/zlib)
+# There seems to be no provision to use a system zlib
```
Mno-hime: Someone [here](https://build.opensuse.org/package/view_file/home:lorenz:idris/ChezScheme/use_native_zlib.patch?expand=1) may got it to work (or to build at least).

components/runtime/chezscheme/Makefile
```diff
@@ -0,0 +1,87 @@
...
+COMPONENT_ARCHIVE_HASH_3=	\
+  sha256:bf784ca46aaca9b665b7eb0c39f04f6a695aa40e99b11d8a6d4440648c1bf40e
+
+COMPONENT_LICENSE=	Apache
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+PATH=$(PATH.gnu)
```
Mno-hime, line 70: Put it after `COMPONENT_ARCHIVE_URL` so it's clear it belongs to the main component not to bundles.

components/runtime/chezscheme/Makefile
```diff
@@ -0,0 +1,87 @@
...
+COMPONENT_SRC_3= $(COMPONENT_NAME_3)-$(COMPONENT_VERSION_3)
+COMPONENT_ARCHIVE_3= $(COMPONENT_SRC_3).tar.gz
+COMPONENT_ARCHIVE_URL_3= https://github.com/dybvig/stex/archive/v$(COMPONENT_VERSION_3).tar.gz
+COMPONENT_ARCHIVE_HASH_3=	\
+  sha256:bf784ca46aaca9b665b7eb0c39f04f6a695aa40e99b11d8a6d4440648c1bf40e
+
+COMPONENT_LICENSE=	Apache
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
```
Mno-hime: For new components use `common.mk`.

components/runtime/chezscheme/chezscheme.license
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
```
Mno-hime: If there's a file in the source code we can point to, that's preferable do `COMPONENT_LICENSE_FILE=...` to copying the text out of sources.

components/runtime/chezscheme/Makefile
```diff
@@ -0,0 +1,72 @@
...
+COMPONENT_CLASSIFICATION=Development/Other Languages
+COMPONENT_SRC=		ChezScheme-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+  sha256:3a370fdf2ffd67d6a0ccbb993dfab1cbaf4a0a97983c869cfaab40528c33c48b
+COMPONENT_ARCHIVE_URL= \
+  https://github.com/cisco/ChezScheme/archive/v$(COMPONENT_VERSION).tar.gz
+
+COMPONENT_LICENSE_FILE=	$(COMPONENT_SRC)/LICENSE
+COMPONENT_LICENSE=	Apache
```
Mno-hime: Apache-2.0

components/runtime/chezscheme/Makefile
```diff
@@ -0,0 +1,72 @@
...
+COMPONENT_FMRI=		runtime/chezscheme
+COMPONENT_CLASSIFICATION=Development/Other Languages
+COMPONENT_SRC=		ChezScheme-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+  sha256:3a370fdf2ffd67d6a0ccbb993dfab1cbaf4a0a97983c869cfaab40528c33c48b
+COMPONENT_ARCHIVE_URL= \
+  https://github.com/cisco/ChezScheme/archive/v$(COMPONENT_VERSION).tar.gz
+
+COMPONENT_LICENSE_FILE=	$(COMPONENT_SRC)/LICENSE
```
Mno-hime: It's not found if it's just `COMPONENT_LICENSE_FILE=LICENSE`? The build system looks for the file in few directories, it should be detected in `COMPONENT_SRC`.

Mno-hime (commented):
Since we have quite some files in `files/`, please copy appropriate license file (`$(COMPONENT_SRC)/LICENSE`?) to the `files/` directory.

richlowe:
The tree contains a mix of "Apache-2.0" and "Apache v2.0", and at least one  "Apache 2.0 License", and "Apache 2.0", and some "Apache" I think are 2.0.   I used the version you asked for, but you might want to clean up the rest.

richlowe:
Oh, and as mentioned above I don't _want_ to have files in files/, I want someone with permissions to make them a separate tarball for the build, as @alarcher tells me is sometimes done, since they're big, unkind to version control, and would get a whole new version each time.

Mno-hime:
From the POV of the build system it's better if we are able to get sources (e.g. a tarball) from a network source, then side-loading tarball to the build server. Ideally we create a GitHub repo like "chezscheme-bootstrap-openindiana" in OpenIndiana GitHub project from which we can get tagged/released tarballs. As a bonus it's reproducible outside the OpenIndiana project. I understand VC is still involved but git should do for this well.

richlowe:
That's fine too (I can't arrange for that either, though...).

The reason I bring up git being an issue, is that the bootstrap files don't delta at all (the largest are gzip compressed), so they can easily balloon the size of the repo (they're why the ChezScheme upstream is a couple of a gigabytes).  Doing this to a separate repo is certainly better than doing it to oi-userland (though obviously it's likely to get far fewer updates even over the long term)

Mno-hime:
> That's fine too (I can't arrange for that either, though...).

Nor can I but I am sure that if you agree to this solution @pyhalov or @alarcher can create one for you.

> The reason I bring up git being an issue, is that the bootstrap files don't delta at all (the largest are gzip compressed), so they can easily balloon the size of the repo (they're why the ChezScheme upstream is a couple of a gigabytes). Doing this to a separate repo is certainly better than doing it to oi-userland (though obviously it's likely to get far fewer updates even over the long term)

Understood. I push for a hosted solution because unfortunately our infrastructure has problems from time to time and I observed that infra is the last thing any of us wants to spend time with. Perhaps the repo can be made "[binary-only](https://stackoverflow.com/a/11162362)", if that helps in any way.

components/runtime/chezscheme/Makefile
```diff
@@ -0,0 +1,72 @@
...
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2019 Richard Lowe
+#
+
+BUILD_BITS= 64
+BUILD_STYLE= configure
```
Mno-hime: `configure` is the default `BUILD_STYLE`, no need to set it.

Mno-hime:
> The tree contains a mix of "Apache-2.0" and "Apache v2.0", and at least one "Apache 2.0 License", and "Apache 2.0", and some "Apache" I think are 2.0. I used the version you asked for, but you might want to clean up the rest.

Did you push it? I still see "Apache" without the version.

alarcher:
It this is sufficient: https://github.com/OpenIndiana/chezscheme-bootstrap/releases/tag/9.5.2

richlowe:
That last commit unsets `BUILD_STYLE` and switches to the tarball provided @alarcher.

Mno-hime (approved):

PR #5102, opened 2019-06-23
Update GNU Guile to 2.2.7

Mno-hime:
Won't publish due to [11274 Crash in elf.so during REQUIRED_PACKAGES target](https://www.illumos.org/issues/11274).

Bug about `accept4(3socket)` detection handling reported:
* https://debbugs.gnu.org/cgi/bugreport.cgi?bug=36342
* https://lists.gnu.org/archive/html/bug-guile/2019-06/msg00022.html

Guile rebuilds: https://github.com/OpenIndiana/oi-userland/pull/5259.

@richlowe FYI.

PR #5259, opened 2019-08-23
Guile 2.2.7 rebuild

Mno-hime:
Requires https://github.com/OpenIndiana/oi-userland/pull/5102/ first.


richlowe:
How are they failing? slib doesn't even run the schemes it supports when it's being packaged.

richlowe:
I pulled these into my branch, and don't see any failures, of builds or tests except for 'slib' which somehow doesn't notice `NO_TESTS` and tries to run `make check`, which doesn't exist.

When you just run `/bin/guile` and `/bin/amd64/guile` do they start up and get you to the repl?
Whether they do or not, do they show a bunch of warnings about sources newer than objects?  maybe the heavy-handed approach I took to fixing that was still not heavy-handed enough?

richlowe:
Ok, I see the failures in slib now.  I must have checked with the old slib rather than the update you added.  The update does not build, either with Guile 2.2, or the original Guile 1.8.

In both cases it seems to assume it is already installed when building itself and looks either in (`/usr/share/slib`), or the guile 2 site directory

(`/usr/share/slib` isn't even where the file the old guile is looking for would be if it worked, it'd be in `/usr/lib/slib`)

richlowe:
https://gist.github.com/richlowe/7cbcd1281f58aa3b84374b6b5bce3ce9

Fixes the updated slib build with either version of guile, building the color database using guile just cannot possibly have ever worked (it doesn't anticipate open throwing an exception, rather than returning `#f` on failure), and didn't set `SCHEME_LIBRARY_DIRECTORY` so assumed it could load pieces of slib without respect to `DESTDIR`

Mno-hime:
gmake tests work fine. autogen should be updated to the latest version to gain Guile 2.2 support, but 64-bit fails with:
```
/export/home/newman/ws/oi-userland/components/developer/autogen/build/amd64/agen5/autogen -L/export/home/newman/ws/oi-userland/components/developer/autogen/autogen-5.18.16/autoopts/tpl -L../autoopts/tpl -MFstamp-agman -Tagman-cmd.tpl /export/home/newman/ws/oi-userland/components/developer/autogen/autogen-5.18.16/columns/opts.def
../autoopts/tpl/mdoc2man: line 35: package: command not found
../autoopts/tpl/mdoc2man: line 36: use: command not found
../autoopts/tpl/mdoc2man: line 37: use: command not found
../autoopts/tpl/mdoc2man: line 38: use: command not found
../autoopts/tpl/mdoc2man: line 39: syntax error near unexpected token `('
../autoopts/tpl/mdoc2man: line 39: `use lib dirname(__FILE__);'
Killing AutoGen 12869
FAILURE REASON:  ../autoopts/tpl/mdoc2man failed in /export/home/newman/ws/oi-userland/components/developer/autogen/build/amd64/columns
AutoGen aborting on signal 15 (Terminated) in state EMITTING
processing template ../autoopts/tpl/agman-cmd.tpl
            on line 51
       for function EXPR (14)
make[2]: *** [Makefile:822: stamp-agman] Hangup
make[2]: Leaving directory '/export/home/newman/ws/oi-userland/components/developer/autogen/build/amd64/columns'
make[1]: *** [Makefile:547: all-recursive] Error 1
make[1]: Leaving directory '/export/home/newman/ws/oi-userland/components/developer/autogen/build/amd64'
make: *** [Makefile:454: all] Error 2
gmake: *** [/export/home/newman/ws/oi-userland/make-rules/configure.mk:183: /export/home/newman/ws/oi-userland/components/developer/autogen/build/amd64/.built] Error 2
main: line 34: kill: (12869) - No such process
main: line 36: kill: (12869) - No such process
```
I think it's because perl shebang is destroyed in generation:
```patch
--- build/i86/autoopts/tpl/mdoc2man	2019-08-24 22:01:14.127480356 +0000
+++ build/amd64/autoopts/tpl/mdoc2man	2019-08-24 22:02:04.867188763 +0000
@@ -1,4 +1,4 @@
-#! /usr/bin/perl
+#!
 
 ## mdoc2man.pl -- Convert mdoc tags to man tags
 ##
```
Probably boils down to 64-bit can't find the 64-bit Perl 5.24? Even I can't find it :).

Thanks for the SLIB patch will have a look at it.

Mno-hime:
@richlowe here are two slib patches for you to considerate:

- someone was unsure about license to the colors code: https://build.opensuse.org/package/view_file/devel:languages:misc/slib/remove-resene-colors.diff?expand=1
- https://src.fedoraproject.org/rpms/slib/blob/master/f/slib-guile2.patch

richlowe:
I don't know enough about slib to be sure of either of the patches.  Rebuilding the color name database is exactly what guile can't do (because the slib code is broken), so patching it depends on whether they patch the result or the input (if only the input, it won't actually do anything for us...)

The array changes say it was broken in 3b4, did it not get fixed in 3b5?

(Honestly, I didn't know people still used slib _or_ that OI shipped slib)

richlowe:
Is there anything more I can do to help with this?

Mno-hime:
> Is there anything more I can do to help with this?

`autogen` build fails on amd64 because it wrongly generates perl shebang in the file above. We have different perls for 32-bit and 64-bit. I am sure @pyhalov has some ideas :).

Mno-hime:
autogen is fixed. Ready for review.

pyhalov:
Also have to check swig with guile 2.2

PR #5270, opened 2019-08-27
Font renames

alarcher:
This fixes some renames due to mispublished packages with fmri system/font/foo instead of system/font/truetype/foo.
In all the case the component revision taken for the rename was the last in the repository plus one.
This also fixes the gnome-fonts package which refers to mispublished fmris instead of the correct ones.

### DIFF
--- a/components/fonts/arphic-ukai/history
+++ b/components/fonts/arphic-ukai/history
@@ -1,0 +1,1 @@
+system/font/arphic-ukai@0.2.20080216.1-2018.0.0.1 system/font/truetype/arphic-ukai
--- a/components/fonts/baekmuk/history
+++ b/components/fonts/baekmuk/history
@@ -1,1 +1,2 @@
-system/font/baekmuk-ttf@2.2-2018.0.0.1 system/font/baekmuk
+system/font/baekmuk-ttf@2.2-2018.0.0.2 system/font/truetype/baekmuk
+system/font/baekmuk@2.2-2018.0.0.2 system/font/truetype/baekmuk
--- a/components/fonts/bitstream-vera/history
+++ b/components/fonts/bitstream-vera/history
@@ -1,0 +1,1 @@
+system/font/bitstream-vera@1.10-2018.0.0.2 system/font/truetype/bitstream-vera 
--- a/components/fonts/freefont/history
+++ b/components/fonts/freefont/history
@@ -1,1 +1,2 @@
-system/font/freefont-ttf@0.20120503-2018.0.0.1 system/font/freefont
+system/font/freefont-ttf@0.20120503-2018.0.0.2 system/font/truetype/freefont
+system/font/freefont@0.20120503-2018.0.0.2 system/font/truetype/freefont
--- a/components/fonts/arphic-ukai/Makefile
+++ b/components/fonts/arphic-ukai/Makefile
@@ -19,6 +19,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		arphic-ukai
 COMPONENT_VERSION=	0.2.20080216.1
+COMPONENT_REVISION=     1
 COMPONENT_SUMMARY=	CJK Unicode TrueType font Ming style
 COMPONENT_PROJECT_URL=	https://www.freedesktop.org/wiki/Software/CJKUnifonts
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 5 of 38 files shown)


components/fonts/baekmuk/history
```diff
@@ -1 +1,2 @@
-system/font/baekmuk-ttf@2.2-2018.0.0.1 system/font/baekmuk
+system/font/baekmuk-ttf@2.2-2018.0.0.1 system/font/truetype/baekmuk
+system/font/baekmuk@2.2-2018.0.0.2 system/font/truetype/baekmuk
```
Mno-hime, line 2: Shouldn't it be both `2018.0.0.2` given that we want to replace `2018.0.0.1` and older?
```
$ pkg info -r system/font/baekmuk-ttf system/font/baekmuk
          FMRI: pkg://openindiana.org/system/font/baekmuk@2.2-2018.0.0.1:20181106T081715Z
          FMRI: pkg://openindiana.org/system/font/baekmuk-ttf@2.2-2018.0.0.1:20190720T075302Z
```

components/fonts/bitstream-vera/history
```diff
@@ -0,0 +1 @@
+system/font/bitstream-vera@1.10-2018.0.0.1 system/font/truetype/bitstream-vera
```
Mno-hime: `2018.0.0.2`?
```
FMRI: pkg://openindiana.org/system/font/bitstream-vera@1.10-2018.0.0.1:20181106T081711Z
```

components/fonts/freefont/history
```diff
@@ -1 +1,2 @@
-system/font/freefont-ttf@0.20120503-2018.0.0.1 system/font/freefont
+system/font/freefont-ttf@0.20120503-2018.0.0.1 system/font/truetype/freefont
```
Mno-hime: `2018.0.0.2`?
```
FMRI: pkg://openindiana.org/system/font/freefont@0.20120503-2018.0.0.1:20181106T081715Z
```

Mno-hime (commented):
It seems to me that sometimes you put the current FMRI there instead of the `+1` in the branch part. I used `pkg info -r` to have a look at those obsoleted FMRIs.

components/fonts/baekmuk/history
alarcher: Probably yes... I forgot that they had been republished... not sure what happens when there is a rename to a fmri that does not exist... because in the previous revision only system/font/truetype/baekmuk@2.2-2018.0.0.1 was actually published and actually provided content... better bump.

components/fonts/freefont/history
alarcher: Same here... when this was republished only system/font/truetype/freefont@0.20120503-2018.0.0.1 provided content... but bumping as well for safety.

components/fonts/baekmuk/history
Mno-hime: Hmm, I don't know... Maybe we should wait for @pyhalov.

components/fonts/arphic-ukai/history
```diff
@@ -0,0 +1 @@
+system/font/arphic-ukai@0.2.20080216.1-2018.0.0.1 system/font/truetype/arphic-ukai
```
AndWac, line 1: Sorry, I am irritated.
From this line I would expect that you want to obsolete system/font/arphic-ukai. But
font.mk sets the default COMPONENT_CLASSIFICATION to System/Font and there is nothing in the Makefile. So why are you creating a new revision just to obsolete the package?
Furthermore, I expect that you'll need to increment the revision in the history file again because the new package revision is already 1 and thus, in order to obsolete it here it must be incremented again.

alarcher: I want to obsolete system/font/arphic-ukai which was published by mistake to enforce the correct FMRI. In this PR system/font/arphic-ukai@0.2.20080216.1-2018.0.0.1 is not published since it is not the correct FMRI, but system/font/truetype/arphic-ukai@0.2.20080216.1-2018.0.0.1 together with the rename rules. The latest revision in the repo is 0. Maybe I misunderstood? You can check that COMPONENT_FONT_TYPE = $(PKGFONTCLS) which is part of the FMRI.

AndWac: Ah, ok. I misread font.mk. You are right.

Mno-hime (approved):
Now it matches my understanding of package obsolescence.

PR #5278, opened 2019-08-29
Update libgcrypt to 1.8.5, fix libgpg-error

Mno-hime:
### libgcrypt
**Changelogs**
- 1.8.5: https://dev.gnupg.org/T4683
- up to and including 1.8.4: https://abi-laboratory.pro/index.php?view=changelog&l=libgcrypt&v=1.8.4

I removed the `-O2` workaround as it seem to me it's not needed with GCC 6 (the indicative test cases passed with the default `-O3`).

ABI checker [clean](https://abi-laboratory.pro/index.php?view=timeline&l=libgcrypt).

**Testing**
- internal test suite did not regress (and even passed with one skip when run as root)
- will keep it for a while in my environment: DONE
- libgcrypt pkg-config module: OK

### libgpg-error

Fix the `pg-error-config-test.sh` test.

### DIFF
--- a/components/library/libgcrypt/Makefile
+++ b/components/library/libgcrypt/Makefile
@@ -21,37 +21,29 @@
 # Copyright 2013 Colin Ellis. All rights reserved.
 # Copyright 2011 Audioboo Ltd. All rights reserved.
 #
+
+BUILD_BITS=		32_and_64
+
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libgcrypt
-COMPONENT_VERSION=	1.8.1
+COMPONENT_VERSION=	1.8.5
 COMPONENT_FMRI=		system/library/security/libgcrypt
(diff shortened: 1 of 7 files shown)


components/library/libgpg-error/patches/02-include-amd64-search-path.patch
```diff
@@ -0,0 +1,14 @@
+This includes /usr/lib/amd64 to `Libs:` in the amd64 .pc file and thus breaks
```
pyhalov: heals?

Mno-hime: Right... Fixed in https://github.com/OpenIndiana/oi-userland/pull/5278/commits/26bf75f863930f824ae83d51950c666dd9e88b70#diff-9dcdfaaa584909b13eb99142fa1eccf6.

PR #5279, opened 2019-08-29
nodejs-12: update to 12.9.1

AndWac:


Mno-hime (commented):
`build`, `install`, and `test` targets should be removed from the `Makefile`. `TEST_TARGET` set to `$(NO_TESTS)` before `common.mk`. `BUILD_BITS` set to `64`.

alarcher (approved):

PR #5285, opened 2019-08-31
kvm: import OS-7960 patch

AndWac:
Simple copyright changes from 2018 to 2019 have been removed because our files are older and the patches failed.


alarcher:
@AndWac Do you want me to merge it right away since Alexander is on vacation?

AndWac:
@alarcher I think it's ok to merge. Alexander is back but has no time for several days.

alarcher (approved):

PR #5296, opened 2019-09-07
Golang 112

Toasterson:
I needed to get rid of all binaries in the testdata as they kept failing pkg's vaildations. Disabling the checks did not work.


components/developer/golang-112/golang.p5m
```diff
@@ -23,4 +23,4 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-depend type=require fmri=developer/golang-111
+depend type=require fmri=developer/golang-112
```
Mno-hime, line 26: Why this? Perhaps you just wanted to `cp` the file and you `git mv` it?

Mno-hime (commented):
Thanks for doing this!

Maybe it's also time to ditch some of the old, unsupported golangs? Namely 1.8 and 1.10.

Toasterson:
We can drop 1.8 and 1.10

components/developer/golang-112/golang.p5m
Toasterson: i wanted to set golang 1.12 as the default thus i moved the golang package manifest to it's new folder like back when i added go 1.10

Mno-hime:
Did you rebuild sample manifest? I get several updates in that file and they should be possibly added to the golang-112.p5m file: [sample-manifest.diff.txt](https://github.com/OpenIndiana/oi-userland/files/3587036/sample-manifest.diff.txt).

Toasterson:
I did but I had problems with it. I'll try to get the same result when rebuilding. sample-manifest did at first not like my locale. I'll see what i can do.

Mno-hime (approved):

alarcher:
Argh... again same issue merging on the phone... did not see that it was not squashed...

PR #5315, opened 2019-09-21
Update open-vm-tools to 11.0.5

Mno-hime:
**Release notes**
- https://github.com/vmware/open-vm-tools/blob/stable-11.0.0/ReleaseNotes.md
- https://github.com/vmware/open-vm-tools/blob/stable-11.0.5/ReleaseNotes.md

**Blog**
- https://blogs.vmware.com/vsphere/2019/09/vmware-tools-11-0-out-now.html

I'd appreciate if some more versed in C checked my changes to the `70-dilos.patch` (compilation used to fail dou to comparison of pointer and integer, probably because illumos changed the `NULL` definition).

**Testing**
- tested on VMware ESXi 6.7

### DIFF
--- a/components/sysutils/open-vm-tools/Makefile
+++ b/components/sysutils/open-vm-tools/Makefile
@@ -14,12 +14,12 @@
 # Copyright 2019 Michal Nowak
 #
 
-PREFERRED_BITS=		64
+BUILD_BITS=		64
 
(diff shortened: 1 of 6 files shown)


Mno-hime:
I did the testing on ESXi 6.7, all clear.

PR #5322, opened 2019-09-25
Update GNU Texinfo to 6.7

Mno-hime:
Release notes: https://lists.gnu.org/archive/html/bug-texinfo/2019-09/msg00018.html

**Testing**
- test suite passed
- GCC 4.9, 3.4, and illumos-gcc build fine.
- `info info` works

### DIFF
--- a/components/text/texinfo/Makefile
+++ b/components/text/texinfo/Makefile
@@ -24,21 +24,21 @@
 # Copyright (c) 2019, Michal Nowak
 #
 
+BUILD_BITS=		64
+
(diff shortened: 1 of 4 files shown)


alarcher:
Do we really need the 32-bit version?

Mno-hime:
> Do we really need the 32-bit version?

Dunno, lets find out! (Will keep in my system as 64-bit-only and will see.)

Mno-hime:
Didn't spot any problems.

PR #5324, opened 2019-09-25
11655 top(1) computes compressed ARC ratio incorrectly

pyhalov:


Mno-hime (approved):

tsoome (changes requested):
- 3 prhaps:)

tsoome (approved):

Mno-hime (approved):
The ratio is good now.

PR #5326, opened 2019-09-25
Update PHP 7.3 to 7.3.9, enable FreeType

Mno-hime:
Changelog: https://www.php.net/ChangeLog-7.php#7.3.9

Enabled FreeType per @Toasterson request:
```
 newman   php_7.3_enable_freetype  …  web  php  php-7_3  ldd build/prototype/i386/usr/php/7.3/extensions/gd.so | grep freetype
	libfreetype.so.6 =>	 /usr/lib/64/libfreetype.so.6
```

**Testing**
- I get fails on top of what is recorded, but no regression

### DIFF
--- a/components/web/php/php-7_3/patches/02-jpeg.h.patch
+++ b/components/web/php/php-7_3/patches/02-jpeg.h.patch
@@ -1,0 +1,13 @@
+--- php-7.3.9/ext/gd/config.m4	2019-08-28 00:52:41.000000000 +0000
++++ php-7.3.9/ext/gd/config.m4	2019-10-16 20:40:20.850333627 +0000
+@@ -101,8 +101,8 @@ AC_DEFUN([PHP_GD_WEBP],[
+ AC_DEFUN([PHP_GD_JPEG],[
+   if test "$PHP_JPEG_DIR" != "no"; then
+ 
+-    for i in $PHP_JPEG_DIR /usr/local /usr; do
+-      test -f $i/include/jpeglib.h && GD_JPEG_DIR=$i && break
++    for i in $PHP_JPEG_DIR; do
++      test -f $i/jpeglib.h && GD_JPEG_DIR=$i && break
+     done
+ 
+     if test -z "$GD_JPEG_DIR"; then
--- a/components/web/php/php-7_3/Makefile
+++ b/components/web/php/php-7_3/Makefile
@@ -119,7 +119,7 @@ CONFIGURE_OPTIONS  +=	--with-pdo-pgsql=shared,$(PG_BINDIR)
 # Enabled Extensions
 #CONFIGURE_OPTIONS     +=	--with-libxml-dir=/usr
 CONFIGURE_OPTIONS     +=	--with-freetype-dir=/usr
-CONFIGURE_OPTIONS     +=	--with-jpeg-dir=/usr
+CONFIGURE_OPTIONS     +=	--with-jpeg-dir=/usr/include/libjpeg8-turbo
 #CONFIGURE_OPTIONS     +=	--with-png-dir=/usr
 #CONFIGURE_OPTIONS     +=	--with-zlib-dir=/usr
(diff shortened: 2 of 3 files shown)


Toasterson:
It works for me when more than one jpeg library is installed. It turns out libjpeg6-ijg delivers  /usr/include/jpeglib.h -> libjpeg6-ijg/jpeglib.h i think we will have to mediate this link for the configure script of php to pass without /usr/include/jpeglib.h existing. I think php's configure script is not intelligent enough to recognize subdirectories under include....

For this Package LGTM though. Unless you want to patch the configure script to be less stupid about presence checks.

pyhalov:
Should we fix '--with-jpeg-dir=/usr'?

Mno-hime:
Building with `--with-jpeg-dir=/usr/include/libjpeg8-turbo` works too because it looks into `/usr/include/libjpeg.h` anyway:
```shell
if test "$PHP_JPEG_DIR" != "no"; then

  for i in $PHP_JPEG_DIR /usr/local /usr; do
    test -f $i/include/jpeglib.h && GD_JPEG_DIR=$i && break
  done

  if test -z "$GD_JPEG_DIR"; then
    as_fn_error $? "jpeglib.h not found." "$LINENO" 5
  fi
```

I am afraid `--with-jpeg-dir=...` works with anything... Perhaps fix it by removing the `include` portion?

I guess `/usr/include/jpeglib.h` should be mediated or removed?:
```
 newman   php_7.3_enable_freetype  …  web  php  php-7_3  ll /usr/include/jpeglib.h 
lrwxrwxrwx   1 root     other         22 Jun  5  2018 /usr/include/jpeglib.h -> libjpeg6-ijg/jpeglib.h
```

Toasterson:
@Mno-hime do we want a seperate pr about /usr/include/jpeglib.h and then merge this? I vote for mediating

pyhalov:
@Mno-hime I think that proper fix would be to fix php-7.3.7/ext/gd/config.m4 either to honour PHP_JPEG_INCLUDE_DIR (or similar macros), simple fix - just avoid checking for /usr/include/jpeglib.h existence. Mediating include file in /usr/include... Have we ever done this? I'm not sure if it's a good idea.

Mno-hime:
@pyhalov @Toasterson not sure I nailed it in 989a0e3756bcf80badb06b65987e1d0d5965d675. Please check.

Toasterson:
@Mno-hime Yeah that makes sense. and works with only libjpeg8-turbo installed. Approved

Mno-hime:
> @Mno-hime Yeah that makes sense. and works with only libjpeg8-turbo installed. Approved

Thanks for the testing!

Toasterson (approved):

components/web/php/php-7_3/patches/02-jpeg.h.patch
```diff
@@ -0,0 +1,13 @@
+--- php-7.3.9/ext/gd/config.m4	2019-08-28 00:52:41.000000000 +0000
++++ php-7.3.9/ext/gd/config.m4	2019-10-16 20:40:20.850333627 +0000
+@@ -101,8 +101,8 @@ AC_DEFUN([PHP_GD_WEBP],[
+ AC_DEFUN([PHP_GD_JPEG],[
+   if test "$PHP_JPEG_DIR" != "no"; then
+ 
+-    for i in $PHP_JPEG_DIR /usr/local /usr; do
+-      test -f $i/include/jpeglib.h && GD_JPEG_DIR=$i && break
++    for i in $PHP_JPEG_DIR; do
```
pyhalov, line 9: It seems that later code tries to use -L $PHP_JPEG_DIR/lib, which is incorrect (so not suitable for upstreaming), but likely it's fine enough for us.

Mno-hime: I take it as your approval :).

pyhalov: OK

PR #5327, opened 2019-09-26
xorg-server: introduce x11-init service

pyhalov:


components/x11/xorg-server/files/x11-init
```diff
@@ -0,0 +1,42 @@
...
+. /lib/svc/share/smf_include.sh
+
+PATH=/bin
+
+IFS=";"
+for rec in $DIRS; do
+   dir=`echo $rec | cut -d ":" -f 1`
+   owner=`echo $rec | cut -d ":" -f 2`
+   group=`echo $rec | cut -d ":" -f 3`
+   mode=`echo $rec | cut -d ":" -f 4`
```
jimklimov: ETOOMANYFORKS :)

````
$ ksh -c 'DIRS="/tmp/.ICE-unix:root:root:1777;/etc/qwe:jim:staff:555"; echo "$DIRS" | tr ";" "\n" | while IFS=":" read dir owner group mode in $rec ; do echo "$dir = $owner = $group = $mode" ; done'
````
Could also define DIRS as a multiline string right away and so no `tr`, and even can do no `echo`:

````
DIRS="/tmp/.ICE-unix:root:root:1777
/tmp/.ICE-unix:root:root:1777"

while IFS=":" read dir owner group mode in $rec ; do
    echo "$dir = $owner = $group = $mode"
done << EOF 
$DIRS
EOF
````

File: components/x11/xorg-server/files/x11-init [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+   
+   if [  -e "$dir"  -a ! -d "$dir" ]; then
+         mv "$dir" "$dir.$$" 
+   fi
+   /bin/mkdir -p -m "$mode" "$dir"
+# If directory exists, mkdir will not change its permissions
+   chmod "$mode" "$dir"
+   chown $owner:$group "$dir"
```
jimklimov: Doublequotes against spaces in usernames? :)

components/x11/xorg-server/files/x11-init.xml
```diff
@@ -0,0 +1,87 @@
...
+            type='service'>
+                <service_fmri value='svc:/system/filesystem/local'/>
+        </dependency>
+
+        <dependency name='single-user'
+            grouping='require_all'
+            restart_on='error'
+            type='service'>
+                <service_fmri value='svc:/milestone/single-user:default'/>
+        </dependency>
```
jimklimov: Indentation is jumpy, and should this be required by an X11 service, e.g. lightdm, to surely start before?

jimklimov: Ok, I see the dependency is covered from the other side in #5328

jimklimov:
FWIW, a quote from IRC discussion lest it gets lost (there are possible security uncertainties with current code):

-----

(15:08:24) leoric: jimklimov: hm... after thinking a bit, can you look at possible races?
(15:09:12) leoric: I mean, if we replace directory with a link, mkdir fails, so we don't chmod anything suspicious...
(15:09:51) leoric: and if it doesn't fail, we have plain directory, possibly, owned by non-root
(15:11:07) leoric: if mkdir succeeds, we have root-owned empty directory, so no harm in chmod/chown
(15:21:13) jimklimov: with `/usr/bin/sh -e` the erroneous line (resulting not in code 0) should kill the script and not proceed beyond failed mkdir?
(15:21:32) jimklimov: maybe explicit `set -e` would better survive possible future editions though
(15:21:48) jimklimov: or `mkdir ... || exit`
(15:22:43) leoric: yes, sh -e will error out
(15:24:35) jimklimov: checked: symlink to existing dir is a [ -d ... ] success in ksh
(15:27:06) jimklimov: but `mkdir -p -m ... symlink` does not fail, and does not change the rights (neither link nor dir)
(15:47:42) jimklimov: so there may be a security problem with somebody making a symlink to whatever and then this service mangles the innocent FS object
(15:48:01) jimklimov: maybe also test `-L` and remove symlinks?
(15:49:51) jimklimov: and also checked that existing symlink to non-existant directory fails, differently in builtin and /bin cases
(15:50:03) jimklimov: jim@jimoi:/tmp$ mkdir -p -m "700" yyyyy ; echo $?                               
(15:50:03) jimklimov: mkdir: cannot create directory ‘yyyyy’: File exists
(15:50:03) jimklimov: 1
(15:50:03) jimklimov: jim@jimoi:/tmp$ /bin/mkdir -p -m "700" yyyyy ; echo $?                          
(15:50:03) jimklimov: mkdir: "yyyyy": Exists but is not a directory
(15:50:03) jimklimov: 2
(15:51:03) jimklimov: hm, for symlink to existing dir builtin is ok, /bin fails
(15:53:08) jimklimov: however if you end the argument with a slash, it is ok for both if the referenced dir existed or not (will get created if missing, then even with rights set)
(15:55:06) jimklimov: so `mkdir d/` is ok where "d" is a directory or a symlink resolving to directory or a dangling symlink (but not a symlink to existing non-directory object)

Mno-hime (approved):

jimklimov (approved):
Thanks, LGTM better now :)

pyhalov:
Ouch, forgot to merge commits, sorry...

PR #5330, opened 2019-10-01
Remove 7661 backout and integrated 11755

Mno-hime:
Now that Perl is supposed to be fixed in the gate and 11755 got integrated our gate build [fails](https://hipster.openindiana.org/logs/illumos-gate/latest/openindiana.illumos-gate.publish.log) and we can backout the 7661 backout and the 11755 patch:
```
error: Your local changes to the following files would be overwritten by merge:
	usr/src/boot/Makefile.version
	usr/src/boot/sys/boot/common/module.c
	usr/src/cmd/perl/Makefile.perl
Please commit your changes or stash them before you merge.
Aborting
```
Will merge once my local build publishes.


pyhalov:
We also likely want to set `BUILDPERL64='#'` in illumos.sh

Mno-hime:
Build passed.

PR #5066, opened 2019-06-07
convert gst-plugins-good to use libjpeg8

timmooney:
I've converted `library/audio/gstreamer/plugin/good` to use libjpeg8-turbo and done other component update tasks, but I'm wondering if perhaps the component should be marked obsolete instead.

Only `mate_install` meta-package requires `library/audio/gstreamer/plugin/good` (the depend is in includes/desktop_common).  Perhaps `mate_install` should be referencing the gstreamer1 versions of components instead?  I know that some components still depend on the older gstreamer 0.x components.

As far as my changes to library/audio/gstreamer/plugin/good, I
  * bumped `COMPONENT_REVISION`
  * added the `CFLAGS/CXXFLAGS/LDFLAGS` boilerplate to build with libjpeg8-turbo
  * updated the `PATH` to use `$(PATH.gnu)`
  * ran `gmake REQUIRED_PACKAGES` and updated the `Makefile`.  Only the JPEG dependency changed.
  * re-ran `gmake sample-manifest` to verify that there were no file changes

PR #5335, opened 2019-10-02
remove gstreamer/plugin/good from mate_install

timmooney:
`library/audio/gstreamer/plugin/good` can be removed from the `mate_install` meta-package, since no other current components from MATE or `desktop_common` depend upon it.

Note that there are external dependencies (like SFW libreoffice) that still depend upon `library/audio/gstreamer/plugin/good`, but nothing in current MATE does.

See PR #5066 for more discussion.


pyhalov:
We still can not do it. Several components still use old gstreamer version (like totem) and so need gstreaemer plugins it.

PR #5337, opened 2019-10-05
Update nvidia driver to version 390

Mno-hime:
Deliver the nvidia driver version 390 to support newer GPUs.

The changes in the old driver are meant to keep the old driver around so users with and very old card not supported in the version 390 driver can manually (with some hassle) install it. But not sure this is possible at all? Or perhaps we don't want the version 340 around at all?

I bypassed dependency generation of `mics/gfx_private` kernel module for the `nvidia` driver as we seems to deliver `gfx_private` via `system/kernel/platform` package and I am not sure which property to alter in the `nvidia` module to find it automatically. And perhaps altering the module is not allowed in the first place?

**Testing**
- Update from v340:
```
 newman  ~  pfexec pkg update -vn pkg://userland/driver/graphics/nvidia
...
       Create boot environment:       Yes
     Activate boot environment:       Yes
Create backup boot environment:        No
          Rebuild boot archive:       Yes

Changed packages:
openindiana.org -> userland
  driver/graphics/nvidia
    0.340.107-2018.0.0.1 -> 0.390.129-2019.0.0.0
userland
  driver/graphics/nvidia-390
    None -> 0.390.129-2019.0.0.0
...
```
- Can't test it on my own as I don't have nvidia GPU


AndWac:
Superceded by #6060

PR #5344, opened 2019-10-14
Update sudo to 1.8.28, CVE-2019-14287

Mno-hime:
Changes: https://www.sudo.ws/stable.html#1.8.28

One change specifically relevant to us: "Sudo will now only set PAM_TTY to the empty string when no terminal is present on Solaris and Linux. This workaround is only needed on those systems which may have PAM modules that misbehave when PAM_TTY is not set."

**Testing**
- test suite did not regress


pyhalov:
Sorry, duplicated this in #5346

PR #5350, opened 2019-10-15
Update golang 1.13.0 to 1.13.1 and fix illumos tools permissions

Toasterson:
needed to also replace COMPONENT_VERSION im manifests as that does not match anymore. Thus the large diff


Mno-hime (dismissed):
@Toasterson `sample-manifest` target rebuild shows some changes for me, see b5dec3369c4913ca02676c65b200f9fbbd0f9ad8.

Tests look good as well except for https://github.com/golang/go/issues/34191.

Toasterson:
@Mno-hime good question. if they are not reproducible the same everytime then it does not make sense to ship them. We simply would not be able to produce the same exact files as the buildserver. I'll remove gocache stuff again. I thought it was the same every time.

Toasterson:
I have updated golang-113.p5m so gocache is not delivered anymore as the README contains information about that cache i have left it in.

Mno-hime (approved):

PR #5358, opened 2019-10-17
X11libs updates

alarcher:
@Mno-hime If you have time to test...

PR #5359, opened 2019-10-18
Update x11-protocols to 2019.2, do not ship XKBgeom.h & vldXvMC.h

Mno-hime:
https://github.com/OpenIndiana/oi-userland/pull/5358#issuecomment-543709074

https://github.com/OpenIndiana/oi-userland/pull/5358 delivers XKBgeom.h & vldXvMC.h files which are also delivered in `x11-protocols` 2019.1 package. In `x11-protocols` 2019.2 those two files were were put to "legacy" mode, so we won't deliver them as part of `x11-protocols` package although "legacy" mode is still on.

Merging now as it unblocks update, but later review is appreciated.


alarcher:
@Mno-hime Perfect. The release note was a bit misleading. Thank you!

PR #5365, opened 2019-10-24
[WiP] make-rules/meson.mk for components that use meson

timmooney:
This needs careful review and should not be committed until all the frequent contributors to oi-userland are comfortable that it's the right approach.

Many packages in the past couple years have switched from `configure` to [meson](https://mesonbuild.com) to generate their build/install/test rules.  I don't personally like it, but if we want to be able to build many recent GNOME components or things like MESA, having support for meson in oi-userland would be beneficial.

To that end, I

- copied `configure.mk` to `meson.mk`
- added `MESON`, `MESON_BUILDTYPE`, and `NINJA` settings, since those are not currently set in `shared-rules.mk`
- overrode `COMPONENT_TEST_CMD` and `COMPONENT_TEST_TARGETS` from `shared-rules.mk`, since meson uses ninja (by default) instead of gmake and the test target is `meson-test` or `test`, not `check`
- removed `CONFIG_SHELL` from `CONFIGURE_ENV`.  Setting `CONFIG_SHELL` is useful for running an `autoconf`-generated `configure` script, but not needed for meson.
- removed `CC`, `CXX`, `F77`, etc. from `CONFIGURE_OPTIONS`.  For meson, those continue to be passed in the environment, but should not be passed on the command line.
- removed `DESTDIR=$(PROTO_DIR)` from `COMPONENT_INSTALL_ARGS`, and added the same setting to `COMPONENT_INSTALL_ENV`.  To install to a `DESTDIR`, meson wants that setting in the environment.
- commented out (but did not remove) various settings related to override `ac_cv_*` cache-values that configure apparently gets wrong for OI.  These aren't relevant for meson.
- changed the `.configured` target to essentially call `meson setup --buildtype=$(MESON_BUILDTYPE) build_dir source_dir`.  With meson, when you call it to generate the ninja build rules, you must specify the build (output) directory and the source directory.
- changed the `.built` target to use `ninja` instead of `gmake`
- changed the `.installed` target to use `ninja` instead of `gmake`
- added `REQUIRED_PACKAGES +=` for both `meson` and `ninja`

**Most importantly**, I left all the settings with a prefix of `CONFIGURE`.  I don't know if it would be better or more logical to name them all `MESON_<whatever>` or stay with `CONFIGURE_<whatever>`.  We won't want to change it after `meson.mk` is used by a few components, so it's important to get it right at the start.

I also did not remove any of the other code for ccache, parfait, etc.  I've never done component builds using either, so I'm not qualified to determine what should be done there.   It should probably all be removed from `meson.mk`, unless one of you can test it and verify it's beneficial to keep.

I've tested using this `meson.mk` on a few simple components that use meson, and have been able to generate build rules for ninja (configure), build both 32 and 64 bit, install both 32 and 64 bit, and run test targets.   Nothing "advanced" has been tested.

Please review, @pyhalov , @alarcher , @Mno-hime , and tag any others that should also review.

### DIFF
--- a/make-rules/meson.mk
+++ b/make-rules/meson.mk
@@ -1,0 +1,257 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+# Copyright (c) 2010, 2016, Oracle and/or its affiliates. All rights reserved.
+# Copyright 2011 EveryCity Ltd. All rights reserved.
+#
+
+#
+# Rules and Macros for building open source software that uses the meson
+# build system, https://mesonbuild.com/
+#
+# This is based heavily upon configure.mk.  It uses ninja to build
+# the components and supports buliding multiple version (32/64 bit)
+# from a shared source.
+#
+# To use these rules, include $(WS_MAKE_RULES)/meson.mk in your Makefile
+# and define "build", "install", and "test" targets appropriate to building
+# your component.
+# Ex:
+#
+# 	build:		$(SOURCE_DIR)/build/$(MACH32)/.built \
+#	 		$(SOURCE_DIR)/build/$(MACH64)/.built
+#
+#	install:	$(SOURCE_DIR)/build/$(MACH32)/.installed \
+#	 		$(SOURCE_DIR)/build/$(MACH64)/.installed
+#
+#	test:		$(SOURCE_DIR)/build/$(MACH32)/.tested \
+#	 		$(SOURCE_DIR)/build/$(MACH64)/.tested
+#
+# Any additional pre/post configure, build, or install actions can be specified
+# in your make file by setting them in on of the following macros:
+#	COMPONENT_PRE_CONFIGURE_ACTION, COMPONENT_POST_CONFIGURE_ACTION
+#	COMPONENT_PRE_BUILD_ACTION, COMPONENT_POST_BUILD_ACTION
+#	COMPONENT_PRE_INSTALL_ACTION, COMPONENT_POST_INSTALL_ACTION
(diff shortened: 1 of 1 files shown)


Mno-hime:
Thanks for taking the lead in this @timmooney. Did you notice that solaris-userland also gained [meson support](https://github.com/oracle/solaris-userland/commit/d9d8dfb5ee3b4d5bd14a76f55d87c378791c9d7d) in the build system recently? Both your and Oracle approach to the problem is very similar. Would you care to have a look at the Solaris change and see if it works for us as well? Given our resources, I think we should share code where possible.

timmooney:
I hadn't looked at Oracle's meson code but I had seen Alan's name in some meson bug reports and other areas while researching meson, so I knew he had been working on it.

It may be some time before I have time to test his `meson.mk` and `ninja.mk` with components in oi-userland.  If someone else has time and interest before I get to it, they shouldn't wait for me.

Looking at his diffs, the main differences seem to be

- they use `BUILD_TYPE` to a greater extent than we appear to.
- they split the settings for the generator (meson) from the tool used to build (ninja).  That's probably a very good idea.
- they have a default for building PIE that can be overridden.  That's also probably a good idea.
- they conditionally set `MESON_OPTIMIZATION`.  I don't know why that's useful, especially when using the `plain` build type, but Alan may have found something in his research that I didn't.

Overall, using their `meson.mk` and `ninja.mk` look like worthwhile goals.

The appear to also have a `gnome-component.mk` that may be worth looking at.

alarcher:
It would be nice to integrate this so that we can update mesa and the graphic stack in the coming weeks. I had a shot at meson.mk earlier but I'd rather follow the solaris-userland recipe.

pyhalov:
This would be good, but someone should check that our customizations to configure.mk are preserved.

timmooney:
As I said back in October, I don't know when I would have time to do extensive testing with Oracle's meson.mk and ninja.mk, but I do think what Alan developed is better than my first draft.  I think if anyone works on meson before I get back to it, they should start with his meson.mk and ninja.mk.

timmooney:
In some brief testing of just Oracle's `meson.mk` and `ninja.mk` (not testing their full patch, with updated component meson), I see two problems:

* they default to building PIE, which doesn't work for the components I tested.
* something about their rules breaks the `gmake clobber` target, so cleanup is not done of the build directory.  I haven't had time to look to see what might be causing that.

timmooney:
If you make 2 changes to Oracle's `meson.mk`, it builds/tests/publishes all the components I had previously tried with my `meson.mk`:

* set `MESON_BUILDPIE = false` on line 50
* add (anywhere, probably near the bottom, could also probably go in `ninja.mk`):
```
# added for OI, must additionally clean up these:
clean::
    $(RM) -r $(BUILD_DIR) $(PROTO_DIR)
```

As I said, I haven't tried any of the rest of Oracle's patch.  I've only tested `meson.mk` and `ninja.mk` with our current version of meson and ninja.

AndWac:
I think we can close this as we have #5423 merged...

Mno-hime:
#5423 got in, closing.

PR #5372, opened 2019-10-28
Update VirtualBox to 6.1

Mno-hime:
Note that `VBOX_OGLLIBS` are missing from the build, due to [`VBoxVGA` 3D support removal](https://www.virtualbox.org/pipermail/vbox-dev/2020-January/015409.html).

Changelog: https://www.virtualbox.org/wiki/Changelog

**Testing**
- [x] OpenIndiana GUI works
- [x] Seamless with OpenIndiana works
- [x] audio works (MN: Does not work for me; Espen Martinsen: OK.)


Mno-hime:
Additional testing by following users:

Predrag Zečević: "Finally I have managed to test VB 6.1. After installing it into separate BE, rebooting, adding VBOX Additions, I have booted my win10 guest. Seamless mode works as before (not perfect, but usable). There is nothing else which makes guest to run worse then in VB 6.0."

Espen Martinsen: "Scaled mode works. Full screen works. Seamless on w10 with VboxSVGA  do not work, I've never used it som I dont' know if it has ever worked. It alwost works, can toggle on/off, windows is coming in theri own window, but Win-menu cannot be opened. Seamless on debian-variant (4.19) and VMSVGA  works fine. Both w10 and Kali (the debian variant) was tested with additions first 6.0.4 and then upgraded to 6.1.0. Both w10 and Kali plays "Merry Christmas Mr Bean" from youtube very well. I'm sorry I cannot test microphone, it have never even worked in host-OS (hardware-error)."

components/sysutils/virtualbox/virtualbox-additions.p5m
```diff
@@ -16,7 +16,7 @@
 #
 
 set name=pkg.fmri value=pkg:/system/virtualbox/virtualbox-additions@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
-set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=pkg.summary value="VirtualBox Guest Additions - provide host/guest integration (install only in VirtualBox guest)"
```
pyhalov: Please, remove (install only) comment - we can't avoid all user errors by moving documentation to package summaries.

components/sysutils/virtualbox/Makefile
```diff
@@ -67,7 +68,6 @@ VBOX_FS_MODULES= vboxfs
 
 VBOX_BINS= VBoxControl VBoxService VBoxClient
 VBOX_SBINS= vboxmslnk
-VBOX_OGLLIBS= VBoxOGL.so VBoxEGL.so
```
pyhalov, line 71: Perhaps, 11-fix-vbox_vendor_select.patch is no longer necessary, likely it's not needed as EGL libs has gone. The same stands for 12-disable-vbox-opengl.patch. Please, look them through.

PR #5376, opened 2019-11-02
samba: readd patch for avahi propagation.

chincheta0815:
Sorry, but this readded patch is still necessary...
Hope you can add that before release 10.2019


chincheta0815:
I close this pull request as there is one with a newer version.

PR #5385, opened 2019-11-06
arphic fonts have incorrect version

pyhalov:


alarcher:
Maybe we could cook some version like only keeping 2.x because 0.5.11 does not mean anything for this package?

pyhalov:
You are welcome ;) 
Or just add IPS human-version attribute.

PR #5401, opened 2019-11-22
Barman update

pyhalov:


pyhalov:
Testing:
Configured local barman backup using streaming replication and rsync. Restored from barman backups.

PR #5405, opened 2019-11-24
mozilla-nspr: update to 4.23; mozilla-nss: update to 3.47.1

AndWac:
According to https://abi-laboratory.pro/index.php?view=timeline&l=nspr 4.23 is 100% compatible with 4.21


AndWac:
According to https://abi-laboratory.pro/index.php?view=timeline&l=nss version 3.47.1 is 100% backwards compatible to 3.44 (it only adds 2 new functions).
I also included fixes for SPARC (https://www.illumos.org/issues/11925) here.

components/library/mozilla-nspr/Makefile
```diff
@@ -73,9 +69,9 @@ COMPONENT_TEST_ENV_CMD=
 COMPONENT_TEST_ARGS=
 COMPONENT_TEST_TARGETS=
 
-test:		$(TEST_32_and_64)
-
+# Build dependencies
 REQUIRED_PACKAGES += text/xmlto
 
 # Auto-generated dependencies
+REQUIRED_PACKAGES += SUNWcs
```
Toasterson, line 76: Did we not want to deprecate this?

SUNWcs is huge and pulls in system/library and everything else to a total of 1GiB dependencies.

AndWac: This part is auto-generated so there seems to be a dependency yet. I am not sure how to proceed...

Toasterson: hmm unless @Mno-hime has an idea how to take it out then leave it in. I thought Auto generated would ignore it. But that discussion was a long time ago. It won't hurt us right now.

Mno-hime: I don't even remember that discussion :). I always keep it. I doubt we can avoid it in reality?

components/library/mozilla-nss/Makefile
```diff
@@ -51,9 +51,9 @@ COMPONENT_TEST_ENV += $(COMPONENT_BUILD_ENV)
 COMPONENT_TEST_ENV += HOST=127.0.0.1
 COMPONENT_TEST_DIR = $(BUILD_DIR_$(BITS))/nss/tests
 
-include $(WS_TOP)/make-rules/prep.mk
-include $(WS_TOP)/make-rules/justmake.mk
-include $(WS_TOP)/make-rules/ips.mk
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
```
Mno-hime, line 56: Can we use `include $(WS_MAKE_RULES)/common.mk` and `BUILD_STYLE=justmake`?

AndWac: I have tried that, but with it the build failed for me. So I left most of the Makefile untouched...

components/library/mozilla-nspr/Makefile
Toasterson: Oh wow that must have been long ago then. Well in that case SUNWcs will be linked into every package and we will have to look into if we want to get rid of it like omnios has. Carry on then. LGTM

pyhalov: Not package, just Makefile. REQUIRED_PACKAGE = build dependencies + packages which used to resolve dependencies . It's not a hard-coded list of dependencies.

Mno-hime:
> I also included fixes for SPARC (https://www.illumos.org/issues/11925) here.

So, I have this right: Do those changes come from your build on SPARC or you just applied the patch from the bug tracker?

AndWac:
As I don't have a SPARC build environment yet I just applied the patch.

PR #5411, opened 2019-12-09
Update to tmux-3.0a

rapenne-s:


components/sysutils/tmux/Makefile
```diff
@@ -11,20 +11,21 @@
...
 # Copyright 2019, Michal Nowak
+# Copyright 2019, Solene Rapenne
 #
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		tmux
-COMPONENT_VERSION=	2.9
+COMPONENT_VERSION=	3.0
 COMPONENT_FMRI=		terminal/tmux
```
AndWac, line 26: There is a special treatment for version numbers pkg cannot cope with:
COMPONENT_VERSION=   3.0a
IPS_COMPONENT_VERSION= 3.0.1
(See openssl for another example)

components/sysutils/tmux/Makefile
```diff
@@ -11,20 +11,21 @@
...
 COMPONENT_SUMMARY=	tmux terminal multiplexer
 COMPONENT_CLASSIFICATION= Applications/System Utilities
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)a
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_HASH=	\
-	sha256:839d167a4517a6bffa6b6074e89a9a8630547b2dea2086f1fad15af12ab23b25
+	sha256:4ad1df28b4afa969e59c08061b45082fdc49ff512f30fc8e43217d7b0e5f8db9
 COMPONENT_ARCHIVE_URL= \
 	https://github.com/tmux/tmux/releases/download/$(COMPONENT_VERSION)a/$(COMPONENT_ARCHIVE)
 COMPONENT_LICENSE_FILE=	COPYING
```
AndWac, line 36: Remove the trailing "a" when you changed the COMPONENT_VERSION

components/sysutils/tmux/Makefile
```diff
@@ -11,20 +11,21 @@
 #
 # Copyright 2011, 2013, EveryCity Ltd. All rights reserved.
 # Copyright 2019, Michal Nowak
+# Copyright 2019, Solene Rapenne
 #
```
AndWac, line 16: Please try to adapt to the new style (see runtime/node12 as an example):
Insert

BUILD_BITS=    64

before importing shared-macros.mk

Later replace

with
TEST_TARGET=    $(NO_TESTS)

and remove

build:		$(BUILD_64)
install:	$(INSTALL_64)
test:		$(NO_TESTS)

AndWac (changes requested):

AndWac:
Ok, final question: Did you check whether the sample-manifest changed? 
gmake sample-manifest will create a new one, so you can check whether there are changes.
If so, you will need to adapt the .p5m file accordingly.
When everything is done please squash the commits into a single one.

rapenne-s:
I did check for changes in the manifest before and there are none

AndWac:
This is good, but you should check again after the switch to the new style. There might be some incompabilities. I don't expect any here, though, because the Makefile is quite simple.

rapenne-s:
I did compare the manifest after the new syntax change of course (and building has been tried too).

components/sysutils/tmux/Makefile
```diff
@@ -11,41 +11,40 @@
...
-include $(WS_MAKE_RULES)/prep.mk
-include $(WS_MAKE_RULES)/configure.mk
-include $(WS_MAKE_RULES)/ips.mk
+TEST_TARGET=	$(NO_TESTS)
+include $(WS_MAKE_RULES)/common.mk
 
 CFLAGS +=	-I/usr/include/ncurses
 
-CONFIGURE_BINDIR.64 = $(CONFIGURE_BINDIR.32)
+#CONFIGURE_BINDIR.64 = $(CONFIGURE_BINDIR.32)
```
Mno-hime, line 44: I think this should have been dropped when moved to the `BUILD_BITS=64` style.

AndWac:
Am 11.12.19 um 09:07 schrieb Michal Nowak:
>
> *@Mno-hime* commented on this pull request.
>
>
> In components/sysutils/tmux/Makefile
> <https://github.com/OpenIndiana/oi-userland/pull/5411#discussion_r356449427>:
>
> >
>   CFLAGS +=	-I/usr/include/ncurses
>
> -CONFIGURE_BINDIR.64 = $(CONFIGURE_BINDIR.32)
> +#CONFIGURE_BINDIR.64 = $(CONFIGURE_BINDIR.32)
>
> I think this should have been dropped when moved to the
> |BUILD_BITS=64| style.
>
Yes, you are right. Sorry, I missed it.

Mno-hime:
> Yes, you are right. Sorry, I missed it.

No problem, we will fix it when tmux gets a new release. Thanks for working on having this merged.

PR #5413, opened 2019-12-12
asterisk: better use option --with-postgres

grueni:


pyhalov:

PR #5418, opened 2019-12-18
update to samba 4.11.4

chincheta0815:


components/network/samba/Makefile
```diff
@@ -128,6 +128,12 @@ COMPONENT_POST_INSTALL_ACTION += \
 	(sed -f $(COMPONENT_DIR)/Solaris/smbconf.sed \
 		> $(PROTO_DIR)$(ETCDIR)/samba/smb.conf.default $(BUILD_DIR_64)/examples/smb.conf.default) 
 
+# samba needs gmake for testing, not waf.
+COMPONENT_TEST_CMD = gmake
```
Mno-hime: Better make it
```suggestion
COMPONENT_TEST_CMD = $(GMAKE)
```

chincheta0815: Just removed the test lines completely as there is no test implemented into the Makefile yet. I have some issues making that compile. Upstream is informed.

PR #5424, opened 2019-12-19
Add ChallengeResponseAuthentication to sshd_config

jlevon:
Let's make it clear you normally need to disable this too
for the PasswordAuthentication disabling to be meaningful.


pyhalov:
Please, bump COMPONENT_REVISION.

Mno-hime (approved):
The second part is from the upstream [sshd_config](https://github.com/openssh/openssh-portable/blob/master/sshd_config#L60).

PR #5425, opened 2019-12-19
meson: fix rpath handling

pyhalov:


pyhalov:
The issue is that meson deliberately spoils runpaths at build time and fixes them at install time. pkgsrc patches removed second part of this logic. This leads to strange runpaths...

Have to see how it works with gcc runtimes, when links are necessary (e.g.. some c++ projects).

pyhalov:
This is incorrect, with this patch I have invalid c++ programs (without RUNPATH).

pyhalov:
The issue is that meson adds obviously wrong runpaths to binaries... Likely, we should preserve these patches and fix this 'auto-runpath' mess.
https://github.com/mesonbuild/meson/issues/3882

pyhalov:
At least this version doesn't mess gcc runtime runpaths.

PR #5426, opened 2019-12-19
glib: update to 2.62

pyhalov:
Depends on #5425  amd #5423 .
Needs better testing.
This also changes python used by glib tools to python3.5.


Mno-hime:
I'll put it into my system to see how things play out but... should we rebuild package requiring glib2? There's a backward incompatibility in 2.59.1, see https://abi-laboratory.pro/index.php?view=timeline&l=glib.

pyhalov:
It seems, there isn't (if we did everything correctly and haven't messed up library defines ;)). 
The ABI tracker tells us about g_slice_debug_tree_statistics(), but it was present only in debug build - `elfdump /usr/lib/libglib-2.0.so|grep g_slice_debug_tree_statistics`  shows nothing, also look at https://gitlab.gnome.org/GNOME/glib/issues/1607 .
However, there could be issue that older apps fail to build with new glib, and this should be checked.

Mno-hime:
I see, we might get away with it :). So far looks good.

pyhalov:
@Mno-hime Please, push your changes again, I've accidentally pushed more than I wanted.

Mno-hime:
> @Mno-hime Please, push your changes again, I've accidentally pushed more than I wanted.

Just pushed 9355702bcb0cd55f2434ffe8092f644ac5a195a8.

pyhalov:
Not sure, why test output differs.

PR #5427, opened 2019-12-19
update nodejs-8, nodejs-10 and nodejs-12

AndWac:


components/runtime/node10/nodejs.p5m
```diff
@@ -99,26 +100,22 @@ link path=usr/share/man/man1/sshpk-conv.1 target=../../../node/10/lib/node_modul
...
+link path=usr/share/man/man7/config.7 target=../../../node/10/lib/node_modules/npm/man/man7/config.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/developers.7 target=../../../node/10/lib/node_modules/npm/man/man7/developers.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/disputes.7 target=../../../node/10/lib/node_modules/npm/man/man7/disputes.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/orgs.7 target=../../../node/10/lib/node_modules/npm/man/man7/orgs.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/registry.7 target=../../../node/10/lib/node_modules/npm/man/man7/registry.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/removal.7 target=../../../node/10/lib/node_modules/npm/man/man7/removal.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/scope.7 target=../../../node/10/lib/node_modules/npm/man/man7/scope.7 mediator=nodejs mediator-version=10
+link path=usr/share/man/man7/scripts.7 target=../../../node/10/lib/node_modules/npm/man/man7/scripts.7 mediator=nodejs mediator-version=10
```
Mno-hime: Is this configurable, so we are back to `removing-npm.7` instead of just `removing.7`?

Those man pages names are so generic I wonder if all node packages install without a conflict with already installed packages? We have some sort of check during publication process for file conflicts in packages, but I'd rather catch it earlier.

AndWac: Not to my knowledge. I will investigate next week but I think we need to do it ourselves.

AndWac: Does it look better now? I have prefixed the relevant links with "npm-" again.

AndWac: Can I merge this or are there any objections?

Mno-hime: Way better, go ahead and merge.

PR #5428, opened 2019-12-20
Update Vim to 8.2.0024

Mno-hime:
Release notes: https://www.vim.org/vim-8.2-released.php


components/editor/vim/vim.p5m
```diff
@@ -148,1694 +148,1713 @@ link path=usr/share/man/ru.UTF-8/man1/rvim.1 target=vim.1
...
-file path=usr/share/vim/vim81/autoload/clojurecomplete.vim
-file path=usr/share/vim/vim81/autoload/context.vim
-file path=usr/share/vim/vim81/autoload/contextcomplete.vim
-file path=usr/share/vim/vim81/autoload/csscomplete.vim
-file path=usr/share/vim/vim81/autoload/decada.vim
-file path=usr/share/vim/vim81/autoload/dist/ft.vim
-file path=usr/share/vim/vim81/autoload/getscript.vim
-file path=usr/share/vim/vim81/autoload/gnat.vim
-file path=usr/share/vim/vim81/autoload/gzip.vim
-file path=usr/share/vim/vim81/autoload/haskellcomplete.vim
```
pyhalov, line 165: Should we use some var for 81/82/etc?

Mno-hime: I'd rather not to. Having variable in vim.p5m makes it a mess for comparison with `vimdiff` when new new stuff gets added to sample-manifest, because the variable is not in the sample-manifest. (I understand that it's a messy in version changes like this one, though still far easier for me to do the update right.)

PR #5439, opened 2019-12-30
libgpg-error: update to 1.36

pyhalov:
Seahorse fix number 1.
ABI-compatible update.

### DIFF
--- a/components/library/libgpg-error/Makefile
+++ b/components/library/libgpg-error/Makefile
@@ -21,36 +21,29 @@
 # Copyright 2013 Colin Ellis. All rights reserved.
 # Copyright 2011 Audioboo Ltd. All rights reserved.
 #
+BUILD_BITS = 32_and_64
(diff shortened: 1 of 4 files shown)


Mno-hime:
@pyhalov Without https://github.com/OpenIndiana/oi-userland/pull/5278/commits/a3a79fd467449a4c023be0e7be4ede02b70899d7#diff-79c447b5596720ee03e48e6edc989235 the `gpg-error-config-test.sh` test fails. Do we want this patch? (Otherwise this PR is the same as https://github.com/OpenIndiana/oi-userland/pull/5278/commits/a3a79fd467449a4c023be0e7be4ede02b70899d7.)

pyhalov:
Likely

Mno-hime:
The fix is in https://github.com/OpenIndiana/oi-userland/pull/5278.

PR #5440, opened 2019-12-30
gpgme: update to 1.13.1

pyhalov:
ABI-incompatible update

PR #5441, opened 2019-12-30
seahorse: fix build with new gpg

pyhalov:


pyhalov:
Seahorse fix number 3
Requires #5440

Mno-hime (approved):

PR #5442, opened 2019-12-30
Rebuild components due to gpgme update

pyhalov:
Merge this and #5418 after #5440

### DIFF
--- a/components/editor/geany-plugins/Makefile
+++ b/components/editor/geany-plugins/Makefile
@@ -17,6 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= geany-plugins
 COMPONENT_VERSION= 1.34
+COMPONENT_REVISION= 1
 COMPONENT_CLASSIFICATION=System/Text Tools
 COMPONENT_FMRI=editor/geany/plugins
 COMPONENT_SUMMARY= Plugins for Geany editor
--- a/components/library/gmime/Makefile
+++ b/components/library/gmime/Makefile
@@ -17,7 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= gmime
 COMPONENT_VERSION= 2.5.10
-COMPONENT_REVISION= 1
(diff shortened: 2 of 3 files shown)


Mno-hime:
It seems that Samba also depends on GPGME?

Mno-hime (approved):

gc-hpc-admin:
Sorry, I do not know how to ask elsewhere:
Is there a reason why oi-userland does not provide a "prep-git.mk" as solaris-userland does (see: https://github.com/oracle/solaris-userland/blob/master/make-rules/prep-git.mk)

It would be quite helpful to have it since downloading the tarball from git very often does somenasty things with the directory name etc.

Simply clone the repo and checking out would help a lot.
If there is no reason against having prep-git.mk I would wok on that and create a pull request.

chincheta0815:
Sorry, I do not know how to ask elsewhere:
Is there a reason why oi-userland does not provide a "prep-git.mk" as solaris-userland does (see: https://github.com/oracle/solaris-userland/blob/master/make-rules/prep-git.mk)

It would be quite helpful to have it since downloading the tarball from git very often does somenasty things with the directory name etc.

Simply clone the repo and checking out would help a lot.
If there is no reason against having prep-git.mk I would work on that and create a pull request.

Mno-hime:
> Sorry, I do not know how to ask elsewhere:

The oi-dev mailing list seems to be a right place for this type of conversation :).

> Is there a reason why oi-userland does not provide a "prep-git.mk" as solaris-userland does (see: https://github.com/oracle/solaris-userland/blob/master/make-rules/prep-git.mk)
> ...
> If there is no reason against having prep-git.mk I would work on that and create a pull request.

Probably no reason other than we didn't fee the need to bring it over. Feel free to go ahead and port it.

We use a similar technique in some components which are just a shell around projects we host on GitHub and don't make proper tarball releases, e.g. https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/openindiana/pkg/Makefile#L20.

PR #5451, opened 2020-01-01
Update ISC DHCP to 4.4.2

Mno-hime:
**Release notes**: https://downloads.isc.org/isc/dhcp/4.4.2/dhcp-4.4.2-RELNOTES

Added the `keama` tool for Kea migration.

**Testing**: IPv4 DHCP service starts fine.


AndWac:
The release notes state that there is a new tool named Keama. Doesn't it have an impact on the sample-manifest and the p5m file?

Mno-hime:
I think it needs to be built separately from the main code.

It's not yet clear if it stays in the source tarball for the time of the release or moves to Kea tarball, because the tool is actually more useful for Kea users performing the DHCP->Kea migration, than for DHCP users.

Lets see when the stable 4.4.2 release gets out.

Mno-hime:
The `keama` tool was kept in the final release, I added it to the Makefile.

PR #5456, opened 2020-01-02
Update Hexchat to 2.14

Mno-hime:
Disabled plugins because they fail to build like this:
```
[73/86] Linking target plugins/checksum/checksum.so.
FAILED: plugins/checksum/checksum.so 
/usr/gcc/6/bin/gcc  -o plugins/checksum/checksum.so 'plugins/checksum/c016833@@checksum@sha/checksum.c.o' -z nodefs -shared -fPIC -Wl,--start-group -Wl,-soname,checksum.so -Wl,-z,now -m32 -O3 -m32 /usr/lib/libgio-2.0.so /usr/lib/libgobject-2.0.so /usr/lib/libglib-2.0.so -Wl,--end-group
Undefined			first referenced
 symbol  			    in file
__stack_chk_fail_local              plugins/checksum/c016833@@checksum@sha/checksum.c.o  (symbol scope specifies local binding)
ld: fatal: symbol referencing errors. No output written to plugins/checksum/checksum.so
```

### DIFF
--- a/components/desktop/hexchat/Makefile
+++ b/components/desktop/hexchat/Makefile
@@ -10,47 +10,42 @@
 
 #
 # Copyright 2016 Marcel Telka <marcel@telka.sk>
+# Copyright 2020 Michal Nowak
 #
+
+BUILD_BITS=		32
+BUILD_STYLE=		meson
 
 include ../../../make-rules/shared-macros.mk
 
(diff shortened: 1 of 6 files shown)


pyhalov:
something uses -fstack-protector, but later links without -lssp

PR #5457, opened 2020-01-03
Update GNU grep to 3.4

Mno-hime:
**Release notes**: https://lists.gnu.org/archive/html/info-gnu/2020-01/msg00001.html

libsigsegv detection won't work unless lib patch fixed (to what it was in GNU grep 3.3).

**Testing**: Test suite passed.

### DIFF
--- a/components/text/ggrep/Makefile
+++ b/components/text/ggrep/Makefile
@@ -20,14 +20,15 @@
 
 #
 # Copyright (c) 2011, 2017, Oracle and/or its affiliates. All rights reserved.
-# Copyright (c) 2018, Michal Nowak
+# Copyright (c) 2020, Michal Nowak
 #
+
+BUILD_BITS=		64
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		grep
-COMPONENT_VERSION=	3.3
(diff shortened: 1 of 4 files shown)


components/text/ggrep/Makefile
```diff
@@ -62,22 +60,15 @@ unexport SHELLOPTS
...
-test:		$(TEST_64)
+	'-n ' \
+	'-e "/TOTAL/p" ' \
+	'-e "/SKIP/p" ' \
+	'-e "/PASS/p" ' \
+	'-e "/FAIL/p" ' \
+	'-e "/ERROR/p" '
 
+# Since version 3.4, GNU grep is needed for build.
+REQUIRED_PACKAGES += text/gnu-grep
```
pyhalov, line 71: Very unfortunate... I think about ABI incompatibilities in ggrep dependencies...

Mno-hime: Can we somehow check that? I find it hard to imagine what might break, it's just a binary...?

pyhalov: I think about some ABI-incompatible changes in libsigsegv or pcre. Anyway, it's not a reason for blocking PR.

PR #5466, opened 2020-01-08
Update nvidia driver to 340.108

Mno-hime:
**Release notes**: https://www.nvidia.com/Download/driverResults.aspx/156165/en-us

**Testing**: None (I don't have any nvidia device).


pyhalov:
Seems to be working here.

PR #5477, opened 2020-01-14
Remove Python2 illumos-gate modules

pyhalov:


pyhalov:
It seems to be the last step to python3-only minimal installation.

Mno-hime:
Also drop `REQUIRED_PACKAGES += runtime/python-27`?

PR #5479, opened 2020-01-14
add prep-git.mk for pulling source from git repositories.

chincheta0815:
added prep-git.mk from github.com/oracle/solaris-userland.git and adapted prep.mk as done in solaris-userland.

### DIFF
--- a/make-rules/prep.mk
+++ b/make-rules/prep.mk
@@ -25,6 +25,7 @@
 
 include $(WS_MAKE_RULES)/prep-download.mk
 include $(WS_MAKE_RULES)/prep-hg.mk
+include $(WS_MAKE_RULES)/prep-git.mk
 include $(WS_MAKE_RULES)/prep-svn.mk
 include $(WS_MAKE_RULES)/prep-unpack.mk
 include $(WS_MAKE_RULES)/prep-patch.mk
--- a/make-rules/prep-git.mk
+++ b/make-rules/prep-git.mk
@@ -1,0 +1,122 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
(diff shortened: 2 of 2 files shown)


chincheta0815:
@pyhalov and @Mno-hime Would you be so kind and review the lines? I fell over some lines in prep.mk , e.g. $(eval $(call eval-git-rules)) that needed to be added to make the prep-git.mk work.
It calls a "define eval-git-rules" in the prep-git.mk.

I also added the ones defined in the prep-*.mk files that were not mentioned in prep.mk.
Hope that was okay. 

In general that does not seem consistent to me as some prep-*.mk do not have a call-... function defined...

Thanks for your help.

make-rules/prep.mk
```diff
@@ -25,11 +25,16 @@
 
 include $(WS_MAKE_RULES)/prep-download.mk
 include $(WS_MAKE_RULES)/prep-hg.mk
+include $(WS_MAKE_RULES)/prep-git.mk
 include $(WS_MAKE_RULES)/prep-svn.mk
 include $(WS_MAKE_RULES)/prep-unpack.mk
 include $(WS_MAKE_RULES)/prep-patch.mk
 
+$(eval $(call eval-download-rules))
```
pyhalov: Why do we need these? It's already called in prep-download.mk.
Perhaps, missing some solaris-userland commit?

chincheta0815: Okay. Then I think I understand...
BUT then there should be a change in prep-hg-rules.mk calling the respective function there directly as it is done in prep-download.mk, or not?

make-rules/prep.mk
```diff
@@ -25,11 +25,12 @@
 
 include $(WS_MAKE_RULES)/prep-download.mk
 include $(WS_MAKE_RULES)/prep-hg.mk
+include $(WS_MAKE_RULES)/prep-git.mk
 include $(WS_MAKE_RULES)/prep-svn.mk
 include $(WS_MAKE_RULES)/prep-unpack.mk
 include $(WS_MAKE_RULES)/prep-patch.mk
 
-$(eval $(call eval-hg-rules))
+$(eval $(call eval-git-rules))
```
pyhalov: Do we still need this?

pyhalov:
Try to build some component using this mk file. This is the only way to find out if it's working ;)

chincheta0815:
Okay. This turned out to work with a component I am working on.
The eval-hg-rules can then also be remove after adapting prep-hg-mk I think...
I will do that in another pr.

pyhalov (approved):

chincheta0815:
Sorry. Stupid me.

chincheta0815:
@pyhalov @Mno-hime If everything is okay with this PR would you mind merging it. After that I want to PR the "cleanup" for prep-hg.mk.

PR #5488, opened 2020-01-20
[WIP] lms: started to work.

chincheta0815:
Right now I decided to start creating a package for the Logitech Media Server. This media server is implemented in perl, but uses some patched standard modules that might conflict with the ones installed in the system. This has to be respected when building it.

For building it two git repos are needed: the base code (lms-base) and the perl modules (lms-vendor). The stuff in lms-vendor has to be build and then being copied into lms-base. Finally there will be one directory containing all stuff needed.

The tricky/important thing is to put it thouroughly into the system. 
My best idea so for is installing it in e.g. something like /opt/lms because of the special character of all stuff needed.

Could anybody give some thoughts on how to make a package out of it?

### DIFF
--- a/components/multimedia/lms/Makefile
+++ b/components/multimedia/lms/Makefile
@@ -1,0 +1,78 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2011, 2015, Oracle and/or its affiliates. All rights reserved.
+# Copyright 2020 Rouven Weiler
+#
+
+BUILD_BITS=			32
+BUILD_STYLE=			archive
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		lms
+COMPONENT_FMRI =		service/multimedia/
+COMPONENT_VERSION =		7.9
+COMPONENT_PROJECT_URL =		https://github.org/Logitech/slimserver.git
+COMPONENT_SUMMARY =		Logitech Media Server
+COMPONENT_LICENSE =		GPLv2
+COMPONENT_LICENSE_FILE =	$(COMPONENT_NAME).license
+COMPONENT_CLASSIFICATION =	System/File System
+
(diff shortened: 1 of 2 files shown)


pyhalov:
1) I suppose that common pattern is to put packages under /usr, like /usr/squid, /usr/apache2/2.4. and so on. 
2) Likely you have to base you component on justmake.mk, but insert custom make and install rules, look at components/runtime/openjdk-8/Makefile for example.

chincheta0815:
Okay thanks. Then I will stick to the packages file scheme.

So far I tried the archive.mk which seems to suit quite well as there is no Makefile or anything...

Another thing I have to ask: How would you handle the perl bindings? One build which simply uses one perl? Or one base package (lms-base) that needs one out of lms-vendor-perl522 or lms-vendor-perl524? As e.g. emacs is built... Is there any mechanism that simply runs the build for all existing perls?

pyhalov:
There's such logic for makemaker, but not for usual packages. I'd stick to system perl by default (now 5.22).

chincheta0815:
Okay. Thanks. I think this will help for the first steps.

components/multimedia/lms/Makefile
```diff
@@ -0,0 +1,78 @@
...
+COMPONENT_SUMMARY =		Logitech Media Server
+COMPONENT_LICENSE =		GPLv2
+COMPONENT_LICENSE_FILE =	$(COMPONENT_NAME).license
+COMPONENT_CLASSIFICATION =	System/File System
+
+COMPONENT_SRC =			$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE =		$(COMPONENT_SRC).tar.gz
+GIT_HASH =			sha256:520c0d1633d707a03e3e7d9fd741b7019592660b648f697e0ca1083053edec9f
+GIT_REPO =			https://github.com/Logitech/slimserver.git
+GIT_BRANCH =			public/7.9
```
Mno-hime, line 45: @chincheta0815 Can you discuss why should we use their Git repo instead of [tarball releases](http://downloads.slimdevices.com/LogitechMediaServer_v7.9.2/)?

chincheta0815: LMS is releasing code in an (in my opinion) uncommon way. The tarballs represent a version freeze after one year, e.g.: 

- 7.9.2 is right now end-of-development and end-of-support release. All the plugins that are existing and addons are not updated for that version too. In terms of samba this would represent the 4.9 release series.


- The 7.9.3 branch is the right now stable and bugfix release. All addons are updated/bugfixed and kept working against that version. I personally have very good experience with that kind of release. In terms of samba this is the 4.11 release series.

Speaking about the tarball releases there exists just a tarball for the 7.9.2 version. As a side effect it might be confusing for a maintainer not dealing each day with lms which one of these to use (noCPAN, CPAN, ...). For the desired version 7.9.3 there is no taball existing. The 7.9.3 version resides in the github.com branch named public/7.9. So I decided to use that branch form the repo.

In addition, the needed vendor/perl modules are only available in a repository on github. So it seemed just a good scheme to use the git repo(s). And, very important, there is not tarball for all the vendor/perl modules that is valid and working for version 7.9.2. So for the vendor/perl modules one has to use the repo.

Using the repo tarballs was something I decided not to do. The tarballs one gets by downloading them from github extract to a directory called slimserver-... So I ended up with "lms-base/slimserver/public/7.9" or "lms-vendor/slimserver-vendor/oi-userland/7.9" which would have caused lines of code in the Makefile that I simply could avoid by adding prep-git.mk and use that.

I hope I could explain my thoughts well and can stick to my code ;o)

PR #5489, opened 2020-01-20
open-vm-tools: fix build

pyhalov:


Mno-hime:
Forgot to add the patch. Sorry about that.

PR #5493, opened 2020-01-21
Twisted deps

pyhalov:


pyhalov:
Twisted-web2 doesn't work with latest twisted and seems to be abandoned.
Coherence is also seems to be more dead than alive.

components/meta-packages/history/history
```diff
@@ -880,6 +880,9 @@ library/python/python-gnome-extras-27@2.25.3,5.11-2018.0.0.5
 library/python/python-gnome-extras@2.25.3,5.11-2018.0.0.5
 library/python-2/argparse@1.2.1,5.11-2016.0.1.0
 library/python-2/argparse-26@1.2.1,5.11-2016.0.1.0
+library/python-2/coherence-26@0.6.6.2-2016.0.1.2
```
Mno-hime, line 883: The latest one I see is `FMRI: pkg://openindiana.org/library/python-2/coherence-26@0.6.6.2-2016.0.1.2:20180211T133129Z`, should this be **2016.0.1.3**?

components/meta-packages/history/history
```diff
@@ -912,13 +915,20 @@ library/python-2/python-notify-24@0.5.11,5.11-2015.0.2.0
 library/python-2/python-pigment-26@0.5.11,5.11-2015.0.2.0
 library/python-2/python-twisted-24@0.5.11,5.11-2015.0.2.0
 library/python-2/python-twisted-web2-24@0.5.11,5.11-2015.0.2.0
+library/python-2/python-twisted-web2-26@8.1.0,5.11-2016.0.1.0
```
Mno-hime, line 918: **2016.0.1.1**?

Mno-hime (commented):
Except for those two history changes, LGTM (builts and installs fine).

components/meta-packages/history/history
pyhalov: No, as we just move definition from coherence history file to global history file (the package is already obsolete).

components/meta-packages/history/history
pyhalov: The same here.

PR #5495, opened 2020-01-21
avahi: update to 0.7

pyhalov:
Upstream avahi 0.7 doesn't support Python 3, but HEAD has some commits to fix this. I also added more patches, which need testing.


components/network/avahi/avahi.p5m
```diff
@@ -35,7 +36,7 @@ depend type=conditional fmri=pkg:/system/network/avahi/gui \
     predicate=pkg:/x11/server/xserver-common
 
 file path=etc/dbus-1/system.d/avahi-dbus.conf
-file path=usr/bin/$(MACH64)/avahi-bookmarks
+#file path=usr/bin/$(MACH64)/avahi-bookmarks
```
pyhalov, line 39: Depends on twisted, needs more work. Was not shipped in OpenSolaris.

pyhalov:
@tsoome I suppose, you can be interested in testing this.

pyhalov:
Tested that avahi-browse detects samba start/stop

pyhalov:
If I change kDNSServiceType_A6 to kDNSServiceType_AAAA in 05-resolve-service.patch, avahi-browse starts prefering interface-local ipv6 addresses to ipv4 addresses. Likely it's not desired...

Mno-hime (approved):
Was able to discover some services on my local network and spawn SSH connection to them.

PR #5504, opened 2020-01-25
Add minidlna 1.2.1

pyhalov:
Based on https://github.com/omniosorg/omnios-extra/tree/master/build/minidlna


components/encumbered/minidlna/patches/03-monitor.c.patch
```diff
@@ -0,0 +1,41 @@
+diff -ur minidlna-1.2.1/monitor.c minidlna-1.2.1-mod/monitor.c
```
pyhalov, line 1: @citrus-it - it seems this patch is more sane, as is_dir() and is_reg() are defined on struct dirent*, not char* and always return -1 on illumos.

PR #5508, opened 2020-01-27
11224 nightly and bldenv should be usable directly

pyhalov:


pyhalov:
@Mno-hime feel free to merge it if it looks good to you

Mno-hime:
Hmm, I didn't know we had such a workaround.

PR #5517, opened 2020-01-29
ccsm: update to 0.8.16, switch to Python 3.5

pyhalov:


pyhalov:
Depends on #5515

Mno-hime:
Can we drop `ccsm-01-desktop.patch`? I rather think user wants the tool to be accessible from the _Preferences_ menu, looks to me like inherited from OpenSolaris.

Mno-hime:
Other than that looks good.

PR #5524, opened 2020-01-30
emacs: disable aslr

alarcher:


components/editor/emacs/Makefile
```diff
@@ -51,7 +51,7 @@ INSTALL_32 = $(VARIANTS:%=$(BUILD_DIR)/$(MACH32)-%/.installed)
 # build with the distribution preferred libjpeg implementation
 CFLAGS   += $(JPEG_CPPFLAGS) $(JPEG_CFLAGS)
 CXXFLAGS += $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS)
-LDFLAGS  += $(JPEG_LDFLAGS)
+LDFLAGS  += $(JPEG_LDFLAGS) $(LD_Z_ASLR)
```
pyhalov: I'm a bit confused.. We do not enable it explicitly. Also I read that 
```
-z aslr[=state]

           Specify whether the executable's address space should be randomized
           on execution.  If state is "enabled" randomization will always
           occur when this executable is run (regardless of inherited
           settings).  If state is "disabled" randomization will never occur
           when this executable is run.  If state is omitted, ASLR is enabled.

           An executable that should simply use the settings inherited from
           its environment should not use this flag at all.
```
Where aslr environment is coming from?
If it's a common thing, why just emacs is affected?

alarcher: @pyhalov ASLR is disabled on other platforms for emacs due to spurious segfaults, I do not know more... Probably related to the memory model in Lisp...

pyhalov: I mean, we don't explicitly enable aslr. So something likely turns it on.  This can be zone config or service description, IIRC. So I'm interested, how this issue can be reproduced in default settings.

AndWac (approved):

components/editor/emacs/Makefile
alarcher: Not setting it will use the system default, therefore it should be enabled explicitly in LD_OPTIONS. According to richlowe LD_EXEC_OPTIONS is not supported on illumos so our variable in shared-macros.mk is bogus.
You can inspect the current setting for your shell with psecflags $$
Default setting is to have aslr active.

PR #5527, opened 2020-02-02
Update libtasn1 to 4.16.0

Mno-hime:
**Release notes**: https://lists.gnu.org/archive/html/info-gnu/2020-02/msg00002.html

**ABI Tracker**: [not there yet](https://abi-laboratory.pro/index.php?view=timeline&l=libtasn1)

**Testing**
- test suite passed
- `gnutls-3`, and `p11-kit` build fine, their test suites are fine too


AndWac (commented):
API/ABI tracker has 4.16.0 already and it says it has only added 1 symbol.
So I think we can re-open and integrate this.

pyhalov:
One should check that 
a) headers don't change after update (I mean, 64-bit headers don't override 32-bit)
b) that dependencies still build fine.
Once it's done, let's merge it.

AndWac:
I don't understand a): There is only one header file in the package (as it was before).
For b): Mno-hime already wrote that he built gnutls-3 and p11-kit with it and tested them fine. The only missing dependant package seems to be rdesktop.

alarcher:
ABI compatible: https://abi-laboratory.pro/index.php?view=timeline&l=libtasn1

PR #5124, opened 2019-07-14
Add Wine 4.17 (experimental)

oicontrib:
Hi all.
I am using wine on OpenIndiana for some time and would like to see it integrated into oi-userland.
Please review the request.
I am ready to do any amendments if needed to meet oi-userland requirements.
Thanks.

PR #5533, opened 2020-02-04
Introduce libgloox at version 1.0.23

jimklimov:
Following up from FOSDEM challenge to try adding support for 0ad game to OI.

The gloox project is one of missing required dependencies, it provides Jabber/XMPP messaging capabilities.

There were issues with `make test` hanging for GnuTLS (with our `library/gnutls-3`) which seem to be due to some mismatch of the priorities string set in `src/tlsgnutlsclient.cpp` : 
````
#if GNUTLS_VERSION_NUMBER >= 0x020600
    int ret = gnutls_priority_set_direct( *m_session, "SECURE128:+PFS:+COMP-ALL:+VERS-TLS-ALL:-VERS-SSL3.0:-VERS-TLS1.3:+SIGN-ALL:+CURVE-ALL", 0 );
````
...vs what our library build supports/expects, so the test could not make the handshake.

I am no expert in this bit so for now I disabled the GnuTLS-aware builds in the recipe.

Another issue is with `gmake test` failing for "ConnectionBOSH" and "ConnectionTCPServer" steps; but I did not do anything further about this:
````
ClientBase: OK
ConnectionBOSH: 1 test(s) failed
test 'listen' failed
connect() returned: 7
test 'connect' failed
connect() returned: 9
test 'accept' failed
connect() returned: 18
ConnectionTCPServer: 3 test(s) failed
DataForm: OK
````

Beside that, gcc also reports a fair share of warnings about type casting, "old-style coding", format strings (IIRC), etc. that I did not even begin to address.

So in short, here is a code drop that would hopefully help that challenge move forward, but I have no idea whether it is functional. We currently have no consumers for it in the gate, and hopefully 0ad team's progress on trying to integrate the game would find and fix the issues. Can't promise I'd have time soon for another pass at this :\ 

I'll try to contact the library's developers though, to pass the patches. Appears they are not developing on github (there are many "forks" of various obsoletion though), and there's only a read-only SVN for downloads referenced from https://camaya.net/gloox/download/ page, so not sure how far emailing would go.


alarcher:
Could you use BUILD_BITS and common.mk in this makefile and subsequent recipes? Also maybe we do not need 32-bit version of this library since we do not have any backward compatibility to assure. @pyhalov What do you think?

pyhalov:
Not sure about if wee need it at all. As always, there's question of support.
I don't understand why bothering porting games to OI, when one can better help with #5124 :)
But if someone wants to support it, why not.

jimklimov:
For one, hopefully it makes the platform more attractive to more users, and in this case also developers that can come and tinker and learn about OI, and perhaps eventually help improve our landscape.

jimklimov:
Patched the codebase recipes a bit more: calling `$(MAKE)` instead of `make` "fixed" the tests so that the build succeeds. But some specific tests are logged as failed. Go figure :\ Author was informed.

````
ClientBase: OK
ConnectionBOSH: 1 test(s) failed
test 'listen' failed
connect() returned: 7
test 'connect' failed
connect() returned: 9
test 'accept' failed
connect() returned: 18
ConnectionTCPServer: 3 test(s) failed
DataForm: OK
````

alarcher:
> For one, hopefully it makes the platform more attractive to more users, and in this case also developers that can come and tinker and learn about OI, and perhaps eventually help improve our landscape.

To attract users we should first improve the support of new hardware... updating nvidia and drm code could be a priority.

PR #5534, opened 2020-02-04
Introduce (lib)enet version 1.3.14

jimklimov:
Another part of dependencies for the FOSDEM challenge of building 0ad in OI (see also #5533).


components/library/enet/Makefile
```diff
@@ -0,0 +1,53 @@
...
+COMPONENT_NAME=		enet
+COMPONENT_VERSION=	1.3.14
+COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.gz
+COMPONENT_FMRI=		library/security/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Libraries
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	http://enet.bespin.org
+# Development is on GitHub, but releases are not :(
+#COMPONENT_PROJECT_URL=	https://github.com/lsalzman/enet
+#COMPONENT_ARCHIVE_URL=	https://github.com/lsalzman/$(COMPONENT_NAME)/releases/download/$(COMPONENT_NAME)-$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
```
Mno-hime: I don't think this fact is worth documenting.

components/library/enet/Makefile
```diff
@@ -0,0 +1,53 @@
...
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:98f6f57aab0a424469619ed3047728f0d3901ce8f0dea919c11e7966d807e870
+COMPONENT_LICENSE=	MIT
+COMPONENT_LICENSE_FILE= LICENSE
+COMPONENT_DESCRIPTION=	ENet's purpose is to provide a relatively thin, simple and robust network communication layer on top of UDP (User Datagram Protocol).The primary feature it provides is optional reliable, in-order delivery of packets. ENet omits certain higher level networking features such as authentication, lobbying, server discovery, encryption, or other similar tasks that are particularly application specific so that the library remains flexible, portable, and easily embeddable.
+COMPONENT_SUMMARY=	ENet is a relatively thin, simple and robust network communication layer on top of UDP
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/configure.mk
+include $(WS_MAKE_RULES)/ips.mk
```
Mno-hime: Make this `common.mk` style et al.

components/library/enet/Makefile
```diff
@@ -0,0 +1,53 @@
...
+        $(RM) config.h )
+
+COMPONENT_PRE_CONFIGURE_ACTION = \
+        ($(CLONEY) $(SOURCE_DIR) $(@D))
+
+CONFIGURE_OPTIONS += --disable-static
+
+build: $(BUILD_32_and_64)
+
+install: $(INSTALL_32_and_64)
```
Mno-hime: What about test? Are they present? Do they work?

Mno-hime (changes requested):
> Another part of dependencies for the FOSDEM challenge of building 0ad in OI.

OK, but blocking before 0 A.D. PR is submitted, I don't think we are interested in libenet itself.

components/library/enet/Makefile
jimklimov: Done

components/library/enet/Makefile
jimklimov: I added the test target to our Makefile, but it seems there are none defined in the original codebase so it is a no-op.

components/library/enet/Makefile
jimklimov: Well... in case this component does get picked up, and in case future releases would go through Github, then maintainers of this recipe would have an easier life looking for current sources :)

jimklimov:
elsewhere > Could you use BUILD_BITS and common.mk in this makefile and subsequent recipes? Also maybe we do not need 32-bit version of this library since we do not have any backward compatibility to assure.

Done here too :)

components/library/enet/Makefile
Mno-hime: If there are no tests then add `TEST_TARGET=            $(NO_TESTS)` before `common.mk` include.

components/library/enet/Makefile
Mno-hime: It was trivial for me to find this information on their site. We have several such components with development site/download site dichotomy and we don't document this. Drop it, please.

components/library/enet/Makefile
jimklimov: ack, done

components/library/enet/Makefile
jimklimov: ack, done

components/library/enet/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,41 @@
...
+file path=usr/include/enet/utility.h
+file path=usr/include/enet/win32.h
+link path=usr/lib/$(MACH64)/libenet.so target=libenet.so.7.0.2
+link path=usr/lib/$(MACH64)/libenet.so.7 target=libenet.so.7.0.2
+file path=usr/lib/$(MACH64)/libenet.so.7.0.2
+file path=usr/lib/$(MACH64)/pkgconfig/libenet.pc
+#link path=usr/lib/libenet.so target=libenet.so.7.0.2
+#link path=usr/lib/libenet.so.7 target=libenet.so.7.0.2
+#file path=usr/lib/libenet.so.7.0.2
+#file path=usr/lib/pkgconfig/libenet.pc
```
Mno-hime, line 41: How come that 32-bit library gets build with `BUILD_BITS=64`? Perhaps you just need to rebuild sample manifest and update enet.p5m accordingly?

PR #5535, opened 2020-02-06
Revert "Add prep-git.mk for pulling source from git repositories"

pyhalov:
Breaks components using GIT_REPO and prep.mk.
This reverts commit 04bc6857ee9c7a1540f8e6b054a7e3e5cf83045f.


pyhalov:
@chincheta0815

chincheta0815:
Could you tell me how it breaks those components? I have not seen issues so far.

chincheta0815:
@pyhalov Do you have an error message or similar?

pyhalov:
The issue is that prep-git.mk misinterprets GIT_REPO variable, detects that it points to github and tries to fetch master branch of the repo as git archive. Given that some GIT_REPOs are set in non-suitable forms, it tries to fetch, for example, no-existent https://github.com/OpenIndiana/slim_source.git/archive/oi/hipster.tar.gz and fails.
So, if we want to fix prep-git.mk, we should check what variables are already used in oi-userland and rename them here or there.

pyhalov:
Try, for example, build  slim_source.

chincheta0815:
> So, if we want to fix prep-git.mk, we should check what variables are already used in oi-userland and rename them here or there.

When I understand you correctly: This simple rename (here or there) should be okay to include the git-prep.mk again. Some things concerned with github/gitlab/etc. would be nice but not compulsory in the first place. Is that correct?

If yes then you would check for the renaming stuff (here or there).

pyhalov:
I think, yes.

PR #5542, opened 2020-02-10
[WIP] Logitech Media Server.

chincheta0815:
Hej there,

I started add the Logitech Media Server as a package. The reason is that zfs and a music library fit very well. LMS is also very modular and has great features. It come for various distors out of the box, but there is no one for Illumos/OpenIndiana.

Nevertheless, this thing is quite special:
- it uses ffmpeg. Thus I want to add it to "encumbered".
- it uses a lot of "special" stuff, e.g. static libs only during build, ffmpeg with some special compile flags, patched components
- the main program is totally written in perl with its own modules.
- maybe some other unusual stuff.
- comes from two git repos hosted by Logitech on github..

For building it one needs a base package and a vendor packages. The vendor packages contains all  things that need to be compiled (e.g. for perl 5.22) and then copied into the base package.

So far I made a plain Makefile, etc.
It gets all sources from git, gets one small patch, tree is cleaned up, gets compiled, compiled modules get copied over and all is installed into the proto dir. Servie manifests etc. for starting it exist, but I have not uploaded them yet. So far it shall be installed under "/opt/lms-VERSION" due to its special and self-containing nature.

I hope in a first place "accept" the new package.
In a second of course it needs to be made complying to the structure (encumbered, installation place, ...). Later on it might be possible to get rid of some special packages (e.g. ffmpeg) as they already exist as packages, but that can be something very tricky. For now I want to leave it as other distros do the packaging...

I hope you accept the idea and can give me some valuable hints and help.

Thanks in advance.


chincheta0815:
@pyhalov and @Mno-hime If you have time, would you be so kind and check the concept of that idea? Having LMS as a (conforming) package would be great ...

gc-hpc-admin:
@pyhalov Could you give me some first feedback so that I can see whether I am on the right track etc. Maybe you also say that there is not way or need for such a package at all...

chincheta0815:
@pyhalov Could you give me a short feedback whether I am on the right way? Maybe you also say that there is not need for such a package at all which I would regret...

components/encumbered/lms/lms.p5m
```diff
@@ -0,0 +1,23 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2018 <contributor>
+#
+
+<include lms.proto-objects>
```
pyhalov: Why do you split proto objects to separate manifest? It's not typical and confusing.

components/encumbered/lms/lms.proto-objects
```diff
@@ -0,0 +1,2582 @@
...
+
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/faad.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/flac.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/grant.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/mac.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/mppdec.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/sls.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/socketwrapper.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/sox.exe
+file path=opt/lms-$(COMPONENT_VERSION)/Bin/MSWin32-x86-multi-thread/wmadec.exe
```
pyhalov: Why do we ship non-native binaries?

components/encumbered/lms/Makefile
```diff
@@ -0,0 +1,119 @@
+#
+# CDDL HEADER START
+#
```
pyhalov, line 5: Please, use new copyright header (look for example at templates/justmake.mk)

components/encumbered/lms/Makefile
```diff
@@ -0,0 +1,119 @@
...
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2011, 2015, Oracle and/or its affiliates. All rights reserved.
```
pyhalov: Or is it based on Oracle component?

components/encumbered/lms/Makefile
```diff
@@ -0,0 +1,119 @@
...
+# Copyright 2020 Rouven Weiler
+#
+
+BUILD_BITS=			32
+BUILD_STYLE=			archive
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		lms
+COMPONENT_FMRI =		service/multimedia/
```
pyhalov: Something is clearly wrong here

components/encumbered/lms/Makefile
```diff
@@ -0,0 +1,119 @@
...
+PERL_VERSION =			5.22
+
+BUILD_TARGET =			$(BUILD_DIR)/$(COMPONENT_NAME)/.built $(BUILD_DIR)/$(COMPONENT_NAME)-vendor/.built
+
+INSTALL_TARGET =		$(BUILD_DIR)/$(COMPONENT_NAME)/.installed
+
+include $(WS_MAKE_RULES)/common.mk
+
+BUILDME_ENV +=                  MAKE=$(GMAKE)
+BUILDME_ENV +=			PATH=$(dir $(CC)):$(SPRO_VROOT)/bin:$(PATH)
```
pyhalov: Why do we need $(SPRO_ROOT)/bin here?

components/encumbered/lms/Makefile
```diff
@@ -0,0 +1,119 @@
...
+LMS_GROUP =			lmsd
+
+COMPONENT_POST_UNPACK_ACTION =	($(FIND) $(COMPONENT_DIR)/$(COMPONENT_REPO_NAME)-$(COMPONENT_GIT_COMMIT_HASH)/CPAN/arch \
+                                    -mindepth 1 -maxdepth 1 ! \( -name "5." -o -name "$(PERL_VERSION)" \) -type d -exec rm -rf {} + ; \
+				$(FIND) $(COMPONENT_DIR)/$(COMPONENT_REPO_NAME)-$(COMPONENT_GIT_COMMIT_HASH)/CPAN/arch \
+                                    -mindepth 1 -name "*-*" -type d -exec rm -rf {} + ; \
+                                $(FIND) $(COMPONENT_DIR)/$(COMPONENT_REPO_NAME)-$(COMPONENT_GIT_COMMIT_HASH)/Bin \
+                                    -mindepth 1 -maxdepth 1 ! \( -name "$(PERL_BUILD_ARCH)" \) -type d -exec rm -rf {} +) 
+
+COMPONENT_POST_INSTALL_ACTION +=	$(SBINDIR)/groupadd lmsd ; \
```
pyhalov: build system shouldn't add users or groups, they are delivered as manifest actions.

components/encumbered/lms/lms.proto-objects
chincheta0815: This is an error. They should not be there and will be removed.

components/encumbered/lms/Makefile
chincheta0815: I did not know there is a new one. Sorry.

components/encumbered/lms/Makefile
chincheta0815: Well, work in progress... This was a "detail", but when one does not change it at the beginning one forgets it...

components/encumbered/lms/Makefile
chincheta0815: Well, that was the copy from makemaker.mk to make it run. I would be glad to delete it... and hoped so. I thought it was necessary for perl to work well...

components/encumbered/lms/Makefile
chincheta0815: So as I understand correctly: I do not need to add lmsd explicitly. As long as the manifest action knows this new user and group it will automatically added by the routine(s)?

components/encumbered/lms/lms.p5m
chincheta0815: Well, I thought it would be more "automated" to have the stuff coming from sample-manifest in one file as proto.objects beiing included in the more "static" customized, handmade-part containing base p5m.

But I changed it.

components/encumbered/lms/lms.proto-objects
chincheta0815: All that stuff came from the downloaded component tarball. I set a find-and-delete for removing all non-native stuff.

components/encumbered/lms/Makefile
chincheta0815: Took the one from justmake.mk.

components/encumbered/lms/Makefile
chincheta0815: Corrected that.

components/encumbered/lms/Makefile
chincheta0815: I took service/network.
Hope that is okay. I found nothing more adequate in 'https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/categories-layout.md'

components/encumbered/lms/Makefile
chincheta0815: Removed $(SPRO_VROOT).

components/encumbered/lms/Makefile
chincheta0815: Hope I got the correct lines for making the manifest magic do that.
I chose uid=99 and gid=99. Seems to be free according to 'https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/reserved_uids_and_gids.md'

components/encumbered/lms/Makefile
pyhalov: How does COMPONENT_FMRI become service/multimedia/lms when it's defined as service/multimedia/?

components/encumbered/lms/Makefile
pyhalov: build system can make everything it wants on build host, but only actions from manifest are applied when package is installed.

chincheta0815:
@pyhalov I got the first version that passed until gmake publish. Next step will be testing etc. I also added a lot of your hints.

alarcher:
@chincheta0815 Added the testing_needed label will prevent automatic close by the bot.

AndWac:
Is there still interest in this? If nobody steps up I will close it.

Toasterson:
It looks like there is no interest, closing PR for a newer one.

PR #5543, opened 2020-02-11
Trousers: Fix service start

Mno-hime:
`svc:/application/security/tcsd` fails to start because it expects `/usr/lib/64/tcsd` to be spawned, but we did not deliver it.


components/library/trousers/trousers.p5m
```diff
@@ -165,13 +162,3 @@ file path=usr/share/man/man3/Tspi_TPM_StirRandom.3
...
-
-file dist/system.data.auth path=var/tpm/system/system.data.auth mode=0400 owner=root group=sys
-file dist/system.data.noauth path=var/tpm/system/system.data.noauth mode=0400 owner=root group=sys
-license trousers.license license=CPL
-link path=var/tpm/system/system.data target=system.data.auth
-#
-# libtspi requires 646%UTF-8.so in package utf-8. This is for function
-# Trspi_Native_To_UNICODE() which calls iconv_open() to open this library.
-#
-#depend fmri=pkg:/system/library/iconv/utf-8 type=require
```
pyhalov, line 177: Does comment still stand true?

Mno-hime: Not entirely true but not entirely wrong either. I added comment based on [solaris-userland](https://github.com/oracle/solaris-userland/blob/master/components/trousers/trousers.p5m) text in e31f4595334baf1b00b884ac381af5e99476f252.

pyhalov (approved):

AndWac (approved):
Can you squash the commits before merging?

PR #5549, opened 2020-02-15
Update Bash 5.0 with patches 12..16

Mno-hime:
**Changes**: mostly related to problems with history invocation

### DIFF
--- a/components/shell/bash/Makefile
+++ b/components/shell/bash/Makefile
@@ -21,16 +21,16 @@
 
 #
 # Copyright (c) 2011, 2016, Oracle and/or its affiliates. All rights reserved.
-# Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2020, Michal Nowak
 #
 
-PREFERRED_BITS=		64
+BUILD_BITS=		64
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		bash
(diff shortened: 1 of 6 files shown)


components/shell/bash/Makefile
```diff
@@ -124,16 +122,12 @@ COMPONENT_TEST_MASTER		= \
 # avoid permission issues for other users running these tests
 COMPONENT_POST_TEST_ACTION =	rm -f /tmp/xx
 
-build: $(BUILD_64)
-
 install: $(INSTALL_64)
```
pyhalov: Move it to COMPONENT_POST_INSTALL_ACTION ?

Mno-hime: Changed `COMPONENT_POST_TEST_ACTION` to `COMPONENT_POST_INSTALL_ACTION`.

pyhalov: Wait, I haven't meant it.
I've meant that lines `(cd $(PROTOUSRSHAREDIR)/locale ; cp -Rp "en@boldquot" en ;...` could be moved to COMPONENT_POST_INSTALL_ACTION, and so we don't need `install: $(INSTALL_64)...` rule

alarcher:
@Mno-hime Any issue blocking this one?

PR #5558, opened 2020-02-19
libpeas: avoid dependency on Python 3.4

pyhalov:
We still need to rebuild pluma, eom, rhythmbox and do something with totem to avoid shipping libpeas Python 2.7 module by default.

PR #5559, opened 2020-02-19
Fix libpeas consumers

pyhalov:


Mno-hime:
Later we can merge it to https://github.com/OpenIndiana/oi-userland/pull/5557.

pyhalov:
Still have to find out what to do with totem - either we have to merge several years of work on python modules or finally fix latest version... Don't know what is easier. So far testing first way...

pyhalov:
Should be merged after #5558 .

PR #5560, opened 2020-02-19
make-rules: strip only autopyc-matching pycache files

pyhalov:


pyhalov:
Due to this issue we miss pycache files for example, for pluma modules.

PR #5570, opened 2020-02-25
Add rapidjson package

citrus-it:


components/python/python-rapidjson/Makefile
```diff
@@ -0,0 +1,67 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
```
pyhalov, line 4: Would you mind using illumos standard header?
It can be found in $(WS_TOP)/templates/python.mk

pyhalov:
@citrus-it I think we can merge it in its current state, however, if there's no reasons to preserve old-style headers, I'd changed them to new-style one.

PR #5593, opened 2020-03-11
Fix components

alarcher:
Some minor fixes.


components/cluster/pcs/Makefile
```diff
@@ -67,6 +67,8 @@ test:           $(TESTS_32)
 # Build dependencies
 REQUIRED_PACKAGES += library/ruby/bundler
 REQUIRED_PACKAGES += system/font/truetype/liberation
+REQUIRED_PACKAGES += application/cluster/corosync
```
pyhalov: Whu is it necessary? Builds fine here without them

pyhalov:
Otherwise looks good.

components/cluster/pcs/Makefile
alarcher: There are missing dependencies, the component does not build on a fresh system and for some reason this seemed to help. I will remove this commit until we find the correct dependency.

PR #5595, opened 2020-03-11
python35: update to 3.5.9, import early pep-538 variant

pyhalov:
Here we import earlear PEP 538 variant, based on Fedora Python 3.6 backport.
There's significant difference in behavior with later Python versions (tested with Python 3.8, likely the same is seen with Python 3.7): when LC_ALL is set to C, it overrides all other settings. When LC_ALL is not set, LC_CTYPE is set to C.UTF-8.
Likely, we need efficiently set LC_CTYPE to C.UTF-8 when LC_ALL is set to C...

### DIFF
--- a/components/python/python35/Makefile
+++ b/components/python/python35/Makefile
@@ -29,14 +29,14 @@ BUILD_BITS=	64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		Python
-COMPONENT_VERSION=	3.5.8
+COMPONENT_VERSION=	3.5.9
 COMPONENT_PROJECT_URL=	https://www.python.org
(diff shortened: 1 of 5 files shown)


pyhalov:
@citrus-it At least this helps when LC_ALL is not explicitly set (like after zlogin).

pyhalov:
It seems we also need [PEP 540](https://www.python.org/dev/peps/pep-0540/)

pyhalov:
After looking at PEP-540 implementation - we likely can't port it without breaking ABI and backporting all configuration infrastructure.

pyhalov:
Tested that latest pkg with this version of python3.5 can install and verify packages with UTF-8 files when LC_ALL is not set and LC_COLLATE is "C".

PR #5603, opened 2020-03-13
libvdpau: bump to 1.3

alarcher:


pyhalov:
Is it an ABI breakage?

pyhalov:
What testing was done?

alarcher:
Tested with Nvidia only, no ABI breakage mentioned in the release notes, the so number is the same.

PR #5615, opened 2020-03-16
12388 GLib macroses are incorrect for std=cxx17

pyhalov:


alarcher (approved):
Seems consistent with Linux headers.

PR #5620, opened 2020-03-17
libglade: drop libglade-convert

pyhalov:


pyhalov:
libglade-convert is (broken in some cases) Python 2.7 script to convert `*.glade` files to `*.ui` files. We don't need it. Unfortunately, we still need libglade for gui-installer.

PR #5623, opened 2020-03-18
glade: update to 3.22.2

pyhalov:


pyhalov:
glade is delivered as both 32 and 64-bit in solaris-userland. But I've found only one consumer (besides end users ;)) here - libpeas. It doesn't link to glade, just checks if it is available to deliver one xml file. Now this check passes only in 64-bit libpeas build, but the same file is delivered.

PR #5625, opened 2020-03-18
xml-sax-base: adapt recipe from solaris-userland and bump to 1.09

alarcher:


components/perl/xml-sax-base/xml-sax-base-PERLVER.p5m
```diff
@@ -0,0 +1,37 @@
...
+
+file path=usr/perl5/$(PERLVER)/man/man3/XML::SAX::Base.3
+file path=usr/perl5/$(PERLVER)/man/man3/XML::SAX::BuildSAXBase.3
+file path=usr/perl5/$(PERLVER)/man/man3/XML::SAX::Exception.3
+file path=usr/perl5/vendor_perl/$(PERLVER)/XML/SAX/Base.pm
+file path=usr/perl5/vendor_perl/$(PERLVER)/XML/SAX/BuildSAXBase.pl
+file path=usr/perl5/vendor_perl/$(PERLVER)/XML/SAX/Exception.pm
+file path=usr/perl5/vendor_perl/$(PERLVER)/$(PERL_ARCH)/auto/XML/SAX/Base/.packlist
+
+depend type=conditional fmri=$(COMPONENT_FMRI)-$(PLV) \
```
pyhalov: Does this line belong to GENFRAG?

alarcher: Not sure how it works but that's how it is done in the solaris-userland recipe.

pyhalov: It's clearly wrong in solaris-userland.
We have *-PERLVER package, which should just depend on library/perl-5/xml-sax-base (which is stated in comment) 
and *-GENFRAG package which should depend on any version depending on installed perl version (this is what this dependency actually means)

pyhalov: If runtime/perl-$(PLV) is installed, depend on library/perl-5/xml-sax-base-$(PLV)

alarcher: A similar component would be https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/perl/DBI/DBI-PERLVER.p5m , should I do the same?

alarcher: No Perl component actually seem to declare a GENFRAG manifest.

pyhalov: Yes, it is generated correctly.

PR #5642, opened 2020-03-26
TESTING illumos-gate: use -mtune=generic

alarcher:
Just in case someone is interested in testing this.


Toasterson:
Any reason this has been posted to oi-userland rather than illumos-gate? I would like not to carry any patch that we can upstream.

alarcher:
@Toasterson Because it is for testing and will never be accepted as it is into illumos-gate without proper evaluation.

Toasterson:
True but there is illumos gerrit for that process. I am still confused. Can we keep PR's here for things we would like to have in OI in production?

alarcher:
@Toasterson The point was to make the patch readily available to people using OI for building illumos-gate in userland and testing. illumos gerrit is not suitable for that.

PR #5644, opened 2020-03-27
add perl module Parse::Yapp in version 1.21

chincheta0815:
Could you please check for consistency with other perl modules.
This is my first perl module I add.

I stumbled on the REQUIRED_PACKAGES. They are hardcoded. Maybe there is a mechnism automating the perl versions...

### DIFF
--- a/components/perl/Parse-Yapp/Makefile
+++ b/components/perl/Parse-Yapp/Makefile
@@ -1,0 +1,40 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 Rouven Weiler
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=         Parse-Yapp
+COMPONENT_VERSION=      1.21
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= \
+  sha256:3810e998308fba2e0f4f26043035032b027ce51ce5c8a52a8b8e340ca65f13e5
+COMPONENT_ARCHIVE_URL= \
+  https://cpan.metacpan.org/authors/id/W/WB/WBRASWELL/$(COMPONENT_ARCHIVE)
+COMPONENT_PROJECT_URL=  https://metacpan.org/release/Parse-Yapp/source/lib/Parse/Yapp.pm
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/makemaker.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+COMPONENT_TEST_TARGETS = test
+
+build:          $(BUILD_32_and_64)
+
+install:        $(INSTALL_32_and_64)
+
+test:           $(TEST_32_and_64)
+
+REQUIRED_PACKAGES += runtime/perl-522
+REQUIRED_PACKAGES += runtime/perl-524
--- a/components/perl/Parse-Yapp/Parse-Yapp-PERLVER.p5m
+++ b/components/perl/Parse-Yapp/Parse-Yapp-PERLVER.p5m
@@ -1,0 +1,38 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 Rouven Weiler
+#
+
+set name=pkg.fmri value=pkg:/library/perl-5/parse-yapp-$(PLV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="Perl extension for generating and using LALR parsers"
+set name=info.classification value=org.opensolaris.category.2008:Development/Perl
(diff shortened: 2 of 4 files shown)


components/perl/Parse-Yapp/Parse-Yapp-PERLVER.p5m
```diff
@@ -0,0 +1,35 @@
...
+set name=pkg.fmri value=pkg:/library/perl-5/parse-yapp-$(PLV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="Perl extension for generating and using LALR parsers"
+set name=info.classification value=org.opensolaris.category.2008:Development/Perl
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license Parse-Yapp.license license="Artistic"
+
+file path=usr/perl5/$(PERLVER)/bin/yapp
```
pyhalov, line 28: Should there be usr/bin/yapp => usr/perl5/$(PERLVER)/bin/yapp mediated symlink? Is it designed to be used by end user?

components/perl/Parse-Yapp/Makefile
```diff
@@ -0,0 +1,39 @@
...
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2015 Josef 'Jeff' Sipek <jeffpc@josefsipek.net>
```
pyhalov: Please, fix copyright.

components/perl/Parse-Yapp/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+include $(WS_TOP)/make-rules/prep.mk
+include $(WS_TOP)/make-rules/makemaker.mk
+include $(WS_TOP)/make-rules/ips.mk
+
+build:          $(BUILD_32_and_64)
+
+install:        $(INSTALL_32_and_64)
+
+test:           $(NO_TESTS)
```
pyhalov: Is there a testsuite for the module?

components/perl/Parse-Yapp/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+#
+# Copyright 2015 Josef 'Jeff' Sipek <jeffpc@josefsipek.net>
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=         Parse-Yapp
+COMPONENT_VERSION=      1.21
+COMPONENT_SUMMARY=      Perl extension for generating and using LALR parsers
```
pyhalov: COMPONENT_SUMMARY is unused here

components/perl/Parse-Yapp/Parse-Yapp-PERLVER.p5m
chincheta0815: Concerning the mediator I am new to that. What is the difference between "system-perl" and "perl"?

If "system-perl" means "the actually used one in the system", would that work:
```
link path=usr/bin/yapp target=../perl5/$(PERLVER)/bin/yapp \
    mediator=system-perl mediator-version=$(PERLVER)
```

components/perl/Parse-Yapp/Makefile
chincheta0815: I got the test working. But shouldn't it run, when I do "gmake publish"? Doing a "gmake test" executes it.

Is the check for a passed test done automatically by some magic?

components/perl/Parse-Yapp/Makefile
chincheta0815: I deleted the COMPONENT_SUMMARY here.

components/perl/Parse-Yapp/Makefile
chincheta0815: Changed.

chincheta0815:
@pyhalov Please find attached a new version. I respected all your requests. Only thing I stumbled a little was the transform as I want the link to  the "yapp" script being changed when the perl mediator link is changed. Did I do that right?

Just some remark for the purpose of this package: It is needed for the latest samba version.

chincheta0815:
@pyhalov after checking the perl-522 and perl-524 p5m-files I saw that /usr/perl5/bin is in it's whole mediated. The yapp binary is in there and should be mediated as the tools of libwww* are. Making it conformal with all other perl modules I think that is okay to go without a script in /usr/bin, having the script in /usr/perl5/bin ...

components/perl/Parse-Yapp/Parse-Yapp-PERLVER.p5m
pyhalov: system-perl is one used by illumos-gate scripts, i.e. /usr/perl5/bin/perl. 
perl is usr/bin/perl. The difference is that one can't just change system-perl and expect things to be working. Please, use perl mediator.

components/perl/Parse-Yapp/Makefile
pyhalov: no, currently tests are run only by devs - an easy way to find out what's broken after update.

components/perl/Parse-Yapp/Parse-Yapp-PERLVER.p5m
chincheta0815: I looked into libwww-perl.PERLVER.p5m (https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/perl/libwww-perl/libwww-perl-PERLVER.p5m). It also adds scripts to /usr/perl5/bin, but does not use a mediator.

Analog to this I removed the mediator for yapp. This then solve this issue here. I think that is okay as it make things transparent.

pyhalov: As I understand, yapp is intended to be used by end user, so it makes sense to make it available in /usr/bin.

chincheta0815: Just added:
```
link path=usr/bin/yapp target=../perl5/$(PERLVER)/bin/yapp \
    mediator=perl mediator-version=$(PERLVER)
```
I hope I understood correctly to use `mediator=perl` instead of `mediator=system-perl`.

>

PR #5648, opened 2020-03-28
marco: fix dependencies and enable compositor by default

alarcher:
Need someone to test this on Intel.


AndWac (approved):
I build a new marco binary and copied it over to my workstation. It seems to run fine there.

PR #5651, opened 2020-03-29
shared-macros.mk: add PERL5BINDIR as a macro

chincheta0815:
this should help to use scripts that are located in PERL5BINDIR depending on the perl version. It will be needed e.g. to build samba-4.12.0 which needs yapp localte in /usr/perl5/5.24/bin ...


pyhalov:
@alarcher How do you feel about this? 
I'm not sure why this is needed - /usr/perl5/bin is used only as search path for internal perl tools like pod2man.

pyhalov:
Ah, OK. It's intended to find correct perl supplemental binary , yapp in our case.

pyhalov (approved):

alarcher (approved):

PR #5652, opened 2020-03-29
update to samba-4.12.0

chincheta0815:
Please commit this PR after PR #5651 and PR #5644 since these two are prereqs for this one.


components/network/samba/Makefile
```diff
@@ -156,10 +160,12 @@ REQUIRED_PACKAGES += library/jansson
 REQUIRED_PACKAGES += library/libarchive
 REQUIRED_PACKAGES += library/ncurses
 REQUIRED_PACKAGES += library/openldap
+REQUIRED_PACKAGES += library/perl-5/parse-yapp-524
```
pyhalov: Should it be in BUILD_DEPENDENCIES section?

chincheta0815: Due to https://wiki.samba.org/index.php/Package_Dependencies_Required_to_Build_Samba#Mandatory

It should only be necessary for building. Same for perl. I changed that in the Makefile.

pyhalov:
Also, please fix 
```
link path=usr/lib/security/pam_winbind.so \
    target=../samba/security/$(MACH64)/pam_winbind.so
```
Should have been 
```
link path=usr/lib/security/pam_winbind.so \
    target=../samba/security/pam_winbind.so
```

chincheta0815:
@pyhalov Okay. Hope I got it.

components/network/samba/Makefile
```diff
@@ -160,7 +168,7 @@ REQUIRED_PACKAGES += library/print/cups-libs
 REQUIRED_PACKAGES += library/readline
```
pyhalov, line 164: library/icu is also needed (it's used if latest version is found on build host).

pyhalov:
Also, are you sure that python3.4 is still needed to build samba if we configure 32-bit version without python? Perhaps, we could remove all dances around it?

chincheta0815:
@pyhalov Hej, I just swa that you were faster ;o) Thanks. Just a question on the lines you added concerning the 32bit comaptiblity of nss_winbind:
Why is that "trick" needed: https://github.com/OpenIndiana/oi-userland/blob/46b7fa8ccc8ccc0ac3f81ec24771e0ba9c339903/components/network/samba/Makefile#L62 and following line plus https://github.com/OpenIndiana/oi-userland/blob/46b7fa8ccc8ccc0ac3f81ec24771e0ba9c339903/components/network/samba/Makefile#L69 and following line.

I thought that LDFLAGS.32 and LDFLAGS.64 might work, but I could not get that working... Maybe I am just too dumb???

pyhalov:
Do you means something like 
```
LDFLAGS.32 =  -R/usr/lib/samba -R/usr/lib/samba/private
LDFLAGS.64 = -R/usr/lib/samba/$(MACH64) -R/usr/lib/samba/private/$(MACH64)
LDFLAGS += $(LDFLAGS.$(BITS))
```
Yes, this could also work.

PR #5656, opened 2020-04-01
sqlite: update to 3.31.1

pyhalov:


components/database/sqlite/files/mapfile
```diff
@@ -27,6 +27,25 @@
 #
 $mapfile_version 2
 
+SYMBOL_VERSION sqlite_3.31.1 {
+    global:
+        sqlite3_column_origin_name;
```
pyhalov, line 32: We miss a lot of public symbols in libsqlite3, introduced between 3.19 and 3.30 (and perhaps, some earlier).
I put what I've found here.

PR #5660, opened 2020-04-09
Libreoffice

pyhalov:


components/desktop/libreoffice/patches/02-pdfium.patch
```diff
@@ -0,0 +1,38 @@
+diff -urN libreoffice-6.4.2.2/external/pdfium-1/illumos.patch libreoffice-6.4.2.2/external/pdfium/illumos.patch
```
pyhalov, line 1: Anyway, pdfium fails as it knows nothing about illumos/solaris

alarcher:
This commit takes into account recent userland dependencies: https://github.com/alarcher/oi-userland/commit/4205d077b9785c20035a0e5bd457648005e64893

pyhalov:
I've updated libreoffice to use more external libraries which are present in OI.

AndWac:
Despite missing scrollbars in the document windows everything seems to work as expected with  my (locally compiled) package...

iigs:
Do you want to add the following packages to the build dependencies in the Makefile?  I had to install these packages before I could compile libreoffice.

developer/gperf
developer/icu
developer/cppunit

Geoff

pyhalov:
@iigs Thanks, added.
So far we have issue with scrollbars (for now no time to debug it).

iigs:
I tried to build libreoffice with the following, but it didn't help with the scrollbar problem or the other widget problems identified below:
CONFIGURE_OPTIONS += --enable-gtk3
CONFIGURE_OPTIONS += --enable-introspection

There are many more gtk3 themeing problems than the missing scrollbars.  I have identified these as well:
* font selection widget doesn't pop-up with font choices or text size choices
* sidebars do not show widget for checkbox or radio button
* menu button widget in sidebars does not display (menu shows up if clicked)
* sheet selection on spreadsheet is underlines
* icons are not displayed with menu items

Changing the MATE gtk3 theme from nimbus to Menta or TradtionalOK doesn't help.

Other platform have a package called libreoffice-gtk3 "to enable GTK3 theme support".  This package contains these files:
/usr/lib/libreoffice/program/liblosessioninstalllo.so
/usr/lib/libreoffice/program/libvclplug_gtk3lo.so

If I remove these two files, then libreoffice uses its native Motif-like theme instead of GTK3, but widgets display/work correctly.  Scrollbars and the other widget issues I found are no longer an issue.  It just doesn't match OI's look and feel (I'll use it, but it looks ugly).


Reviewing config.log during compile, I am concerned about this:
checking whether and how to use Xinerama... no

I tracked this to line 38349 and lines 38397-38406 in libreoffice-6.4.3.2/configure
It think "$_os" = "SunOS" needs to be added to line 38349, but I don't know what OI would need for XINERAMA_EXTRA_LIBS

This led to me to check other places where _os is checked.  I am wondeirng if line 27069 needs SunOS added.

pyhalov:
It seems I don't have enough time to debug GTK3 plugin issues for now, so I suggest merging LO with default VCL plugin set to `gen`.

pyhalov:
@AndWac @iigs

AndWac:
Merging now is OK for me. General functionality is given, problems can be solved later.
We should explain the degraded functionality, though.

iigs:
Built and ran with the latest two git changes--no regressions found when compared to sfe's 5.2. Unless my google foo is bad, the look and feel of this libreoffice seems like a retro StarOffice 5.1 running on Solaris.

PR #5666, opened 2020-04-10
no unaligned memory access on sparc

tsoome:
memory access on sparc must be aligned.

### DIFF
--- a/components/library/zlib/Makefile
+++ b/components/library/zlib/Makefile
@@ -28,7 +28,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		zlib
 COMPONENT_VERSION=	1.2.11
-COMPONENT_REVISION=	2
+COMPONENT_REVISION=	3
 COMPONENT_SUMMARY=	The Zip compression library
 COMPONENT_PROJECT_URL=	http://www.zlib.net/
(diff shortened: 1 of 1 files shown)


components/library/zlib/Makefile
```diff
@@ -57,7 +57,7 @@ gcc_OPT="-O2"
 
 # Also, the x86 architecture does not require alignment for multi-byte
 # loads, so we can define UNALIGNED_OK for x86
-CFLAGS.32 += -DUNALIGNED_OK
+CFLAGS.i386.32 += -DUNALIGNED_OK
```
pyhalov, line 60: I'm not sure this will work as expected - we don't have CFLAGS+=$(CFLAGS.$(MACH).$(BITS)) anywhere

pyhalov:
Also, please bump COMPONENT_REVISION while touching component and not updating COMPONENT_VERSION.

components/library/zlib/Makefile
```diff
@@ -57,7 +57,7 @@ gcc_OPT="-O2"
...
 
 CFLAGS += $(CC_PIC)
 CFLAGS += $(CFLAGS.$(BITS))
```
pyhalov: Perhaps, you should also change this line.

PR #5667, opened 2020-04-10
ruby-23: fix sparc build

tsoome:


pyhalov:
Please, add `COMPONENT_REVISION=1` after COMPONENT_VERSION line.
Commit message should mention affected component, like 
`ruby-23: fix sparc build` or `Update X to A.B`

PR #5689, opened 2020-04-22
gcm : update to match current Git HEAD and drop most interim patches

jimklimov:
Better fixes than those made by local patches got into upstream, so use them :)


pyhalov:

PR #5694, opened 2020-04-27
Smalltalk

cstes:
Added a Makefile in components/smalltalk for the "Squeak" Smalltalk Classic VM (see http://squeak.org and http://squeakvm.org).


components/smalltalk/squeak4/Makefile
```diff
@@ -0,0 +1,67 @@
...
+#
+
+#
+# Package Makefile for the "Classic VM" Squeak Smalltalk V4 system
+# See http://squeak.org and http://squeakvm.org
+#
+
+# squeak4 uses cmake and a custom configure script (not GNU autoconf)
+BUILD_STYLE=		configure
+
```
AndWac, line 18: configure is the default BUILD_STYLE, so you don't have to set it here.
Instead you should add BUILD_BITS=32 here.

components/smalltalk/squeak4/Makefile
```diff
@@ -0,0 +1,67 @@
...
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak4
+COMPONENT_VERSION=	4.16.7
```
AndWac, line 22: Why do you have the major version number in the name? If you later want to add newer versions you can make use of the mediator system.

components/smalltalk/squeak4/Makefile
```diff
@@ -0,0 +1,67 @@
...
+COMPONENT_FMRI=		runtime/squeak4
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+SVN_REPO=		http://squeakvm.org/svn/squeak/trunk/
+SVN_REV=		3775
+SVN_HASH=  sha256:497a8cc645a75fc049f9467dab6fff2bd06726e09c39d1a4d1a42164f820a102
+SQUEAK_REV = $(COMPONENT_VERSION)-$(SVN_REV)
+SQUEAK_VERSION = squeak-$(SQUEAK_REV)
+COMPONENT_LICENSE=	MIT
+COMPONENT_LICENSE_FILE=	usr/share/doc/$(SQUEAK_VERSION)/LICENSE
+
```
AndWac, line 34: We don't use the target path for the license file. Typically we copy the appropriate license file into the package folder (that is, make it an explicit part of your commit) and just name it here, like
COMPONENT_LICENSE= LICENSE
You can even omit the copying if the file is in the top level folder of the $COMPONENT_ARCHIVE.

components/smalltalk/squeak4/Makefile
```diff
@@ -0,0 +1,67 @@
...
+REQUIRED_PACKAGES += x11/library/libx11 
+REQUIRED_PACKAGES += x11/library/libxext
+REQUIRED_PACKAGES += x11/library/libxrender
+REQUIRED_PACKAGES += library/glib2 
+REQUIRED_PACKAGES += library/libffi 
+REQUIRED_PACKAGES += library/audio/gstreamer 
+REQUIRED_PACKAGES += library/audio/pulseaudio 
+REQUIRED_PACKAGES += library/desktop/pango 
+REQUIRED_PACKAGES += library/desktop/cairo 
+
```
AndWac, line 67: Is the list of required packages generated by "gmake REQUIRED_PACKAGES"? If so, why did you remove the generated comment line? If not, do it :)

components/smalltalk/squeak4/squeak4.p5m
```diff
@@ -0,0 +1,49 @@
+set name=pkg.fmri value=pkg:/runtime/squeak4@$(IPS_COMPONENT_VERSION)
+set name=pkg.description value="Squeak V4"
+set name=pkg.summary value="The Squeak V4 Smalltalk System"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+
```
AndWac, line 7: It looks like you have completely hand-crafted this .p5m file.
Usually you should run "gmake sample-manifest" that generates a stub file in the manifests folder.
This file can be used as an input to pkgmogrify (which does some automatic transformations),
optionally add some more transformation rules to apply and finally piped into pkgfmt.
This will generate an almost complete p5m file which typically only need a few manual adaptions.

components/smalltalk/squeak4/squeak4.p5m
```diff
@@ -0,0 +1,49 @@
...
+file usr/lib/squeak/$(SQUEAK_REV)/so.SqueakFFIPrims group=bin mode=0555 owner=root path=usr/lib/squeak/$(SQUEAK_REV)/so.SqueakFFIPrims
+file usr/lib/squeak/$(SQUEAK_REV)/so.vm-display-custom group=bin mode=0555 owner=root path=usr/lib/squeak/$(SQUEAK_REV)/so.vm-display-custom
+file usr/lib/squeak/$(SQUEAK_REV)/so.XDisplayControlPlugin group=bin mode=0555 owner=root path=usr/lib/squeak/$(SQUEAK_REV)/so.XDisplayControlPlugin
+file usr/lib/squeak/$(SQUEAK_REV)/so.FT2Plugin group=bin mode=0555 owner=root path=usr/lib/squeak/$(SQUEAK_REV)/so.FT2Plugin
+file usr/bin/squeak.sh group=bin mode=0555 owner=root path=usr/bin/squeak.sh
+file usr/bin/squeak group=bin mode=0555 owner=root path=usr/bin/squeak
+
+file usr/share/man/man1/squeak.1 group=bin mode=0444 owner=root path=usr/share/man/man1/squeak.1
+file usr/share/doc/$(SQUEAK_VERSION)/LICENSE group=bin mode=0444 owner=root path=usr/share/doc/$(SQUEAK_VERSION)/LICENSE
+
```
AndWac, line 49: manifests/sample-manifest.p5m is missing from your commit.

AndWac (commented):
Can I expect another package from you with the latest OpenSmalltalk-VM so I can run Pharo ;-)

components/smalltalk/squeak4/Makefile
cstes: I have another package called squeak5,  squeak5 is indeed built from "opensmalltalk", like "pharo", as you mention.   I did not yet build pharo, but squeak5 seems to be close to it.  The motivation here to use squeak4 and squeak5 is because currently it seems there is no such VM as "squeak" but it has forked in two different build systems : squeak4 uses "cmake" and squeak5 uses "autotools" (as in my understanding, squeak 3.x did).

components/smalltalk/squeak4/Makefile
cstes: ok thanks, I'll give this a try

components/smalltalk/squeak4/Makefile
cstes: The list was handcrafted by reviewing the pkgdepend resolv messages.

components/smalltalk/squeak4/squeak4.p5m
cstes: The p5m is not entirely hand-crafted.   I started from a "pkgsend generate" output of the prototype directory, which I then started to edit in the p5m file until I had something that "gmake publish" accepts.

components/smalltalk/squeak4/squeak4.p5m
cstes: I've noticed that indeed many components have a manifests directory.  I'm not sure I fully understand what this directory is supposed to be (and yes, I've tried to read the doc/packaging.txt !).  In my case I guess I could add one by copying my squeak4.p5m manifest to manifests/sample-manifest.p5m ?

In any case thanks for reviewing and the suggestions.

AndWac:
Ok, some more remarks:
- You should name you PR more intention-revealing, like "Add squeak4" instead of "Smalltalk".
- We have some documentation located inside the doc folder of oi-userland. You should at least read packaging.txt and current-tasks.md. It will give you more ideas about the build system.
Of course we have more documentation on https://docs.openindiana.org.

components/smalltalk/squeak4/Makefile
AndWac: Fine, when you start to integrate it you should read about the mediator concept of pkg.
You'll need to add some mediator lines into the .p5m file manually.
But until we only have one squeak you can omit them.
So, with squeak5 or pharo in the pipeline it's perfectly ok to name your new package squeak4.

cstes:
> Can I expect another package from you with the latest OpenSmalltalk-VM so I can run Pharo ;-)

There exist many dialects I believe of Smalltalk, due to Smalltalk's rich and interesting history.  One reason to use a components directory "smalltalk" and to use the name "squeak4" (it can only run Squeak V4 images) is that it can be populated with all of the wonderful dialects such as squeak4, squeak5,  pharo , maybe  "gnu-smalltalk"  and many others.  In theory IPS packaging of Smalltalk modules (ST code) could also be there, although that Smalltalk systems presumably have their own packaging systems.

components/smalltalk/squeak4/Makefile
AndWac: You should run "gmake REQUIRED_PACKAGES". It will add some lines to your Makefile. Afterwards you should remove all entries that have been detected automatically by this process and add additionally needed (undetected) packages in a separate area with a small comment above (like # Manually added build dependencies)

components/smalltalk/squeak4/squeak4.p5m
AndWac: It's better to rely on "gmake sample-manifest" and add/change whatever is needed then.

components/smalltalk/squeak4/squeak4.p5m
AndWac: Whenever you create a new version you will run gmake sample-manifest again.
The diffs between the new sample-manifest.p5m and the old (git diff manifest/sample-manifest.p5m) will tell you what you'll need to change in your .p5m package file.
Otherwise you will have more work to find the differences and merge them into the p5m file.

AndWac:
When you search for old Squeak and Pharo mailing lists you will find me there.
I have also been to ESUG Smalltalk conference 2010 in Barcelona :)
Plus I am located not thus far from Potsdam so I was able the attend a presentation of Alan Kay at Hasso Plattner Institute some years ago.

cstes:
> When you search for old Squeak and Pharo mailing lists you will find me there.
> I have also been to ESUG Smalltalk conference 2010 in Barcelona :)
> Plus I am located not thus far from Potsdam so I was able the attend a presentation of Alan Kay at Hasso Plattner Institute some years ago.

According to the webpage of http://www.vpri.org/  Alan Kay was or is involved in that.  Also note that in my squeak4.p5m license action I try to include the LICENSE file which refers to VPRI among other.

I would appreciate it, if you could build the package (squeak4) in userland, and you (or any other volunteer) could test running it with a Squeak image like Squeak 4.6 and SqueakV4sources from http://squeak.org,  then feedback is certainly appreciated,  whether it works on OpenIndiana.  For me I built it on OpenIndiana Hipster and with my limited knowledge on Smalltalk, I'm satisfied with it.

It depens on how far you go with that, they have internal Test FrameWorks.  I ran a few hundred tests which ran OK, but some tests fail for the moment in squeak5  (related to mmap memory allocation of +1GB segments, which for some reason "works" in squeak4 or doesn't raise an error , better said).

As I said, squeak5 is also working for me, and that's much more modern with full 64bit support.

In any case, I expect the squeak4 package to be fairly static, as not much development seems to be going on there for the last 10 years.

But the squeak4 package remains valuable to test some old "Squeak V4" Smalltalk images.

I expect a lot more changes and updates for the squeak5 package, which also works on OpenIndiana.  So the next package I could submit is squeak5 in 64 bit mode.

components/smalltalk/squeak4/Makefile
AndWac: Additionally you'll need to add
TEST_TARGET= $(NO_TESTS)
before you include common.mk.

components/smalltalk/squeak4/Makefile
AndWac: Running gmake REQUIRED_PACKAGES detects all but two dependencies:
ogl-select and dbus.

components/smalltalk/squeak4/squeak4.p5m
AndWac: Running gmake sample-manifest generates a simpler file which you can take as a starting point.
Copying over some additional lines from your original p5m file would be easy.

components/smalltalk/squeak4/patches/01-aio.patch
```diff
@@ -0,0 +1,69 @@
+*** squeak4-3775/platforms/unix/vm/aio.c	Sun Apr 26 21:11:52 2020
```
pyhalov, line 1: Please, convert patches to unified format.

cstes:
Close pull request , will create a new request specifically for squeak4.
Note that squeak4 may work on SPARC, I've only tested it on intel (32 bit).
Squeak5 supports 64bit but it is a different type of virtual machine and different format for Smalltalk images.

AndWac:
Am 28.04.20 um 18:12 schrieb David Stes:
>
> Close pull request , will create a new request specifically for squeak4.
> Note that squeak4 may work on SPARC, I've only tested it on intel (32
> bit).
>
Yeah, I once had squeak running on my Blade 2000. If I remember
correctly it was squeak 3.
Maybe I can find the package in my archives. It must have been SVR
package format.

> Squeak5 supports 64bit but it is a different type of virtual machine
> and different format for Smalltalk images.
>
>
Yes, Cog brought vast speed improvements. I guess FFI needs to be
tested. Did you test FFI with your squeak4 port?
BTW: impressive work, especially if you were able to get the new
OpenSmalltalkVM runnable on OpenIndiana...

Regards
Andreas

cstes:
An old SPARC version can still be found at http://squeakvm.org/unix/release/Squeak-3.7-7.sparc-sun-solaris2.9.tar.gz  ... that is why I think that squeak4 may still compile on SPARC, I don't know.

I didn't test FFI (Foreign Function Interface).  Also my understanding is that squeak4 is not really "active" any longer, all development seems to be going on in Pharo, Squeak Cog Spur etc. 

The Squeak5 images from http://squeak.org cannot be ran with the squeak4 vm.   So the more active development is going on in squeak5.

PR #5695, opened 2020-04-28
gcc-9: sparc fix is missing sparcv9 for g++

tsoome:
While building gate, found another mistake in packaging.


components/developer/gcc-9/gcc-9.p5m
```diff
@@ -924,6 +924,29 @@ file path=usr/gcc/9/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/stdc++.
 file path=usr/gcc/9/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/stdtr1c++.h
 file path=usr/gcc/9/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/time_members.h
 file path=usr/gcc/9/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/ext/opt_random.h
+file path=usr/gcc/9/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/atomic_word.h
```
pyhalov: Have we missed atomic_word.h on both sparc and x86?

PR #5696, opened 2020-04-28
Add squeak4

cstes:
Tested on Intel 32bit with Squeak4.6-15102.zip and SqueakV46.sources.zip.  This vm will not work or run 64bit images like Squeak5.3-19431-64bit.image.  For that I plan to submit a squeak5 package.  Note that squeak5 is not likely to work on SPARC.   squeak4 may still work on SPARC in 32bit (I haven't tested that).  squeak v3 and earlier was available in the past on SPARC.   The squeak4 is not a modern Smalltalk system, but may still be of use as reference implementation.


cstes:
Note that the .p5m manifest is now created with "gmake sample-manifest".

However it sets the mode 0444 on virtually all files

# ls -l /usr/lib/squeak/4.16.7-3775/so.vm-sound-pulse
-r--r--r--   1 root     bin        18292 Apr 28 16:08 /usr/lib/squeak/4.16.7-3775/so.vm-sound-pulse

For the "ckformat" and "squeakvm" (also located by design of Ian Piumarta in /usr/lib) this does not work , so I've changed the sample-manifest to set mode 0555.

I'm not sure whether that is required on the so.vm-sound-pulse and other Plugins.  They seem to be loaded

$ pldd `pgrep squeakvm` | grep pulse
/usr/lib/squeak/4.16.7-3775/so.vm-sound-pulse

so I've left the plugins mode 0444 as it seems to work .

Finally also note that my manifest has a line

# see desktop/rss-glx/rss-glx.p5m:
<transform file -> add pkg.depend.bypass-generate libGL\.so\.1>

this is because I encountered a problem

/scratch/clone/components/smalltalk/squeak4/build/manifest-i386-squeak4.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libGL.so.1 \
        pkg.debug.depend.reason=usr/lib/squeak/4.16.7-3775/so.vm-display-X11 \

cstes:
by the way squeak4 may still compile on SPARC ... squeak v3 and earlier worked on SPARC
squeak5 is probably not compiling on SPARC

cstes:
The old SPARC version can still be downloaded from http://squeakvm.org/unix/release/Squeak-3.7-7.sparc-sun-solaris2.9.tar.gz   

I think that was work of Andreas Raab or Ian Piumarta ?

I don't think it was tested later in Squeak4 but there is a document in unix/doc where Ian Piumarta wrote

3.3. SunOS/Solaris
The support hasn't been tested in ages.  I haven't a clue if it still works.

On OpenIndiana Intel with "pulseaudio" pulseaudio sound seems to work for me;
perhaps there may be issues on SPARC that there is no audio ... or perhaps it is not even compiling, but it is possible that it compiles.

components/runtime/squeak4/Makefile
```diff
@@ -0,0 +1,67 @@
...
+COMPONENT_SUMMARY=	The Squeak V4 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_ARCHIVE_URL=  http://squeakvm.org
+COMPONENT_FMRI=		runtime/squeak4
+COMPONENT_CLASSIFICATION=	Development/Smalltalk
+SVN_REPO=		http://squeakvm.org/svn/squeak/trunk/
+SVN_REV=		3775
+SVN_HASH=  sha256:1f40ae6a9ad05739c07d63f4e67bfd6afacbe76c4deb2802bba1be7332f373a7
+COMPONENT_LICENSE=      LICENSE
+
```
AndWac, line 40: You still need to set correct COMPONENT_LICENSE and COMPONENT_LICENSE_FILE.
For COMPONENT_LICENSE I am not sure what to set it to as the license file doesn't give a name to the licenses. Probably something like "various Open Source, see license file and individual soure files".

When everything is done please rebase your repository to latest OpenIndiana/oi-hipster and then squash all your commit belonging to this PR into one. Maybe alp has some more comments.

cstes: Based on http://wiki.squeak.org/squeak/159 and http://wiki.squeak.org/squeak/933  I have to use "MIT" as LICENSE.   Regarding license files, I can also copy those URL's as COMPONENT_LICENSE_URL if that exists ??

AndWac: No, our build system doesn't know about COMPONENT_LICENSE_URL but only COMPONENT_LICENSE and COMPONENT_LICENSE_FILE. You can manually download and add the license information to your PR, though. If you do it set COMPONENT_LICENSE_FILE accordingly.

PR #5700, opened 2020-04-29
Add squeak5

cstes:
This is a build for the "Cog Spur Squeak" VM from http://opensmalltalk.org.

This is a modern 64-bit Smalltalk VM.

It works for me with the image from http://squeak.org  Squeak5.3-19431-64bit.zip and SqueakV50.sources.zip ; it also worked for Squeak 6.0 alpha

Note that there is an issue with the /usr/bin/squeak script : it cannot find /usr/lib/amd64 and looks for a binary in /usr/lib which is not the OpenIndiana convention for 64bit.

The workaround is not to use the "wrapper script" /usr/bin/squeak but to launch squeak via 

/usr/lib/amd64/squeak/5.0-202004291145/squeak Squeak5.3-19431-64bit.image

Also I'm submitting to the (apparently very active development) of OpenSmalltalk so there may be many subsequent changes to this pull request .


components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+#
+
+#
+# Package Makefile for the "OpenSmalltalk Cog Spur" Squeak Smalltalk V5 system
+# See http://squeak.org and http://opensmalltalk.org
+#
+
+BUILD_BITS=64
+COMPILER=gcc
+
```
AndWac: Remove  COMPILER=gcc, it's the default.

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Package Makefile for the "OpenSmalltalk Cog Spur" Squeak Smalltalk V5 system
+# See http://squeak.org and http://opensmalltalk.org
+#
+
+BUILD_BITS=64
+COMPILER=gcc
```
AndWac: Do you only support 64 Bits?

components/smalltalk/squeak5/squeak5.p5m
```diff
@@ -0,0 +1,69 @@
...
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 David Stes
+# (with help from Andreas Wacknitz and Alexander Pyhalov)
+#
```
AndWac: Just remove us helpers here, you are the copyright holder. We just give some tips for integration ;-)

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+TEST_TARGET= $(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+PATH=$(PATH.gnu)
+
+# The ugly hack with update-publish target is necessary to update
+# source from git repository on each "gmake publish".
+publish: update-publish
+
```
AndWac: Why is it necessary to update the sources from git on each "gmake publish"? This would create a new build on every Jenkins run on our build server, thus, whenever a change in our git repository happens.

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+BUILD_BITS=64
+COMPILER=gcc
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak5
+COMPONENT_VERSION=	5.0.6
+COMPONENT_SUMMARY=	The Squeak V5 Smalltalk System
+COMPONENT_PROJECT_URL=	http://www.squeak.org
```
AndWac: Probably "The Squeak V5 Smalltalk Virtual Machine" fits better.
Maybe you can even think of a better name of the package and its summary if Pharo Smalltalk is also running on this VM?

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += library/audio/pulseaudio
+REQUIRED_PACKAGES += library/desktop/cairo
+REQUIRED_PACKAGES += library/desktop/pango
+REQUIRED_PACKAGES += library/glib2
+REQUIRED_PACKAGES += library/security/openssl
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += x11/library/libx11
+REQUIRED_PACKAGES += x11/library/libxrender
```
AndWac: Something seems to be wrong or missing: "gmake build" fails on my build machine:
libtool: link: ( cd ".libs" && rm -f "UUIDPlugin.la" && ln -s "../UUIDPlugin.la" "UUIDPlugin.la" )
: XDisplayControlPlugin.la
libtool: link: /usr/gcc/7/bin/gcc -shared  -fPIC -DPIC -Wl,-z -Wl,text -Wl,-h -Wl,SqueakSSL.so -o .libs/SqueakSSL.so  .libs/SqueakSSL.o .libs/sqUnixSSL.o   -L/usr/local/lib -lm -lsocket -lnsl -lcrypto -lssl -ltls  -g -pthread -m64 -m64 -Wl,-z -Wl,now   -pthread
: UUIDPlugin.la
/oibuild/oi-userland/components/smalltalk/squeak5/opensmalltalk-vm/platforms/unix/config/verstamp version.c /usr/gcc/7/bin/gcc
ld: fatal: library -ltls: not found
ld: fatal: file processing errors. No output written to .libs/SqueakSSL.so
make[2]: *** [Makefile:169: SqueakSSL.la] Error 1
make[1]: *** [Makefile:541: SqueakSSL.la] Error 2

AndWac (changes requested):

components/smalltalk/squeak5/Makefile
cstes: Ideally I could build from an 'archive' or 'release' from codeload.github.com  (a tar.gz package).  I tested that and it is better.  But unfortunately I think the opensmalltalk build setup currently more or less implies that you have to build from git clones (not unversion-ed sources).  So that's why I  "git clone" the sources.   The "update-publish" target is not doing anything for the moment in my case.   I copied and modified that code from "./openindiana/time-slider/Makefile" ; I'll check this later ...

components/smalltalk/squeak5/Makefile
cstes: Looks like a configure issue where configure is perhaps enabling --with-tls on your server.

I did not on my system.

From looking at the configure script it is checking for a header file tls.h.

Do you :  find /usr/include/openssl -name tls.h on your system ?

I'll check tomorrow but it built for me and best guess is that somehow the openssl configure logic of "opensmalltalk" is enabling the tls support (by mistake) either by detecting the header file or due to some other reason ...

cstes: What is the result on your system of

pkg search -af tls.h

??   is there perhaps a package installed that offers a tls.h ?

My guess is that the opensmalltalk-vm  configure script is confused and enables --with-tls in your case.

To avoid this (if it is the problem) the Makefile would have to "exclude" some packages that may not be installed I guess.

cstes: One possibility is perhaps

./tcltls/tcltls.p5m:file path=usr/include/tls.h

tcltls is it installed for you ?  anyway I'll check tomorrow

also depending on your point of view this can be considered a problem with the opensmalltalk-vm configure script  -- or a problem with the build server - or the Makefile of squeak5.p5m that could perhaps (if possible) exclude packages that may not be installed

cstes: Possibly adding --without-libtls  helps ... 
I've just made that modification to the Makefile but I will have to test whether it then still builds for me.   For me it built without that option.

AndWac: I have almost any package installed on my build system.
─➤  pkg search -l   /usr/include/tls.h
INDEX      ACTION VALUE             PACKAGE
path       file   usr/include/tls.h pkg:/runtime/tcl-8/tcl-openssl@1.6.7-2020.0.1.0

--without-libtls fixed the problem.

components/smalltalk/squeak5/Makefile
AndWac: I think alp should have a look at it. I am not experienced enough in this area.

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+COMPONENT_VERSION=	5.0.6
+COMPONENT_SUMMARY=	The Squeak V5 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_FMRI=		runtime/squeak5
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+COMPONENT_LICENSE=	LICENSE
+COMPONENT_SRC=		opensmalltalk-vm
+
+GIT=			git
+GIT_REPO=		https://github.com/cstes/opensmalltalk-vm.git
```
AndWac: Can you add --depth=1 for cloning the repository? Cloning the whole history is taking a lot of time and space.

components/smalltalk/squeak5/squeak5.p5m
```diff
@@ -0,0 +1,68 @@
...
+
+file mode=0555 path=usr/bin/squeak
+file path=usr/doc/squeak/COPYING.gz
+file path=usr/doc/squeak/COPYRIGHT.gz
+file path=usr/doc/squeak/LICENSE.gz
+file path=usr/doc/squeak/README.Contributing.gz
+file path=usr/doc/squeak/README.Keyboard.gz
+file path=usr/doc/squeak/README.Sound.gz
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/B3DAcceleratorPlugin.a
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/B3DAcceleratorPlugin.so
```
AndWac: Please remove all .a files from squeak5.p5m.
We do not deliver static libraries.

cstes: Good point.  I just did that from the manifest.
I am not sure why they (opensmalltalk) have them installed with the install target.

components/smalltalk/squeak5/Makefile
cstes: I can do that.   I think working with a .tar.gz tarball (COMPONENT_ARCHIVE_URL ) will in the end be easier than working with "git" (in fact git offers the feature with 'releases'  and then downloading it from codeload.github.com) .   But as I wrote before currently I have to build from the git clone and cannot build from an exported .tar.gz which is due to the upstream opensmalltalk-vm build scripts.   I'll test the --depth=1 tomorrow.

components/smalltalk/squeak5/squeak5.p5m
cstes: By the way "pharo" from http://pharo.org/  is not only a different Smalltalk image but I think also a different vm (virtual machine).   As far as I know, squeak5 cannot run pharo images.  But this may be incorrect understanding from my point of view.   The pharo VM itself is built from similar sources as squeak5 so it is not much work I think to add pharo as well.  But I didn't do it so far, I may be wrong.

components/smalltalk/squeak5/Makefile
AndWac: Yes, working with COMPONENT_ARCHIVE_URL is the preferred way. We have a files server where we put the files the build server fetches so we have an alternative path if the original is not accessible.

components/smalltalk/squeak5/squeak5.p5m
AndWac: From the homepage (you should probably change COMPONENT_PROJECT_URL?) opensmalltalkvm.org:
"The cross-platform virtual machine for Squeak, Pharo, Cuis, and Newspeak."
I don't know how the Pharo guys build their VM, though.
But I think Pharo and CUIS would be good additions.
Probably CUIS runs out of the box with this VM because it's a straight descendent from Squeak.

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,95 @@
...
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak5
+COMPONENT_VERSION=	5.0.6
+COMPONENT_SUMMARY=	The Squeak V5 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_FMRI=		runtime/squeak5
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+COMPONENT_LICENSE=	LICENSE
+COMPONENT_SRC=		opensmalltalk-vm
```
AndWac: COMPONENT_LICENSE= MIT
COMPONENT_LICENSE_FILE=squeak5.license

components/smalltalk/squeak5/Makefile
AndWac: I have checked the Github project pages. They have releases with source files, eg. https://github.com/OpenSmalltalk/opensmalltalk-vm/releases with https://github.com/OpenSmalltalk/opensmalltalk-vm/archive/202003021730.tar.gz
Maybe you can use this? Taking sources from a release and not the latest revision may also be wiser and less error prone.

components/smalltalk/squeak5/squeak5.p5m
```diff
@@ -0,0 +1,53 @@
...
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/UUIDPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/UnicodePlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/UnixOSProcessPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/XDisplayControlPlugin.so
+file mode=0555 path=usr/lib/$(MACH64)/squeak/5.0-202004291145/squeak
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-display-X11.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-display-null.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-sound-null.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-sound-pulse.so
+file path=usr/share/man/man1/inisqueak.1
```
AndWac: When I run gmake sample-manifest I'll get more libraries:
╰─➤  diff sample-manifest.p5m sample-manifest.p5m.orig
59,62d58
< file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-sound-OSS.a
< file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-sound-OSS.so
< file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-sound-Sun.a
< file path=usr/lib/$(MACH64)/squeak/5.0-202004291145/vm-sound-Sun.so

Do you have the audio headers installed?
─➤  pkg list header-audio
system/header/header-audio                        0.5.11-2020.0.1.19654      i--

cstes: You have a build server that has more packages than my system, I think.  What happens is, the "configure" script detects some audio headers and then enables the build of ALSA, OSS, Sun etc. depending on the header files it can find.   The configure script has an option -without-OSS but unfortunately it does not have an option --without-Sun so when it finds the sun audio header, it builds that plugin (which worked fine in your case but I suspect the plugin compiles, but does not work, it's old SPARC Sun audio I think).   Note that my personal preference would be to ship only the vm-sound-pulse.so module.   The squeakvm will check for modules in the following order: OSS, Sun, pulse.   This order is the reverse of what I want.   I want it to check first for pulse, then for OSS, then for Sun audio.    But by having only the vm-sound-pulse.so in the manifest, and only packaging pulse,  the problem is solved because the squeakvm will only find the pulse plugin so it will use that, which is fine.   If I'd package OSS, Sun and pulse plugins, then the "search order" of the squeakvm is doing the "wrong thing".   Wrong in the sense of not using the pulse-server unless the VM is started with the -vm-sound-pulse command line switch.   This is something that most users will not know about, in fact, it's something I discovered by experimenting a bit.   Note that on some other Linux distributions they also use "pulse audio" and some Squeak users also would like the search order to be changed so that it uses pulse if it can find it, but the VM is not doing that at this moment, and some Linux distributions that do not use pulseaudio object to the change.

components/smalltalk/squeak5/Makefile
cstes: I started by doing that, before I used the "git clone".   So I tested it but it does not seem to work.   The problem is - it seems - that scripts/updateSCCSVersions must be ran before configure (configure checks and aborts if it did not run).   Now that script seems to assume that it can issue git commands that fail if it is running on the unversion-end tree.   But I agree that this is something that can be fixed (and should not be so difficult).  And it will be a big win for this package Makefile.   That's true.   I tried working with COMPONENT_ARCHIVE_URL and it first the oi-userland much better, that's true.  I will try to do that if I can convince the "upstream" opensmalltalk of this.    The releases they have on their github.com page have this problem for me ...   I don't know how the Microsoft VisualC guys (who do not use UNIX "configure") use those releases but for the world of UNIX "configure" it does not currently work - unfortunately.    I agree that they should fix that.

components/smalltalk/squeak5/Makefile
cstes: done

components/smalltalk/squeak5/squeak5.p5m
AndWac: Valid reasoning, but you should have the libs in the sample-manifest because it's not only the base for the package's p5m file but also let you (or another developer) know whether something has changed during an update of the package. If there is something missing now it might be visible next time and then included into squeak5.p5m creating all the hassle you describe.

cstes: So basically short story : it is intentional that there is only vm-sound-pulse in the manifests of squeak4 and squeak5.  I'm not going to change that.

components/smalltalk/squeak5/squeak5.p5m
cstes: I could have a try for compiling the Pharo VM , or CUIS or Newspeak.  My understanding is that those are smalltalk dialects. I think there also exist GNU smalltalk and many other implementations, they can all be IPS packaged - if there is interest in that.   This is because Smalltalk originated in 1972 or so , so it has a long history and a history that knew many dialects and variants.   Maybe there exist more Smalltalk variants than that there exist UNIX variants ! Probably it is not a lot of effort (I may be wrong) to compile some of those VMs.   But they can be in separate packages : for example an IPS package "pharo" for the "Pharo Virtual Machine".

components/smalltalk/squeak5/Makefile
cstes: I tested : git clone --depth 1 --branch sun
That works but it is not much faster and it is not really checking out the tag sun-v5.0.6
Maybe I'm imaging things but what seems to work and what seems faster is
git clone --depth 1 --branch sun-v5.0.6  (which is not the branch but the tag)
So I'll use that.

cstes: No sorry it does not work;   I think I'll leave the "git clone" as it is right now.   I agree with you that I should work on making COMPONENT_ARCHIVE_URL work .  That fits the oi-userland framework best.   I think it can be done ...   it's currently not how their opensmalltalk-vm "configure" works, but should be possible.

cstes: I just commited the original code which was using COMPONENT_ARCHIVE_URL.   I think we both agree it 's best.   It's much faster than GIT and using git does not make sense for the oi-userland download. You'll notice that it currently breaks the build, but it should be possible to fix the opensmalltalk-vm configure script, so I know it's currently broken but I'm working on the fix for that.

AndWac: perfect

cstes: The build should be much faster now because it does not "git clone" any longer,  it works for me ...  The solution to use "patches" is more work for me for subsequent updates (e.g. for version 5.0.7) but the advantage is that it is much faster for rebuilds for userland, I think .   Also, the opensmalltalk project may be interested and they may have a solution on the long term that avoids this problem.   For now, I currently work with 'patches' which is a solution to search a "balance" between the requirements of the userland framework and the requirements of the opensmalltalk project.

AndWac: Yes, this sounds by far better. BTW: You should change to either IRC #oi-dev or mail oi-dev@openindiana.org for communication. IRC is logged by echelog.com so you can read older messages. Albeit the service will close soon and there are discussions on #illumos to provide our own service for it.

AndWac: This seems to work a lot better. Now that you have a fixed version you need to update your sample-manifest.p5m because it has changed.

components/smalltalk/squeak5/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,50 @@
...
+
+file path=usr/bin/squeak
+file path=usr/doc/squeak/COPYING.gz
+file path=usr/doc/squeak/COPYRIGHT.gz
+file path=usr/doc/squeak/LICENSE.gz
+file path=usr/doc/squeak/README.Contributing.gz
+file path=usr/doc/squeak/README.Keyboard.gz
+file path=usr/doc/squeak/README.Sound.gz
+file path=usr/lib/$(MACH64)/squeak/$(PLUGIN_REV)/B3DAcceleratorPlugin.so
+file path=usr/lib/$(MACH64)/squeak/$(PLUGIN_REV)/FileAttributesPlugin.so
```
AndWac: You seem to have edited the sample-manifest.p5m. That prohibits its intention. If you run gmake sample-manifest again it will create a different file and thus it looks like something has to be changed in squeak5.p5m.
So, generate this file and leave it untouched. Do all necessary changes in the squeak5.p5m (as you already did). So it will be possible later to check for real changes by comparing the old and the new sample-manifest.p5m. Unfortunately this is not perfect as you can see with your $(PLUGIN_REV) variable use.

components/smalltalk/squeak5/Makefile
```diff
@@ -0,0 +1,80 @@
...
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak5
+COMPONENT_VERSION=	5.0.6
+GIT_TAG=		sun-v$(COMPONENT_VERSION)
+PLUGIN_REV=		5.0-202005010835-sun
+COMPONENT_SUMMARY=	The Squeak V5 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_FMRI=		runtime/squeak5
+COMPONENT_CLASSIFICATION=	Development/Other Languages
```
AndWac: For me it's ok that you have introduced the new category "smalltalk" in the file system hierarchy. We need to wait for alp's decission whether the new package should be moved to the runtime folder. But I would like to have the COMPONENT_FRMI match the category (so in the end the squeak5 folder should either be moved into the runtime folder or the COMPONENT_FRMI should be changed to smalltalk/squeak5.

Remember, when everything is done you should squash all commits into a single one by means of git rebase.

components/smalltalk/squeak5/manifests/sample-manifest.p5m
cstes: Ok, I'll regenerate the sample-manifest file, but I'll do that for a new version , say 5.0.7.
It is a good time to create 5.0.7 because I can merge new changes from opensmalltalk-vm into a new package (so the new package will effectively be a more recent squeakvm).  The build process will be the same,  there will be a .tar.gz for the 5.0.7 so it will then build from the tar.gz for the 5.0.7 release.   Finally I'd like to add a few words about "pharo".   Pharo VM compiles fine ;  it works.     I tried with "Pharo8.0-SNAPSHOT-64bit-a153e04.image" .

components/smalltalk/squeak5/Makefile
cstes: The fmri runtime/squeak5 was chosen randomly without much thought.

I like the fmri smalltalk/squeak5 better.

Also I have been thinking about just calling the package "smalltalk/squeak" but the problem that I have there, is that the "build" system of squeak4 and squeak5 is different and that perhaps a mediator package could be made called smalltalk/squeak.

I'll change it to use smalltalk/squeak5 but if I understand correctly the decision in the end depends on the folder that will be used in the end ...

Perhaps it makes sense to have a folder 'smalltalk' given the number of packages that could be stored there, given the number of different implementations and dialects.

components/smalltalk/squeak5/manifests/sample-manifest.p5m
AndWac: Have you tried something that uses their new UFFI? The image relies heavily on it as far as I remember.

components/smalltalk/squeak5/Makefile
AndWac: You cannot randomly choose it. Eg. the file hierarchy is documented in doc/categories-layout.md

pyhalov: Please, don't introduce new top-level FMRI pkg:/smalltalk/blah. 
A new component's category (not related to FMRI) is justified if there's a dozen sub-components (e.g, if you plan to add several smalltalk libraries). If these are just compilers, they can live in developer/ (with corresponding FMRI) for compilers and  in runtime/ (with corresponding FMRI) for interpreters.

pyhalov: Sorry, so far I don't have time to look at this PR.

AndWac:
There one more thing to add: SqueakV50.sources file is exptected in usr/lib/$(MACH64)/squeak/$(PLUGIN_REV)/.
So best is you'll add it into a files/ folder in your project and add a new line to the squeak5.p5m file:
file files/SqueakV50.sources path=usr/lib/$(MACH64)/squeak/$(PLUGIN_REV)/SqueakV50.sources

components/smalltalk/squeak5/Makefile
cstes: In my judgement the Squeak VM is an "interpreter" although https://squeak.org/features/ states (probably quite rightfully so) that it provides "just-in-time compilation, stack-to-register mapping, and aggressive in-line message caching yield efficiency in executing Smalltalk byte code".     But basically given pyhalov's remark to either use developer/ or runtime/  I think runtime/ was the right choice.   This is what I chose initially.   Then AndWac's remark is just that I have to move the component to the components/runtime/ directory and get rid of the components/smalltalk directory.   I thought that adding a components/smalltalk directory made sense, but it is also true that this is not compatible with the "fmri" name.

cstes:
> There one more thing to add: SqueakV50.sources file is exptected in usr/lib/$(MACH64)/squeak/$(PLUGIN_REV)/.
> So best is you'll add it into a files/ folder in your project and add a new line to the squeak5.p5m file:
> file files/SqueakV50.sources path=usr/lib/$(MACH64)/squeak/$(PLUGIN_REV)/SqueakV50.sources

No that is not true.   Basically you can just download SqueakV50.sources from http://squeak.org and save it in your home directory or any other directory, and then also unzip the image and changes file , so then you have three files : sources, changes, images and start squeak there (with the image).    If you then choose "Save as" in the Squeak GUI then you save your own "dialect" or "version" of the image.   It is the very nature of Smalltalk that people work on their own "images" and you get thousands of different images.   It makes no sense in my opinion to start IPS (Solaris Package System) the images and change files.   Smalltalk has its own package management tools for that.   Basically what I know is "Squeak Map" and "Monticello".    I used Squeak Map and Monticello succesfully with this port of Squeak to OpenIndiana.   But my understanding is that the Squeak system has thus its own mechanisms to exchange and distribute packages, and there is no need to IPS package the source, change and image files.

cstes:
Also if I'd package the SqueakV50.sources file then I'd have to go through the process of defining  a COMPONENT_ARCHIVE_URL for it to indicate where I download the source file (what its URL is, or the origin of that file).   My preference is just to package the VM (virtual machine) and it's up to the users to figure out that they have to download the other 3 components : source, change, image from http://squeak.org.    This is not nice as it is then not working "out of the box" but it sufficient to indicate in some documentation (that is done on http://squeak.org) that that's how it works.

components/smalltalk/squeak5/Makefile
cstes: I am going to close the squeak4 and squeak5 pull requests.   And submit a new pull request called "squeak" containing the 32-bit squeak4 and 64-bit squeak5 packages in one pull request: called "Add squeak".    I'll add the packages to the components/runtime folder.   Also in the future I will modify the squeak5 package to support both 32-bit and 64-bit , and I can also modify squeak4 to support both 32-bit and 64-bit, but currently squeak4 is 32-bit only and squeak5 is 64-bit only.   So I will do a new pull request now for squeak5 version 5.0.7 (going up one level and this will also incorporate new changes from opensmalltalk-vm).

AndWac:
Read again what I wrote: you add a files folder to your package folder and put the SqueakV50.sources file there. That is, make it part of your commit. Add a single line to the package manifest (squeak5.p5m). Sure, it's just for convenience for the user. But the package is named squeak5, so it's adequate to add the sources file for exactly these kind of images. If somebody runs a different dialect he/she can add the appropriate sources file to the image's folder.

components/smalltalk/squeak5/Makefile
AndWac: That is ok, but you should be aware that you'll open an new can of worms. Eg. mediator is not a separate package but part of every package that is an alternative. See for an example the different node packages in the runtime folder. So if you put everything in one package you cannot use the mediator feature to let users choose the default Smalltalk they want to use. They always need to add the corresponding path.

components/smalltalk/squeak5/manifests/sample-manifest.p5m
cstes: No but I can imagine that that would perhaps not work.  I haven't tried it.   But we have to start somewhere ... It is a complex issue because on the one hand there is the VM (virtual machine) and on the other hand there's the development of the Smalltalk "image" which is let's call it "application level programming".

components/smalltalk/squeak5/Makefile
cstes: I am for the moment moving the packages to components/runtime.    Smalltalk is like "ruby" and "python" in the sense that Smalltalk may provide lots of "modules" (interpreted by the "vm" or virtual machine).    But I'm not convinced that a folder "smalltalk" is required.   We could change it later again.   The main point of "AndWac"'s remark is that component folder and FMRI in the package name should match, if I understand correctly.

cstes:
> Read again what I wrote: you add a files folder to your package folder and put the SqueakV50.sources file there. That is, make it part of your commit. Add a single line to the package manifest (squeak5.p5m). Sure, it's just for convenience for the user. 

But SqueakV50.sources doesn't change between squeak5-5.0.6 and squeak5-5.0.7 or subsequent versions squeak5-5.0.8 etc.   

So SqueakV50.sources can be delivered in a separate package  "squeak5-sources".  That would by the way have an FMRI like  smalltalk/squeak5-sources because it is a "text file" with Smalltalk code in it.

For example:
$ file SqueakV46.sources
SqueakV46.sources:	ascii text

$ strings SqueakV46.sources| more
'From Squeak4.6 of 8 July 2015 [latest update: #15102] on 8 July 2015 at 10:05:08 pm'!

components/smalltalk/squeak5/Makefile
AndWac: Not exactly. We have some packages which differ but for new packages I prefer it (easier to find the package sources).

AndWac:
I know how it works. I am not sure whether putting the sources file into a speparate package is better. This sources package would only hold a single file. If you do it then make the squeakv5 package dependant on squeakv5-sources package.

cstes:
> I know how it works. I am not sure whether putting the sources file into a speparate package is better. This sources package would only hold a single file. If you do it then make the squeakv5 package dependant on squeakv5-sources package.

For the moment I am not going to add the SqueakV5.0.sources or Squeak4.6.sources to the packages.  Also I'm not going to package them in a different package.

I have to see in the future what feedback of any possible users (if there are any) will be.

Note that some of the source code of the VM is generated with a VMMaker tool, which has its own configuration file.   I have no experience with that.   But I can imagine that there are configuration file for VMMaker , image, source, changes etc. that would have to be included as well.

Anyway for the moment I'll stick with the manifest as it is now.

If you (or any other user) uses Squeak Smalltalk --  from this discussion I get the impression that you are familiar with Squeak Smalltalk, perhaps more than I am -- then you can later on still send feedback or explain how I should improve it.

cstes:
The classification for those 2 packages : squeak4 and squeak5 is 

COMPONENT_CLASSIFICATION=       Development/Other Languages

I picked the "Other Languages" category just by lack of inspiration.

Can I put there something like "Development/Smalltalk" as classification please ?

For example there exists Development/Python and Development/Ruby and Development/Perl.

So can I simply put in there a name of myself Development/Smalltalk please , or does this have to be approved first ?

AndWac:
Classification Development/Smalltalk is ok for me.

cstes:
I am encountering now a problem : it is apparently not allowed to simply choose a new classification:

ERROR pkglint.manifest008.6       info.classification attribute in pkg:/runtime/squeak5@5.0.7,5.11-2020.0.1.0 does not contain one of the values defined for the section Development: C,C++,Databases,Distribution Tools,Editors,Fortran,GNOME and GTK+,GNU,High Performance Computing,Integrated Development Environments,Java,Objective C,Observability,Other Languages,PHP,Perl,Python,Ruby,Source Code Management,Suites,System,X11 from /usr/share/lib/pkg/opensolaris.org.sections

AndWac:
Sorry, I didn't know about it. So we would need to change pkg (which is in  a separate repository) to allow for a new classification. Too much of a hassle at the moment.

cstes:
Ok, I have classified the packages again as "Other Language", for the moment.   On my machine both squeak4 and squeak5 are building.  I did not "git rebase" anything so far, also I'm not sure why I should do that.   In any case it seems safest to still wait a while to see whether anybody else wants something changed ...

PR #5702, opened 2020-05-01
Add bhyve firmware package

citrus-it:
This is the firmware that will be needed to boot VMs under bhyve, when bhyve is merged to illumos-gate. Work is currently ongoing against https://github.com/illumos/ipd/blob/master/ipd/0015/README.md

Adding it to OpenIndiana will make bhyve usable when it lands and will facilitate testing in advance.
I've tested each firmware file by booting a VM on OmniOS, which already has bhyve. 

This is a pretty old firmware version but it's the one currently shipped with SmartOS and OmniOS.
There is work being done to update the bhyve firmware in FreeBSD to a much newer version and that will be brought over to https://github.com/illumos/uefi-edk2 when it's ready.

### DIFF
--- a/components/openindiana/bhyve-fw/Makefile
+++ b/components/openindiana/bhyve-fw/Makefile
@@ -1,0 +1,104 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 OmniOS Community Edition (OmniOSce) Association.
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		uefi-edk2
+COMPONENT_VERSION=	20180309
+COMPONENT_FMRI=		system/bhyve/firmware
+COMPONENT_CLASSIFICATION=System/Core
+COMPONENT_SUMMARY=	UEFI-EDK2(+CSM) firmware for bhyve
+COMPONENT_SRC=		uefi-edk2-il-udk2014.sp1-1
+COMPONENT_ARCHIVE=	il-udk2014.sp1-1.tar.gz
+COMPONENT_ARCHIVE_HASH= \
+	sha256:d894d1ca61beeb0a1b76c2d6bbd137b02feb4783e4307f36e76efa29d415f6a8
+COMPONENT_ARCHIVE_URL= \
+	https://github.com/illumos/uefi-edk2/archive/$(COMPONENT_ARCHIVE)
+COMPONENT_PROJECT_URL=	https://github.com/illumos/uefi-edk2
+COMPONENT_LICENSE=	simplified-BSD
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_BUILD_ENV =
+COMPONENT_BUILD_ENV += GCCVER=4.4.4
+COMPONENT_BUILD_ENV += GCCPATH=/opt/gcc/4.4.4
+COMPONENT_BUILD_ENV += GMAKE=/usr/bin/gmake
+COMPONENT_BUILD_ENV += GAS=/usr/bin/gas
+COMPONENT_BUILD_ENV += GAR=/usr/bin/gar
+COMPONENT_BUILD_ENV += GLD=/usr/bin/gld
+COMPONENT_BUILD_ENV += GOBJCOPY=/usr/bin/gobjcopy
+
+# Four variants of the bhyve firmware are built;
+# DEBUG and RELEASE, and with and without CSM.
+VARIANTS =	DEBUG DEBUG.csm RELEASE RELEASE.csm
+
+BUILD_64 =	$(VARIANTS:%=$(BUILD_DIR)/%/.built)
+INSTALL_64 =	$(VARIANTS:%=$(BUILD_DIR)/%/.installed)
+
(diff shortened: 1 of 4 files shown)


components/openindiana/bhyve-fw/bhyve-fw.p5m
```diff
@@ -0,0 +1,30 @@
+#
```
pyhalov, line 1: Can you replace bhyve-fw.p5m with sample-manifest (adding your copyright there)?

components/openindiana/bhyve-fw/Makefile
```diff
@@ -0,0 +1,69 @@
...
+COMPONENT_BUILD_ENV += GCCVER=4.4.4
+COMPONENT_BUILD_ENV += GCCPATH=/opt/gcc/4.4.4
+COMPONENT_BUILD_ENV += GMAKE=/usr/bin/gmake
+COMPONENT_BUILD_ENV += GAS=/usr/bin/gas
+COMPONENT_BUILD_ENV += GAR=/usr/bin/gar
+COMPONENT_BUILD_ENV += GLD=/usr/bin/gld
+COMPONENT_BUILD_ENV += GOBJCOPY=/usr/bin/gobjcopy
+
+$(BUILD_DIR_64)/.built: $(SOURCE_DIR)/.unpacked
+	$(MKDIR) -p $(BUILD_DIR_64)
```
pyhalov: Should $(BUILD_DIR_64)/.built target just depend on individual  build targets ? Like in https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/editor/emacs/Makefile

components/openindiana/bhyve-fw/Makefile
```diff
@@ -0,0 +1,69 @@
...
+COMPONENT_ARCHIVE_URL= \
+	https://github.com/illumos/uefi-edk2/archive/$(COMPONENT_ARCHIVE)
+COMPONENT_PROJECT_URL=	https://github.com/illumos/uefi-edk2
+COMPONENT_LICENSE=	simplified-BSD
+
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_BUILD_ENV =
+COMPONENT_BUILD_ENV += GCCVER=4.4.4
```
pyhalov, line 37: Why do we have to use GCC 4?

components/openindiana/bhyve-fw/bhyve-fw.p5m
citrus-it: The sample one is missing the directory entries and the links, as well as the 555 permissions on the files themselves. Can you let me know what I need to do?

components/openindiana/bhyve-fw/Makefile
citrus-it: Unfortunately this only builds with up to gcc 4.9, and SmartOS and OmniOS are building it with the illumos 4.4.4 compiler. It's building standalone UEFI firmware binaries so it's quite a delicate process.

The new branch which will be coming later in the year builds with more recent compilers.

components/openindiana/bhyve-fw/Makefile
citrus-it: I can restructure it that way if you prefer - are you thinking of a DEBUG and RELEASE target?
You can probably tell that I'm not familiar with the build framework - I had to add the .unpacked dependency to get it to actually download and unpack the archive.
I'm happy to restructure it in whatever way you think is best but I'll need a couple of extra pointers.

components/openindiana/bhyve-fw/bhyve-fw.p5m
pyhalov: 1) dir actions are not compulsory, if directories are not empty.
2) links should be added separately, before auto-generated section from sample-manifest (Have I pointed to https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/best-practices.md ?)
3) mode is better set with transform (or in individual action if you prefer).

components/openindiana/bhyve-fw/Makefile
pyhalov: Yes. Usually you have `$(BUILD_DIR)/%/.built:  $(SOURCE_DIR)/.prep` dependency from one of the specific makefiles (justmake.mk, configure.mk, etc). Perhaps, you can use justmake.mk, overwrite COMPONENT_INSTALL_ACTION, COMPONENT_BUILD_ACTION (perhaps COMPONENT_COPY_ACTION  if you need) and perhaps defining two variants - like in https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/editor/vim/Makefile#L52. 
This also could be ugly, but perhaps avoids using so much shell code.
Or just split current build action in two - debug and release build.

components/openindiana/bhyve-fw/Makefile
pyhalov: OK.

components/openindiana/bhyve-fw/Makefile
pyhalov: I'd looked at it myself, but for now I just have no time.

citrus-it: No problem, thanks for the feedback, I'll take a look.

components/openindiana/bhyve-fw/Makefile
```diff
@@ -0,0 +1,100 @@
...
+	COMPONENT_POST_INSTALL_ACTION = $(LN) -s $(COMPONENT_INSTALL_FILE) \
+	$(PROTO_DIR)/usr/share/bhyve/firmware/BHYVE_CSM.fd
+
+build:		$(BUILD_64)
+
+install:	$(INSTALL_64)
+
+test:		$(NO_TESTS)
+
+# Auto-generated dependencies
```
pyhalov, line 104: Please, add developer/acpi to REQUIRED_PACKAGES above this line like this 
```
# Build dependencies
REQUIRED_PACKAGES += developer/acpi

```

citrus-it: done

pyhalov (approved):

PR #5705, opened 2020-05-03
Preparare for 2020.04 snapshot

pyhalov:


components/openindiana/release/files/release
```diff
@@ -1,4 +1,4 @@
-             OpenIndiana Hipster 2019.10 (powered by illumos)
+             OpenIndiana Hipster %%DISTRIBUTION_VERSION%% (powered by illumos)
```
pyhalov: Note, first line in motd file format is used at least by installer to determine URL to release notes, by openindiana-welcome to get distribution version and perhaps by some other software, so we can't just change it to '$(DISTRIBUTION_NAME) $(DISTRIBUTION_VERSION)'

make-rules/shared-macros.mk
```diff
@@ -100,6 +100,9 @@ space := $(empty) $(empty)
 
 ROOT =			/
 
+# Distribution version
+DISTRIBUTION_NAME = OpenIndiana
```
pyhalov: So, question is if we should include 'Hipster' in DISTRIBUTION_NAME :)

Toasterson: Yes I would say so. It makes us stand out from the older Release based OpenIndiana versions. Or we could follow OmniOS here.

Toasterson:
Will you spin images once this is merged? I will spin Vagrant boxes as soon as we have images.

pyhalov:
Yes, if testing them doesn't find any problems, I'll announce snapshot.

AndWac:
Am 03.05.20 um 14:42 schrieb Alexander Pyhalov:
>
> Yes, if testing them doesn't find any problems, I'll announce snapshot.
>
If you can provide a link I can test iso's, too.

Regards
Andreas

pyhalov:
So far I have no isos, have to rebuild illumos-gate after this PR

AndWac:
Is see. If you want me to test just write me an email.

What do you think about #5706? Despite the fact that he still needs to
squash it seems to be ok for me. I still need to check both VMs again,
tough.

Am 03.05.20 um 18:25 schrieb Alexander Pyhalov:

PR #5706, opened 2020-05-03
Add Squeak

cstes:
Add squeak4 version 4.16.7 and squeak5 version 5.0.7.

### DIFF
--- a/components/runtime/squeak4/Makefile
+++ b/components/runtime/squeak4/Makefile
@@ -1,0 +1,86 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 David Stes
+#
+
+
+#
+# Package Makefile for the "Classic VM" Squeak Smalltalk V4 system
+# See http://squeak.org and http://squeakvm.org
+#
+# The squeak4 vm can only run Squeak4 images; it will not work with 
+# Squeak5 images; squeak5 uses  different technology called "Cog Spur"
+#
+# squeak4 uses cmake and a custom configure script (not GNU autoconf)
+# in order to build this package, "cmake" must be installed
+#
+
+# squeak4 can be built both in 32-bit and 64-bit 
+# but generally SqueakV4 images are 32bit
+BUILD_BITS=32
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak-4
+COMPONENT_VERSION=	4.16.7
+COMPONENT_SUMMARY=	The Squeak V4 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_ARCHIVE_URL=  http://squeakvm.org
+COMPONENT_FMRI=		runtime/squeak-4
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+SVN_REPO=		http://squeakvm.org/svn/squeak/trunk/
+SVN_REV=		3775
+SVN_HASH=  sha256:918f3576a148bc41ae10a929f3ac53dfa6984208604a9c28d4a712f58db55748
+
+# See http://wiki.squeak.org/squeak/933
+# See http://wiki.squeak.org/squeak/159
+# there's 2 license lines in the manifest, the MIT squeak.license and this one
+COMPONENT_LICENSE=      Squeak4
+COMPONENT_LICENSE_FILE= squeak4.license
+
+TEST_TARGET= $(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+PATH=$(PATH.gnu)
+
+# squeak4 configure looks for .svn or svnversion file, but
+# prep-svn.mk did a svn export and created an unversioned copy of the sources
+COMPONENT_PRE_CONFIGURE_ACTION = \
+( echo $(SVN_REV) > $(SOURCE_DIR)/platforms/unix/svnversion )
+
+CONFIGURE_SCRIPT= $(SOURCE_DIR)/platforms/unix/cmake/configure
+CONFIGURE_OPTIONS +=	--src=$(SOURCE_DIR)/src
+
+# Makefile install/strip doesn't work
+COMPONENT_POST_INSTALL_ACTION = \
+( $(STRIP) $(PROTOUSRLIBDIR32)/squeak/$(COMPONENT_VERSION)-$(SVN_REV)/squeakvm )
+
+REQUIRED_PACKAGES += x11/library/mesa
+REQUIRED_PACKAGES += system/library/dbus
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += library/audio/gstreamer
+REQUIRED_PACKAGES += library/audio/pulseaudio
+REQUIRED_PACKAGES += library/desktop/cairo
+REQUIRED_PACKAGES += library/desktop/pango
+REQUIRED_PACKAGES += library/glib2
+REQUIRED_PACKAGES += library/libffi
+REQUIRED_PACKAGES += library/security/openssl
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/freetype-2
+REQUIRED_PACKAGES += system/library/libdbus
+REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += x11/library/libx11
+REQUIRED_PACKAGES += x11/library/libxext
+REQUIRED_PACKAGES += x11/library/libxrender
--- a/components/runtime/squeak4/squeak-4.p5m
+++ b/components/runtime/squeak4/squeak-4.p5m
@@ -1,0 +1,67 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 David Stes
+#
+
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license squeak.license license='MIT'
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform file -> add pkg.depend.bypass-generate libGL\.so\.1>
+
+file usr/bin/squeak path=usr/bin/squeak4
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/ckformat mode=0555
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.AioPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.B3DAcceleratorPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.ClipboardExtendedPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.DBusPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.FT2Plugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.FileCopyPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.GStreamerPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.HostWindowPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.ImmX11Plugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.KedamaPlugin2
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.Mpeg3Plugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.RomePlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.ScratchPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.Squeak3D
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.SqueakFFIPrims
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.SqueakSSL
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.UUIDPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.UnicodePlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.UnixOSProcessPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.XDisplayControlPlugin
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.vm-display-X11
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.vm-display-custom
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.vm-display-null
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.vm-sound-custom
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.vm-sound-null
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/so.vm-sound-pulse
+file path=usr/lib/squeak/$(COMPONENT_VERSION)-3775/squeakvm mode=0555
+file path=usr/share/doc/squeak-$(COMPONENT_VERSION)-3775/LICENSE
+file usr/share/man/man1/squeak.1 path=usr/share/man/man1/squeak4.1
+
+# the squeak script uses test -L (test for symbolic link) - use hardlink
+
+hardlink path=usr/bin/squeak target=squeak4 mediator=squeak \
+    mediator-version=4
+hardlink path=usr/share/man/man1/squeak.1 target=squeak4.1 mediator=squeak \
+    mediator-version=4
+
--- a/components/runtime/squeak5/Makefile
+++ b/components/runtime/squeak5/Makefile
@@ -1,0 +1,103 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 David Stes
+#
+
+
+#
+# Package Makefile for the "OpenSmalltalk Cog Spur" Squeak Smalltalk V5 system
+# See http://squeak.org and http://opensmalltalk.org
+#
+
+# opensmalltalk-vm can be built both in 32 and 64bit
+# but generally SqueakV5 images are 64bit
+BUILD_BITS=64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak-5
+COMPONENT_VERSION=	5.0.7
+GIT_TAG=		sun-v$(COMPONENT_VERSION)
+PLUGIN_REV=		5.0-202005021055
+COMPONENT_SUMMARY=	The Squeak V5 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_FMRI=		runtime/squeak-5
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+
+# add profile Squeak for pthread_setschedparam failed: Not owner issue
+# run squeak via 'pfexec squeak' if 'userattr profiles' includes Squeak
+PKG_PROTO_DIRS += $(COMPONENT_DIR)/files
+
+# See http://wiki.squeak.org/squeak/933
+# See http://wiki.squeak.org/squeak/159
+# there's 2 license lines in the manifest, the MIT squeak.license and this one
+COMPONENT_LICENSE=	Squeak5
+COMPONENT_LICENSE_FILE=	squeak5.license
+
+COMPONENT_SRC=		opensmalltalk-vm-$(GIT_TAG)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:50c180b174048acd98a7d9705f0a40a3dc8bdc62b42d88eb7e20e71834184e4c
+COMPONENT_ARCHIVE_URL=	https://codeload.github.com/cstes/opensmalltalk-vm/tar.gz/$(GIT_TAG)
+
+TEST_TARGET= $(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+PKG_OPTIONS += -D PLUGIN_REV="$(PLUGIN_REV)"
+
+PATH=$(PATH.gnu)
+
+# opensmalltalk configure script checks for plugins files in builddir
+COMPONENT_PRE_CONFIGURE_ACTION = ( cp plugins.ext plugins.int $(BUILD_DIR_64) )
+
+# the default CFLAGS.gcc will be -m64 -O3 but this does not work for us
+# opensmalltalk code is not compatible with the gcc optimizer -O2 or higher
+# also the -g (debug) flag is intentional, perhaps omit-frame-pointer issues
+gcc_OPT=		-g -DNDEBUG -DDEBUGVM=0
+
+CONFIGURE_SCRIPT= $(SOURCE_DIR)/platforms/unix/config/configure
+CONFIGURE_OPTIONS +=	--with-vmversion=5.0
+CONFIGURE_OPTIONS +=	--disable-dynamicopenssl
+CONFIGURE_OPTIONS +=	--without-libtls
+CONFIGURE_OPTIONS +=	--with-src=spurstack64src
+CONFIGURE_OPTIONS +=	--disable-cogit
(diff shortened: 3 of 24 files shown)


components/runtime/squeak4/Makefile
```diff
@@ -0,0 +1,72 @@
...
+PATH=$(PATH.gnu)
+
+# squeak4 configure looks for .svn or svnversion file, but
+# prep-svn.mk did a svn export and created an unversioned copy of the sources
+COMPONENT_PRE_CONFIGURE_ACTION = ( echo $(SVN_REV) > $(SOURCE_DIR)/platforms/unix/svnversion)
+
+CONFIGURE_SCRIPT= $(SOURCE_DIR)/platforms/unix/cmake/configure
+CONFIGURE_OPTIONS +=	--src=$(SOURCE_DIR)/src
+
+REQUIRED_PACKAGES += service/opengl/ogl-select
```
pyhalov: Why is ogl-select in this list? Should it be mesa?

components/runtime/squeak4/Makefile
```diff
@@ -0,0 +1,72 @@
...
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak4
+COMPONENT_VERSION=	4.16.7
+COMPONENT_SUMMARY=	The Squeak V4 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_ARCHIVE_URL=  http://squeakvm.org
+COMPONENT_FMRI=		runtime/squeak4
```
pyhalov: If it's a compiler, it should be developer/squeak4.

components/runtime/squeak4/squeak4.p5m
```diff
@@ -0,0 +1,59 @@
...
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# see desktop/rss-glx/rss-glx.p5m:
```
pyhalov: Remove comment, it's misleading.

components/runtime/squeak5/squeak5.p5m
```diff
@@ -0,0 +1,58 @@
...
+
+# see desktop/rss-glx/rss-glx.p5m:
+<transform file -> add pkg.depend.bypass-generate libGL\.so\.1>
+<transform file -> add pkg.depend.bypass-generate vm-display-X11\.so>
+
+# for the sound plugins, we only package vm-sound-pulse
+# we do not package vm-sound-OSS and vm-sound-Sun
+# on my machine, vm-sound-Sun compiles, but it does not work
+
+file mode=0555 path=usr/bin/squeak
```
pyhalov: You don't need mode=0555, for bin files it's by default

pyhalov:
These two packages can't coexist. This contradicts OI policy. Well, it should have been written somewhere.... But one of the goals was to make it possible to install `pkg:/*`. Here we have duplicated files in different packages. You should install them to different places and use mediator to select implementation.
Look at "Delivering Multiple Implementations of a Given Application" chapter in http://docs.openindiana.org/dev/pdf/ips-dev-guide.pdf 
It's highly desirable  that these compilers could detect corresponding libraries even if started by real name when mediator is set to another version (like you can run gcc-6 without caring what is current gcc mediator set to).

components/runtime/squeak4/Makefile
AndWac: It's a virtual machine, similar to JVM or DOT.NET runtime. Together with appropriate image triple it is a very capable environment. The compiler is just a part of it. So, either it should either have its own category or runtime would be the best fit.

pyhalov: OK.

components/runtime/squeak4/Makefile
cstes: To be checked ...  I believe that "gmake REQUIRED_PACKAGES" did not find or discover mesa.

components/runtime/squeak4/squeak4.p5m
cstes: ok

components/runtime/squeak5/squeak5.p5m
cstes: ok

cstes:
The best solution would be to use 'mediators'.   Ideally I want to be able to install both packages at the same time.

components/runtime/squeak4/Makefile
cstes: Changed to mesa.

components/runtime/squeak4/squeak4.p5m
cstes: Removed comment.

components/runtime/squeak5/squeak5.p5m
cstes: Removed mode=0555.  Note that it is essential that the mode=0555 remains in there for /usr/lib/.../squeak   .  Basically the /usr/bin/squeak script is a wrapper script which is doing some setup and then executes the /usr/lib/ <squeak plugin directory>/squeak executable.  The wrapper script must tell the squeak executable under /usr/lib where to search for the plugins.  It has not builtin plugin directory.  By default it searches for plugins in the current working directory.

cstes:
I've implemented the mediator 'squeak'.  I can install both packages which is important in order to run Smalltalk V4 images and Smalltalk V5 images at the same time.

# pkg list squeak4 squeak5
runtime/squeak4 (userland)                        4.16.7-2020.0.1.0          i--
runtime/squeak5 (userland)                        5.0.7-2020.0.1.0           i--

# pkg mediator squeak
MEDIATOR            VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
squeak              system    5       system

pyhalov:
Otherwise looks good.

cstes:
> We usually have clang-9 or gcc-9, not clang9, gcc9. Is there a established practice to use squeak4/squeak5 names or should we correct them to squeak-4/squeak-5?

For pyhton I see :  components/python/python27/python-27.p5m
note that this is a mix python27 directory and with a hyphen in the manifest name.
and for the binary

link path=usr/bin/python target=python2.7 mediator=python mediator-version=2.7

for gcc the situation is different as it has:

link path=usr/bin/gcc target=../gcc/9/bin/gcc mediator=gcc mediator-version=9


Anyway what I could do is follow the python example.   I rename the manifests to squeak-4.p5m and the component FMRI name to squeak-4.   But just as in the python case, the folder name remains squeak4 (without hyphen) and the executable name remains squeak4 (without hyphen).   That's how it is for python2.7 (without hyphen),   I believe apache is also doing it like that.

For example the folder  web/apache24  has no hyphen.   The manifest has : apache-24.p5m

The binary has file path=usr/apache2/2.4/bin/httpd

Stylistic comments are welcome, I'm open to them.   If there's a style guide, that would even be better ...

pyhalov:
OK, there's a mess... There's /usr/bin/ruby23 and /usr/bin/ruby26, but /usr/bin/gcc-6 and /usr/bin/gcc-7.
But having name-VERSION seems to be a bit more common. 
As for guide, there's https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/best-practices.md

cstes:
The squeak.sh seems to use "kdialog".  I'll remove it.

cstes:
Please note that I'm packaging these scripts,  I don't write them.  The 'squeak' system goes back to Smalltalk-72 perhaps.   It has a long history of almost 50 years, I guess.   There's stuff in there that is not applicable at all to OpenIndiana.   Some of the shell scripts also have a tendency to be rather Linux or MacOS oriented.

cstes:
I'm now following the good style of mixing hyphen and non-hyphen names.   Like apache-24 and apache24 , python27 and python-27, I introduced a mix of squeak-4 and squeak4 etc.     I hope it's ok now.

cstes:
I've added a "mediator-implementation" so that multiple implementations of "squeak-5" can be installed.

# pkg mediator squeak
MEDIATOR            VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
squeak              system    5       system     stack-spur

In any case for the squeak-4 and squeak-5 packages the "mediator-version" is the differentiator.  And because most active development is going on in squeak-5 it can be expected that multiple implementations of sqeuak-5 exist (they do in fact, there exist many variants of the Smalltalk VM) these can all have version 5.

cstes:
I've discovered that in order to build those packages, you also have to have pkg:/developer/build/pkg-config installed on the server.   I believe this is normally installed.  Is it recommended to add pkg:/developer/build/pkg-config to the REQUIRED_PACKAGES in the Makefile ?  Or set REQUIRED_TOOL ?  There is a variable REQUIRED_TOOL but it doesn't seem to do anything.   Anyway pkg-config is only needed to configure, it is not required for pkgdepend (resolve the package dependencies can be done without pkg-config).

cstes:
For the moment I'm not adding pkg:/developer/build/pkg-config to the REQUIRED_PACKAGES although that pkg-config should be installed to build the packages.   But adding pkg-config is like adding "gmake" or "gcc" to the REQUIRED_PACKAGES, it is implied that the necessary build tools have to be installed so those are not true REQUIRED_PACKAGES as in the sense of required for "pkgdepend".

AndWac:
@pyhalov: Are you fine with this now? I have rebuild and reinstalled both and they seem to work fine. I'd like to merge if it's ok for you.

components/runtime/squeak4/squeak-4.p5m
```diff
@@ -0,0 +1,59 @@
...
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 David Stes
+#
+
+
+<transform file -> add pkg.depend.bypass-generate libGL\.so\.1>
```
pyhalov: Please, format manifest according to https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/doc/best-practices.md rule 10, this one is pkgfmt'd , but unreadable.

components/runtime/squeak4/Makefile
```diff
@@ -0,0 +1,83 @@
...
+
+CONFIGURE_SCRIPT= $(SOURCE_DIR)/platforms/unix/cmake/configure
+CONFIGURE_OPTIONS +=	--src=$(SOURCE_DIR)/src
+
+# squeak4 installs squeak.sh and wrapper script squeak in /usr/bin
+# the squeak.sh script uses "kdialog" and we don't include it
+# rename the scripts and the manpage squeak4.sh and squeak4
+# to avoid a conflict with the squeak-5 package (mediator "squeak')
+COMPONENT_POST_INSTALL_ACTION = ( \
+  $(CP) $(PROTOUSRBINDIR)/squeak $(PROTOUSRBINDIR)/squeak4; \
```
pyhalov: Likely we want $(MV) here to avoid confusing next person, looking on sample-manifest

components/runtime/squeak5/Makefile
```diff
@@ -0,0 +1,100 @@
...
+COMPONENT_INSTALL_ARGS=  	ROOT=$(PROTO_DIR)
+COMPONENT_INSTALL_TARGETS= 	install-squeak install-doc install-plugins
+
+# opensmalltalk-vm installs a wrapper script squeak in /usr/bin
+# rename those scripts and the manpage to squeak5
+# to avoid conflict with the squeak-4 package (mediator "squeak')
+# also their old Makefile installs in /usr/doc (for SunOS the location)
+# move that to $(PROTOUSRSHAREDOCDIR) /usr/share/doc as it is nowadays
+COMPONENT_POST_INSTALL_ACTION = ( \
+  $(CP) $(PROTOUSRBINDIR)/squeak $(PROTOUSRBINDIR)/squeak5; \
```
pyhalov: Likely we want $(MV) here to avoid confusing next person, looking on sample-manifest

components/runtime/squeak5/squeak-5.p5m
```diff
@@ -0,0 +1,60 @@
...
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
```
pyhalov, line 12: Again, please, reformat the manifest

components/runtime/squeak5/Makefile
alarcher: @pyhalov In such cases I find it easier to address renaming directly in the manifest, like:
`file usr/bin/squeak path=usr/bin/squeak4`

AndWac:
Am 07.05.20 um 12:00 schrieb Aurélien Larcher:
>
> *@alarcher* commented on this pull request.
>
>
> In components/runtime/squeak5/Makefile
> <https://github.com/OpenIndiana/oi-userland/pull/5706#discussion_r421387073>:
>
> > +CONFIGURE_OPTIONS +=	--without-vm-display-fbdev
> +CONFIGURE_OPTIONS +=	--without-npsqueak
> +
> +CONFIGURE_ENV +=	TARGET_ARCH="-m64"
> +
> +# the opensmalltalk Makefile does not use DESTDIR
> +COMPONENT_INSTALL_ARGS=  	ROOT=$(PROTO_DIR)
> +COMPONENT_INSTALL_TARGETS= 	install-squeak install-doc install-plugins
> +
> +# opensmalltalk-vm installs a wrapper script squeak in /usr/bin
> +# rename those scripts and the manpage to squeak5
> +# to avoid conflict with the squeak-4 package (mediator "squeak')
> +# also their old Makefile installs in /usr/doc (for SunOS the location)
> +# move that to $(PROTOUSRSHAREDOCDIR) /usr/share/doc as it is nowadays
> +COMPONENT_POST_INSTALL_ACTION = ( \
> +  $(CP) $(PROTOUSRBINDIR)/squeak $(PROTOUSRBINDIR)/squeak5; \
>
> @pyhalov <https://github.com/pyhalov> In such cases I find it easier
> to address renaming directly in the manifest, like:
> |file usr/bin/squeak path=usr/bin/squeak4|
>
|This is a nice idea, but I fear it would be easily overlooked when
upgrading later when the sample-manifest has changed. Especially when
it's done
by somebody else than the original author.
|

cstes:
The idea is to use a "mediator"  squeak to allow the system administrator to set /usr/bin/squeak as the user prefers.

# pkg mediator squeak
MEDIATOR            VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
squeak              local     4       system     

# ls -l /usr/bin/squeak
lrwxrwxrwx   1 root     root           7 May  7 10:37 /usr/bin/squeak -> squeak4

So squeak-4 and squeak-5 can be installed simultaneously.  

In the original source code , the system simply installs a script "/usr/bin/squeak", not a symbolic link.  We change in the Makefile with the $(MV) command the script to squeak4 or squeak5, but the "pkg mediator" logic sets up an appropriate symbolic link.

alarcher:
> The idea is to use a "mediator" squeak to allow the system administrator to set /usr/bin/squeak as the user prefers.
> # pkg mediator squeak
> 
> MEDIATOR VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
> squeak local 4 system
> # ls -l /usr/bin/squeak
> 
> lrwxrwxrwx 1 root root 7 May 7 10:37 /usr/bin/squeak -> squeak4
> 
> So squeak-4 and squeak-5 can be installed simultaneously.
> 
> In the original source code , the system simply installs a script "/usr/bin/squeak", not a symbolic link. We change in the Makefile with the $(MV) command the script to squeak4 or squeak5, but the "pkg mediator" logic sets up an appropriate symbolic link.

Using the line I mentioned avoids adding $(MV) command in post install steps but achieves the same.

cstes:
I'm fine with either approach ... if it is required to change it, I can give it a try.

Basically I've just discovered that the mediator approach has its own problems.

For example if I install squeak-4 and then squeak-5, the link /usr/bin/squeak works fine.
But if I change back to squeak-4 using "pkg set-mediator -V 4 squeak" it stops working,
perhaps the underlying script of squeak-4 has issues .  I get the following error:

squeak Squeak4.6-15102.image 
expr: syntax error
/usr/bin/squeak[17]: test: argument expected
cannot find executable file: squeakvm

But this does not seem a show-stopper to me,  as launching the Smalltalk V4 image using the command : squeak4 Squeak4.6-15102.image works fine.

Presumably in order to have the above syntax error fixed, I have to debug it under the mediator framework (the original author is completely unaware of that) and have the script fixed by the original author -- if that would ever happen, as these sources are 10-15 years old.

3eka:
Hi @cstes,

this error:
```
expr: syntax error
/usr/bin/squeak[17]: test: argument expected
```
sounds like wrong shell (#/bin/sh ?)... Just shot in the dark. I did not checked anything..

Regards.

cstes:
The squeak4 script uses the "test -L" and perhaps has an issue with the symbolic link set by the mediator.

does not work:
$ stat /usr/bin/squeak
  File: /usr/bin/squeak -> squeak4

the following seems to be working fine:

# stat /usr/bin/squeak
  File: /usr/bin/squeak -> /usr/bin/squeak4

I'm going to try to change the manifest to do this, to set the symbolic link with a full expansion if that is possible.

cstes:
I've made the mediator "squeak" use "hardlinks" so that it works with both the squeak-4 and squeak-5 package as the squeak-4 script has some code in it in the case of symbolic links (test -L) that does not seem to work.    If there's other suggestions , please feel free to add ...

cstes:
I have further simplified the Makefile and manifests and removed the COMPONENT_POST_INSTALL_ACTION, which is not strictly needed as Aurélien Larcher pointed out.   For me the important issue is just that both packages can be installed simultaneously.

# pkg list squeak-4 squeak-5
runtime/squeak-4 (userland)                       4.16.7-2020.0.1.0          i--
runtime/squeak-5 (userland)                       5.0.7-2020.0.1.0           i--

In order to start Smalltalk V4 images "squeak4" can be used, and in order to start Smalltalk V5 and V6-alpha images, "squeak5" can be used.   The mediator can be set by the user/system administrator as they want to have 'squeak' link to either squeak4 or squeak5.

cstes:
By the way, currently the classification is Development/Other Languages which is fine,  maybe in the far future (months, years) it may be Development/Smalltalk.   Given the fact that Smalltalk exists since about 1972, after those 50 years it can easily live for a while in the category "Development/Other Languages" for some years.  There's no hurry in the classification.

components/runtime/squeak4/Makefile
```diff
@@ -0,0 +1,75 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
```
pyhalov, line 9: Please, add author here

pyhalov: I suppose, you missed
```
#
# Copyright 2020 David Stes
#
```

components/runtime/squeak5/Makefile
```diff
@@ -0,0 +1,89 @@
...
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
```
pyhalov, line 11: The same copyright issue here

components/runtime/squeak5/Makefile
```diff
@@ -0,0 +1,89 @@
...
+
+PATH=$(PATH.gnu)
+
+# opensmalltalk configure script checks for plugins files in builddir
+COMPONENT_PRE_CONFIGURE_ACTION = ( cp plugins.ext plugins.int $(BUILD_DIR_64) )
+
+# the default CFLAGS.gcc will be -m64 -O3 but this does not work for us
+# opensmalltalk code is not compatible with the gcc optimizer -O2 or higher
+# also the -g (debug) flag is intentional, perhaps omit-frame-pointer issues
+CFLAGS=			-g -DNDEBUG -DDEBUGVM=0
```
pyhalov: If you want to overwrite only -03, you can use ` gcc_OPT = blah`

components/runtime/squeak4/Makefile
cstes: What exactly do you mean by add author ? The Makefile author ?  I'm not sure I understand what you mean by "you missed".   Also you have quoted a piece of code that states that a full copy of the CDDL should accompany the source.  I can copy CDDL.license into the directory (of the Makefile) but obviously the CDDL is not related to the squeak application.  Please clarify as I don't fully understand the reasoning.

components/runtime/squeak5/Makefile
cstes: I think you refer to a copyright issue of the Makefile.    (in the oi-userland framework).   There is also the license issue of the squeak work itself.   I have found an additional license file which I downloaded from http://www.squeak.org if you click on "LICENSE" it also has some info.   I'm going to include that as well ... However this point may be 'moot' as I think your remark on 'copyright issue' is essentially about the oi-userland Makefile.

components/runtime/squeak5/Makefile
cstes: done

cstes:
I've included an extra license with an interesting copyright in it:

Copyright (c) 1981-1982 Xerox Corp. All rights reserved.

Essentially as far as I understand the Squeak project is a far descendant of the Xerox Smalltalk team.

pyhalov:
Doesn't run for me.
```
Couldn't open Squeak image file `squeak.image`
```

AndWac:
This is because it's just the VM (runtime system).
Typically you will need a triplet of files: squeak.image, squeak.changes and squeak.sources (you can omit the sources file but then you'll get a warning message).
You need to download an appropriate triplet, eg. from here:
https://squeak.org/downloads. Then you can run it with squeak <image-name> from within the folder holding the three (or two) files.

pyhalov:
It seems, FreeBSD delivers both these files - https://github.com/freebsd/freebsd-ports/blob/master/lang/squeak/Makefile . I suppose, we also should do it.

AndWac:
Yes, but IMO that's just for convenience.
Derived from the same sources (probably just a different branch or some different flags during compilation and linking) you can create a VM for https://pharo.org.
Smalltalk has some unique properties. Usually you have a fixed sources file, eg. squeakV50.sources and then a lot of pairs of <file>.image and <file>.changes.
The .image file holds a binary blob of your Smalltalk "world" (the Smalltalk system and your project(s) / changes). The .changes file holds your source code changes against the .sources file (which changes only very seldom, eg. with a new major release).
Typical Smalltalkers have a few to several dozens of .image/.changes files and only a very few .sources files.
BTW: Squeak was one of the foundations of the OLPC (one laptop per child) project. Squeak has been created for educating children (eg. eToys), while Pharo is a fork for professional/business oriented development.

cstes:
> It seems, FreeBSD delivers both these files - https://github.com/freebsd/freebsd-ports/blob/master/lang/squeak/Makefile . I suppose, we also should do it.

You can go to http://files.squeak.org/  on their file server.   In the folder 4.6 you find

Squeak4.6-15102.zip	2015-07-09 14:25	7.9M	File
SqueakV46.sources.zip	2015-07-09 14:27	6.1M	File

Those are the actual Smalltalk systems.

In my opinion only the C written "virtual machine" (the VM) should be IPS packaged.

There are so many files on http://files.squeak.org.

It makes no sense to IPS package them.   Besides Squeak has its own "Packaging Systems" like "Squeak Map" and "Monticello".

cstes:
squeak is an interpreter,  only the runtime portion (the VM) should be in pkg:/runtime/

If somebody wants to package in IPS all of the Smalltalk images (I don't) they can have a category like pkg:/library/smalltalk or something similar for the images (which are ran by the "interpreter").

cstes:
is it ok to have two license lines in the manifest ?

I think it is.  I have included a squeak.license and another squeak4.license which gives credit to VPRI (the organisation www.vpri.org that was involved in the Squeak4 effort).   For squeak5 there is also a requirement I think for the squeak.license and on the other hand the "OpenSmalltalk" team squeak5.license.

It is less certain that two lines with the name license='MIT' are ok.
But the pkgllnt of OpenIndiana accepts it, and pkg info --license prints the 2 licenses.

cstes:
I've also tested
license squeak.license license=MIT
license squeak4.license license=Squeak4
that works as well.  But I like the 2 lines with license=MIT better.
Feel free to change the Makefile and change
-COMPONENT_LICENSE=      MIT
+COMPONENT_LICENSE=      Squeak4
if that would be required.   In the past , many years ago , they had a "Squeak license" but apparently the Squeak community changed to the MIT license during the Squeak3-Squeak4 time.

cstes:
squeak-5 is also capable of running the "Cuis" smalltalk image : http://cuis-smalltalk.org

If you download CuisV5.sources and Cuis5.0-4112.image from http://cuis-smalltalk.org

Then : squeak Cuis5.0-4112.image    will run "Cuis".

However I don't think the image files and source files of Cuis should be IPS packaged.

It's easier to just download those from GIT.

For the binary "sqeuak" package it makes a lot of sense to use IPS,  however for the image files and sources  - it theoretically can be done via IPS - but Smalltalk users will start to interchange , copy , save and modify images locally anyway, it doesn't make so much sense to "IPS package" the images, this will just be an extra level of "administration".

cstes:
I've added a Squeak profile:
/etc/security/exec_attr.d/squeak
/etc/security/prof_attr.d/squeak
for squeak-5 to run the VM using 'pfexec squeak'  (with the proc_priocntl privilege)
I've also tested classification Development/Smalltalk which works now in the new pkg@0.5.11-2020.0.1.5537.   However I don't see a reason to change the current Development/Other Languages to Development/Smalltalk until the new pkg is compatible with the next 'entire' and 'userland-incorporation'.  And in addition it seems it's easy to change the classification of a package, as in a small test, I see no difficulty in upgrading from squeak-5 in the classification 'Development/Other Language' to a future version of the squeak-5 package in the classification 'Development/Smalltalk'.

alarcher:
@cstes Thank you I'll look at your PR over the week-end and we can merge it if you feel it is ready.

cstes:
Yes thank you, in my perception it is ready.  By "ready" I mean that I'm satisfied with the 'mediator' implementation which allows to have both Squeak 4 images and Squeak 5 images (and Cuis Smalltalk images) ran by the VM.   Also I may change in the future the package Category from "Development/Other Languages" to "Development/Smalltalk" which is possible with the latest pkg , but I think it makes sense to start with the category that is accepted by pkglint on most OpenIndiana systems right now, and only after a few months, when the new file  /usr/share/lib/pkg/opensolaris.org.sections (which has now the Smalltalk entry) is propagated.

cstes:
Basically what I propose is to submit the packages right now in the category "Development/Other Languages"  as classification.   The change to classification "Development/Smalltalk" has already been tested (and it works) but I don't see the point in submitting that now, as it can be done in a few months again.   Also as far as I can see in a test, there is no issue with a change of classification when doing a "pkg update".  See previous comment about the latest "pkg" package and the  /usr/share/lib/pkg/opensolaris.org.sections file in it.

alarcher:
@cstes No problem ;) I tested publishing and merged. Could you add some documentation at https://github.com/OpenIndiana/oi-docs about how to use this package? Thank you for your contributions!

cstes:
The best documentation will be to refer to http://wiki.squeak.org/squeak and the world of Smalltalk documentation.  But I can do that by adding some documentation at https://github.com/OpenIndiana/oi-docs about that.   Some OpenIndiana specific things like "mediator" and "pfexec" (profile exec) within the Squeak profile can be added both in the oi-docs  and on the wiki.squeak.org.   In fact I'm asking in the Squeak community to update their docs.

PR #5708, opened 2020-05-03
WIP: salt 3000

Toasterson:
I am currently getting this although the set action is valid as far as I can understand it.

```
ERROR pkglint.manifest006         duplicate set actions on org.opensolaris.smf.fmri in pkg:/system/management/salt-27@3000.2,5.11-2020.0.1.0
```


Toasterson:
Oh dear this seems completely broken... need msgpack and backport and pycrypto updates

Toasterson:
Note to self currently need to package the following pip commands:
```
pfexec pip install --global-option=build_ext --global-option="-I/usr/include/gmp" pycrypto
pfexec pip install backports-abc==0.5
pfexec pip install 'msgpack!=0.5.5,<1.0.0,>=0.5'
```

And for some reason I get:
```
/usr/lib/python2.7/site-packages/Crypto/Util/number.py:57: PowmInsecureWarning: Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.
  _warn("Not using mpz_powm_sec.  You should rebuild using libgmp >= 5 to avoid timing attack vulnerability.", PowmInsecureWarning)
```
Which makes no sense since we have pkg:/library/gmp@6.1.2-2018.0.0.2

pyhalov:
```
set name=org.opensolaris.smf.fmri value=svc:/application/salt:minion value=svc:/application/salt:master
```
line in manifest is auto-generated from smf file, so you have duplicate here.

pyhalov:
If we update to 2019.2.4, will it go easier?

Toasterson:
Nope. It gets further in the installation but fails on dependencies later when trying to run the master or minion. My suggestion: remove the package from userland and point people to pip to install salt. This has never worked properly. So we have no users. Otherwise, they fixed things and did not let us know. It's definitely broken, and I can't remember that I got it working properly for myself.

`pip install --global-option=build_ext --global-option="-I/usr/include/gmp" salt` works best.

PR #5715, opened 2020-05-09
Add gst  (GNU smalltalk)

cstes:


cstes:
The name "gst" is what the package uses as the script interpreter (launch as gst).

# pkg info gst
             Name: runtime/gst
          Summary: GNU Smalltalk
         Category: Development/Other Languages

There exist for example gstd "GStreamerD" and other such daemons.  However thanks to IPS packaging it should not be a problem, if there is a conflict.

AndWac:
I am against merging gst.
It seems to be a dead project as it hasn't seen any updates for several years.

cstes:
There's multiple smalltalk implementations and I like the approach of GNU smalltalk which places Smalltalk code (classes and method implementation) in plain text files and which is definitely a more "batch" or "script" oriented implementation, although it also supports some form of images.

Also GNU smalltalk gives credit to the reference book "Smalltalk 80 language and implementation" the reference by Adele Goldberg and David Robson,  so as far as I'm concerned I find GNU smalltalk a fine implementation, which has its purpose and its own use.

Personally I prefer the package name "gst" which is the name of their interpreter (gst where 'g' stands for GNU and 'st' stands for smalltalk) anyway, and also because the "squeak" system would be a better package for carrying the name "smalltalk", because "squeak" is closer to the original team that developed Smalltalk.

Whereas 'squeak' is definitely more graphical browser oriented for class definitions,  the GNU smalltalk approach also has its advantages.

cstes:
I've tested the classification Development/Smalltalk, it works, but I don't see a reason to change the "Other Languages" classification for the next few months, until a new release of OpenIndiana is made.   Also upgrade of the package seems to work if the classification changes.

# pkg list pkg
package/pkg                                       0.5.11-2020.0.1.5537       i--

# pkg info gst
             Name: runtime/gst
          Summary: GNU Smalltalk
###          **Category: Development/Smalltalk**
            State: Installed
        Publisher: userland
          Version: 3.2.91
           Branch: 2020.0.1.0

cstes:
I can temporarily close this pull request, although that my impression is that the gst interpreter by itself is quite an interesting project and package.   In any case there exists a 'manifest' and 'Makefile' for it in my oi-userland fork , should this be requested someday by someone ...

alarcher:
@cstes The main problem is maintaining such old software in our userland tree. For instance I have fixed more than 90 packages just because of bad coding practices that throw an error with gcc-10. I still have about 60 packages to fix and most people would contributed them in the first place may not be able to help... As far as I am concerned you can keep the PR open but with a pending tag.

cstes:
Feel free to assign the "Pending" label.   Also note that this package is quite sensitive to GCC upgrades (I think) because it has its own libtool configure script and is quite sensitive to GCC upgrade I think.   I only tested with gcc 7.5.   So feel free to assign 'Pending' tag to it, or to close it.

alarcher:
@cstes If you have time to test with gcc-10 and fixup the possible failures we can consider it after the next migration.

cstes:
If I install gcc-10,  and then modify the make-rules/shared-macros.mk to set:

-GCC_VERSION =  7
+GCC_VERSION =  10
 GCC_ROOT =     /usr/gcc/$(GCC_VERSION)

i.e. make it use gcc-10 (GCC_VERSION 10) then "gnu smalltalk" (gst) builds and installs and publishes with gcc-10. 

Note that this works for me because in my Makefile I set 
BUILD_BITS=32

A difference is that with gcc-10 the "config.guess" will report a different value than with gcc-7 (default 32bit 64bit output).    So gst builds either way, 32bit and 64bit, but there is an issue in the 64bit case (which also happens with gcc-7).

Anyway I can build and install gst with gcc-10 in the 32bit case.   So basically the same as in the gcc-7 case.

cstes:
Actually as far as I can see, "gst" would be a nice addition to OpenIndiana Hipster.

cstes:
If I remove the BUILD_BITS or set it to 64, then the 'gmake build' fails with gcc-10.  However this is basically the same issue as with gcc-7.   Gst compiles and works in 64bit mode, but some optional packages do not install in the 64bit case.  Because those optional packages are not so easily removed (they are not truly 'optional') that's an issue for 64bit support.   However this is not just a gcc-10 issue.   Basically gcc-10 works fine for gst.

cstes:
Temporarily close, will submit again in the future since "gst" could be good addition.

PR #5717, opened 2020-05-11
qt5.8: fix crash with multidisplay environment

cgrzemba:
These two patches fix crash problem if QT application is started on other display than :0
Fix for openindiana issue 12301

### DIFF
--- a/components/library/qt5/patches/qtbase-0005-qglxconvience-qtbug81904.patch
+++ b/components/library/qt5/patches/qtbase-0005-qglxconvience-qtbug81904.patch
@@ -1,67 +1,35 @@
-diff --git a/qtbase/src/platformsupport/glxconvenience/qglxconvenience.cpp b/qtbase/src/platformsupport/glxconvenience/qglxconvenience.cpp
-index 4225beb..ba593fe 100644
+qglxconvenience: Avoid null pointer dereference
+glXGetVisualFromFBConfig according to documentation can return NULL [1].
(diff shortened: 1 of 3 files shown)


alarcher:
@cgrzemba Are you able to add a link to the corresponding ticket in each patch? I'll merge right after.

cgrzemba:
Do you mean the QT tickets?

alarcher:
> Do you mean the QT tickets?

Yes if it is possible, just to be able to track the issues when we update Qt. Thank you.

components/library/qt5/patches/qtbase-0005-qglxconvience-qtbug81904.patch
```diff
@@ -1,67 +1,35 @@
...
++        if (pointer)
++            XFree(pointer);
+     }
+ };
+ 
+@@ -202,6 +203,8 @@ GLXFBConfig qglx_findConfig(Display *display, int screen , QSurfaceFormat format
              GLXFBConfig candidate = configs[i];
  
              QXlibPointer<XVisualInfo> visual(glXGetVisualFromFBConfig(display, candidate));
--
```
alarcher: Did you drop this part of the patch on purpose? Just to confirm.

cgrzemba: yes, it is now the original patch frm QT 5.9

PR #5720, opened 2020-05-12
Workaround for libssp dependency

alarcher:
Since gcc-9 libssp is provided in gcc-X-runtime as it was misplaced in g++-X-runtime.
As a consequence some REQUIRED_PACKAGES had a bogus dependency to GXX_RUNTIME_PKG even for pure C software.
GXX_RUNTIME_PKG has a dependency to GCC_RUNTIME_PKG but it is not resolved when the publish target is executed.
The workaround is to add the GCC_RUNTIME_PKG dependency beside GXX_RUNTIME_PKG.
After the transition to gcc-10 the bogus GXX_RUNTIME_PKG dependency can be removed.


components/network/tor/Makefile
```diff
@@ -18,6 +18,7 @@ BUILD_BITS=		64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		tor
+COMPONENT_REVISION=	1
 COMPONENT_VERSION=	0.4.1.9
```
AndWac, line 21: Just a nitpick: you put the COMPONENT_REVISION before COMPONENT_VERSION here

AndWac (approved):

components/network/tor/Makefile
alarcher: Fixed :)

PR #5722, opened 2020-05-12
p7zip: fix error code mess with gcc-10

alarcher:
The function MyFormatMessage takes a DWORD as argument (unsigned) while error codes are declared as LONG and possibly negative... the comparison is unlikely to succeed.
Not sure what to do really but since comparison always involve DWORD (unsigned long) or LONG (signed long) I converted error code to the former (wider range).
Tests passed as before.

### DIFF
--- a/components/archiver/p7zip/Makefile
+++ b/components/archiver/p7zip/Makefile
@@ -28,7 +28,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		p7zip
 COMPONENT_VERSION=	16.2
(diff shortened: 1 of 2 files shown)


AndWac (approved):
Don't you want to change to the new
BUILD_STYLE= justmake
also?

alarcher:
> Don't you want to change to the new
> BUILD_STYLE= justmake
> include $(WS_MAKE_RULES)/common.mk
> also?

The install target has some extra step so I did not want to mess with it right now :S

AndWac:
Good point.

PR #5727, opened 2020-05-14
pacemaker: fix gcc-10 build

alarcher:


AndWac (approved):
How can we proceed with #5246?

alarcher:
> How can we proceed with #5246?

I do not know how pacemaker works so I would follow your decision. If you prefer I can build pacemaker 2.0 in the gcc-10 zone and patch if needed. We could then close this PR.
As you like.

AndWac:
I don't use it yet but I'd like to do so in the future (another item on my todo list). Maybe grueni will answer here...

alarcher:
> I don't use it yet but I'd like to do so in the future (another item on my todo list). Maybe grueni will answer here...

I guess we'll merge this one for now, just to make my build one is happy ;)

grueni:
Sorry for to be unresponsive for such a long time.
The new version of pacemaker 2.0.2 and 2.0.3 work without problems.

PR #5728, opened 2020-05-14
mariadb-101: fix build with gcc-10

alarcher:


components/database/mariadb-101/patches/07-mariabackup-duplicate-symbol.patch
```diff
@@ -0,0 +1,10 @@
+--- mariadb-10.1.44/extra/mariabackup/xbstream.c.orig	2020-05-14 06:57:33.351536835 +0000
++++ mariadb-10.1.44/extra/mariabackup/xbstream.c	2020-05-14 07:00:09.782764562 +0000
+@@ -45,7 +45,6 @@
+ datasink_t datasink_xbstream;
+ datasink_t datasink_compress;
+ datasink_t datasink_tmpfile;
+-datasink_t datasink_buffer;
+
```
AndWac, line 8: I don't understand this. What was the failure?

alarcher: datasink_buffer was declared extern in one header then declared in two different source files of the same library. This patch is consistent with upstream.

AndWac (approved):

PR #5741, opened 2020-05-15
notify-python: fix build with gcc-10

alarcher:


AndWac (approved):
Wow, this package seems to be old. The latest version available on the project's homepage is 0.5.2. And that one is from 2006?
It looks like a good candidate to drop: it's python 2.x only, doesn't seem to have dependant packages and is quite old.

alarcher:
> Wow, this package seems to be old. The latest version available on the project's homepage is 0.5.2. And that one is from 2006?
> It looks like a good candidate to drop: it's python 2.x only, doesn't seem to have dependant packages and is quite old.

Probably yes, it seems supported in other distributions due to desktop tools... not sure we need them...

PR #5747, opened 2020-05-16
12692: tmux: should install mdoc man page and enable libutempter

alarcher:
@ypankov Tell me if it is suitable then we'll merge.


ghost (approved):
Looks good, thank you!

PR #5748, opened 2020-05-16
gcc does not need to emulate thread-local storage

alarcher:
Issue found in the gcc-10 build zone while building illumos-gate.
Thanks go to @citrus-it for pointing to the relevant OmniOS patches.


alarcher:
> Issue found in the gcc-10 build zone while building illumos-gate.
> Thanks go to @citrus-it for pointing to the relevant OmniOS patches.

illumos-gate now builds in the gcc-10 build zone \o/

PR #5759, opened 2020-05-17
proftpd: bump to 1.3.6 (solaris-userland)

alarcher:
No idea how to test this...


components/network/proftpd/Makefile
```diff
@@ -20,45 +20,47 @@
...
 
+BUILD_BITS=64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		proftpd
-COMPONENT_VERSION=	1.3.5
-COMPONENT_REVISION=	5
-COMPONENT_SUBVERSION=	d
+COMPONENT_VERSION=	1.3.6
+COMPONENT_REVISION=	6
```
pyhalov: You can remove it as COMPONENT_VERSION increased.

components/network/proftpd/proftpd.pam
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+#
+# PAM stack customized for FTP comes without pam_fm_notify.
+#
+session definitive	pam_user_policy.so.1
```
pyhalov: We don't have pam_user_policy or pam_fm_notify. So, this makes session case default one, and we don't need this customized file.
BTW, I'm not sure it works as desired anyway.

pyhalov (changes requested):

components/network/proftpd/Makefile
alarcher: Fixed.

components/network/proftpd/proftpd.pam
alarcher: Removed.

pyhalov:
I didn't have time to look at it attentively, but I'd say push it if it works for you.

PR #5761, opened 2020-05-18
librecad: update for use of qt5

cgrzemba:
update librecad so that qt5 library will be used


alarcher:
> update librecad so that qt5 library will be used

- patches/0001-add-custom.pro.patch sets paths to qt 4.8, did you forget to remove it?
- Also PATH.qt5 should be placed before PATH.illumos otherwise Qt 4.8.7's qmake is picked.
- You are missing a dependency on boost.

alarcher (commented):
I pushed one commit with suggested fixes.
You can test them, squash the commits and ping me when you are ready to merge this PR.
Thank you!

alarcher:
@cgrzemba Somehow I am not allowed to push the squashed branch... Opening a new PR with your commit.

PR #5762, opened 2020-05-18
Bump to VirtualBox 6.1.8

jimklimov:
New vbox-6.1.8 was released, changelog indicates that a couple of recent regressions were addressed that bit me (I hope to check today if they really were), so it may be worthwhile to bump the one in OI if not yet. Namely, for guest additions: it should (still/again) work with clipboard, and ("newly" via mew mechanisms after old hacks were dropped) support resizing; for hosts: should not block into slow teletype mode when the TCP listener of a serial port is not actually listened.

Unfortunately, I fail to consistently build, install and package the recipe and not sure why. So feel free to pick up the updated version/hash data and make a usable package :) I would not have time to proceed on this in the nearby weeks, unless it were a quick win :(


components/sysutils/virtualbox/Makefile
```diff
@@ -152,9 +156,10 @@ CONFIGURE_OPTIONS+= --nofatal
 CONFIGURE_OPTIONS+= --enable-webservice
 CONFIGURE_OPTIONS+= --enable-vnc
 
-CONFIGURE_OPTIONS.32 += --only-additions
+CONFIGURE_OPTIONS.32 = --only-additions
 CONFIGURE_OPTIONS.32 += --target-arch=x86
-CONFIGURE_OPTIONS.64 += --enable-qt5
+
```
pyhalov: Please, remove empty line.

alarcher: Fixed and added a patch for gcc-10.

alarcher:
> New vbox-6.1.8 was released, changelog indicates that a couple of recent regressions were addressed that bit me (I hope to check today if they really were), so it may be worthwhile to bump the one in OI if not yet. Namely, for guest additions: it should (still/again) work with clipboard, and ("newly" via mew mechanisms after old hacks were dropped) support resizing; for hosts: should not block into slow teletype mode when the TCP listener of a serial port is not actually listened.
> 
> Unfortunately, I fail to consistently build, install and package the recipe and not sure why. So feel free to pick up the updated version/hash data and make a usable package :) I would not have time to proceed on this in the nearby weeks, unless it were a quick win :(

I noticed that the parallel build can exhaust 32GB of RAM and fail.

alarcher (approved):
Add gcc-10 patch.

PR #5765, opened 2020-05-19
WIP: valgrind: bump to 3.16rc1

alarcher:


cstes:
Hi I am opening a new PR for valgrind 3.17.0

Note that there is an issue with valgrind that it deals with epoll, epoll_ctl, epoll_wait as Linux specific and valgrind has code already to deal with these calls in syswrap-linux.c   .  However the syswrap-solaris.c does not have that,  so in some sense because illumos supports epoll now, the call is no longer linux specific so perhaps we can copy/paste the linux code into a patch for openindiana.

AndWac:
Obsoleted by #6664

PR #5769, opened 2020-05-19
bass-o-matic: do not use pager in git log

alarcher:
@xen0l So in my gcc-10 build zone bass-o-matic would always get stuck in incremental mode for some jobs.
I notice that the git command may output to a pager, then that the output of the git command was quite big. I tried to modify bass-o-matic to address these two aspects. It seems to work OK as the incremental build works fine for the same job that failed earlier.


alarcher:
Not enough...

pyhalov (approved):

PR #5795, opened 2020-05-24
groovy-24: fix build recipe

alarcher:


AndWac (commented):
Don't you want to add COMPONENT_REVISION and update to use common.mk?

alarcher:
> Don't you want to add COMPONENT_REVISION and update to use common.mk?

I did not bump the COMPONENT_REVISION since it does not change the package but fixes only republishing. I can do it and use common.mk

alarcher:
> Don't you want to add COMPONENT_REVISION and update to use common.mk?

Updated.

AndWac (approved):

PR #5812, opened 2020-05-26
userland: update templates

alarcher:


templates/cmake.mk
```diff
@@ -26,12 +28,11 @@ COMPONENT_ARCHIVE_URL=
 COMPONENT_ARCHIVE_HASH=
 COMPONENT_LICENSE=
 
-include $(WS_MAKE_RULES)/prep.mk
-include $(WS_MAKE_RULES)/cmake.mk
-include $(WS_MAKE_RULES)/ips.mk
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
```
AndWac, line 31: Does this work as expected? Shouldn't TEST_TARGET only be set if not yet set?

templates/ant.mk
```diff
@@ -26,12 +28,5 @@ COMPONENT_ARCHIVE_URL=
 COMPONENT_ARCHIVE_HASH=
 COMPONENT_LICENSE=
 
-include $(WS_MAKE_RULES)/prep.mk
-include $(WS_MAKE_RULES)/ant.mk
-include $(WS_MAKE_RULES)/ips.mk
+include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 31: What about TEST_TARGET?

templates/configure.mk
```diff
@@ -26,12 +27,10 @@ COMPONENT_ARCHIVE_URL=
 COMPONENT_ARCHIVE_HASH=
 COMPONENT_LICENSE=
 
-include $(WS_MAKE_RULES)/prep.mk
-include $(WS_MAKE_RULES)/configure.mk
-include $(WS_MAKE_RULES)/ips.mk
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
```
AndWac, line 30: Again: Set only if not yet done?

templates/configure.mk
```diff
@@ -12,6 +12,7 @@
 # Copyright 2020 <contributor name>
 #
 
+BUILD_BITS=64 # for binaries or 32_and_64 for libraries
```
AndWac, line 15: And what about libraries? Shouldn't we set BUILD_BITS only if not yet done?

templates/python.mk
```diff
@@ -26,12 +28,9 @@ COMPONENT_ARCHIVE_URL=
 COMPONENT_ARCHIVE_HASH=
 COMPONENT_LICENSE=
 
-include $(WS_MAKE_RULES)/prep.mk
-include $(WS_MAKE_RULES)/setup.py.mk
-include $(WS_MAKE_RULES)/ips.mk
+TEST_TARGET=$(NO_TESTS)
```
AndWac: Unless there aren't any python tests and will never be, shouldn't this also only be set if not yet defined?

templates/cmake.mk
alarcher: This has the same effect as `test: $(NO_TESTS)` which was previously in the file. Do you intend to change the default?

templates/ant.mk
alarcher: It is already set in common.mk

templates/configure.mk
alarcher: Tests are enabled by default for `configure` recipes.

templates/configure.mk
alarcher: The problem with defaults is that they may not hold anymore... that's why I thougt about adding the comment. Is it the wording that bothers you? Should I write "for executables-only packages or 32_and_64 for libraries"?

templates/python.mk
alarcher: For python packages most are without tests so we could disable them by default.

alarcher: But then we should fix all the packages using already BUILD_STYLE if needed:
```
narval> ggrep -l "BUILD_STYLE" */Makefile | cut -f 1 -d '/'
argcomplete
argh
attrs
automat
barman
ccsm
constantly
hamcrest
hyperlink
hypothesis
incremental
kafka-python
netifaces
pillow
psutil
psycopg2
pycups
python-dateutil
python-memcached
python-xlib
pyxdg
redis-py
twisted
zope-interface

```

templates/cmake.mk
AndWac: No, I misunderstood how it works

templates/ant.mk
AndWac: Ah Ok

templates/configure.mk
AndWac: No, it's ok.

pyhalov (approved):

templates/python.mk
pyhalov: Enabling test target by default is a good practice ;)

alarcher: OK I will change it then :)

PR #5813, opened 2020-05-26
openjade: fix broken code

alarcher:
*Blurb*

### DIFF
--- a/components/developer/openjade/patches/openjade-05-static-iid.patch
+++ b/components/developer/openjade/patches/openjade-05-static-iid.patch
@@ -1,0 +1,73 @@
+Symbols of static variables are deleted from object files and anyway can be replaced by plain static inline calls
+
+$ jade
+ld.so.1: openjade: fatal: relocation error: file /usr/lib/libostyle.so.0: symbol _ZN6OpenSP7LocNode3iidE: referenced symbol not found
+Killed
+
+--- openjade-1.3.2/spgrove/GroveBuilder.cxx.orig	2020-05-26 16:57:10.055077738 +0000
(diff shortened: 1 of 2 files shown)


components/developer/openjade/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2017 <contributor>
+# Copyright 2018 <contributor>
```
AndWac: Haha, when did you do this?

AndWac (approved):

components/developer/openjade/patches/openjade-05-static-iid.patch
```diff
@@ -0,0 +1,73 @@
+Symbols of static variables are deleted from object files and anyway can be replaced by plain static inline calls
```
pyhalov, line 1: Is it a known issue?

alarcher: I have not seen anyting like that but other distros maintain their own fork...
Probably a peculiarity of our linker, at least for the virtual function it is clear.
In any case the use of static members is ugly.

components/developer/openjade/manifests/sample-manifest.p5m
alarcher: I travelled back in time :)

components/developer/openjade/patches/openjade-05-static-iid.patch
alarcher: If you are happy with it, feel free to merge.

PR #5818, opened 2020-05-28
gcc-10: fix sparc build

tsoome:
tested by building bash:)


alarcher (approved):
Damn I still missed some of these variants... :(
Should I work on allowing split sample manifests for each MACH to ease the diffs?

PR #5629, opened 2020-03-19
icu: bump to 66.1

Toasterson:
In preparation for firefox esr 68

PR #5827, opened 2020-05-30
harfbuzz: bump to 2.6.6

alarcher:
After #5629.


components/library/harfbuzz/manifests/sample-manifest.p5m
```diff
@@ -61,38 +64,31 @@ file path=usr/include/harfbuzz/hb-version.h
 file path=usr/include/harfbuzz/hb.h
 file path=usr/lib/$(MACH64)/cmake/harfbuzz/harfbuzz-config.cmake
 link path=usr/lib/$(MACH64)/libharfbuzz-icu.so \
-    target=libharfbuzz-icu.so.0.$(COMPONENT_VERSION).0
+    target=libharfbuzz-icu.so.0.20600.6
```
AndWac, line 67: I have a question here: As you can see the old sample-manifest made use of $(COMPONENT_VERSION), the new one does not. Did we loose some power in our gmake sample-manifest or had this been changed manually before?

AndWac (approved):

components/library/harfbuzz/manifests/sample-manifest.p5m
alarcher: It was actually wrong in the previous version... still that bug that does not escape strings properly.

PR #5829, opened 2020-05-30
ICU 66.1 rebuild

alarcher:
If you could check that all dependent components have been listed.
The list I obtained is at #5629


Toasterson:
List looks good compared to output of pkg. It's a bit strange though that php-73/php-common appears but not php-70/php-common. Should we bump that as well just to be sure?

Toasterson (approved):

alarcher:
Unfortunately the ICU update breaks desktop/gnome3/evolution-data-server, library/webkitgtk, network/samba, and web/php/php-7_0 ... the fun continues ...

alarcher:
@pyhalov @AndWac @Toasterson The PRs are ready and can be merged in 3 batches:

- #5629 
- #5827 and #5829
- #5828, #5831, and #5830

AndWac (approved):

PR #5830, opened 2020-05-30
libreoffice: rebuild after icu 66.1 update

alarcher:
After #5829 since dependent libraries are used.


AndWac (approved):

iigs:
COMPONEN_REVISION needs to be COMPONENT_REVISION in Makefile.  Another option to consider is to bump libreoffice to 6.4.4.2 (Makefile change would instead be COMPONENT_VERSION= 6.4.4 ).

alarcher:
> COMPONEN_REVISION needs to be COMPONENT_REVISION in Makefile. Another option to consider is to bump libreoffice to 6.4.4.2 (Makefile change would instead be COMPONENT_VERSION= 6.4.4 ).

Fixed. The LibreOffice update is probably too much work, I'd rather focus on updating rust and firefox for now.

PR #5833, opened 2020-05-31
tigervnc: disable epoll

alarcher:


pyhalov:
Sorry for late response, but why?

alarcher:
@pyhalov Because epoll is known to break the Xorg 19.x code base for us, the mouse and the keyboard freeze. It was already disabled in xorg-server. Apparently it did not help... the user still reported issues with our package.

PR #5834, opened 2020-05-31
userland: prepend bit-aware binary paths to PATH

alarcher:
Until now we assumed that 32-bit binaries were located in /usr/bin while 64-bit binaries were in /usr/bin/$(MACH64).
With the current transition this does not hold anymore and the general rule should be that for a $(BITS) environment variable (32 or 64) binaries are first searched in /usr/bin/$(MACH$(BITS)) then in /usr/bin.
This change allows transitioning to PREFERRED_BITS=64 without breaking components the configuration of which relies on a foo-config file.


make-rules/shared-macros.mk
```diff
@@ -26,8 +26,8 @@
 # whenever PATH is to be defined there:
 #     PATH = $(PATH.illumos)
 #     PATH = $(PATH.gnu)
-PATH.illumos=$(USRBINDIR):$(GNUBIN):$(USRSBINDIR):$(PERL5BINDIR)
-PATH.gnu=$(GNUBIN):$(USRBINDIR):$(USRSBINDIR):$(PERL5BINDIR)
+PATH.illumos=$(USRBINDIR$(BITS)):$(USRBINDIR):$(GNUBIN):$(USRSBINDIR$(BITS)):$(USRSBINDIR):$(PERL5BINDIR)
+PATH.gnu=$(USRBINDIR$(BITS)):$(GNUBIN):$(USRBINDIR):$(USRSBINDIR$(BITS)):$(USRSBINDIR):$(PERL5BINDIR)
```
AndWac: Shouldn't $(GNUBIX) first (before $(USRBINDIR$(BITS)))? Otherwise GNU wouldn't be used...

alarcher: Yes. Also we have unused 64-bit binaries in /usr/gnu/bin and that's annoying. But let us leave it for later.

alarcher: Fixed.

AndWac (approved):

PR #5840, opened 2020-06-02
libjpeg6: libjpeg-turbo is now the default

alarcher:

PR #5841, opened 2020-06-02
libjpeg-turbo: bump to 2.0.4

alarcher:
After #5840 

ABI compatible https://abi-laboratory.pro/index.php?view=timeline&l=libjpeg-turbo

### DIFF
--- a/components/library/libjpeg-turbo/libjpeg-turbo.p5m
+++ b/components/library/libjpeg-turbo/libjpeg-turbo.p5m
@@ -25,22 +25,34 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 # Require package with default JPEG API
 depend fmri=image/library/libjpeg8-turbo type=require
 
+# Dependency on libjpeg6 with removed docs and symlinks
+depend fmri=image/library/libjpeg6@6.0.2-2020.0.1.1 type=optional
+
 # Only shipped by libjpeg-turbo
 file path=usr/share/doc/libjpeg-turbo/LICENSE.md
 file path=usr/share/doc/libjpeg-turbo/README.ijg
 file path=usr/share/doc/libjpeg-turbo/README.md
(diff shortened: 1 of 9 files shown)


components/library/libjpeg-turbo/libjpeg-turbo.p5m
```diff
@@ -29,18 +29,27 @@ depend fmri=image/library/libjpeg8-turbo type=require
...
-#file path=usr/share/man/man1/wrjpgcom.1
+# Previously shipped by libjpeg6-ijg
+file path=usr/share/man/man1/cjpeg.1
+file path=usr/share/man/man1/djpeg.1
+file path=usr/share/man/man1/jpegtran.1
+file path=usr/share/man/man1/rdjpgcom.1
+file path=usr/share/man/man1/wrjpgcom.1
+
+# Add symlinks to default binaries
+link path=usr/bin/cjpeg target=../lib/libjpeg8-turbo/bin/$(MACH64)/cjpeg
```
AndWac, line 50: These links were already provided by libjpeg6-ijg. Do we need a mediator?

alarcher: I think we just need to forget about libjpeg6-ijg and choose our default implementation.
Since the previous PR drops the links in libjpeg6-ijg and qdds a require dependency on libjpeg-turbo it seems fine, I could update without any conflict.

AndWac (approved):

components/library/libjpeg-turbo/libjpeg-turbo.p5m
pyhalov, line 50: If you want to make it in a proper way, add optional dependency on libjpeg6-jig version with removed files.

alarcher: Thank you :) I'll do that.

PR #5842, opened 2020-06-02
userland: bass-o-matic recurses in illumos-gate and openssl-0.9.8 directories

alarcher:
If there is no *.p5m file then bass-o-matic recurses in the directory to find manifests.
It takes quite a long time in illumos-gate source and build trees.

### DIFF
--- a/components/components.ignore
+++ b/components/components.ignore
@@ -2,6 +2,8 @@
 
 # skip openindiana/pkg/pkg/src/pkg/manifests directory
 /^openindiana\/pkg\/pkg\/src\/pkg\/manifests$/d
+# skip openindiana/illumos-gate, it should be built directly
+/^openindiana\/illumos-gate$/d
 # skip openindiana/kvm, it should be built with illumos-gate
 /^openindiana\/kvm$/d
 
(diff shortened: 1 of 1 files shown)


AndWac (approved):

pyhalov:
This will likely make both components real components and oi-userland build will build them. And we don't want this on the build server.

pyhalov (changes requested):
This will make full oi-userland build include these components, but we don't want this.

alarcher:
> This will make full oi-userland build include these components, but we don't want this.

OK thanks I missed that :(
Then we should add them to an exclusion list because the rule for finding a component path is to recurse until a .p5m is found.
Otherwise I'll leave it until we move forward on the changes to bass-o-matic.

pyhalov:
I'm fine with adding them to exclusion list. Check that illumos-gate build job is not affected.

alarcher:
> I'm fine with adding them to exclusion list. Check that illumos-gate build job is not affected.

OK. I added them to components.ignore to make sure they do not appear in COMPONENT_DIRS.
They both publish correctly, I will check the Jenkins build job in my build zone.

xen0l (approved):

alarcher:
Tested Jenkins build job in the gcc-10 zone.

PR #5843, opened 2020-06-02
gcc-10: fixup manifest

alarcher:
@tsoome @AndWac Let us see if it works for both worlds :)


components/developer/gcc-10/gcc-10.p5m
```diff
@@ -908,29 +908,29 @@ file path=usr/gcc/10/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/stdc++
 file path=usr/gcc/10/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/stdtr1c++.h
 file path=usr/gcc/10/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/bits/time_members.h
 file path=usr/gcc/10/include/c++/$(COMPONENT_VERSION)/$(GNU_TRIPLET)/ext/opt_random.h
```
AndWac, line 910: I wonder whether the files about should also have a variant.arch, no?

alarcher: Probably not since @tsoome did not need to flag them as i386-specific nor added them for the sparc build.

AndWac (approved):

xen0l (approved):

PR #5846, opened 2020-06-04
Use pkgmogrify -r for intermediate manifests

alarcher:


alarcher:
Only required if https://github.com/OpenIndiana/pkg5/pull/84 makes sense, which is not clear at all.

PR #5863, opened 2020-06-16
squeak-5c: Add cog implementation of squeak vm

cstes:
This package is offering a slightly different implementation of squeak-5.   It is the so-called "cog" implementation, which in my understanding is slightly faster, but perhaps also less portable than the stack implementation.   To my knowledge, there exists no "sparc" version of the cog implementation although that the stack implementation has ran in the past on sparc.    The package can be installed at the same time as the default implementation thanks to OpenIndiana (IPS) pkg mediator.  It uses the 'implementation' feature of mediator to set a different implementation of the same Smalltalk virtual machine.   The version is set to 5.0.7 because it is built from the same opensmalltalk source code as the stack vm version 5.0.7.    By the way, I initially compiled the "cog" implementation on OpenIndiana but so far I have not seen major performance improvement compared to the classical "stack" implementation, but others may disagree.

### DIFF
--- a/components/runtime/squeak5c/Makefile
+++ b/components/runtime/squeak5c/Makefile
@@ -1,0 +1,102 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 David Stes
+#
+
+
+#
+# Package Makefile for the "OpenSmalltalk Cog Spur" Squeak Smalltalk V5 system
+# See http://squeak.org and http://opensmalltalk.org
+#
+
+# opensmalltalk-vm can be built both in 32 and 64bit
+# but generally SqueakV5 images are 64bit
+BUILD_BITS=64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		squeak-5c
+COMPONENT_VERSION=	5.0.7
+GIT_TAG=		sun-v$(COMPONENT_VERSION)
+PLUGIN_REV=		5.0-202005021055-cog
+COMPONENT_SUMMARY=	The Cog Implementation of the Squeak V5 Smalltalk Virtual Machine
+COMPONENT_PROJECT_URL=	http://www.squeak.org
+COMPONENT_FMRI=		runtime/squeak-5c
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+
+# See http://wiki.squeak.org/squeak/933
+# See http://wiki.squeak.org/squeak/159
+# there's 2 license lines in the manifest, the MIT squeak.license and this one
+COMPONENT_LICENSE=	Squeak5
+COMPONENT_LICENSE_FILE=	squeak5.license
+
+COMPONENT_SRC=		opensmalltalk-vm-$(GIT_TAG)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:50c180b174048acd98a7d9705f0a40a3dc8bdc62b42d88eb7e20e71834184e4c
+COMPONENT_ARCHIVE_URL=	https://codeload.github.com/cstes/opensmalltalk-vm/tar.gz/$(GIT_TAG)
+
+TEST_TARGET= $(NO_TESTS)
+
+# some builds seem to be broken with -Bdirect
+LD_B_DIRECT=
+
+include $(WS_MAKE_RULES)/common.mk
+
+PKG_OPTIONS += -D PLUGIN_REV="$(PLUGIN_REV)"
+
+PATH=$(PATH.gnu)
+
+# opensmalltalk configure script checks for plugins files in builddir
+COMPONENT_PRE_CONFIGURE_ACTION = ( cp plugins.ext plugins.int $(BUILD_DIR_64) )
+
+# the default CFLAGS.gcc will be -m64 -O3 but this does not work for us
+# opensmalltalk code is not compatible with the gcc optimizer -O2 or higher
+# also the -g (debug) flag is intentional, perhaps omit-frame-pointer issues
+gcc_OPT=		-g -DNDEBUG -DDEBUGVM=0 -DCOGMTVM=0
+
+CONFIGURE_SCRIPT= $(SOURCE_DIR)/platforms/unix/config/configure
+CONFIGURE_OPTIONS +=	--without-npsqueak
+CONFIGURE_OPTIONS +=	--with-vmversion=5.0
+CONFIGURE_OPTIONS +=	--disable-dynamicopenssl
+CONFIGURE_OPTIONS +=	--without-libtls
+CONFIGURE_OPTIONS +=	--with-src=spur64src
+
+CONFIGURE_ENV +=	TARGET_ARCH="-m64"
+
+# the opensmalltalk Makefile does not use DESTDIR
+COMPONENT_INSTALL_ARGS=  	ROOT=$(PROTO_DIR)
+COMPONENT_BUILD_TARGETS= 	getversion squeak plugins
+COMPONENT_INSTALL_TARGETS= 	install-squeak install-doc install-plugins
+
+# Makefile has no strip target
+COMPONENT_POST_INSTALL_ACTION = \
+( $(STRIP) $(PROTOUSRLIBDIR64)/squeak/$(PLUGIN_REV)/squeak )
+
+REQUIRED_PACKAGES += x11/library/mesa
+REQUIRED_PACKAGES += system/library/dbus
+REQUIRED_PACKAGES += library/audio/gstreamer
+REQUIRED_PACKAGES += library/libffi
+REQUIRED_PACKAGES += system/library/freetype-2
+REQUIRED_PACKAGES += x11/library/libxext
+REQUIRED_PACKAGES += x11/library/libxevie
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += library/audio/pulseaudio
+REQUIRED_PACKAGES += library/desktop/cairo
+REQUIRED_PACKAGES += library/desktop/pango
+REQUIRED_PACKAGES += library/glib2
+REQUIRED_PACKAGES += library/security/openssl
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += x11/library/libx11
+REQUIRED_PACKAGES += x11/library/libxrender
--- a/components/runtime/squeak5c/manifests/sample-manifest.p5m
+++ b/components/runtime/squeak5c/manifests/sample-manifest.p5m
@@ -1,0 +1,50 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 <contributor>
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/squeak
+file path=usr/doc/squeak/COPYING
+file path=usr/doc/squeak/COPYRIGHT
+file path=usr/doc/squeak/LICENSE
+file path=usr/doc/squeak/README.Contributing
+file path=usr/doc/squeak/README.Keyboard
+file path=usr/doc/squeak/README.Sound
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/B3DAcceleratorPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/FileAttributesPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/ImmX11Plugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/LocalePlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/Squeak3D.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/SqueakFFIPrims.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/SqueakSSL.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/UUIDPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/UnicodePlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/UnixOSProcessPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/XDisplayControlPlugin.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/squeak
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/vm-display-X11.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/vm-display-null.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/vm-sound-null.so
+file path=usr/lib/$(MACH64)/squeak/5.0-202005021055-cog/vm-sound-pulse.so
+hardlink path=usr/share/man/man1/inisqueak.1 target=squeak.1
+file path=usr/share/man/man1/squeak.1
+file path=usr/squeak
--- a/components/runtime/squeak5c/patches/01-Makefile-amd64.patch
+++ b/components/runtime/squeak5c/patches/01-Makefile-amd64.patch
@@ -1,0 +1,15 @@
+--- opensmalltalk-vm-sun-v5.0.7/platforms/unix/config/Makefile.in	Sat May  2 12:55:37 2020
++++ p0/opensmalltalk-vm-sun-v5.0.7/platforms/unix/config/Makefile.in	Tue Jun 16 06:53:28 2020
+@@ -77,10 +77,10 @@
+ 
+ # unix launch scripts
+ squeak.sh: ${cfgdir}/squeak.sh.in getversion
+-	sed "s|.expanded_relative_imgdir.|lib/squeak/`./getversion VERSION_TAG`|" <${cfgdir}/squeak.sh.in >squeak.sh
++	sed "s|.expanded_relative_imgdir.|lib/amd64/squeak/`./getversion VERSION_TAG`|" <${cfgdir}/squeak.sh.in >squeak.sh
+ 
+ bin.squeak.sh: ${cfgdir}/bin.squeak.sh.in getversion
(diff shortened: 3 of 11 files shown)


components/runtime/squeak5c/CDDL.license
```diff
@@ -0,0 +1,377 @@
+COMMON DEVELOPMENT AND DISTRIBUTION LICENSE Version 1.0
```
pyhalov: Where is it used?

cstes: The Makefile(s) of the squeak(-4,-5,-5c) packages are contributed and use that license.

pyhalov: There's no need to add it explicitly here. There's reference to it in Makefile and manifests.

pyhalov: It shouldn't appear in component's directory.

cstes:
On Intel, which is the main interest of OpenIndiana, both the 'stack' and the 'cog' implementations of the 'squeak' package work.

One can install either 'squeak-5' and it sets 'stack-spur' as mediator
implementation and 'squeak-5c' and it sets 'cog-spur' as mediator for the
squeak package.

So the squeak package has those 2 implementations and possibly more.

For somebody who wishes to use the 'Cog' implementation,
installing squeak-5c is sufficient.   Somebody who prefers the stack impl.
can install squeak-5.  It is even possible to install both and compare them.

Is it possible to integrate the 'cog-spur' implementation as well now ??

Thanks.

AndWac:
I have created a package of it and installed it locally. Something seems to be wrong with this version.
Unlike the Spur version of Squeak5 this one creates garbage graphics when moving windows around.
At first everything seems to be ok but when moving windows or doing anything that draws on the screen the problems are getting obvious. This happens with both, squeak and Cuis Smalltalk images.

cstes:
Try rebuilding please after removing -g from the Makefile cflags.

gcc_OPT=                -g -DNDEBUG -DDEBUGVM=0 -DCOGMTVM=0

change that 
gcc_OPT=                -DNDEBUG -DDEBUGVM=0 -DCOGMTVM=0

cstes:
I tested in my case with -g and I thought it was OK, but for some other builds I've seen the issue that you described.  This is also what I mean in the Makefile with "the -g is intentional".   The Cog Makefile on Linux has the -g and it is definitely required for some setups.

cstes:
Also the gcc version may be different.   The author of 'opensmalltalk' admits that with gcc some of the VMs are 'broken' but this is a discussion between on the one hand the author of opensmalltalk and on the other hand the gcc maintainers/packagers.

cstes:
By the way, saying "with gcc" is little bit suggestive, as I have observed the same sort of issue with other compilers like "clang".   The Linux version of Cog smalltalk has migrated from gcc to clang, but as far as I can see I have similar issues with clang as compared to gcc,   so another way to look at it, is that it is related to the cog /spur/opensmalltalk vm itself.    Let's say that sometimes it does not work if I remove -g and sometimes it only seems to work with -g.  And in any case with -O2 or higher it results in a VM that acts as you described ...

AndWac:
Compiler version and flags can have important influences on the runtime, especially if the runtime uses a jitter like Cog is. So you typically have a working combination of compiler, flags and Cog version. All others are untested and might result in misbehaving executables. If the original project is using clang we should try to use the same. We have clang-8 and -9 available on Hipster and these can probably be used.

AndWac:
I have the same results with the debugging flag removed as you proposed.

cstes:
The original project is moving towards clang but that is not enforced. 
For example the comment they have in their Makefile is "# Some gcc versions create a broken VM using -O2"
However this is open to interpretation.
The behavior that you describe is something that I have reported and seen many times, not only with "cog" but also with "stack" implementation.   I don't think it is specific to Cog.
Are you sure that it does not help to remove -g ?  I have pushed the gcc_OPT which I explicitely set because I realise all too well that compile options have a big impact,  that is the reason why I am resetting gcc_OPT in the first place ...

cstes:
By the way, I also compiled OpenSmalltalk Squeak with clang on OpenIndiana and I reported to you that I think that the oi-userland framework is not providing a way to set clang as the compiler for oi-userland builds.  Can you remember this , I asked whether it was possible to use 'clang' as the oi-userland compiler.   Anyway I think personally that it should work with gcc.
We have to investigate this further.

The issue is not specific to "Cog".   I have seen it on 'stack implementation' as well.

Also there is currently in the upstream sources a discussion on the use of clang's option
COGDEFS:=$(COGDEFS) -mno-stack-arg-probe
so they are reporting currently a broken VM with clang as well,  I don't think clang is the miracle answer to the issue.

Because it works for me, I suspect maybe a different 'debug' library is present in your case.

But I can confirm that I have also experienced the same issue.

Have you double checked please that if you do a 

make clobber  (make clean)
and 

rebuild you still have the redraw issue ?

cstes:
Also please provide the gcc --version and whether gcc-10 is installed.  I believe I have gcc-10 installed for testing with that compiler as well.

cstes:
As far as I can see in the oi-userland/make-rules/
There is no support currently for clang:
bash-4.4$ grep clang shared-macros.mk 

there is no clang_OPT for example like I use gcc_OPT to exclude the -O2/-O3.

Anyway I have built cog/stack with clang on OI and it works, but in my opinion I have the same issues, so it does not look to me a solution.

Actually while I mentionned gcc-10,  I believe I have set 

bash-4.4$ grep GCC_VERSION shared-macros.mk 
GCC_VERSION =	10

on my build server, I have to double check.   In your case the GCC_VERSION is set to 7 in the shared-macros.mk definition ?

cstes:
If it is possible to build packages with oi-userland and set COMPILER=clang instead of COMPILER gcc (or studio) let me know please.
As said, squeak builds with clang on OI which is also nice to know, but as far as I can see it is subject to the same -O optimizer and -g debug issues with clang, so I'd rather forget about clang in this issue.

alarcher:
One can easily imagine adding a COMPILER variable and adding definitions to support clang as well in shared-macros.mk, there is no problem with that.

cstes:
I understand that it is possible - but it is not done yet.   I would actually like to do it, I mean extend and contribute the oi-userland with a COMPILER=clang option, I can give it a try.    Note however that for squeak/opensmalltalk, I have indications that similar "strange redraw issues" with the smalltalk vm exist with clang and other compilers, given the fact that squeak contains a lot of near-assembly code in the vm, this is not such a surprise.    So basically in this particular case I think to focus first on figuring out what the exact problem with the gcc was.   I have also reported that also in the 'stack-implementation' I have the same issue, and work around it with setting gcc_OPT.   I override the gcc_OPT in the Makefile because the default oi-userland gcc_OPT produces a vm for me with the 'redraw' issues that Andreas reports.

cstes:
Please simply add squeak-5c version 5.0.7 as well.   I will add the -g flag again , I double checked and for me on my machine, I don't have the issue.   Please note that I believe Andreas,  I am experiencing similar issues on both stack and cog implementations (with different compilers by the way, as I am also compiling with several different compilers).

I am not asking for COMPILER=clang support in oi-userland.

If you do it, fine, I could use that.   But I'm not asking for it, as it should work with gcc as well , and frankly to me the issue does not look like a C compiler issue, it is more a code issue with opensmalltalk itself.

The fact that sometimes somebody is 'lucky' with a certain compiler is not a proof that it is always working with that compiler.

In my case I have gcc-10 installed

   # pkg list | grep gcc
developer/gcc-10                                  10.1.0-2020.0.1.1          i--
developer/gcc-7                                   7.5.0-2020.0.1.3           i--
developer/illumos-gcc                             4.4.4-2020.0.1.8           i--
system/library/gcc-10-runtime                     10.1.0-2020.0.1.1          i--
system/library/gcc-3-runtime                      3.4.3-2020.0.1.3           i--
system/library/gcc-7-runtime                      7.5.0-2020.0.1.3           i--

However I double checked and I compiled the package with GCC_VERSION=7 not with GCC_VERSION=10 although that I have done some experiments with GCC_VERSION=10 as well.

So for me:

   # pkg list | grep userland
consolidation/userland/userland-incorporation     0.5.11-2020.0.1.13117      i--
runtime/squeak-5c (userland)                      5.0.7-2020.0.1.0           i--

I don't have the redraw problem that Andreas saw.

However I can confirm that the same redraw problem also is in the 'stack' version squeak-5.

You can thoroughly test squeak-5 (stack) and who knows, you'll find a lot of bugs, I am sure.

This software is only a few commits behind the Cog mainline pre-release opensmalltalk VM ...

So please siimply add cog squeak-5c 5.0.7 and I can try to test a "pkg install " from the OI repository.

I am actually curious whether the binaries produced by you will give me the same (good) results on my machine.


In any case the COMPILER=clang issue is a "moot issue", there is no COMPILER=clang support in oi-userland  and I am NOT asking for it.

Also I have here an upcoming version squeak-5 and squeak-5c of the 5.0.8 which is again adding new commits from the mainline Squeak opensmalltalk, so as I said all those are pre-release as they are working to a new stable release.   FYI  it seems also on other platforms they have similar issues.

cstes:
working version for me is 

         # type squeak
squeak is hashed (/usr/bin/squeak)
         # pkg mediator squeak
MEDIATOR            VER. SRC. VERSION IMPL. SRC. IMPLEMENTATION
squeak              system    5       system     cog-spur

          # squeak -version
5.0-202005021055-cog  Tue Jun 16 08:30:24 CEST 2020 /usr/gcc/7/bin/gcc 7.5.0 [Production Spur 64-bit VM]
CoInterpreter VMMaker.oscog-eem.2743 uuid: 6fa490eb-27b9-471e-9e46-6aa4ce54e382 Jun 16 2020
StackToRegisterMappingCogit VMMaker.oscog-eem.2743 uuid: 6fa490eb-27b9-471e-9e46-6aa4ce54e382 Jun 16 2020
VM: 202005021055-cog stes@deimos:src/opensmalltalk
Date: Sat May 2 12:55:37 2020 CommitHash: 5cd355762
Plugins: 202005021055-cog stes@deimos:src/opensmalltalk
SunOS deimos 5.11 illumos-b30f02def0 i86pc i386 i86pc illumos
plugin path: /usr/bin/../lib/amd64/squeak/5.0-202005021055-cog [default: /usr/lib/amd64/squeak/5.0-202005021055-cog/]

Note that in the squeak -version output it clearly states I compiled with gcc 7.5 as well.

I also did a test with gcc-10 by the way.

So please simply add squeak-5c as it is, and I can continue to work (perhaps) on contributing COMPILER=clang code to the oi-userland framework !

components/runtime/squeak5c/Makefile
```diff
@@ -0,0 +1,97 @@
...
+CONFIGURE_ENV +=	TARGET_ARCH="-m64"
+
+# the opensmalltalk Makefile does not use DESTDIR
+COMPONENT_INSTALL_ARGS=  	ROOT=$(PROTO_DIR)
+COMPONENT_INSTALL_TARGETS= 	install-squeak install-doc install-plugins
+
+# Makefile has no strip target
+COMPONENT_POST_INSTALL_ACTION = \
+( $(STRIP) $(PROTOUSRLIBDIR64)/squeak/$(PLUGIN_REV)/squeak )
+
```
AndWac, line 83: Stripping is a contradiction to your debug flag (-g)! as stripping removes (among other things) the debug information!
Exerpt from man strip:
strip - strip symbol table, debugging and line number information from
       an object file.

cstes: There is no contradiction, as the man page explains it 's perfectly normal to strip debugging and line number information from an object file.  The 'strip' command is not a contradiction.   Somebody can produce an object with -g and then strip it, perfectly fine.   By the way there exists other software in OpenIndiana like 'firefox' that is also stripped.

cstes:
Andreas,  how well did you test the stack implementation ?   I'm sometimes experiencing 'redraw' issues in the stack implementation.   In your specific case, when you experienced the 'redraw' issue with the Smalltalk image with the cog implementation, did you test the 'sqeuak mediator' technology to install the 'stack' implementation please (squeak-5).  Did you have the same issue with the 'old' classical implementation, which is derived from code that is more classical than the cog implementation ?  I don't think the redraw problem is actually a 'cog' problem, but it is interesting with IPS on OI that it is possible to compare various squeak mediator implementations.

AndWac:
I have installed both variants on my system:

╭─andreas@saturn ~
╰─➤  pkg list -a|grep squeak
runtime/squeak-4 4.16.7-2020.0.1.0          ---
runtime/squeak-5 5.0.7-2020.0.1.0           i--
runtime/squeak-5c 5.0.7-2020.0.1.1           i--

This one (Spur version) runs fine:

╭─andreas@saturn
~/Squeak6.0alpha-17873-64bit-All-in-One.app/Contents/Resources
╰─➤  /usr/lib/64/squeak/5.0-202005021055/squeak
Squeak6.0alpha-17873-64bit.image&


The cog version has the problems: (it's published without -g but
independant on the debug flag as the previous version had the same
problems):

╭─andreas@saturn
~/Squeak6.0alpha-17873-64bit-All-in-One.app/Contents/Resources
╰─➤  /usr/lib/64/squeak/5.0-202005021055-cog/squeak
Squeak6.0alpha-17873-64bit.image&


I have the same results when running an actual Cuis Smalltalk.


Am 28.06.20 um 16:01 schrieb David Stes:

cstes:
Is it possible to provide binaries for squeak-5c as built by your build server please.

I am curious because my build works,  perhaps if I "pkg install" the published binaries (which do not work well for you) they could work for me ...

And in any case, we know that we can ask any other potential user to try the stack implementation ...

cstes:
I'm not asking to provide "broken" software in OI.   The package that you publish may not work for you, but it could for me ... The simple fact is that it may work on some hardware's and on some not ...    So basically please publish squeak-5c  and then I can have a look at it and in addition, most likely I will simply submit a squeak-5@5.0.8 and squeak-5c@5.0.8 because I'm following closely the new commits in mainline opensmalltalk repository ...    In any case the Makefile and build setup for 5.0.8 will be the same as for 5.0.7 (but the actual sources will be upgrade to a new snapshot of Eliott Miranda's repo).

cstes:
squeak-5c package was upgraded to 5.0.8

cstes:
Can you please have a look at this again ?  I've changed the Makefile so that LD_OPTIONS does not contain -Bdirect.   This may help ...   In my case for squeak-5c  it was not necessary to use -Bdirect, that version ran fine for me, but I've seen the issue that Andreas reports and the squeak builds may be sensitive to options such as -z ignore and -Bdirect.

cstes:
Actually some of the binaries that are built directly from Eliott Miranda's repository (and that do not set LD_OPTIONS -Bdirect) are also broken.  So removing -Bdirect from LD_OPTIONS is not a solution that will solve all problems, but just like adding -g (debug) it seems to have an impact and it may be sufficient in this case to get a working binary.    Also note that as far as I've seen, this issue is present in both the Stack Spur and the Cog Spur binaries.

cstes:
Also I'd like to add that I'm not against using "-Bdirect" in the oi-userland framework.   Ideally this should be used.   The documentation on "-Bdirect" says "Unanticipated, and erroneous bindings that result from multiple instances of the same symbol frequently result in hard to diagnose process failures."   This seems to be the case for me, so ideally I get the package working with the -Bdirect and without -g.    But for the moment we just fiddle around with those options to get the VM working.

cstes:
I've gone back to 5.0.7 because as written before the Cog implementation ran fine for me, and there's no need to have 5.0.8 before adding 5.0.7 to the oi-hipster ...    I suspect that the LD_OPTIONS change may help, also for 5.0.7,  and in my case it works with both -Bdirect and without -Bdirect but my suspicion is that this may be a sensible explanation of the weird behavior I observe where some builds work and then again not ... and of the strange difference between your setup and my setup (where apparently you have the 'redraw' issue where I don't)  ... it may be related to some sort of 'search order' problems with the linker (direct link issues).    Please let me know whether this Makefile for the cog VM 5.0.7 works for you.  And if so feel free to merge that one, I'll upgrade it then later to 5.0.8 (which consists of new commits from the main OpenSmalltalk repo).

AndWac:
This version still has graphics problems on my system.
I'll let alp decide whether to integrate this or not. At least it looks like this version has no problems re-running gmake publish unlike squeak-5 Spur.

cstes:
Thanks for testing.   It would be good to merge this if possible ...  The basic setup where 2 implementations are provided (squeak 5 for the squeak stack spur and squeak 5c for the squeak cog spur) will remain the same.   In my case the 5.0.7 ran / runs ok but I've found out that on some version which has the redraw problem in my case, I can fix it by removing -Bdirect.  That is an interesting fix because it points towards multiple defined symbols, overlap perhaps of some squeak internal symbols with some libraries  (see manpage of "lari" the tool lari explains how to debug this).    I think this is a sensible explanation, it provides a hint where to look ... but it will not be so easy to find the 'offending' wrong defintions and it will take a lot of work to get OpenSmalltalk upstream to fix it.    Therefore I would like to plead for merging / incorporation squeak-5c please.   I will follow this up with a new version squeak-5c 5.0.8 and there is also an upcoming squeak 5.0.9  ....     So basically I would like to plead for including it, and please remember that any potential user can choose between the stack and the cog implementation (with pkg mediator) which is useful in OpenIndiana -- although I suspect that the 'redraw' problem is linker (-Bdirect related) rather than cog/stack related.

cstes:
@pyhalov   this version 5.0.7 works for me without the so-called 'graphics problem'  ... it would be nice if it could be merged into oi-hipster.   Note that the Makefile now sets LD_B_DIRECT variable (FEATURE_MACRO) to disable -Bdirect.    Merging squeak-5c will help me to continue providing fix/support for the squeak cog version , which is a faster implementation then the more classical squeak stack version.  Also I have updates 5.0.8 and 5.0.9 coming up and then I can update the squeak-5 and squeak-5c simultaneously for 5.0.9 so providing updates to the 2 packages at the same time.

cstes:
When debugging with "lari" the stack spur version of squeak  (the squeak-5 package which is already merged into oi-hipster) it can be seen for example:

$ lari -D lari.dbg.03935 | more
[2:0]: copysign(): /lib/64/libm.so.2
[2:0P]: copysign(): /usr/lib/squeak/5.0-202007040853/squeak

So squeak is redefining "copysign()" of libm.so.2 : there is a definition in libm and a definition in squeak.   I have to take this up with the Squeak developer to figure out whether this is desired.

I think it is not desired.  But such a redefine, although not desired, is not a tragedy ...
It's interesting to continue to work on this, but it is not a show-stopper.

You have already merged squeak-5  and the second package squeak-5c should be considered as a companion package of squeak-5 : the user has a choice with "pkg mediator" to select either the stack-spur or the cog-spur implementation and see which works (if any at all).

Currently I can say that squeak-5c is usable (version 5.0.7) and it would be helpful if you can merge this in oi-hipster.

This is so I can continue to debug the -Bdirect (LD_DEBUG issue) and further improve the squeak OpenIndiana package in subsequent versions, where I will gradually continue to resolve any issues with plugins.

Note that for Squeak 5.0.9 I would like to simultaneously update the stack squeak-5 and cog squeak-5c package.  

The Squeak 5.0.9 has new Smalltalk plugins for cryptography (MD5 etc.)  so Squeak-5 is a "moving target" it is actively being developed.

Please go ahead and merge squeak-5c 5.0.7.   The basic layout of the components squeak-5 and squeak-5c is OK I think, and it provides a framework for updates and fixes for OpenIndiana in subsequent versions, while the current binaries are already very usable.

cstes:
For more info on Cog : [http://www.mirandabanda.org/cogblog/about-cog/](http://www.mirandabanda.org/cogblog/about-cog/)

"benchmarks from the computer language shootout is in the range of 4 to 6 times faster than the interpreter"

That is the claim from the Cog developer ...   My understanding is that you basically have a choice of a more classical stack VM and this Cog VM.

I find it interesting that OpenIndiana has a "mediator" technology that allows 2 different implementations of the same package,  as I'd personally rather have an option to also have a more classical (perhaps slower) VM.

cstes:
Note that a lot of packages in oi/hipster oi-userland are setting LD_B_DIRECT= empty to override -Bdirect.   For example bacula, wireshark, tcpdump, perl, ruby and a bunch of others they all set LD_B_DIRECT= empty.  I would like to use -B direct but I have the impression that squeak (both stack and cog) work slightly better without -B direct.

I think further investigation in this direction (using the "lari" tool and LD_DEBUG) is useful .    Squeak plugins and the internals of the VM tend to duplicate a lot of code (namespace pollution / overlapping symbols/ interpose issues with loaded objects).   For example it can be seen that the SoundCodec plugin of squeak defines the same symbols as the library/libsndfile library (Squeak compiles that library into the VM).   For the libm (math) there is overlap with the FloatMathPlugin.

I have meanwhile a version of squeak-5 @5.0.9 with a patch for the libm (math library) collision "copysign()" :

--- a/platforms/Cross/plugins/FloatMathPlugin/ieee754names.h
+++ b/platforms/Cross/plugins/FloatMathPlugin/ieee754names.h
@@ -26,3 +26,5 @@

 #define tanh __ieee754_tanh

+#define copysign __ieee754_copysign


I have added this patch in version 5.0.9.

However please merge version 5.0.7 of squeak-5c first.

This version is working decently for me.   I have no redraw issue with 5.0.7.

As I said, I believe Andreas that he has the issue, but I still think the squeak-5c package 5.0.7 can still be merged.

cstes:
Thanks this will allow me to update the 5.0.9 components at the same time in one commit ...  Basically offering the user 2 options for the VM.

PR #5865, opened 2020-06-16
Building in zones

xen0l:
This PR adds support for building components in build zones.


make-rules/environment.mk
```diff
@@ -48,6 +52,63 @@ component-environment-prep::
...
+component-zone-build:
+	$(call separator-line)
+	$(call separator-line,Create $(ZONENAME))
+	$(PFEXEC) $(ZONE) --prefix $(ZONENAME_PREFIX) spawn-zone --id $(ZONENAME_ID)
+	$(call separator-line,Boot $(ZONENAME))
+	@while $$(true); do \
+		echo "Waiting for zone $(ZONENAME) to boot..."; \
+		$(PFEXEC) /usr/sbin/zlogin -l $${USER} $(ZONENAME) \
+				/bin/true >/dev/null 2>&1 && break; \
+		sleep 10; \
```
pyhalov, line 74: should we check for zone boot failure or some timeout?

make-rules/environment.mk
```diff
@@ -48,6 +52,63 @@ component-environment-prep::
...
+	done
+	# FIXME:
+	# - remove once we figure out a better way how to enable zoneproxy-client inside the nlipkg brand
```
pyhalov, line 77: Optionally we could teach sysding (or pkgcreatezone) to perform necessary actions. Perhaps, sysding is general enough. I don't insist on doing it now, just proposing.

make-rules/environment.mk
```diff
@@ -48,6 +52,63 @@ component-environment-prep::
...
+		$(PFEXEC) /usr/bin/pkg -R $${ZONEROOT} set-property use-system-repo True && \
+	while $$(true); do \
+		echo "Waiting for sysrepo to be ready..." && \
+		$(PFEXEC) /usr/bin/pkg -R $${ZONEROOT} publisher | \
+		$(GNU_GREP) syspub >/dev/null 2>&1 && break; \
+		sleep 10; \
+	done
+	$(call separator-line,Build in $(ZONENAME))
+	$(PFEXEC) /usr/sbin/zlogin -l $${USER} $(ZONENAME) \
+		"cd $(COMPONENT_DIR); gmake install"
```
pyhalov, line 106: why always do install, not publish (or custom target)? What is expected use case?

xen0l: The expected use case is to run `zone-build`, which will do `build install` in the zone. We can discuss changing `build install` to `publish`. However, the issue is that zone has no networking, so `publish` will fail. It's expected to run `download` in GZ with working networking. Once built, we can run gmake publishing globally.

I mean I am open to changing and discusisng this, but would not like to complicate setup with working networkingin the zones.

make-rules/environment.mk
xen0l: pkgcreatezone would be the place to do this, but I am not certain if it's a good idea to do this for all nlipkg zones or we shoudl introduce a new "nlipkg+zone-proxy" brand specific for just this use case.

make-rules/environment.mk
xen0l: I want to keep this light. I will start testing to see, how zones behave about this and if those errors start to appear, i would most like deal with them.

make-rules/environment.mk
AndWac: Hi xen0l, what to do next?

xen0l:
@AndWac functionaly, this is done. However, I haven't done any bigger rebuilds as I don't have HW to do it. Can run some smaller tests in vagrant in needed.

PR #5869, opened 2020-06-17
nginx: bump to 1.19.0

alarcher:


pyhalov:
1.18 is LTS, perhaps it's better suits us.

PR #5874, opened 2020-06-23
Deliver gnome-connection-manager 1.2.1 from github release

jimklimov:
Update of the GCM version and cleanup of the recipe.


components/desktop/gnome2/gnome-connection-manager/Makefile
```diff
@@ -57,7 +43,9 @@ include $(WS_MAKE_RULES)/prep.mk
 include $(WS_MAKE_RULES)/ips.mk
 
 PYTHON_VERSION =	3.5
+PYVER_SHORT    =	$(subst .,,$(PYTHON_VERSION))
```
pyhalov: @alarcher What do you think, should we move this definition to shared-macros.mk and call PYV (as it is called in PYVER manifests)?

jimklimov:
I think it is a good idea, but my earlier PRs with intrusions into the
build system stalled. This can also be moved in a subsequent PR, though if
you prefer the `PYV` name it can be used here right away :)

components/desktop/gnome2/gnome-connection-manager/Makefile
jimklimov: I took a quick look at the oi-userland codebase and see `PYV` indeed used in many recipes but never defined except in this PR's code. It is also referenced in calls to `pkgmogrify`:

````
jim@jimoi:~/shared/oi-userland-gcm$ git grep -E 'PYV[ \t]*='
make-rules/ips.mk:      $(PKGMOGRIFY) -D PYVER=$(1) $(MANIFEST_LIMITING_VARS) -D PYV=$(2)  $$< > $$@
make-rules/ips.mk:      $(PKGMOGRIFY) -D PYV=### $(BUILD_DIR)/mkgeneric-python \
````

So I am not certain if defining it in shared-macros.mk would confuse things or not. The idea is good but potentially can add breakage and need iterations to fix it, so belongs in a different PR than bumping a particular component.

There are also comments about it like:

````
doc/branchid-change-instructions.md:The main issue during this process is to ensure that components, using auto-generated PLV, RUBYV and PYV manifests are republished after BRANCHID change.

transforms/mkgeneric:# name end in "-###PYV###" (typically by setting a pkgmogrify variable to
transforms/mkgeneric:# "###PYV###"), which we also remove here.  This works in conjunction with a set
transforms/mkgeneric:<transform set name=pkg.fmri value=(?:pkg:/)?(.+)-###PYV###@(.*) -> \
transforms/mkgeneric:   edit value -###PYV### "">
````

and a usage that I'm not too sure is right (but did not check):

````
make-rules/ips.mk:PYV_FMRI_VERSION = PYV
````
(sanity check: should this be an expanded `$(PYV)` value?)

pyhalov (approved):

PR #5876, opened 2020-06-28
ekiga, claws-mail: fix ABI breakage

AndWac:


alarcher:
Are you sure that this is really the issue?
All the packages were republished after the gcc-7 migration but as we saw for tcl/tk, there is a problem that sometimes libraries in prototype/mangled are not updated with new libraries, so the ones published are not the ones installed in prototype.
We should probably wipe out prototype whenever the install target is executed.
I would rather check whether this is the issue than assuming it was simply not republished.
@pyhalov @AndWac

AndWac:
I am quite sure as I installed the packages after the update and they
work now. They didn't work before the update.

alarcher:
@AndWac Not really the question I was asking :)

My question is:
1) Was the breakage caused by packages that were not republished at all?
or
2) Was the breakage caused by packages that were republished but due to the issue with prototype/mangled that I mentioned and seems to happen sometimes? In this case new libraries are installed in prototype but old copies in prototype/mangled were published instead in the new package?

If we are in the second case, the new libraries may have been published and it may seem that the problem is fixed but the underlying issue is not solved. In which case we are bound to have the same problem in the future.

AndWac:
Ok, I misunderstood you. Re-publishing fixed the actual problems for
these two packages.

As far as I analyzed the situation, both packages haven't been
re-published after the ABI breakage happened around mid of April.
I don't have access to our build server so my analysis is limited to the
git history.
As I see it we have had some problems in the past because the builds
usually don't run in a clean environment. We also have some problems due
to the parallel nature of the builds.
On different build systems this kind of problem manifests for different
packages (e.g. webkitgtk, ekiga, gdk-pixbuf, libotr and squid to name a
few that have problems on my build servers).
This problem is independant of the other (unclean prototype area) problem.

Alas, I don't know how to fix them.


Regards
Andreas


Am 28.06.20 um 14:37 schrieb Aurélien Larcher:

PR #5885, opened 2020-07-10
Zeromq + Salstack 

pyhalov:
Should be merged in series
- Move czmq to upper level + zeromq: update to 4.3.2
- czmq: update to 4.2.0
- pyzmq: update to 19.0.0
- salt: update to 2019.2.5

As for salt itself, I've checked that master can query minion, but not much else.


pyhalov:
Merged as #5887 #5889 #5892

PR #5888, opened 2020-07-12
WeeChat 2.8

oicontrib:
WeeChat IRC client.


components/network/weechat/weechat.p5m
```diff
@@ -0,0 +1,116 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/weechat
+link path=usr/bin/weechat-curses target=weechat
+file path=usr/bin/weechat-headless
+file path=usr/include/weechat/weechat-plugin.h
+file path=usr/lib/pkgconfig/weechat.pc
+file path=usr/lib/weechat/plugins/alias.a
```
pyhalov, line 32: Please, remove static libraries, they are likely useless (especially as plugins).

components/network/weechat/Makefile
```diff
@@ -0,0 +1,74 @@
...
+CONFIGURE_OPTIONS += --enable-doc
+CONFIGURE_OPTIONS += --enable-man
+CONFIGURE_OPTIONS += --enable-tests
+CONFIGURE_ENV += CA_FILE="/etc/certs/ca-certificates.crt"
+
+COMPONENT_TEST_CMD =    ctest -V
+COMPONENT_TEST_TARGETS =
+
+# common targets
+build:		$(BUILD_32)
```
pyhalov: Why don't build it 64-bit by default?

components/network/weechat/Makefile
```diff
@@ -0,0 +1,74 @@
...
+COMPONENT_ARCHIVE_HASH=	\
+    sha512:3bebb42118fca6b601f1de8f4d52378619a10e5c490f5b46f86d27a9680f00a56d48a1ade729e78d965cae589e6c80647c22414393000c7d3cfcc94a1e41c4df
+COMPONENT_ARCHIVE_URL=	https://weechat.org/files/src/$(COMPONENT_ARCHIVE)
+COMPONENT_SIG_URL=	$(COMPONENT_ARCHIVE_URL).asc
+COMPONENT_FMRI=		network/chat/weechat
+COMPONENT_LICENSE=	GPLv3
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+COMPONENT_CLASSIFICATION=	Applications/Internet
+
+include $(WS_MAKE_RULES)/prep.mk
```
pyhalov: Can you switch to common.mk (like in https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/desktop/ekiga/Makefile)?
Include only common.mk (instead of prep,configure,ips.mk), and you can remove lines 53-59

components/network/weechat/weechat.p5m
oicontrib: done

components/network/weechat/Makefile
oicontrib: I'll try. Right now there are many errors when trying to BUILD_BITS=64

components/network/weechat/Makefile
oicontrib: done

pyhalov:
Publish step fails here:
```
   depend type=require fmri=__TBD pkg.debug.depend.file=liblua5.3.so \
        pkg.debug.depend.reason=usr/lib/weechat/plugins/lua.so.0.0.0 \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=lib \
        pkg.debug.depend.path=usr/gcc/7/lib \
        pkg.debug.depend.path=usr/lib'.
```

components/network/weechat/Makefile
```diff
@@ -0,0 +1,65 @@
...
+COMPONENT_TEST_TARGETS =
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
+REQUIRED_PACKAGES += library/gnutls-3
+REQUIRED_PACKAGES += library/guile
+REQUIRED_PACKAGES += library/ncurses
+REQUIRED_PACKAGES += library/pcre2
+REQUIRED_PACKAGES += library/zlib
+REQUIRED_PACKAGES += runtime/lua
```
pyhalov: Likely needs lua-53 dependency - it's used instead of 5.2 when found.

pyhalov: Otherwise, seems to work correctly

oicontrib: done dependency on lua-53.

As weechat can be bult with lua-5.2.x and with lua-5.3.x
I've tried to set something like
>depend type=require-any \
>     fmri=pkg:/runtime/lua \
>     fmri=pkg:/runtime/lua-53

in weechat.p5m but without success.

@pyhalov, is it possible to define 'require-any' dependency?

pyhalov: Not in this case (or at least not so easy). Application explicitly requires specific library version, against it was compiled (or compatible). pkgdepend finds that it requires liblua5.3.so, and tries to find package, containing it. And it was explicitly renamed - https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/runtime/lua-53/patches/src.Makefile.patch#L36 -  as libula 5.2 and 5.3 interfaces are not compatible.

PR #5858, opened 2020-06-13
WIP: Deprecate Python 3.4

alarcher:

PR #5908, opened 2020-07-17
Obsolete remaining python-34 modules

AndWac:
pkg list -a | grep -- "-34" | ggrep -v -- "--o"
revealed some omitted python-34 modules (only 2 false positives regarding mongodb :))


AndWac:
After this we can close #5858

PR #5911, opened 2020-07-18
autoconf: update to 2.69b

pyhalov:
Rebuilding the world with this autoconf version would be interesting. I'll keep it on my build server for a while...


Toasterson:
Yep. Although I would love to wait for the final release of 2.70 which was announced for in three months, for delivering it. For testing I would agree to test this.

AndWac:
superceded by #6302

PR #5915, opened 2020-07-19
There's no more 32-bit gobject-introspection

pyhalov:


pyhalov:
We have a problem here: python 3.4 was the last 32-bit python 3. Now we don't have one, and this means that only 64-bit gobject-introspection tools work. This can lead to some mess, as now random 32-bit python program can fail. We have to drop 32-bit introspection results and look for broken things. Luckily, our python staff already uses python 3.5.

pyhalov:
Want some quick reviews, as now all touched components are broken.

components/library/libpeas/Makefile
```diff
@@ -23,15 +23,17 @@
 # Copyright (c) 2015, 2017, Oracle and/or its affiliates. All rights reserved.
 #
 
-BUILD_BITS = 32_and_64
+# requires gobject-introspection
```
pyhalov, line 26: Luckily, libpeas is used only by 64-bit applications

AndWac (approved):
I skimmed through the changes and didn't find anything suspicious. I also ran gmake sample-manifest through the list of packages to check and everything went fine despite two differences for tracker (an additional plugin "trackerbird" has been generated because thunderbird has been detected) and pluma (additional .pyc files have been generated because cython is installed on my build machine).

PR #5916, opened 2020-07-19
gkrellm: fix build

pyhalov:


AndWac:
Bad news: gkrellm dumps core on my desktop ("IOT instruction").

PR #5921, opened 2020-07-20
Terminator

pyhalov:


AndWac (commented):
This PR has 2 commits; configobj doesn't look like a dependency of terminator. Shouldn't these be 2 PR's?

pyhalov:
It is a dependency, but only runtime. It's not found automatically, so they can be built in one PR.

PR #5923, opened 2020-07-20
gkrellm: fix crashes while gathering disk and battery info

pyhalov:


pyhalov:
solaris support seems to be rather ugly in gkrellm....

AndWac (approved):
Works on my desktop.

AndWac:
The code is unfinished and looks like it has been written in a hurry.

PR #5934, opened 2020-07-22
Add userland as prefered publisher to test installation in vagrant

Toasterson:
A small addition to vagrant that makes installing locally compiled packages a breeze on the vagrant dev VM.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -83,7 +83,8 @@ Vagrant.configure("2") do |config|
     pfexec pkg install build-essential
 
     cd /vagrant && gmake setup
-    pfexec pkg set-publisher --sticky --search-before=openindiana.org -g /vagrant/i386/repo userland
+    pfexec pkg set-publisher --non-sticky -g file://$HOME/oi-userland/i386/repo userland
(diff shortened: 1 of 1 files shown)


AndWac (approved):

pyhalov:
I disagree with this change. `--search-before` is necessary only when you do mass upgrade. Otherwise it can break normal upgrade paths. Also it breaks  zones (publisher is inaccessible from inside zone). Necessary steps to configure build environment are described in [documentation](http://docs.openindiana.org/dev/userland/). If we cold do it more generic...
What about configuring package server and adding this publisher as non-preferred non-sticky? Also you should do openindiana.org publisher non-sticky.

Toasterson:
So that if people look at the setup they copy something that does not break even in others aswell? Can do.

Toasterson:
I changed the publisher preference and stickyness as asked. I disagree that a whole package server is useful here. Vagrant is already a way to build in a container and I don't think people want to build in zones nor upgrade the VM at all. Vagrant's workflow is more like docker in that it downloads newer images when one updates.

pyhalov (approved):

PR #5936, opened 2020-07-23
rsync: update to 3.2.2

AndWac:
I reworked the file patch but I don't know whether it's really needed (I don't understand its description).
Furthermore there's a warning (already in our actual version):

configure.sh: WARNING: netinet/ip.h: present but cannot be compiled
configure.sh: WARNING: netinet/ip.h:     check for missing prerequisite headers?
configure.sh: WARNING: netinet/ip.h: see the Autoconf documentation
configure.sh: WARNING: netinet/ip.h:     section "Present But Cannot Be Compiled"
configure.sh: WARNING: netinet/ip.h: proceeding with the compiler's result
configure.sh: WARNING:     ## --------------------------------------------------- ##
configure.sh: WARNING:     ## Report this to http://rsync.samba.org/bugzilla.html ##
configure.sh: WARNING:     ## --------------------------------------------------- ##

### DIFF
--- a/components/network/rsync/Makefile
+++ b/components/network/rsync/Makefile
@@ -22,43 +22,44 @@
 #
 # Copyright (c) 2009, 2012, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2018, Michal Nowak
+# Copyright (c) 2020, Andreas Wacknitz
 #
 
+BUILD_BITS=			64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		rsync
-COMPONENT_VERSION=	3.1.3
-COMPONENT_REVISION=	1
+COMPONENT_VERSION=	3.2.2
 COMPONENT_SUMMARY=	Utility that provides fast incremental file transfer and copy
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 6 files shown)


pyhalov:
The patch was needed  to avoid embedding path to build area in compiled binary and in particular in error message.

components/network/rsync/Makefile
```diff
@@ -22,43 +22,41 @@
...
 
 COMPONENT_TEST_TRANSFORMER = $(NAWK)
 COMPONENT_TEST_TRANSFORMS = "'/^PASS|^SKIP|^FAIL|^XFAIL/'"
 
-# common targets
-build:		$(BUILD_64)
-
-install:	$(INSTALL_64)
-
-test:		$(TEST_64)
```
pyhalov, line 61: Is fakeroot needed for tests? If it is, add it as build dependency.
Otherwise looks good.

AndWac: How to use fakeroot? I haven't found any reference to it in the repository.

pyhalov: It's used by rsync testsuite if found. Just add it to REQUIRED_PACKAGES, as test results change if it's not present on the build system.

pyhalov (approved):

PR #5940, opened 2020-07-25
libetpan: update to 1.9.4

AndWac:
claws-mail needs to be rebuild after this

PR #5941, opened 2020-07-25
claws-mail: update to 3.17.6

AndWac:
This new version works on my desktop system (only IMAP account tested).


AndWac:
Should be merged after #5940

pyhalov (approved):

PR #5946, opened 2020-07-26
Upstream more of mate 1.24

mvwieringen:
This is a continuation of the following 3 pull requests: #5555, #5556, #5557 including a patch to make caja compile and
adding the following mater components (it also rebased on top of oi/hipster and removed eye of mate (eom) as
that already seems to be on 1.24 and not 1.22):

- mate-applets
- mate-notification-daemon
- mate-panel
- mate-session-manager
- mate-user-guide
- mozo
- engrampa
- mate-common (1.2.4.1)

The following are more involving due to missing support on OI:

- mate-control-center needs polkit available in solaris-userland but that
  needs mozjs-60 which is much more work but also available in solaris-userland
- atril needs external libtexlive which used to be internal not available in solaris-userland
- caja-extensions
- mate-power-manager
- mate-system-monitor need newer libgtop libs
- mate-utils


AndWac (changes requested):
You'll need to split the PR into several PR's like Mno-hime did it because base packages like mate-common need to be merged before its dependant packages. Otherwise the dependant packages will be build against the old mate versions.

mvwieringen:
> You'll need to split the PR into several PR's like Mno-hime did it because base packages like mate-common need to be merged before its dependant packages. Otherwise the dependant packages will be build against the old mate versions.

Ok given there is nothing new in the old commits I probably just make a new branch with only the changes on top of the 3 other PRs. Although those really should be rebased as they don't apply to the current oi/hipster tip.

AndWac:
Yes, rebasing will be needed. I am not sure whether it's the best idea to base your PR's on Mno-hime's, though.

mvwieringen:
> Yes, rebasing will be needed. I am not sure whether it's the best idea to base your PR's on Mno-hime's, though.

Hmm I will see maybe it makes more sense to close the old ones, I mostly downloaded the original patches and git am-ed the files to my branch. Lets wait and see if people have an idea how to do this as things have been lingering since February and
I mostly did this so I can use the new packages myself.

AndWac:
Just create new PR's based on Mno-hime's (leaving his Copyrights intact).
We can close his PR's after yours have been integrated.

mvwieringen:
Will open new PRs later tonight

PR #5949, opened 2020-07-26
Mate 1.24 part2 (rebased)

mvwieringen:
Rebased version of #5556


components/desktop/mate/mate-panel/Makefile
```diff
@@ -12,22 +12,23 @@
 # Copyright 2016 Till Wegmueller
```
AndWac: mate-panel depends on mate-menus, mate-desktop and libmateweather and thus it needs to be moved into PR part3 (#5950).

mvwieringen: @andwac this is in the last rebase.

PR #5950, opened 2020-07-26
Mate 1.24 part3 (rebased)

mvwieringen:
Rebased version of #5557 with working caja compile.


pyhalov:
How have  you tested  time-slider caja extension?

mvwieringen:
> How have you tested time-slider caja extension?

Nope, not yet. Then again the changes are quite straight forward when you google on on the two errors you find for the first one
a patch for  rhythmbox e.g. https://mail.gnome.org/archives/commits-list/2011-May/msg05912.html e.g. it seems g_date_time_format also supports all old time formats and more. For the second patch it makes no sense to convert the local time to a tm struct and then to the Gdate. As we already get the local unix time passed in via time_raw we can just convert that to a Gdate using g_date_time_new_from_unix_local()

pyhalov:
I'm fine with Mate PRs, but would like to see more testing...

AndWac:
mate-media needs mate-panel which should also be updated (in part2?).

mvwieringen:
> mate-media needs mate-panel which should also be updated (in part2?).

Ok let me move that then from one branch to an other this was however how the original pull requests were structured.

mvwieringen:
> > mate-media needs mate-panel which should also be updated (in part2?).
> 
> Ok let me move that then from one branch to an other this was however how the original pull requests were structured.

Ah its needs to move from 4 to 2. We probably also need to merge the libgtop stuff before some of the mate parts as things
like marco (what in a name :-)) and more are dependent on that, So over the coming times things might get moved around.

I already spend the whole weekend and yesterday so I would appreciate if we move forward with mate merging sooner then
later so not all effort is lost. I'm only afraid that we might need to settle for updating most of mate as some stuff will be really
hard and maybe not worth the trouble. It seems mate is getting also more dependencies on Linux based software.

AndWac:
Yes, tomorrow I will start some tests on my desktop with what I have now. If nothing bad happens I will approve the PR's and let Alexander decide what to do.
I completely uninstalled all mate-1.22 stuff so I found out that mate-media needs mate-panel (for whatever reason).
Alas mate is our only desktop environment.

AndWac:
Bad news: caja immediately crashes (segfault), even with caja-extensions uninstalled.

mvwieringen:
> Bad news: caja immediately crashes (segfault), even with caja-extensions uninstalled.

Ok new attempt could reproduce the crash so took an other approach this time importing the missing function from elsewhere
like what other projects do. This time caja works. Also republished all branches and have now all of mate compiling.

components/desktop/mate/caja-extensions/Makefile
```diff
@@ -11,22 +11,23 @@
 #
```
AndWac: caja-extensions need to be moved to part4 (#5951) because it depends on new caja.

mvwieringen: I will push rebased branches when we are sure we move forward with this. The old mate PRs  got stale again maybe
they should just be closed. In those rebased branches I will also adress moving some patches from one part to an other.

Things have been solid working for over 2 weeks for me and I have not encountered any problems.

PR #5961, opened 2020-07-30
nmap: update to 7.80

pyhalov:


pyhalov:
Unfortunately, making nmap work with py3 means inclusion of at least two big patches - https://github.com/nmap/nmap/pull/2088 and https://github.com/nmap/nmap/pull/1807 . 
Will wait at least for upstream to include them.

PR #5963, opened 2020-07-31
Make pkgdepend ignore GCC runtime libraries in /usr/lib/64

pyhalov:


AndWac (approved):
This fixes problems for mosh when running gmake REQUIRED_PACKAGES

PR #5968, opened 2020-08-01
Gimp

pyhalov:


pyhalov:
This is a big "gimp update" PR. Gimp still requires python 2.7, so no luck here, but at least it can be 64-bit.
PR will be splited in several parts:
1) babl, libmypaint, mypaint-brushes
2) gegl
3) gimp
4) gtkam, sane-fronteds, xsane rebuilds.

AndWac (commented):
You'll need to split this PR because you are adding new libraries which won't be available on the build system. Furthermore there are packages with new .so versions which might also need to be separated from the gimp build.

Sorry, I didn't see your comment before I wrote this. Please ignore it.

components/library/gegl/Makefile
```diff
@@ -10,79 +10,73 @@
 #
```
AndWac, line 10: gegl needs the updated babl:
meson.build:159:0: ERROR: Invalid version of dependency, need 'babl' ['>=0.1.78'] found '0.1.18'.

AndWac (approved):
I have installed gimp, gtkam, and xsane on my desktop.
Alas I don't have a supported scanner and no camera at hand. So I can only confirm the xsane and gtkam started but I wasn't able to further check them.
I opened some pictures (jpeg, tif, gif) and tried some color settings with them. All worked as expected.

pyhalov:
Merged as several separate PRs: #5976 #5977 #5978 #5979

PR #5980, opened 2020-08-03
Add libsynctex

mvwieringen:
Needed for building Mate 1.24 atril version

### DIFF
--- a/components/library/libsynctex/Makefile
+++ b/components/library/libsynctex/Makefile
@@ -1,0 +1,49 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2020 Marco van Wieringen
+#
+
+BUILD_BITS=             64
+BUILD_STYLE=		justmake
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		libsynctex
+COMPONENT_VERSION=      2.0.0
+COMPONENT_FMRI=		library/libsynctex
+COMPONENT_CLASSIFICATION= System/Libraries
+COMPONENT_SUMMARY=	Library for synchronization between TeX files and resulting file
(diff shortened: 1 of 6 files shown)


components/library/libsynctex/Makefile
```diff
@@ -0,0 +1,50 @@
...
+#
+
+BUILD_BITS=             64
+BUILD_STYLE=		justmake
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		libsynctex
+COMPONENT_VERSION=      2.0.0
+COMPONENT_REVISION=	0
```
AndWac: Remove COMPONENT_REVISION. We set it only if needed (starting with 1).

components/library/libsynctex/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_PROJECT_URL=  https://github.com/jlaurens/synctex
+COMPONENT_LICENSE=	MIT
+COMPONENT_LICENSE_FILE=	LICENSE
+
+TEST_TARGET=            $(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+PATH=/usr/gnu/bin:/usr/bin
+
```
AndWac, line 42: PATH=$(PATH.gnu)

AndWac, line 43: PATH=$(PATH.gnu)

pyhalov:
I have 2 questions:
1) Should we provide complete texlive instead?
2) In case we build atril without this library, what is missing?

mvwieringen:
> I have 2 questions:
> 
> 1. Should we provide complete texlive instead?
Honestly I think providing texlive completely is mostly overkill as atril only needs the parser I just packaged just enough so
not needing to have a big dependency.

> 2. In case we build atril without this library, what is missing?
As far as I remember you cannot build without it but I could be wrong. The actual thing is that the parser I now
packaged used to be part of atril as bundled source but they decided as most Linux distributions deliver it as
part of texlive to make it an external dependency. I now also added the latest version from the original author of
the parser which is in a different repo then texlive. (The only disadvantage is that in that repo there is no build
system so I had to hack a rudimentary Makefile but given that the whole parser is 2 source files and some headers
the current solution is the most lightweight and simplest solution for providing atril with its needed now external source
for the parser and not needing to package the full texlive stuff.)

AndWac:
I have been thinking about providing texlive like OmniOSce is doing. It is on my wishlist for some time but I haven't found enough time for it.

mvwieringen:
@AndWac this one also has to merge before mate 1.24 can merge. But you can always drop this stuff and replace it with the full texlive in the future.

PR #5981, opened 2020-08-03
Update libgtop to 2.40.0

mvwieringen:
Needed for mate-1.24 original version is pretty hacked this new version is trying to only import the additional stuff and the original patches don't apply at all due to things having changed quite severely.  This needs more testing as a least the % CPU is not correct. But mate-system-monitor mostly works which is quite a accomplishment already.

libgksu needs rebuilding due to the new version of libgtop.


components/library/libgksu/Makefile
```diff
@@ -10,13 +10,16 @@
 
 #
```
AndWac: If libgksu needs to be rebuilt after the libgtop update, you'll need to put it into a separate PR otherwise libgksu will be build against the old libgtop (new packages get installed after creating all packages of a PR).

components/library/libgtop/Makefile
```diff
@@ -11,31 +11,32 @@
...
+
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= libgtop
-COMPONENT_VERSION= 2.28.5
-COMPONENT_REVISION= 3
-# Some historical artefact in package versioning
-IPS_COMPONENT_VERSION= 2.30.0
+COMPONENT_VERSION= 2.40.0
+COMPONENT_REVISION= 0
```
AndWac: Remove COMPONENT_REVISION

components/library/libgtop/Makefile
```diff
@@ -11,31 +11,32 @@
...
 COMPONENT_SUMMARY= Library to get system specific data
 COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.xz
 COMPONENT_ARCHIVE_HASH= \
-  sha256:c812c174e44a8971a1f33265437cfd10f1e99869c7f5c05f8ee95846a70a3342
+  sha256:78f3274c0c79c434c03655c1b35edf7b95ec0421430897fb1345a98a265ed2d4
 COMPONENT_ARCHIVE_URL= \
-  http://ftp.gnome.org/pub/GNOME/sources/$(COMPONENT_NAME)/2.28/$(COMPONENT_ARCHIVE)
+  http://ftp.gnome.org/pub/GNOME/sources/$(COMPONENT_NAME)/2.40/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL = http://www.gnome.org
```
AndWac, line 30: I recommend to use https instead of http where possible.

components/library/libgksu/Makefile
mvwieringen: There is quite a bit that needs rebuilding. Honestly this should be something for the build system to handle correctly.
For now I just added the non mate ones as those will be rebuild because of the upgrade to 1.24 anyway.

components/library/libgtop/patches/libgtop-01-illumos.patch
```diff
@@ -0,0 +1,1765 @@
+diff -ru --new-file libgtop-2.40.0/aclocal.m4 libgtop-2.40.0.new/aclocal.m4
```
pyhalov: Avoid patching generated files, if possible (aclocal, Makefile.in, configure...).

pyhalov:
Have you looked at https://github.com/oracle/solaris-userland/tree/master/components/gnome/libgtop ?

mvwieringen:
> Have you looked at https://github.com/oracle/solaris-userland/tree/master/components/gnome/libgtop ?

I had not as I didn't find it. I will try importing the patches from there into the existing build. If that works better
probably settle on that. I get the feeling the original patches also came from that source.

components/library/libgtop/patches/libgtop-01-illumos.patch
mvwieringen: Ok need to find a way to run aclocal first as otherwise the build blows up.

pyhalov: Usually this looks like 
`COMPONENT_PREP_ACTION = (cd $(@D) && autoreconf -fi)`

mvwieringen:
Ok tried importing the Solaris patches which partly works but mate-system-monitor now core dumps when selecting the processes tab.

pyhalov:
Also look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/library/libgtop/patches/libgtop-04-ncpu.patch.
Without looking at code and core it's hard to find out what's going on.

mvwieringen:
> Also look at https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/library/libgtop/patches/libgtop-04-ncpu.patch.
> Without looking at code and core it's hard to find out what's going on.

Ok the mentioned patch is already in Solaris userland too. It only seems they increased the number of cpus etc so some
structs are completely different size so need to recompile everything using this new lib. So when I get this all done I probably
update all pull requests.

mvwieringen:
New attempt based on patches from solaris-userland after recompiling mate-system-monitor things work ok now. Only added extra part to libgtop-05-ncpu.patch to be sure GLIBTOP_NCPU is defined.

components/library/libgksu/Makefile
AndWac: Your libgksu change won't have an effect despite modernizing the Makefile. It will be build by jenkins within the same run. You are right that the automatic rebuild of dependant packages is something the build system should handle but at the moment it does not.

mvwieringen: Ok rebased the branch to only include the libgtop changes.

AndWac:
We need a PR for a rebuild of libgksu after this has been integrated.

mvwieringen:
> We need a PR for a rebuild of libgksu after this has been integrated.

I will push a new branch and PR when this integrates containing the patch that was in this branch before.
But lets first get this reviewed and merged.

AndWac:
How can we proceed with this? Can we merge it separately from the other Mate PRs (will it work with old Mate)?

mvwieringen:
> How can we proceed with this? Can we merge it separately from the other Mate PRs (will it work with old Mate)?

I just compiled the old mate sources with the new lib and at least it builds. Its also a dependency  for mate-1.24 e,g, that needs a
certain minimum version which is newer then the one currently available. The only other things that depens on this is libgksu and that builds fine too as seen in the other pull request that is needed anyway after merging this. This version and its patches is also what solaris-userland uses so I think upgrading this is a wise thing to do even when not upgrading mate.

AndWac:
What do we need to rebuild if we merge this? Can you create a separate PR for depending packages?
Next month is a new OI Hipster release due. I want to merge some things before but we also need some time to fix possible breakages.

mvwieringen:
I will add one for the one mate package to the just created pull request for libgksu

mvwieringen:
Ok its quite some work to rebuild all dependent packages. So I put them all in one pull request. Mostly mate ones.

PR #5992, opened 2020-08-09
dmenu: update to 4.9 and drop 32-bit

AndWac:


components/desktop/dmenu/patches/01-config-mk.patch
```diff
@@ -0,0 +1,17 @@
...
++PREFIX = /usr
+ MANPREFIX = $(PREFIX)/share/man
+ 
+ X11INC = /usr/X11R6/include
+@@ -28,4 +28,4 @@
+ LDFLAGS  = $(LIBS)
+ 
+ # compiler and linker
+-CC = cc
++CC = gcc
```
pyhalov: Should we just remove `CC = cc` or set to `CC ?= gcc` to use oi-userland gcc, not any?

pyhalov (changes requested):

AndWac:
Ok, I removed it.

AndWac:
Are you fine with the change?

pyhalov:
Seems to be fine

pyhalov (approved):

PR #6001, opened 2020-08-11
nasm: update to 2.15.3

AndWac:
Tests run fine.
sample-manifest didn't change.

### DIFF
--- a/components/developer/nasm/Makefile
+++ b/components/developer/nasm/Makefile
@@ -14,30 +14,27 @@
 # Copyright 2019, Michal Nowak
 #
 
+BUILD_BITS=			64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		nasm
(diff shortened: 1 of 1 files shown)


pyhalov:
Have you tried to build something, requiring nasm?

AndWac:
No, there doesn't seem to be any dependant packages, at least
pkg search -r -H -o pkg.name "depend::developer/assembler/nasm" only showed build-essential.
I just run the internal test suite.

pyhalov:
Of course, there will be no runtime dependency, but there are build time dependencies,

AndWac:
Ok, I successfully build x265 and libreoffice. I even installed that libreoffice and it worked like before (albeit I don't know where the assembler comes into play so I am not sure whether it really get used when I played with lo writer).

pyhalov (approved):

PR #6009, opened 2020-08-15
couchdb-21: update netaddr, python dependency to 3.5

AndWac:
There doesn't seem to be a direct dependency on the versions. All contained python script just use a python interpreter without an explicit version.


components/database/couchdb-21/couchdb.p5m
```diff
@@ -30,8 +30,8 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 depend type=require fmri=runtime/erlang
-depend type=require fmri=runtime/python-27
-depend type=require fmri=library/python/netaddr-27
+depend type=require fmri=runtime/python-35
+depend type=require fmri=library/python/netaddr-35
```
pyhalov: There's no netaddr-35

PR #6011, opened 2020-08-16
sqlite: update to 3.33.0

AndWac:
- All test run fine.
- Installed on my desktop system; yelp works fine
- According to https://abi-laboratory.pro/index.php?view=timeline&l=sqlite fully backwards compatible upt to 3.32.3 (no later data available)
- Dependency on runtime/tcl-8 added as it is mentioned in sqlite's docutmentation as a build dependency
- We don't ship tcl bindings yet, thus I commented them out.

### DIFF
--- a/components/database/sqlite/Makefile
+++ b/components/database/sqlite/Makefile
@@ -22,17 +22,17 @@
 # Copyright (c) 2020 Michal Nowak
 #
 
-BUILD_BITS=		32_and_64
+BUILD_BITS=			32_and_64
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		sqlite
-COMPONENT_VERSION=	3.31.1
-TARBALL_VERSION=	3310100
+COMPONENT_VERSION=	3.33.0
+TARBALL_VERSION=	3330000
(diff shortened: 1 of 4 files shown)


components/database/sqlite/Makefile
```diff
@@ -88,10 +88,11 @@ COMPONENT_POST_INSTALL_ACTION += $(MKDIR) $(PROTO_DIR)$(USRSHAREMAN1DIR) ; $(CP)
 
 COMPONENT_TEST_TARGETS = test
 
-# Build dependencies
-REQUIRED_PACKAGES += library/libedit
+# Manually added build dependencies
+REQUIRED_PACKAGES += runtime/tcl-8
```
pyhalov: Is it really needed?
We build tcl package in separate component - https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/tcl/tcl-sqlite/Makefile

AndWac: That's what I read out of the documentation (README.md). It's not detected automatically so I can remove it. The changes in the sample-manifest are generated. IMO is doesn't hurt as I commented out the sqlite tcl bindings in the p5m file.

pyhalov: Remove and merge it.

PR #6014, opened 2020-08-17
unrar: update to 5.91

AndWac:
- Checked with an example RAR file: unrar t  and unrar e still work with that file.
- The source code version (5.9.4) is different from what the app itself says (5.91). So I set the IPS_COMPONENT_VERSION to what the app itself states.


pyhalov:
People are strange...

pyhalov (approved):

PR #6019, opened 2020-08-25
net-snmp: 64-bit fixes, build depends on picl headers

jclulow:
I was trying to build the `net-snmp` package by doing, effectively:

```
$ cd /ws/oi-userland/components/sysutils/net-snmp
$ gmake publish
```

First, the build failed because I had not installed the PICL headers; the actual failure mentions a SPARC platform-specific header that of course isn't on my x86 host -- but the cause ends up being that it did not detect the right PICL header during the configure process.

Second, the packaging step appeared to fail because (and I am still new to this bit) it seems like the default posture for binary directories has flipped in the last few months.  We are now defaulting 64-bit binaries (where 32- and 64-bit binaries are both built) into `/usr/bin` instead of `/usr/bin/$MACH`, so the manifest doesn't know where to find them.

I also ran `gmake update-metadata` in this directory to fix the `pkg5` file.  Let me know if I missed anything!

Cheers.

### DIFF
--- a/components/sysutils/net-snmp/Makefile
+++ b/components/sysutils/net-snmp/Makefile
@@ -37,7 +37,7 @@ PATH=$(VPATH):$(PATH.gnu)
 
 COMPONENT_NAME=		net-snmp
 COMPONENT_VERSION=	5.8
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_PROJECT_URL=	http://www.net-snmp.org/
 COMPONENT_FMRI= system/management/snmp/net-snmp
 COMPONENT_SRC=    $(COMPONENT_NAME)-$(COMPONENT_VERSION)
@@ -168,3 +168,5 @@ REQUIRED_PACKAGES += system/library/math
 REQUIRED_PACKAGES += system/management/snmp/net-snmp
 # Bogus dependency due to libssp
 REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
+# Without the picl headers, the build fails on a missing SPARC-specific file.
+REQUIRED_PACKAGES += system/header/header-picl
--- a/components/sysutils/net-snmp/manifests/sample-manifest.p5m
+++ b/components/sysutils/net-snmp/manifests/sample-manifest.p5m
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2018 <contributor>
+# Copyright 2020 <contributor>
 #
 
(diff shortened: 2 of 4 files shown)


components/sysutils/net-snmp/Makefile
```diff
@@ -168,3 +168,5 @@ REQUIRED_PACKAGES += system/library/math
 REQUIRED_PACKAGES += system/management/snmp/net-snmp
```
AndWac, line 168: Please set COMPONENT_REVISION=2

jclulow:
Component revision bumped.  Thanks!

AndWac:
Regarding to your statement about flipping the preferred 64-bit binaries: That happened in PR #5836 .
And this has an additional impact on your PR: Please create a new sample-manifest.p5m by running "gmake sample-manifest" and add the new manifest/sample-manifest.p5m file to your commit.
Background: the sample-manifest.p5m has two purposes:
1. For new packages it is the base for the package's manifest file.
2. It can be used to check whether a new version (or any other change) has an impact on the files a package provides. That is, if the sample-manifest.p5m changes you most probably have to care for the package's manifest file, too. So it's important to update the sample-manifest.p5m in case of any changes, too.

jclulow:
I have updated the sample manifest.  Note that because it appears to be generated using the 32-bit proto area, and we are now using the 64-bit proto area for the binaries, the manifest no longer quite reflects what we're shipping.

jclulow:
## Testing Notes

I installed the package from my local repository:

```
$ pfexec pkg install pkg://userland/system/management/snmp/net-snmp@5.8,5.11-2020.0.1.2:20200902T202731Z
            Packages to update:   2
       Create boot environment:  No
Create backup boot environment: Yes

Completed                                2/2         55/55      3.5/3.5      --

Removing old actions                           10/10
Installing new actions                         11/11
Updating modified actions                      63/63
Updating package cache                           2/2

$ file /usr/bin/snmpget
/usr/bin/snmpget:       ELF 64-bit LSB executable AMD64 Version 1, dynamically linked, not stripped, no debugging information available

$ /usr/bin/snmpget
No hostname specified.
USAGE: snmpget [OPTIONS] AGENT OID [OID]...

  Version:  5.8
  Web:      http://www.net-snmp.org/
  Email:    net-snmp-coders@lists.sourceforge.net

OPTIONS:
  -h, --help            display this help message
  -H                    display configuration file directives understood
  -v 1|2c|3             specifies SNMP version to use
  -V, --version         display package version number
...
```

PR #6020, opened 2020-08-25
python/more-itertools: fix typo in Makefile

jclulow:
Correct a small typo in the Makefile.  Without this, I was missing a package for whichever of the two versions is not the default.


components/python/more-itertools/Makefile
```diff
@@ -29,7 +29,7 @@ COMPONENT_CLASSIFICATION=	Development/Python
 COMPONENT_LICENSE=	MIT
```
AndWac, line 29: Please set COMPONENT_REVISION=2

jclulow:
Component revision bumped!  Thanks.

PR #6021, opened 2020-08-25
python/cherrypy: build depends on setuptools-scm

jclulow:
I was using the (relatively?) new `userland-zone` tool to set up a build zone that was isolated from the network.  I noticed that the build for cherrypy was trying to download additional Python modules from the Internet.

With these added dependencies, the build then proceeds correctly in the isolated zone.


components/python/cherrypy/Makefile
```diff
@@ -59,7 +59,9 @@ REQUIRED_PACKAGES += library/python/nose-35
 
 # Auto-generated dependencies
```
AndWac, line 60: Please set COMPONENT_REVISION=3

jclulow:
Component revision bumped!  Thanks.

PR #6022, opened 2020-08-25
developer/gcc49: need prior GCC 4.9 installed to build

jclulow:
Using an isolated (`userland-zone`) build zone, without most packages installed, I noticed that the GCC 4.9 build would fail without having GCC 4.9 installed already.

### DIFF
--- a/components/developer/gcc49/Makefile
+++ b/components/developer/gcc49/Makefile
@@ -18,7 +18,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= gcc
 COMPONENT_VERSION= 4.9.4
-COMPONENT_REVISION=6
+COMPONENT_REVISION= 7
 MPFR_NAME= mpfr
 MPFR_VERSION=3.1.3
 MPC_NAME=mpc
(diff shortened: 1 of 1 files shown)


components/developer/gcc49/Makefile
```diff
@@ -128,3 +128,5 @@ REQUIRED_PACKAGES += system/library/gcc-4-runtime
 REQUIRED_PACKAGES += system/library/gfortran-4-runtime
```
AndWac, line 128: Please set COMPONENT_REVISION=7

Whenever you change a package you'll have to
- either change its COMPONENT_VERSION (and remove COMPONENT_REVISION if there)
- or change its COMPONENT_REVISION (either set to 1 if not yet there or increment an existing one)

If neither COMPONENT_VERSION (nor IPS_COMPONENT_VERSION) nor COMPONENT_REVISION changes a newly created package won't be deployed.
(Of course here it's not strictly necessary because you just want to fix the build but we usually follow the rules in these cases, too)

jclulow: That's good to know, thanks!

I was trying to understand exactly how that all hangs together -- it seemed like "bass-o-matic" would make some guesses based on looking at the commit history?

If one should always explicitly bump the component version/revision, though, that makes sense to me and I'll be sure to do that in future.

jclulow:
I have bumped the revision!

PR #6045, opened 2020-09-05
Update Postgresql 9.5, 9.6, 10, 11, and 12

AndWac:
- All test run fine
- I commented out .a library files (https://www.illumos.org/issues/13073) and rearranged some link actions


pyhalov:
Remove static libraries only if there are dynamic libraries. Thisis not True for Postgresql server libraries.

AndWac:
Ok, I removed the # and activated the static libraries again. Thanks for the explanation.

pyhalov (approved):

PR #6053, opened 2020-09-10
packaging metadata improvements

jclulow:
I have made a few improvements:

* The order of elements in lists in the metadata files (e.g., `pkg5` files) is dependent on locale at present.  By forcing `sort -u` into the C locale, we get a consistent ordering no matter what locale is used on the build system.
* Python packages generally provide both a _concrete_ package, for each Python runtime version (e.g., `cffi-35` and `cffi-27`); they also provide a _generic_ package (e.g., `cffi`) which uses a conditional dependency to pull in the concrete versions for whichever Python runtimes are actually installed.  The `mapping.json` file does not contain the generic versions in the supplied package FMRI list, so build tools cannot locate the correct component directory based on the generic FMRI.  I have enabled the Makefiles to include the generic FMRI for Python packages.
* In order to make mass-regeneration of the `pkg5` files easier, I have enabled the recursive use of `update-metadata` within the `components/` directory; you can now, e.g., `gmake -s -C components -j8 update-metadata` and it should do the right thing as quickly as possible.
* Added the `dependencies.json` and `mapping.json` auto-generated files to the git ignore list.

In the case of the changes that affect the `pkg5` files, I have kept the change itself separate from the metadata update to make it easier to see what's going on.  I also made the sort order change separate from the Python manifest change, as it makes the diff more relevant there.


alarcher (approved):

alarcher:
@jclulow Thanks for doing this! If you are interested I also have a branch with additional logic to do batch rebuilds in order of dependencies, I can push it later today.
@AndWac If you are fine with it, feel free to merge.

PR #6058, opened 2020-09-12
Rebuilds needed for new libgtop version

mvwieringen:


AndWac:
Ok, do you consider both PR's ready for integration?

mvwieringen:
I can only test the new libgtop with the new mate stuff there its been stable for weeks.
As to how things work out in the older mate no idea think it should be fine but I cannot test it.
You could always try yourself with the new generated packages like what you did with the new mate packages.
I have long blow my older bootenviroments aways as I will run with the newer mate stuff which is fine for my usage.

AndWac:
Ok, I am inclined to merge Mate-1.24. Can you fix the small things I requested earlier?

mvwieringen:
I think I did most things like move things from one to an other. What is still open ?

AndWac:
Ok, I will integrate Mate-1.24. libgksu needs to be rebuild, the other packages from this PR don't. It would be best to remove them. Sorry, for the inconvenience, but you convinced me to merge the newer versions.

mvwieringen:
Ok will drop then them, will push a new mate-1 branch in a moment.

PR #6059, opened 2020-09-12
Mate 1.24 part 5

mvwieringen:
Ok last part with patches mate-power-manager and mate-system-monitor with the polkit stuff removed.


mvwieringen:
@andwac here are the last 2 mate utils which now publish correctly and have polkit stuff removed just as was done partly before but it seems these tools got more deps on polkit in 1.24.

AndWac:
Alas part4 failed in mate-control-center for no apparent reason. At least a manual "gmake build" in the jenkins workspace worked fine but
the next jenkins run failed again.

AndWac:
I commented too early. The last jenkins run went fine but fooled me because it went red during its run and then finished successfully.

mvwieringen:
Yeah I do see it as published.

PR #3673, opened 2017-11-05
[IN DISCUSSION] Provide packages for different NVidia driver types - mediated

jimklimov:
Alternative/follow-up to #3458 with proposed mediation of FS objects delivered by each driver in such a way that they can be installed all at once (or in any combination), and then the user can pick one to be enabled via IPS mediation.

NOTE: Still not quite sure how to properly address the driver_aliases in a way that drivers won't conflict over common cards, and won't claim ones they do not know either. And also having the active driver named `nvidia` is probably still a goal.

Hopefully would supersede the PR #5337 as well ;)

To build this in a local testing environment, you'd probably need to set up a local publisher, `make publish` the `nvidia-select` into it, and then piggy-back on encumbered builds support to keep the `pkglint` satisfied while publishing the actual driver packages and the `nvidia` metapackage (new package, old name). So something like:

````
:; ( cd components/openindiana/nvidia-select && make publish )
:; for N in 304 340 084 387 390 396 ; do ( \
    cd components/openindiana/nvidia-$N && make ENCUMBERED=yes publish ; \
   ) ; done
:; ( cd components/meta-packages/nvidia && make ENCUMBERED=yes publish )
````

and then you should be able to install them into a local test BE which would use this publisher (possibly using the full FMRIs to publisher+package names).

PR #6060, opened 2020-09-13
NVIDIA: provide alternative drivers

AndWac:
This is based on Mno-himes work #5337.
I am not sure what other things are necessary:
- do we need to obsolete the original nvidia package?


Toasterson:
Am I right in the assumption, that we have no more driver/graphics/nvidia right now? Then we need to make one that replaces driver/graphics/nvidia. (rename) or make driver/graphics/nvidia a metapackage, that holds the mediator links for ogl-select service. and the driver/graphics/nvidia-340 and driver/graphics/nvidia-390 have only the files themselves.

AndWac:
Yes, driver/graphics/nvidia will be replaced by driver/graphics/nvidia-340 (for the moment). The new nvidia-390 is added. Later we can add more drivers and change the default.

Toasterson:
Then we need to rename the old package. Meaning publish a new version with the rename action

AndWac:
Like this? Can you check the version information, I am not sure whether to increment it or not.

Toasterson:
I have absolutely no clue how that process works. I just know there needs to be a package driver/graphics/nvidia with a higher version number than before that has a rename action in it. (Special kind of set action) if we generate that I do not know how.

components/openindiana/nvidia-390/driver-graphics-nvidia.p5m
```diff
@@ -0,0 +1,25 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2019 Michal Nowak
+#
+
+set name=pkg.fmri value=pkg:/driver/graphics/nvidia@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
```
pyhalov, line 16: We publish two driver/graphics/nvidia packages as result.

pyhalov:
Can we make these two drivers coexist? Look at https://github.com/OpenIndiana/oi-userland/pull/3673

pyhalov:
While #3673 is hard to support, some simpler mediated  solution would be interesting.

components/openindiana/nvidia-390/driver-graphics-nvidia.p5m
AndWac: What would be the better solution:
- keep this file and remove the history and server_desktop change
or
- remove this manifest and control explicitly?

AndWac:
Yes, my plan is to get a simple solution now, gain some experience, maybe add more driver versions and then try to create something more comfortable.

despair86:
ha
i ran into a nasty set of bugs with the stock installation package cause it _really_ wants to use `gfx_private` from Oct 2007

they still occur but at least when X crashes i can go back to the console (the old driver would leave me at a black screen)

unfortunately i have a fairly recent gpu which requires quadro R435+

despair86:
re: driver errors - i was able to dig up the kmod sources from the other half of the root pkg that isn't installed - comes with the object code for the driver itself + source for the Solaris kernel interfaces - same way the linux and freebsd driver sources are shipped - this object code is _exactly the same_ for all three platforms (!)

was able to recompile the source and link against the closed-source bits (using a similar process to the way the nvidia linux kmod is rebuilt after kernel update) but the new driver triggered a page fault and was unusable

compiling with gcc-9 triggered a million and one relocation errors in krtld (gas or /usr/ccs/bin/as from a 2018 release not the 2009 version)

jpandre:
Not sure this is related, I suspect it could be.

On bootup I see errors related to loading nvidia_modeset. As I do not have any Nvidia graphic card, I wanted to uninstall the nvidia packages, but when I uninstall nvidia-390, nvidia-340 gets installed, and conversely (on a system updated from an older version).

Is nvidia a compulsory package irrespective of Nvidia hardware ?

Toasterson:
Hey @jpandre Merged PR's are not a really good place to ask questions, as not everybody will see it. Can you ask on the mailing list or in the discussion parts of Github?

PR #6065, opened 2020-09-13
Update GTK3 to 3.24.23

mvwieringen:
Think its time after almost 1 year to sync gtk3 to the latest stable. 
From 3.24.12 to 3.24.23 so no major minor change. Also don't think there is any ABI/API breakage.
First tests with mate stuff looks ok.


AndWac (approved):
- "gmake test" runs fine
- I installed it on my desktop system and successfully tested a bunch of applications

PR #6067, opened 2020-09-13
Update GNU gettext to 0.21

mvwieringen:
Ran make test which finished was mostly ok.
One test failed: msgmerge-compendium-6

```
-# TOTAL: 433
-# PASS:  413
-# SKIP:  20
+# TOTAL: 445
+# PASS:  421
+# SKIP:  23
 # XFAIL: 0
 # XFAIL: 0
-# FAIL:  0
+# FAIL:  1
 # XPASS: 0
 # XPASS: 0
 # ERROR: 0
```

### DIFF
--- a/components/text/gettext/Makefile
+++ b/components/text/gettext/Makefile
@@ -22,24 +22,23 @@
 #
 # Copyright (c) 2011, 2015, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2020, Marco van Wieringen
 #
 
+BUILD_BITS=		32_and_64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		gettext
-COMPONENT_VERSION=	0.20.1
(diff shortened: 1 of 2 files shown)


AndWac:
I have diffferent failures for amd64 and i86:
amd64: FAIL: test-access
i86: FAIL: test-access
         FAIL: test-explicit_bzero
Plus there are multiple test suites each with its own statistics.

mvwieringen:
Yeah maybe it makes more sense to forget this. The idea was that this version seems to support ITS which I thought was needed
for mate but as I now fixed that in a other way e.g. drop the polkit policies there is no real urge to upgrade this. For now I just leave it as something possible but I'm not sure if the failures are at all so problematic.

AndWac:
We have several packages with increasing failed tests. Solaris-userland has similar problems and usually patches the failing tests away (where not possible to fix). I don't like this attitude because it's hiding possible problems. My preferred way is to accept failues but make them visible. It's also a bad sign that our systems show different failures.

PR #6093, opened 2020-09-22
bug #9732: pulseaudio always uses the first audio device

jgmills:
With the new patches, pulseaudio handles multiple audio devices in the same way that audioplay and the other audio commands do. The module-detect module does this magic. It first attempts to get the audio device path name from the AUDIODEV environment variable. If this fails, it uses /dev/audio instead. Regardless, the path name must resolve to a character device. It obtains the instance number from this path to set the argument to device=/dev/dspN where N is the instance number, and then loads the module-oss module with this argument. The module-oss module opens and manipulates the correct devices with this instance number. Note that the module-oss module is commented out in /etc/pulse/default.pa, so that this module does not load by itself.


jgmills:
With the new patches, pulseaudio handles multiple audio devices in the same way that audioplay and the other audio commands do.  The module-detect module does this magic.  It first attempts to get the audio device path name from the AUDIODEV environment variable.  If this fails, it uses /dev/audio instead.  Regardless, the path name must resolve to a character device.  It obtains the instance number from this path to set the argument to device=/dev/dspN where N is the instance number, and then loads the module-oss module with this argument.  The module-oss module opens and manipulates the correct devices with this instance number.  Note that the module-oss module is commented out in /etc/pulse/default.pa, so that this module does not load by itself.

jgmills:
With the new patches, pulseaudio handles multiple audio devices in the same way that audioplay and the other audio commands do.  The module-detect module does this magic.  It first attempts to get the audio device path name from the AUDIODEV environment variable.  If this fails, it uses /dev/audio instead.  Regardless, the path name must resolve to a character device.  It obtains the instance number from this path to set the argument to device=/dev/dspN where N is the instance number, and then loads the module-oss module with this argument.  The module-oss module opens and manipulates the correct devices with this instance number.  Note that the module-oss module is commented out in /etc/pulse/default.pa, so that this module does not load by itself.

jgmills:
With the new patches, pulseaudio handles multiple audio devices in the same way that audioplay and the other audio commands do.  The module-detect module does this magic.  It first attempts to get the audio device path name from the AUDIODEV environment variable.  If this fails, it uses /dev/audio instead.  Regardless, the path name must resolve to a character device.  It obtains the instance number from this path to set the argument to device=/dev/dspN where N is the instance number, and then loads the module-oss module with this argument.  The module-oss module opens and manipulates the correct devices with this instance number.  Note that the module-oss module is commented out in /etc/pulse/default.pa, so that this module does not load by itself.

jgmills:
With the new patches, pulseaudio handles multiple audio devices in the same way that audioplay and the other audio commands do.

AndWac (approved):
I have tested it on my desktop. With these patches volume control works now.

AndWac:
Congratiolations and thank you for your contribution!

PR #6094, opened 2020-09-22
add minicom 2.7.1

jclulow:
This adds a package for the minicom serial communications program.  It was not super clear to me if there was a better upstream for the source than the one I chose here.

Things appear to build and install and work correctly:

```
$ pkg list minicom
terminal/minicom                                  2.7.1-2020.0.1.0           i--

$ /usr/bin/minicom --version
minicom version 2.7.1 (compiled Sep 22 2020)
Copyright (C) Miquel van Smoorenburg.

This program is free software; you can redistribute it and/or
modify it under the terms of the GNU General Public License
as published by the Free Software Foundation; either version
2 of the License, or (at your option) any later version.

$ ldd /usr/bin/minicom
        libsocket.so.1 =>        /lib/64/libsocket.so.1
        libncurses.so.5 =>       /usr/lib/64/libncurses.so.5
        libc.so.1 =>     /lib/64/libc.so.1
        libnsl.so.1 =>   /lib/64/libnsl.so.1
        libmp.so.2 =>    /lib/64/libmp.so.2
        libmd.so.1 =>    /lib/64/libmd.so.1
        libm.so.2 =>     /lib/64/libm.so.2
```


components/sysutils/minicom/.gitignore
```diff
@@ -0,0 +1 @@
+/minicom-*/
```
AndWac, line 1: Is this really necessay? We usually only have a .gitignore in the root folder.

jclulow: It would be really great to be able to use `git status` without having to disable showing untracked files.  I think putting a `.gitignore` in each component directory that tracks the locally expected detritus makes a lot of sense.  In particular, if we move or rename a component directory, or just delete a component in future, it will be impossible to forget to update the top-level `.gitignore` as it will move or be removed with the component.

AndWac (approved):

components/sysutils/minicom/.gitignore
AndWac: Ok, I am convinced ;-)

PR #6096, opened 2020-09-24
correct various build dependency issues

jclulow:
I have been running builds of every component in a template zone which does not have Internet access, and which mostly just has `build-essential` installed -- plus whatever is listed in the `REQUIRED_PACKAGES` in the Makefile.  My tool destroys and recreates the zone from the template before starting each new package build, and looks at `pkg5` to see which packages need to be installed in the zone before we boot it to run the build.

I discovered the missing dependencies mostly through trial and error.  Sometimes the package would build, but not produce all files, so publish would fail.  Sometimes `configure` or the subsequent `make` would fail.

With these fixes I have been able to make forward progress!  More fixes are coming as well.


components/desktop/tracker/pkg5
```diff
@@ -13,6 +14,7 @@
         "library/audio/taglib",
         "library/desktop/atk",
         "library/desktop/gdk-pixbuf",
+        "library/desktop/gobject/gobject-introspection",
         "library/desktop/gtk3",
         "library/desktop/libgsf",
         "library/giflib",
```
AndWac, line 20: In my environment tracker's sample-manifest.p5m also changed. Can you check this and possibly merge the changes back into the appropriate manifest file? This might have been overlooked at the last update.

AndWac (commented):
Aurélien was too fast for me...

PR #6097, opened 2020-09-24
cluster/pcs: pre-download all required source

jclulow:
I have been doing builds in a zone without networking access.  I make sure to run `gmake download` in the component directory before entering the zone to run `gmake publish`.

The `cluster/pcs` component downloads a few extra things during the _build_ rather than using the proper artefact management mechanism.  This change moves to specific reproducible versions with hash checks that are completely preloaded by `gmake download`.  It also stops us shipping cached copies of the gem files in the final package.

Test package build and published.


components/cluster/pcs/Makefile
```diff
@@ -26,6 +27,87 @@ COMPONENT_ARCHIVE_URL= https://github.com/ClusterLabs/$(COMPONENT_NAME)/archive/
 COMPONENT_LICENSE= GPLv2
 COMPONENT_SUMMARY= Pacemaker command line interface and GUI.
 
+#
+# The build process tries to use wget to download an archive without even a
+# check for the correct hash.  In order to allow the build to work without
+# networking, we download it first.  This will need to be updated if the
+# version or URL in "pcs/snmp/Makefile" changes.
+#
```
AndWac, line 35: pcs/snmp/Makefule? What has snmp to do with it?

components/cluster/pcs/Makefile
```diff
@@ -26,6 +27,87 @@ COMPONENT_ARCHIVE_URL= https://github.com/ClusterLabs/$(COMPONENT_NAME)/archive/
...
+COMPONENT_ARCHIVE_URL_1= https://github.com/ondrejmular/pyagentx/archive/v$(PYAGENTX_VERSION).tar.gz
+
+#
+# The package includes a private copy of a series of Ruby Gems.  We must
+# download these in advance so that the build can run without network access.
+#
+COMPONENT_ARCHIVE_2= backports-3.11.3.gem
+COMPONENT_ARCHIVE_HASH_2= sha256:57b04d4e2806c199bff3663d810db25e019cf88c42cacc0edbb36d3038d6a5ab
+COMPONENT_ARCHIVE_URL_2= https://rubygems.org/gems/backports-3.11.3.gem
+CACHED_GEM_FILES += $(COMPONENT_ARCHIVE_2)
```
AndWac, line 48: You should use $(COMPONENT_ARCHIVE_2) here instead repeating the name.

components/cluster/pcs/Makefile
```diff
@@ -26,6 +27,87 @@ COMPONENT_ARCHIVE_URL= https://github.com/ClusterLabs/$(COMPONENT_NAME)/archive/
...
+
+COMPONENT_ARCHIVE_3= ethon-0.11.0.gem
+COMPONENT_ARCHIVE_HASH_3= sha256:88ec7960a8e00f76afc96ed15dcc8be0cb515f963fe3bb1d4e0b5c51f9d7e078
+COMPONENT_ARCHIVE_URL_3= https://rubygems.org/gems/ethon-0.11.0.gem
+CACHED_GEM_FILES += $(COMPONENT_ARCHIVE_3)
```
AndWac, line 53: Please use the introduced variables where possible.

components/cluster/pcs/Makefile
jclulow: Within the source we are building, there is a `pcs/snmp/Makefile`.  It is what drives the download of `pyagentx`. See: https://github.com/ClusterLabs/pcs/blob/pcs-0.9/pcs/snmp/Makefile#L1-L2

components/cluster/pcs/Makefile
jclulow: Good point!  I have made this change for all the gem files.

AndWac (approved):

components/cluster/pcs/Makefile
AndWac: Thank you for the explanation. It wasn't obvious for me.

jclulow: You're welcome!

PR #6100, opened 2020-09-25
sdl2: update to 2.0.12

AndWac:
- qt5 still builds (with one new lib)
- apps using qt5 still work


AndWac:
- wxwidgets-3 builds fine

PR #6144, opened 2020-10-10
Bug #12896: QT5 manifest needs mediated symlinks

jgmills:
This PR adds mediated links to the qt5 manifest, making it compatible with the qt4 package which already has them.  Both packages can be installed, but the mediated symlinks are taken from the second one installed.  Nothing else has changed.  Nothing has been deleted.


AndWac (commented):
Did you test this already?

pyhalov (approved):

PR #6148, opened 2020-10-12
freetype: bump to 2.10.3

alarcher:


AndWac:
I have the same commit in my local repository. I didn't create a PR from it because my tests with librecad showed some weird errors/warnings that made me wonder whether we need to rebuild all dependant packages. Alas I went out of time to check further by rebuilding librecad with the new freetype version.
Did you run any tests? librecad was the only one I checked yet.

AndWac:
Superceeded by #6187

PR #6153, opened 2020-10-13
python-35: bump to 3.5.10 final

alarcher:


AndWac (approved):
I run it on my
- build server (build some packages with it)
- desktop
Everything seems to be ok.

PR #6154, opened 2020-10-13
ips: add recipe to generate generic manifests for python

alarcher:
Just an idea to generate a generic manifest which is easier to compare to -PYVER manifests.


pyhalov:
Looks interesting.

AndWac (approved):

PR #6159, opened 2020-10-14
raptor2: patch for new qsort_r(3C)

AndWac:
This is necessary after qsort_r(3C) has landed in illumos-gate.
A similar fix is needed for libreoffice which uses its own raptor2 version.

### DIFF
--- a/components/library/raptor2/Makefile
+++ b/components/library/raptor2/Makefile
@@ -21,10 +21,12 @@
 # Copyright (c) 2013 David Hoeppner. All rights reserved.
 # Copyright (c) 2019 Nona Hansel
 #
+BUILD_BITS=			32_and_64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		raptor2
 COMPONENT_VERSION=	2.0.15
+COMPONENT_REVISION=	1
 COMPONENT_PROJECT_URL=	http://librdf.org/raptor/
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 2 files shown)


alarcher:
> This is necessary after qsort_r(3C) has landed in illumos-gate.
> A similar fix is needed for libreoffice which uses its own raptor2 version.

You can move the patch to the corresponding patch subdirectory. Alternatively you can import LibreOffice's patches to this component and use it for the build. When Alexander packaged LibreOffice we did not have time to package all the dependencies and also LibreOffice applies some patches of their own.

AndWac:
Am 19.10.20 um 10:10 schrieb Aurélien Larcher:
>
>     This is necessary after qsort_r(3C) has landed in illumos-gate.
>     A similar fix is needed for libreoffice which uses its own raptor2
>     version.
>
> You can move the patch to the corresponding patch subdirectory.
> Alternatively you can import LibreOffice's patches to this component
> and use it for the build. When Alexander packaged LibreOffice we did
> not have time to package all the dependencies and also LibreOffice
> applies some patches of their own.
>
Where is the corresponding patch subdirectory for raptor2?
I found out that I need to patch
"build/i86/workdir/UnpackedTarballs/raptor/src/sort_r.h" but I don't
know when. I tried it in COMPONENT_POST_CONFIGURE_ACTION but that seems
too early because the file is not yet there.

pyhalov:
You should tell libreoffice to patch library, not directly patch it. https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/desktop/libreoffice/patches/02-pdfium.patch contains similar logic, don't know if it's applicable to raptor.

PR #6163, opened 2020-10-17
Bug #13187: Libreoffice breaks man

jgmills:
This PR corrects the modules.sh script to remove the lines that set and export MANPATH to the shell environment.  I don't know what the intention was, but setting MANPATH in this way interferes with the man  command.  Curiously, it sets it without mediation by the module command.

I have tested removing those lines from modules.sh .  Doing this has no effect on libreoffice and no effect on other commands.  It does restore the man command back to normal operation.

### DIFF
--- a/components/sysutils/environment-modules/Makefile
+++ b/components/sysutils/environment-modules/Makefile
@@ -17,7 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		environment-modules
 COMPONENT_VERSION=	3.2.10
-COMPONENT_REVISION=     1
+COMPONENT_REVISION=     2
 COMPONENT_FMRI=		package/$(COMPONENT_NAME)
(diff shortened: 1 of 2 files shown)


AndWac:
I don't know why you have installed this package. Depending on it are
╰─➤  pkg search -r -H -o pkg.name "depend:require:package/environment-modules" 

library/openmpi/gcc/modulefiles
library/mpich/gcc/modulefiles

alarcher (approved):

alarcher:
> I don't know why you have installed this package. Depending on it are
> ╰─➤ pkg search -r -H -o pkg.name "depend:require:package/environment-modules"
> 
> library/openmpi/gcc/modulefiles
> library/mpich/gcc/modulefiles

Libroffice may have a dependency on boost (although it should not since it uses a local boost), boost-mpi on mpich, then mpich on environment-module.

PR #6164, opened 2020-10-17
Update qt5 to 5.12.9

despair86:
this is the previous LTS supported until 31 Dec 21 as qt recently issued 5.15 as the final release of qt5


despair86:
the only reason all apps need to be recompiled is due to the fact that qt apps in OI have version-specific DT_RUNPATH
for testing purposes one can use `crle(1)` to change the link-editor's default runpath to the new qt lib dir as qt5 is _intended to be backward compatible across minors_

components/library/qt5/Makefile
```diff
@@ -19,18 +19,19 @@ BUILD_BITS=		32_and_64
...
 
+COMPILER=gcc
 COMPONENT_NAME=		qt
-COMPONENT_VERSION=	5.8.0
-COMPONENT_VERSION_MJR=	5.8
-COMPONENT_REVISION=	12
+COMPONENT_VERSION=	5.12.9
+COMPONENT_VERSION_MJR=	5.12
+COMPONENT_REVISION=	13
 COMPONENT_FMRI=		library/qt5
```
AndWac, line 26: Don't increment COMPONENT_REVISON when you change COMPONENT_VERSION. Remove COMPONENT_REVISION instead.

despair86: I see. I need a little help here with IPS dependency resolution: 
```
/data/oi-userland/components/library/qt5/build/manifest-i386-qt5.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libEGL.so.1 \
        pkg.debug.depend.reason=usr/lib/qt/5.12/plugins/amd64/xcbglintegrations/libqxcb-egl-integration.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=usr/lib/qt/5.12/lib/amd64'.
```
i'm getting this even though we already specify `x11/library/mesa` to satisfy the dependency on GLX (and EGL)

despair86: nvm i accidentally stripped the bypass directives from the old manifest

AndWac: Before resolving this problem, you'll need to run "gmake sample-manifest" and check whether manifests/sample-manifest.p5m has changed. If so, you'll need to incorporate the changes back into qt5.p5m (or re-generate this file if too many things have changed).
After that you should run "gmake REQUIRED_PACKAGES". This will generate the dependencies and you'll have to check/edit the Makefile. In many cases this will solve dependency problems.

despair86: i ended up doing all of that manually because that.....wasn't really documented anywhere (!)

AndWac: Did you read http://docs.openindiana.org/dev/userland/ ?

despair86: ah i'd been lurking here since _before_ the transition to MkDocs. thanks! (this stuff probably wasn't in the old oi handbook on confluence)

AndWac:
This doesn't build on my build system:
...
g++ -m32 -Wl,-R,/usr/clang/9.0/lib -Wl,-z,origin -Wl,-R,\$ORIGIN/../lib -o ../../bin/qdoc .obj/atom.o .obj/clangcodeparser.o .obj/codechunk.o .obj/codemarker.o .obj/codeparser.o .obj/config.o .obj/cppcodemarker.o .obj/cppcodeparser.o .obj/doc.o .obj/editdistance.o .obj/generator.o .obj/helpprojectwriter.o .obj/htmlgenerator.o .obj/location.o .obj/main.o .obj/node.o .obj/openedlist.o .obj/puredocparser.o .obj/qdocdatabase.o .obj/qdoctagfiles.o .obj/qdocindexfiles.o .obj/quoter.o .obj/sections.o .obj/separator.o .obj/text.o .obj/tokenizer.o .obj/tree.o .obj/yyindent.o .obj/webxmlgenerator.o .obj/jscodemarker.o .obj/qmlcodemarker.o .obj/qmlcodeparser.o .obj/qmlmarkupvisitor.o .obj/qmlvisitor.o   -L/usr/clang/9.0/lib -lclang /oibuild/oi-userland/components/library/qt5/build/i86/qtdeclarative/lib/libQt5QmlDevTools.a /oibuild/oi-userland/components/library/qt5/build/i86/qtbase/lib/libQt5Core.so -lpthread -lrt
ld: fatal: file /usr/clang/9.0/lib/libclang.so: wrong ELF class: ELFCLASS64
ld: fatal: file processing errors. No output written to ../../bin/qdoc
make[4]: *** [Makefile:200: ../../bin/qdoc] Error 1

despair86:
I can disable QDoc for the 32-bit build but i have no idea what its flag is
```
CONFIGURE_OPTIONS.32 += [flag to disable qdoc]
```

despair86:
(I'd have gone ahead and tried to port 5.15 but we don't have an openssl 1.1 yet)

AndWac:
Alas I don't know how to disable 32-bit binaries, either.
Maybe we should keep qt-5.8 and add qt-5.12 additionally with only 64-bit. There are some packages that depend explicitly on qt-5.8, eg. desktop/fbreader has QT5_ROOT=/usr/lib/qt/5.8 in its Makefile.

despair86:
i patched the qmake project files instead, i'll push a better change once i finish regenerating the manifest

alarcher:
We should definitely get rid of the 32bit build. If it is easier for you then ditch it and we fix the affected components.

AndWac:
libreoffice would be affected. It doesn't build atm because of the new qsort_r(3C). The problem is in the local raptor2 and I don't know how to apply a patch best. My only idea is to patch the raptor2 archive and put it on the download folder. Do you have a better idea?

despair86:
i've yet to merge the sample manifest back into the original but this should now require LLVM 9 and handle libclang correctly

despair86:
ok this should build normally and removes the dangling reference to encumbered package

despair86:
was able to run my existing Qt5 apps after setting `LD_LIBRARY_PATH` so all that should be needed downstream is to recompile dependent packages

AndWac (approved):

PR #6185, opened 2020-10-21
gnu-make: update to 4.3

AndWac:
- The structure of the sources have changed
- The guile patch doesn't work anymore and doesn't seem to be necessary anymore
- All tests run fine (including guile tests)
- At the moment I am building whole userland with it (will take one more day)


AndWac:
At the moment we have a great amount of problematic packages, most related to python problems.
So far the following packages are problematic with the new gnu-make 4.3:
database/postgresql-11
database/postgresql-12
library/ptlib

PR #6190, opened 2020-10-24
Vim: update to 8.2.1897

balrog-nona:


AndWac:
Nice work. You haven't written about any tests you made. I have installed the new version on my desktop and used both, vim and gvim, a little bit without any problems.

AndWac (approved):

balrog-nona:
Thank you, next time I'll write about how I tested it. In this case, I did the same as you but couldn't test gvim since I don't have GUI.

PR #6191, opened 2020-10-24
Bug #13078: Update breaks sendmail configuration

jgmills:
This PR is somewhat controversial.  It only affects people who use sendmail as their MTA, usually only for outgoing e-mail.  Specifically, it only affects two configuration files: sendmail.cf and submit.cf .  For these two files, it changes the package manifest from `preserve=renameold' (the former parameter) to `preserve=renamenew' (the new parameter).  The effect only appears during a package update.  Formerly, it would change the name of the existing configuration file, replacing it with the new file with the original name.  With this change, it leaves the existing configuration file untouched, but installs the configuration file from the new package with a new name.  It will no longer replace a working configuration file with one that may not work.


AndWac:
I have checked our existing manifests for preserve actions. There doesn't seem to be a common path to either rename new or old configurations.
As you seem to be the only one who cares for the sendmail configuration it's ok to change the default.

AndWac (approved):
If nobody objects in the next two days I will merge this.

tsoome:
> I have checked our existing manifests for preserve actions. There doesn't seem to be a common path to either rename new or old configurations.
> As you seem to be the only one who cares for the sendmail configuration it's ok to change the default.

Not the only one:)

PR #6194, opened 2020-10-26
switch freerdp to OSS

despair86:
from irc:
```
<despair> oh holy hell
<despair> pulseaudio is just so freakishly broken on not-Linux
<despair> audio works fine with one pulseaudio client
<despair> then once you have more than one thing wanting to use audio
<despair> mic input slows to a crawl
<despair> had to rebuild remmina and freerdp to use OSS instead
<rennj> sucks on linux also
<despair> i'm going to put in a patch for that on userland because holy hell
<rennj> on older solaris 11.x on shutdown of desktop it would pause the process because the proc wouldnt die
<rennj> alsa+dmix was fine till i was forced to use pulse when tor-browser,firefox compile against it
<rennj> apulse is not a proper fix
<rennj> i have 2 iso's  i can show the diff on the same hardware, my old 4.x,amd fglrx, vmware 12,alsa vs 5.x amdgpu,vmware 15. pulse
<rennj> my gaming vm i get audio glitching now
<rennj> course lot of different software change in that comparison
<rennj> not all the time, only under heavy load
<rennj> project zomboid java/lua for one
```


despair86:
ideally aside from firefox which has a hard dependency on pulseaudio on not-Linux desktop apps should let the user choose between pulse, oss, or `/dev/audio` and otherwise default to pulse as before

despair86:
my local build has pulse turned off completely because i use freerdp via remmina which doesn't allow you to pick audio _output_ plugin (but will allow you to pick the audio _input_ plugin)

components/desktop/freerdp/Makefile
```diff
@@ -44,8 +44,10 @@ CMAKE_OPTIONS += -DWITH_ALSA=OFF
 CMAKE_OPTIONS += -DWITH_CAIRO=ON
```
AndWac, line 45: You'll need to set or increment COMPONENT_REVISION if you are making changes like this.
Simple rule: with every PR you'll need to either change the COMPONENT_VERSION or COMPONENT_REVISION. Otherwise nobody will gets a package updated!

components/desktop/freerdp/Makefile
```diff
@@ -44,8 +44,10 @@ CMAKE_OPTIONS += -DWITH_ALSA=OFF
...
 CMAKE_OPTIONS += -DWITH_PCSC=OFF
 CMAKE_OPTIONS += -DWITH_CUPS=ON
-CMAKE_OPTIONS += -DWITH_PULSE=ON
-CMAKE_OPTIONS += -DWITH_OSS=OFF
+CMAKE_OPTIONS += -DWITH_OSS=ON
+# mic input slows to a crawl if there is more than one
+# pulse client
+CMAKE_OPTIONS += -DWITH_PULSE=OFF
 CMAKE_OPTIONS += -DWITH_SSE2=ON
```
AndWac, line 52: What are the implications of switching to OSS? To my knowledge it was a decision to use only a single audio system and that was pulse.

I am willing to drop pulseaudio but we'll need something better then. Is OSS working all over? What other components would need a change and do they all work?

despair86: > What are the implications of switching to OSS?

`xfreerdp(1)` accesses the mixer `/dev/dsp` directly which results in lower latency even with multiple apps recording or playing back to the same device. pulseaudio appears to _serialise_ access to the sound card (!) which results in massive recording or playback latency even if one app is exclusively playing back or recording audio (and a second app doing exclusively the opposite)

> What other components would need a change and do they all work?

i'm not currently seeking a full-scale transition away from pulse at the moment, but it is something i'm mulling over and looking through the tree for affected apps. pretty sure by the end the only app left with a hard dependency on pulse is firefox (as a counterexample, vlc _already_ builds both pulse and oss audio backends for itself and users can change this in its configuration)

pyhalov:
When switching something from pulse, check that you can still control application volume with mate volume applet.

despair86:
going to bail out instead since at this point i've mostly set up my own window manager and use `audioctl(1M)` and `ossmix(1M)` directly

PR #6197, opened 2020-10-28
jq: update to 1.6

balrog-nona:
All tests passed (running gmake test).

Regarding patches: 
- both CVE patches have been incorporated into source code
- I kept the first two lines from Makefile.in.patch because it looks like some kind of description to me
- I adapted 01-spurious-macros.patch to new source code version: https://github.com/stedolan/jq/blob/master/src/builtin.c 
As in the previous version of this patch file, the builtin.c file starts with
`#include <sys/time.h>`
`#include <stdlib.h>`
`#include <stddef.h>`
but I'm not sure if it's okay.


AndWac:
Regarding the additional text in Makefile.in.patch: Yes that is an explanation of the purpose of the patch. I consider it good style to give an idia about a patches prupose. Oracle engineers are doing this, too. It's very valuable when something has changed and the patch fails. Without an explanation you are easily lost.

AndWac (approved):

PR #6198, opened 2020-10-29
Bug #13250: No SMF information in named man page

jgmills:
This PR only affects the named man page.  It adds SMF information to the man page by means of a patch.  The executables in the package are not changed, nor is the package manifest.

I tested the changed man page by creating it with gmake install and viewing the resulting man page with the man command.

### DIFF
--- a/components/network/bind/patches/07-man-named.patch
+++ b/components/network/bind/patches/07-man-named.patch
@@ -1,5 +1,5 @@
 --- bind-9.16.6/doc/man/named.8in-orig	Mon Aug 10 04:31:13 2020
-+++ bind-9.16.6/doc/man/named.8in	Wed Oct 28 15:50:54 2020
++++ bind-9.16.6/doc/man/named.8in	Thu Oct 29 08:02:00 2020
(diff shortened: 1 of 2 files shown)


components/network/bind/patches/07-man-named.patch
```diff
@@ -0,0 +1,121 @@
...
+ .UNINDENT
++.SH AUTOMATIC SERVICE MANAGEMENT (SMF)
++.sp
++The \fBDNS\fP service is managed by the service management facility, \fBsmf\fP(5), under the service identifier:
++.INDENT 0.0
++.INDENT 3.5
++.sp
++.nf
++.ft C
++svc:/network/ntp:default
```
pyhalov: svc:/network/dns/server:default

pyhalov:
Fix service name. Otherwise looks good.

jgmills:
Thanks for noticing the error.  It should be corrected now.  I've never done a git correction before, so I'm not certain that it's fixed.

pyhalov (approved):

PR #6210, opened 2020-11-04
Bug #11890: Undefined symbol in build of network/ntp

jgmills:
This PR changes only the Makefile.  The patches and the manifest is unchanged, meaning that the package is also unchanged.  It works by omitting the sntp component from the build.  sntp is an alternate NTP client.  The letter s in the name stands for simple, not for secure.  The sntp files were not included in the manifest anyway.  Nothing is lost by not building the sntp component.

I tested these changes by building, installing, and publishing the package.  All were successful.

### DIFF
--- a/components/network/ntp/Makefile
+++ b/components/network/ntp/Makefile
@@ -28,6 +28,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		ntp
 COMPONENT_VERSION=	4.2.8p13
+COMPONENT_REVISION=	1
 IPS_COMPONENT_VERSION=	$(subst p,.,$(COMPONENT_VERSION))
 COMPONENT_SUMMARY=	Network Time Protocol Daemon v4
 COMPONENT_DESCRIPTION=	Network Time Protocol v4, NTP Daemon and Utilities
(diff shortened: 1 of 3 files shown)


AndWac:
When I run "gmake sample-manifest" the newly created manifests/sample-manifests.p5m file differs from the former one (it lacks the file action for the sntp binary). So you'll need to add the sample-manifest.p5m file to your PR, too. This sample-manifest files have two puproses:
1. the form the basis for the package's manifest file(s). This is needed for new packages or packages that have major differences such that manually changing the manifest file is more work than generating from scratch. (See pkgmogrify and pkgfmt).
2. An indication if the package manifest needs adaption after a change.
So these sample-manifests need to be up-to-date. The only change we will ignore is the change of the copyright year.

AndWac:
We make use of Repology (https://repology.org/projects/n/?inrepo=openindiana&outdated=True). https://repology.org/project/ntp/versions indicates that our package version of ntp is slightly outdated.
If possible you should consider to update our package to the newest version, especially if it's a minor version update.

jgmills:
I've added two more changed files to this PR: manifests/sample-manifest.p5m pkg5 .  Both of them are consequences of my earlier change to Makefile.

I'll leave the update of the NTP product to somebody else.  It should be quite straightforward.  I've already tested this change to Makefile on 4.2.8p12 and ntp-4.2.8p13 .  That should be sufficient.

AndWac (approved):

PR #6211, opened 2020-11-06
grep: update to 3.5

balrog-nona:
Tested it running `gmake test`, all tests passed after adapting them to this new version.
When installed locally, the man page says the right version and it worked when I tested it with `ggrep root /etc/passwd`.


AndWac (approved):
Fun fact: on my machine 3 more tests pass :)

balrog-nona:
Thank you for that comment. It turned out, I wasn't aware of what that tests mean. Next time, I'll put more effort into tests that went from passed to skipped.

PR #6220, opened 2020-11-09
gawk: update to 5.1.0

balrog-nona:
I adapted tests to this new version, ran `gmake test`, and if I interpret it correctly, all tests passed.
When installed locally, I've tested it with `gawk -F: '$3 > 500 { print $1 }' /etc/passwd` which worked and the man page says the right version.

### DIFF
--- a/components/text/gawk/Makefile
+++ b/components/text/gawk/Makefile
@@ -21,15 +21,15 @@
 # Copyright (c) 2011, 2012, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2017, Aurelien Larcher
 # Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2020, Nona Hansel
 #
 
-PREFERRED_BITS=		64
+BUILD_BITS=		64
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=          gawk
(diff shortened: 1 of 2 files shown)


components/text/gawk/manifests/sample-manifest.p5m
```diff
@@ -23,8 +23,8 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 link path=usr/bin/awk target=gawk
-hardlink path=usr/bin/gawk target=gawk-$(COMPONENT_VERSION)
-file path=usr/bin/gawk-$(COMPONENT_VERSION)
+file path=usr/bin/gawk
+hardlink path=usr/bin/gawk-$(COMPONENT_VERSION) target=gawk
 file path=usr/etc/profile.d/gawk.csh
```
AndWac: Please remove the sample-manifest from your PR. The values have been simply swapped and I bet if you run gmake sample-manifest several times you will get both variants.

balrog-nona: Removed, thanks.

AndWac (approved):

PR #6222, opened 2020-11-09
Update gnutls-3 to 3.6.15

despair86:
this update fixes a gnutls bug in 3.6.11 and below https://gitlab.com/gnutls/gnutls/-/issues/841
and re-enables ECC if there are no objections to it


despair86:
since this is a C library it just drops in when replaced 😄

components/library/gnutls-3/manifests/sample-manifest.p5m
```diff
@@ -81,7 +81,7 @@ file path=usr/share/doc/gnutls/gnutls-logo.png
 file path=usr/share/doc/gnutls/gnutls-modauth.png
 file path=usr/share/doc/gnutls/gnutls-x509.png
 file path=usr/share/doc/gnutls/pkcs11-vision.png
-file path=usr/share/info/dir
+#file path=usr/share/info/dir
 file path=usr/share/info/gnutls-client-server-use-case.png
```
AndWac: Please don't edit the generated sample-manifest.p5m.
It's not only a template for creating the package manifest but also used to check if anything has changed for a new version or release. Only do the necessary changes to the real manifest.

AndWac:
Did you run "gmake test"? If so, you should copy the new results to the test folder if anythings has changed (on my machine it has changed a lot) and probably check why a new failure has happened. TBH this part is the hardest.
The new results can be found at build/test/i86/results-32.snapshot and should be copied over test/results-all.master

AndWac:
Furthermore, did you check whether ecc is now working? Do any dependant packages need to be updated in order to make it working? Do all dependant packages still work with the new version?

despair86:
GCC splits the `test_stack` loops into separate functions:
see `gl/tests/test-explicit_bzero.c` for context
the unit test expects `do_secret_stuff()` to be a single function
```c
/* generated by a friendly decompiler */
LABEL_10:
  v6 = 1000;
  do
  {
    /* no! bad! */
    do_secret_stuff_constprop_0();
    v3 += do_secret_stuff_constprop_1();
    --v6;
  }
  while ( v6 );
  v14 = v3;
  printf("test_stack: count = %d\n", v3);
  if ( v3 > 49 )
  {
    v13 = "count < 50";
    v12 = (const char *)164;
    goto LABEL_3;
  }
  return 0;
}
```
```
FAIL: test-explicit_bzero
=========================

/data2/oi-userland/components/library/gnutls-3/gnutls-3.6.15/gl/tests/test-explicit_bzero.c:164: assertion 'count < 50' failed
test_heap: address range is still mapped after free().
test_stack: count = 1000
FAIL test-explicit_bzero (exit status: 134)
```

despair86:
after adding `-fno-ipa-cp-clone` the test passes

despair86:
as for putting ECC back into the build i have no context for why it was removed as gnutls-3 was imported from oracle with that patch verbatim

so far none of the apps we ship is using it directly so it's not hurting anyone for leaving it in, and makes available ECC curves to TLS apps

despair86:
found the patch from upstream, using that instead. also found out that ECC got reinstated later on (sun CR 25373386)

AndWac (approved):

pyhalov (approved):

PR #6236, opened 2020-11-16
fix python metadata dependency problem

AndWac:
Many python packages fail to publish atm. This fixes the problems (taken from OmniOSce). There they have this transformation local for python 3.7.
I am not sure whether activating it globally does any harm.


AndWac:
I have found out that other packages like database/couchdb-21 have problems with this patch.

AndWac:
#6245 fixed the packages separately because the global fix created new problems in other packages.

PR #6237, opened 2020-11-17
Spawn fcgi

alm0ndiga:
Hi!, this is my second PR, please any comments/revisions are welcome !!!


components/web/spawn-fcgi/Makefile
```diff
@@ -22,31 +22,22 @@
...
-COMPONENT_VERSION=	1.6.3
+COMPONENT_VERSION=	1.6.4
 COMPONENT_PROJECT_URL=	http://www.lighttpd.net/
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
-COMPONENT_ARCHIVE_HASH=	\
-    sha256:7507a822995731629149040ee5686279b7582ae46f0296efd64b542ef052a90e
+COMPONENT_ARCHIVE_HASH=	sha256:ab327462cb99894a3699f874425a421d934f957cb24221f00bb888108d9dd09e
 COMPONENT_ARCHIVE_URL=	$(COMPONENT_PROJECT_URL)download/$(COMPONENT_ARCHIVE)
 COMPONENT_BUGDB=	utility/spawn-fcgi
```
AndWac, line 37: TEST_TARGET: $(NO_TESTS)
needs to be added before the include. If not set explicitly, the BUILD_BITS settings will create automatically a test: $(TEST_64) target. If you try to run "gmake test" it will fail.

alm0ndiga:
Thanks, it's done, i've added the TEST_TARGET right tnow.

AndWac:
Re-read my comment and find the difference ;-)

components/web/spawn-fcgi/Makefile
alm0ndiga: Sorry @AndWac i don't sure to understand what you mean. When i add this:

```
TEST_TARGET= $(NO_TESTS)

```
and then do 'gmake test' i see this:

`There are no tests available at this time.`

However, when i delete:

TEST_TARGET= $(NO_TESTS)

and do 'gmake test' without TEST_TARGET line, i can't see any fails. Could you be so kind to tell me what should i see ?

AndWac: Sorry, I was wrong and you are right.
I shouldn't have done several things in parallel.

AndWac (approved):

PR #6239, opened 2020-11-18
Problem: build recipes looking for clang++-(version) find no such name

jimklimov:
Solution: add the symlinks to installation tree and mediated to /usr/bin
    
Signed-off-by: Jim Klimov <jimklimov@gmail.com>

### DIFF
--- a/components/developer/clang-80/clang-80.p5m
+++ b/components/developer/clang-80/clang-80.p5m
@@ -49,6 +49,7 @@ link path=usr/bin/clang-offload-bundler target=../clang/8.0/bin/clang-offload-bu
 link path=usr/bin/clang-refactor target=../clang/8.0/bin/clang-refactor mediator=clang mediator-version=8.0
 link path=usr/bin/clang-rename target=../clang/8.0/bin/clang-rename mediator=clang mediator-version=8.0
 link path=usr/bin/clang++ target=../clang/8.0/bin/clang++ mediator=clang mediator-version=8.0
+link path=usr/bin/clang++-8 target=../clang/8.0/bin/clang++-8 mediator=clang mediator-version=8.0
 link path=usr/bin/diagtool target=../clang/8.0/bin/diagtool mediator=clang mediator-version=8.0
 link path=usr/bin/dsymutil target=../clang/8.0/bin/dsymutil mediator=clang mediator-version=8.0
 link path=usr/bin/git-clang-format target=../clang/8.0/bin/git-clang-format mediator=clang mediator-version=8.0
@@ -116,6 +117,7 @@ file path=usr/clang/8.0/bin/bugpoint
 file path=usr/clang/8.0/bin/c-index-test
 link path=usr/clang/8.0/bin/clang target=clang-8
 link path=usr/clang/8.0/bin/clang++ target=clang
+link path=usr/clang/8.0/bin/clang++-8 target=clang++
 file path=usr/clang/8.0/bin/clang-8
 file path=usr/clang/8.0/bin/clang-check
 link path=usr/clang/8.0/bin/clang-cl target=clang
--- a/components/developer/clang-80/Makefile
+++ b/components/developer/clang-80/Makefile
@@ -24,6 +24,7 @@ LLVM_NAME=		llvm
 CLANG_NAME=		cfe
 
 COMPONENT_VERSION=	8.0.1
+COMPONENT_REVISION=	1
 COMPONENT_MAJOR_VERSION= \
 	$(shell echo $(COMPONENT_VERSION) | $(GSED) -e 's/\([0-9]\+\.[0-9]\+\).*/\1/')
 COMPONENT_PROJECT_URL=	http://llvm.org/
--- a/components/developer/clang-90/Makefile
+++ b/components/developer/clang-90/Makefile
@@ -26,6 +26,7 @@ LLVM_NAME=		llvm
 CLANG_NAME=		clang
 
 COMPONENT_VERSION=	9.0.1
+COMPONENT_REVISION=	1
 COMPONENT_MAJOR_VERSION= \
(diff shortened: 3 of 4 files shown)


jimklimov:
By the way, not sure it makes sense to have symlinks like `/usr/bin/clang-8` mediated. If we have two clang versions installed, each in their tree, why not have the numbered-name explicit access to both?

AndWac:
Have you checked whether clang-8 works with other clang binaries (eg. clang-cpp) from clang 9 and vice versa?

jimklimov:
Not yet, so far the problem was building FOSS with one release of clang
present, but recipe can tune which version to use - and by that effectively
suffixes minus-version (-3.5, -5.0, -9 ...) to base executable name. And
can't find e.g. `clang++-9` in `PATH`.

Changes by this PR solve that use case.

Coexistence of several releases is something to investigate yet.

Jim

AndWac:
Ok, then you should investigate further and create an new PR if you were successful.

AndWac (approved):

jimklimov:
So, as far as investigation goes: installing `clang-80` package after having `clang-90` installed, as defined in the repo today, changes nothing in `/usr/bin`. So as far as common consumers are concerned, clang 8 does not exist in the system (unless I'd activate it with mediator; didn't check but presume that with current p5m files, then clang 9 would "not exist" for consumers who look at PATH).

````
root@jimoi:/root# pkg install --no-backup-be clang-80
           Packages to install:  1
       Create boot environment: No

Completed                                1/1     2310/2310  774.7/774.7  7.4M/s

Installing new actions                     2348/2348
Updating package cache                           6/6 

root@jimoi:/root# ls -la /usr/bin/clang*
lrwxrwxrwx 1 root root 22 Nov  2 15:56 /usr/bin/clang -> ../clang/9.0/bin/clang
lrwxrwxrwx 1 root root 24 Nov  2 15:56 /usr/bin/clang-9 -> ../clang/9.0/bin/clang-9
lrwxrwxrwx 1 root root 28 Nov  2 15:56 /usr/bin/clang-check -> ../clang/9.0/bin/clang-check
lrwxrwxrwx 1 root root 25 Nov  2 15:56 /usr/bin/clang-cl -> ../clang/9.0/bin/clang-cl
lrwxrwxrwx 1 root root 26 Nov  2 15:56 /usr/bin/clang-cpp -> ../clang/9.0/bin/clang-cpp
lrwxrwxrwx 1 root root 37 Nov  2 15:56 /usr/bin/clang-extdef-mapping -> ../clang/9.0/bin/clang-extdef-mapping
lrwxrwxrwx 1 root root 29 Nov  2 15:56 /usr/bin/clang-format -> ../clang/9.0/bin/clang-format
lrwxrwxrwx 1 root root 34 Nov  2 15:56 /usr/bin/clang-import-test -> ../clang/9.0/bin/clang-import-test
lrwxrwxrwx 1 root root 38 Nov  2 15:56 /usr/bin/clang-offload-bundler -> ../clang/9.0/bin/clang-offload-bundler
lrwxrwxrwx 1 root root 31 Nov  2 15:56 /usr/bin/clang-refactor -> ../clang/9.0/bin/clang-refactor
lrwxrwxrwx 1 root root 29 Nov  2 15:56 /usr/bin/clang-rename -> ../clang/9.0/bin/clang-rename
lrwxrwxrwx 1 root root 24 Nov  2 15:56 /usr/bin/clang++ -> ../clang/9.0/bin/clang++
lrwxrwxrwx 1 root root  7 Nov 18 19:59 /usr/bin/clang++-9 -> clang-9

root@jimoi:/root# ls -la /usr/bin/clang* | grep -v 9.0
lrwxrwxrwx 1 root root  7 Nov 18 19:59 /usr/bin/clang++-9 -> clang-9
````

I think packages should deliver all `/usr/bin/clang*-8` and `/usr/bin/clang*-9` as unmediated links (assuming we don't deliver potentially coexisting and separately mediatable sub-versions like 9.0.1 and 9.0.2, or families like 9.0 and 9.5), with default not-numbered names mediated by user preference. And such change probably belongs in a separate PR :)

Following your question, so far I have no idea what preprocessor would a `clang-N` binary call by default - a `clang-cpp` from PATH, from its install-dir (`/usr/clang/9.0/bin/clang-cpp`), smartly "something suffixed like myself from PATH", etc...

Another note is that I added locally the symlinks under `/usr/lib/ccache/` so it would wrap my clang's - but not sure if/how we can set up such symlinks optionally (if clang-8 and ccache are both installed, make those links, otherwise don't bother) other than adding more (optional?) packages depending on both and making a couple of symlinks.

File: components/developer/clang-80/clang-80.p5m [hunk 1]
```diff
@@ -49,6 +49,7 @@ link path=usr/bin/clang-offload-bundler target=../clang/8.0/bin/clang-offload-bu
 link path=usr/bin/clang-refactor target=../clang/8.0/bin/clang-refactor mediator=clang mediator-version=8.0
 link path=usr/bin/clang-rename target=../clang/8.0/bin/clang-rename mediator=clang mediator-version=8.0
 link path=usr/bin/clang++ target=../clang/8.0/bin/clang++ mediator=clang mediator-version=8.0
+link path=usr/bin/clang++-8 target=../clang/8.0/bin/clang++-8 mediator=clang mediator-version=8.0
```
pyhalov, line 52: Do we need mediator here?

jimklimov: I think, absolutely not (see the discussion in the PR) :) 

But as posted here, it follows the precedent set by existing (and IMHO flawed) p5m, where `/usr/bin/clang-8` and `...-9` is also mediated (IMHO needlessly).

Transactionally speaking, I think it makes sense to address (drop) the mediation of numbered-suffix names, but it belongs in a different PR to not stall this small change.

PR #6262, opened 2020-11-29
apr: update to 1.7.0

AndWac:
- apr and apr/documentation have been merged as solaris-userland did
- apache-24 still works with it (but I haven't tested thoroughly)
- the sendfile="0" patch has been removed. With it the tests fail with "This program won't work on this platform because there is no support for sendfile().", without it the tests are ok and apache still seems to work.


alarcher (approved):
If it works for you go ahead.

PR #6269, opened 2020-12-04
vte: enable pty packet mode

RocketMan:
vte uses packet mode to inform the pty master of state changes in the slave.

The current OI patch set comments out the packet mode code in vte, as illumos does not support the TIOCPKT ioctl that vte uses to enable packet mode.  illumos supports packet mode, but it uses STREAMS.

I've added [support for STREAMS packet mode](https://gitlab.gnome.org/GNOME/vte/-/commit/8586e4292e76943a100a8039f5ce9e33e5984bb3) in upstream vte.  This PR backports those changes for OI.

The patch is delivered as a single file, `04-pty-packet-mode.patch`.  It replaces two previous patches, `04-src_pty.cc.patch` and `05-src_vte.cc.patch`.


AndWac (changes requested):
Your PR is missing a Makefile with an incremented COMPONENT_REVISION. If we don't change either the COMPONENT_VERSION or the COMPONENT_REVISION an update won't be delivered to existing installations.

AndWac (approved):

AndWac:
Thank you for your valuable contribution!

RocketMan:
Thanks!

PR #6275, opened 2020-12-05
squeak-4: add inisqueak

cstes:
The new package is stlil called 4.19.2 because the upstream Squeak package is the same , still version 4.19.2.

I'm not sure whether OpenIndiana policy allows to push packages that do not change/upgrade the IPS_COMPONENT_VERSION.

In principle IPS allows to have multiple packages with the same version, but with different timestamps.

I realise this is not ideal, but I can't update the version as long as the subversion upstream version is still the same.  

Note that the Squeak subversion repository for the stable release squeak-4 is only rarely updated (every 6 months or so).

### DIFF
--- a/components/runtime/squeak4/Makefile
+++ b/components/runtime/squeak4/Makefile
@@ -32,6 +32,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		squeak-4
 COMPONENT_VERSION=	4.19.2
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	The Squeak V4 Smalltalk Virtual Machine
 COMPONENT_PROJECT_URL=	http://www.squeak.org
 COMPONENT_ARCHIVE_URL=  http://squeakvm.org
@@ -39,7 +40,7 @@ COMPONENT_FMRI=		runtime/squeak-4
 COMPONENT_CLASSIFICATION=	Development/Other Languages
 SVN_REPO=		http://squeakvm.org/svn/squeak/trunk/
 SVN_REV=		3790
-SVN_HASH=  sha256:60521172ecd055b182813893e18db9adff528ee5d42f4b1ddd0c5df83a33aaa2
+SVN_HASH=  sha256:73a79cc285fe80672442ee8a66caf1b3aa28695de3edcd5c550772d65dc30915
 
 # See http://wiki.squeak.org/squeak/933
 # See http://wiki.squeak.org/squeak/159
--- a/components/runtime/squeak4/inisqueak4
+++ b/components/runtime/squeak4/inisqueak4
@@ -1,0 +1,249 @@
+#!/bin/sh
+# $Id: inisqueak4,v 1.2 2020/12/01 09:44:53 stes Exp $
+# 
+# inisqueak -- setup a directory for use with Squeak
+# 
+#   Copyright (C) 1996-2004 by Ian Piumarta and other authors/contributors
+#                              listed elsewhere in this file.
+#   All rights reserved.
+#
+#   Copyright (C) 2020 by David Stes
+#   All rights reserved.
+#   
+#   This file is part of Unix Squeak.
+# 
+#   Permission is hereby granted, free of charge, to any person obtaining a copy
+#   of this software and associated documentation files (the "Software"), to deal
+#   in the Software without restriction, including without limitation the rights
+#   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+#   copies of the Software, and to permit persons to whom the Software is
+#   furnished to do so, subject to the following conditions:
+# 
+#   The above copyright notice and this permission notice shall be included in
+#   all copies or substantial portions of the Software.
+# 
+#   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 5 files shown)


AndWac:
If you want to change something without touching the COMPONENT_VERSION (or IPS_COMPONENT_VERSION) you'll need to change COMPONENT_REVISION (if not yet there you'll need to set COMPONENT_REVISION=1 otherwise you'll need to increment its value).
If later the COMPONENT_VERSION changes you'll need to remove COMPONENT_REVISION (effectively make it 0).

dcstes:
I've set COMPONENT_REVISION=1 now in the Makefile

The resulting package is indeed different now:

runtime/squeak-4 (userland)                       4.19.2-2020.0.1.1          ---

runtime/squeak-4 (userland)                       4.19.2-2020.0.1.0          i--

the last digit changed from '0' to '1'.

AndWac:
Exactly, that's the COMPONENT_REVISION

AndWac (approved):

AndWac:
Can you check the checksum, please?
When I run "gmake publish" (or gmake download) I will get this:
    validating hash... corruption detected
    expected: sha256:680cfb325a78a561d76d1f06ecdecdc427eae227e17164503b87b2158e40d215
    actual:   sha256:73a79cc285fe80672442ee8a66caf1b3aa28695de3edcd5c550772d65dc30915
    payload:  sha256:5e5e76f8ba1bd1469ecdb034a1714d31c81a93330e24cc9ae89786fa74ff2cb2

AndWac:
It will automatically be fixed by changing the Makefile, though.

dcstes:
I noticed that the checksum was automatically updated, and I added that new checksum to the commit, which I perhaps shouldn't have done.   

I was already wondering why it changed the checksum as this is not logical.

As I've explained, to my knowledge I'm rebuilding a new 4.19.2 package from exactly the same sources as before

Only the COMPONENT_REVISION should be updated and the sources should be exactly the same as for the first build (COMPONENT_REVISION 0).

I'm checking whether a new download reverts it to the old checksum now ...

dcstes:
I think the reason is that in my local archive (source archive) I created a new .tar.bz2 tar ball which was exactly again a new svn export but it changes every time .

Sorry about that I'm fixing it now.   I think the reason is that in the OI source archive the .tar.bz2 is present and it should remain the same for both COMPONENT_REVISION's of course.

Upstream in the Squeak repo nothing changed.

In fact the classic Squeak version numbering dictates 4.19.2 as the 'official' release number.

dcstes:
I've copy pasted the old checksum for SVN_HASH from build 0 back into the Makefile.

I'm not sure why it changes everytime,  perhaps this is something specific to the prep-svn.mk

AndWac:
This is still broken. I cannot merge it because it will most likely create problems on the build server.

dcstes:
I've now copy pasted the checksum you had emailed as 'actual' checksum.

Can you try again please ?

dcstes:
It may be a good idea on the long term to get rid of the 'subversion' Makefile for squeak-4 and use a regular tarball download as I'm doing for squeak-5.   
Remember that for squeak-5 an alternative is to use the 'git' Makefile and checkout from git on the build server, but that was very slow, so we recoded it to use the tarball download

AndWac:
No, still erroring:
    validating hash... corruption detected
    expected: sha256:73a79cc285fe80672442ee8a66caf1b3aa28695de3edcd5c550772d65dc30915
    actual:   sha256:16c54bb2a5a429d056b947c3225a211678321d8b5a1932002683115bd91e99a3
    payload:  sha256:c6b1f418b088ec19872c6de8ca37410a64af65d843e5c27a7bea55f1e61e7d00

dcstes:
Well possibly the COMPONENT_REVISION trick works for a package that uses tools/userland-fetch with a regular source tarball download, but not for the prep-svn.mk makefile.  

I propose that I close then this PR and wait a while until the Squeak developers update the VMmaker release number or update at least the SVN revision number (by doing some commits).

As soon as they do a commit a new source archive name will be used by the userland build system.

My guess is that the problem now comes from the fact that the source archive for the OpenIndiana build server is already populated now with the source archive from the build 0.

So feel free to close this PR.

dcstes:
or what I can also do is try to define COMPONENT_ARCHIVE_HASH.

just a minute, I'll try that.

dcstes:
In order to set COMPONENT_ARCHIVE_HASH= in my Makefile,
I'd have to figure out first what the hash is of the squeak-4-3790.tar.bz file on the OpenIndiana build server.
Remember that this file is not download  , it is created by a "svn export".

Is it possible to run  digest -a sha256 on the squeak-4-3790.tar.bz2 file on the OI build server and email me the result ?  Then I could try to set COMPONENT_ARCHIVE_HASH in my Makefile

I'm no sure whether it makes sense.

The alternative is to just close this PR and I can wait for an update to squeak repo and "avoid the problem" so to speak.

dcstes:
My guess is that closing this PR is the best thing to do.

I can wait until the subversion revision is increased and then the oi build server will automatically download a new source package.

dcstes:
Another idea may be to remove the line SVN_HASH altogether from the Makefile.
Or to delete the source archive squeak-4-3790.tar.bz2 from the source archive on the OI build server.
I have no idea.

Best idea is perhaps to close this PR.  If you decide to close this, no problem for me.

The work is still valid for a future commit.

AndWac:
I want to avoid a hackish solution as it might create more problems in the future.

PR #6278, opened 2020-12-06
lsof: update to 4.94.0

balrog-nona:
Upstream moved to github.
When installed locally, the man page says the right version and the command itself works.

### DIFF
--- a/components/sysutils/lsof/Makefile
+++ b/components/sysutils/lsof/Makefile
@@ -12,60 +12,38 @@
 # Copyright 2017 Gary Mills
 # Copyright 2016 Jim Klimov. All rights reserved.
 # Copyright 2019 Michal Nowak
+# Copyright 2020 Nona Hansel
 #
 
+BUILD_BITS=	64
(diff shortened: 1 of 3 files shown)


AndWac:
Did you try to run "gmake sample-manifest"? I always do that during an upgrade in order to find out if I have to change the manifest.
When I run it on your PR it gives errors, eg.
pkgsend generate: No such file or directory: '/oibuild/oi-userland/components/sysutils/lsof/build/prototype/i386'

balrog-nona:
> Did you try to run "gmake sample-manifest"? I always do that during an upgrade in order to find out if I have to change the manifest.
> When I run it on your PR it gives errors, eg.
> pkgsend generate: No such file or directory: '/oibuild/oi-userland/components/sysutils/lsof/build/prototype/i386'

I did and got the same error. I think `gmake sample-manifest` never worked in this component given the content of `manifest/sample-manifest`. When I run `gmake sample-manifest` for the non-updated version, I get the same error.

AndWac (approved):

PR #6284, opened 2020-12-08
cpuid: update to 1.7.6

balrog-nona:
Sample-manifest didn't change. 
I couldn't adjust Makefile to new template:
```
BUILD_BITS= 64
TEST_TARGET: $(NO_TESTS)
```
and get rid of 
```
build:          $(BUILD_64)

install:        $(INSTALL_64)

test:           $(NO_TESTS)
```
because for some reason the component wouldn't build.

When installed locally, I tested it with `cpuid --help` and `cpuid --version` which worked and said the right version.

### DIFF
--- a/components/sysutils/cpuid/Makefile
+++ b/components/sysutils/cpuid/Makefile
@@ -11,12 +11,15 @@
 #
 # Copyright 2018 Aurelien Larcher
 # Copyright 2018-2019 Michal Nowak
+# Copyright 2020 Nona Hansel
 #
 
+BUILD_BITS=	64
+BUILD_STYLE=	justmake
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		cpuid
-COMPONENT_VERSION=	1.7.2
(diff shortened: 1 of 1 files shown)


AndWac:
It should work with a little additional change:
BUILD_BITS= 64
BUILD_STYLE= justmake
...

Look carefully at the includes: the 2nd wasn't including configure.mk but justmake.mk!
If configure.mk has been included you don't need to set BUILD_STYLE because BUILD_STYLE=configure is the default. Here you'll need justmake.

balrog-nona:
> It should work with a little additional change:
> BUILD_BITS= 64
> BUILD_STYLE= justmake
> include ../../../make-rules/shared-macros.mk
> ...
> 
> Look carefully at the includes: the 2nd wasn't including configure.mk but justmake.mk!
> If configure.mk has been included you don't need to set BUILD_STYLE because BUILD_STYLE=configure is the default. Here you'll need justmake.

Thank you, I added suggested changes.

AndWac (approved):

PR #6286, opened 2020-12-09
vsftpd: add version 3.0.3

cgrzemba:
please add.


AndWac (commented):
- Did you test this in any way?
- Do we need to mediate this with proftpd?
- Why do we need it? It seems to be abandoned (last release is from 2015) while proftpd had its last release in July 2020.

PR #6299, opened 2020-12-14
squeak-4: add inisqueak

cstes:
New attempt for the closed squeak-4: add inisqueak #6275

The difference now is that the SVN_REV is different.

I suspect that the openindiana userland framework and more specifically the svn (subversion) makefile will detect that the revision has changed, and trigger a new checkout and a new SVN_HASH computation.

So the component version is still 4.19.2 and the component revision is 1 (changed from 0) and hopefully with the changed subversoin revision number it will now work.


AndWac:
I still see the problem with a changing SVN_HASH in the Makefile. That will probably trigger a republish of the package on every Jenkins run.

cstes:
I think this was always present; also in the past ...   the SVN_HASH is always being updated after the fetch.   Perhaps this is an issue between prep-svn.mk and the userland-fetch script.      In any case, I suspect it will work now :  the archive file is not present; because the name has changed : this will trigger a new download and then the SVN_HASH will be updated;   my guess is that on subsequent downloads or biulds it will detect that the hash is correct (because it was locally  computed) and then not trigger new publish.

cstes:
Actually you can check on other svn components in the oi-userland/components how it is there for the Makefile's of those components that use prep-svn.mk.   My guess is that for all those components (there are not so many) the SVN_HASH is updated and the Makefile has changed after the build.

PR #6306, opened 2020-12-17
zstd: update to 1.4.7

balrog-nona:
It builds, installs and publishes fine, although runnig `gmake sample-manifest`, it gave me this message
```
if [ "X" != "X" ]; \
then sed  /export/home/nona/oi-userland/components/archiver/zstd/manifests/sample-manifest.p5m \
	| gawk '!seen[$0]++' > /export/home/nona/oi-userland/components/archiver/zstd/manifests/generic-manifest.p5m; fi;
```
which I'm not familiar with. The subsequent `echo $?` gives me 0.

I tested it with `zstd file` and `zstd -d file.zst` and it worked and the man page says the right version.

### DIFF
--- a/components/archiver/zstd/Makefile
+++ b/components/archiver/zstd/Makefile
@@ -10,6 +10,7 @@
 
 #
 # Copyright 2019 Michal Nowak
+# Copyright 2020 Nona Hansel
 #
 
 BUILD_BITS=		64_and_32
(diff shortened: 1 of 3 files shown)


AndWac:
I also have prepared a PR for this. gmake test just finished successfully :)
The unknown message during gmake sample-manifest is due to #6154  that I tested successfully and merged a few days ago.
You can steer the generation of generic manifest with GENERATE_GENERIC_TRANSFORMS

AndWac (approved):

PR #6319, opened 2020-12-20
glib2: update to 2.66.4

AndWac:
Desktop still works


AndWac:
I am not sure whether we should keep 08-gio-trash-only-home.patch.
It's maybe not worth the efforts to keep it and without a context hard to justify.
I will probably remove it in a future version.

PR #6369, opened 2021-01-02
wget: update to 1.21

balrog-nona:
It builds, installs, and publishes fine. All tests passed.
I had to redo one patch, but the change is the same.
Tested it with `wget https://github.com/powerline/powerline/archive/2.8.1.tar.gz` which worked and the man page says the right version.

### DIFF
--- a/components/web/wget/Makefile
+++ b/components/web/wget/Makefile
@@ -20,25 +20,24 @@
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2018, Michal Nowak
+# Copyright (c) 2021, Nona Hansel
 #
+
+BUILD_BITS=64
 
 include ../../../make-rules/shared-macros.mk
 
-PREFERRED_BITS=64
-
 COMPONENT_NAME=		wget
-COMPONENT_VERSION=	1.20.3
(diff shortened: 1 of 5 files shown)


components/web/wget/Makefile
```diff
@@ -20,25 +20,25 @@
...
-COMPONENT_ARCHIVE_URL=	http://ftp.gnu.org/gnu/wget/$(COMPONENT_ARCHIVE)
+	sha256:87ae105e76e5b550e03e009ba94341143c66623a5ecbba047f6ef850608b6596
+COMPONENT_ARCHIVE_URL=	https://ftp.gnu.org/gnu/wget/$(COMPONENT_ARCHIVE)
 COMPONENT_BUGDB=	utility/wget
 
-include $(WS_MAKE_RULES)/prep.mk
-include $(WS_MAKE_RULES)/configure.mk
-include $(WS_MAKE_RULES)/ips.mk
+TEST_TARGET=	$(TEST_64)
+include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 40: TEST_TARGET is not necessary. It is automatically set to this value because of the BUILD_BITS settings.

balrog-nona: Removed it, thanks.

AndWac (approved):

PR #6370, opened 2021-01-02
Add i3

despair86:
finally figured out the patch required to get the status indicator app working (something something `statfs` vs `statvfs`....)
the window mgr itself has not required any patches since 4.17 (i was on 4.16 locally)

....except for the thing where keyboard scan codes on Xsun (and some BSD X servers) are different than what linux decides on

### DIFF
--- a/components/desktop/i3/Makefile
+++ b/components/desktop/i3/Makefile
@@ -28,6 +28,7 @@ COMPONENT_CLASSIFICATION=System/X11
 COMPONENT_ARCHIVE_URL=https://i3wm.org/downloads/$(COMPONENT_ARCHIVE)
 COMPONENT_ARCHIVE_HASH=sha256:aca48b03c0c70607826a1a91333065ff44d61774c152ddc9210fbc1627355872
 COMPONENT_LICENSE=BSD-3
+PATCH_LEVEL=0
 
 TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
 include $(WS_MAKE_RULES)/common.mk
@@ -37,6 +38,7 @@ LDFLAGS+= -lsocket -lnsl
 CONFIGURE_OPTIONS+= -Dmans=true
 
 REQUIRED_PACKAGES += desktop/dmenu
+REQUIRED_PACKAGES += desktop/window-manager/i3/i3status
 REQUIRED_PACKAGES += text/asciidoc
 REQUIRED_PACKAGES += text/xmlto
 # Auto-generated dependencies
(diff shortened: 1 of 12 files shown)


components/desktop/i3/Makefile
```diff
@@ -0,0 +1,59 @@
...
+COMPONENT_PROJECT_URL=https://i3wm.org
+COMPONENT_FMRI=desktop/window-manager/i3
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.xz
+COMPONENT_CLASSIFICATION=System/X11
+COMPONENT_ARCHIVE_URL=https://i3wm.org/downloads/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=sha256:aca48b03c0c70607826a1a91333065ff44d61774c152ddc9210fbc1627355872
+COMPONENT_LICENSE=BSD-3
+
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
```
AndWac, line 33: Your patch needs an additional setting here in the Makefile:
PATCH_LEVEL= 0

AndWac:
We didn't, your patch did. If you wan to avoid it, your patch should be generated from the project folder with eg.
diff -u i3-4.19/etc/config.orig i3-4.19/etc/config >patches/01-fix-config
This way the patch contains the fully qualified path.

despair86:
yeah that patch fixed it, I was confused for a bit earlier

components/desktop/i3status/Makefile
```diff
@@ -0,0 +1,46 @@
+#
```
AndWac, line 1: Please separate the i3 and i3status into separate PR's.
This is needed because i3 depends on i3status and thus i3status has to be merged and installed before i3 can be merged.

AndWac:
Do you need assistance with the separation?
It is really needed to add the dependency first, otherwise the build server will fail because he cannot fulfill the 
REQUIRED_PACKAGES += desktop/window-manager/i3/i3status
line because there is no such package available in the official repositories yet.
Thus, we need to create the i3status package in a separate run.
In order to do so, you'll need to create a new branch, eg. i3status, move the commit for i3status there. Then update your existing branch (thus removing the i3status commits from it), create another PR with the i3status and then we  can merge first the i3status PR and then the i3 PR.

despair86:
sorry about that, closing for now.

PR #6377, opened 2021-01-08
youtube-dl: update to 2021.01.08

balrog-nona:
When installed locally, I tested it with `youtube-dl -o file_name URL` and `youtube-dl -F URL` and it worked.

The option -x (extract audio) is a bit tricky because it requires ffprobe/avprobe and ffmpeg/avconv:
```
ERROR: ffprobe/avprobe and ffmpeg/avconv not found. Please install one.
```
 It gives the extracted audio in format m4a though. Is it okay or should I add these packages?


AndWac:
I'd say it's good enough as it is for now. You can add these new packages in a later PR and rebuilt youtube-dl if needed then.

AndWac (approved):

PR #6378, opened 2021-01-08
python-37: backport fixes

alarcher:


AndWac (approved):
Where did you find this patch? I haven't seen it neither on solaris-userland nor omnios-build repos.
We should check other patches from omnios-build, too,
eg. mod-posix-sched_priority.patch from there seems also appropriate here.

alarcher:
> Where did you find this patch? I haven't seen it neither on solaris-userland nor omnios-build repos.

This patch is a backport from a fix to 3.8.

[Edit: see the link from the patch file https://github.com/python/cpython/pull/23246]

> We should check other patches from omnios-build, too,
> eg. mod-posix-sched_priority.patch from there seems also appropriate here.

I already checked them, the patches are mostly from solaris-userland.
The rest of the patches are fixes to the test suite.

This particular patch was submitted from solaris-userland: https://github.com/python/cpython/pull/22374/files

We should probably use this version.

PR #6383, opened 2021-01-10
python-six: add 3.7 and 3.9

alarcher:
I modified the macros for selecting/ignoring lines in python manifests.
However this could be automated in the generation of generic-manifest.p5m in a near future.


alarcher:
In the meantime I found other required fixes so I merged them together in #6385

PR #6391, opened 2021-01-11
findutils: update to 4.8.0

balrog-nona:
It builds, installs and publishes fine, sample-manifest didn't change.
When installed locally, I tested it with:
`pfexec updatedb`
`locate file_name`
`find -type f -name "*string*" | xargs ls -l`
and it worked.

### DIFF
--- a/components/sysutils/findutils/Makefile
+++ b/components/sysutils/findutils/Makefile
@@ -22,6 +22,7 @@
 #
 # Copyright (c) 2012, 2017, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2021, Nona Hansel
 #
 
 BUILD_BITS=		64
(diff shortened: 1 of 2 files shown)


alarcher:
You may want to add REQUIRED_PACKAGES += locale/zh_cn-extra to ensure test reproducibility.

@AndWac What do you think?

AndWac:
Am 11.01.21 um 17:48 schrieb Aurélien Larcher:
>
> You may want to add REQUIRED_PACKAGES += locale/zh_cn-extra to ensure
> test reproducibility.
>
> @AndWac <https://github.com/AndWac> What do you think?
>
>
Yes, that would be good - together with a short comment that states that
it is needed for one of tests.

Andreas

balrog-nona:
> Am 11.01.21 um 17:48 schrieb Aurélien Larcher:
> You may want to add REQUIRED_PACKAGES += locale/zh_cn-extra to ensure test reproducibility. @AndWac <https://github.com/AndWac> What do you think?
> Yes, that would be good - together with a short comment that states that it is needed for one of tests. Andreas

I added it to test dependencies, but I didn't add any comment - there already was one to introduce several test dependencies. I hope it's okay.

AndWac (approved):

PR #6392, opened 2021-01-11
WIP: Build plan

alarcher:
@xen0l @AndWac @jclulow @Toasterson @olbohlen These were the early steps from last year for building PRs or doing a full rebuild in order of dependencies with intermediate stages. This is very naive and there are wildcards to break cycles...

Not sure if this still runs after the rebase.


Toasterson:
@alarcher have you seen tools/jenkinshelper.sh and Jenkinsfile changes? PR building is currently based on that and active in https://jenkins.openindiana.aurora-opencloud.org

Toasterson:
Also see #8508

alarcher:
> @alarcher have you seen tools/jenkinshelper.sh and Jenkinsfile changes? PR building is currently based on that and active in https://jenkins.openindiana.aurora-opencloud.org

Yes I have see these additions. We had tested userland-zone with Adam before that, good to see that you have implemented the procedure. This PR has a different goal: it is to rebuild components and update them in order of dependencies automatically. The point is too ease rebuilds of dependencies for components that have a large number of consumers and possibly intricate dependencies like ICU. I have tested on a full rebuild last year, it worked well but requires a bit more proofing.

PR #6401, opened 2021-01-13
Update smartmontools Makefile

hoewweken:
.. to add to PR 6400 smartmontools


hoewweken:
maybe we could merge both PR as they are byte-size ?
in any other case i will make a new clen PR later this evening..

PR #6404, opened 2021-01-13
gnupg: update to 2.2.27

balrog-nona:
It builds and installs fine. All tests pass but one - I couldn't find the variable `XTEST_GPGME_SRCDIR` in all oi-userland/ to fix it, so the test is skipped. But this was the case even for the original version.

There is a new file in `sample-manifest.p5m` which can't be included in `gnupg.p5m` because there is no such file on the system. I hope that's correct.

When installed locally, I tested it with `gpg2 --list-keys` and `gpg2 -r email -e sample.txt` and it worked.


components/sysutils/gnupg/gnupg.p5m
```diff
@@ -125,6 +126,7 @@ file path=usr/share/locale/da/LC_MESSAGES/gnupg2.mo
 file path=usr/share/locale/de/LC_MESSAGES/gnupg2.mo
 file path=usr/share/locale/el/LC_MESSAGES/gnupg2.mo
 file path=usr/share/locale/en/LC_MESSAGES/gnupg2.mo
+#file path=usr/share/locale/en/en@boldquot/LC_MESSAGES/gnupg2.mo
 file path=usr/share/locale/en@boldquot/LC_MESSAGES/gnupg2.mo
```
AndWac: Something must be wrong on your build machine. When I generate a new sample-manifest this line is missing.

AndWac: Maybe you started from a polluted build folder?
Try to run
gmake clean && gmake sample-manifest
and see whether the generated sample-manifest is correct.

components/sysutils/gnupg/test/results-64.master
```diff
@@ -0,0 +1,92 @@
...
+PASS: tests/migrations/extended-pkf.scm 
+PASS: tests/migrations/issue2276.scm 
+PASS: tests/gpgsm/setup.scm 
+PASS: tests/gpgsm/import.scm 
+PASS: tests/gpgsm/encrypt.scm 
+PASS: tests/gpgsm/verify.scm 
+PASS: tests/gpgsm/decrypt.scm 
+PASS: tests/gpgsm/sign.scm 
+PASS: tests/gpgsm/export.scm 
+SKIP: Environment variable 'XTEST_GPGME_SRCDIR' not set.  Please point it to a recent GPGME source tree to run the GPGME test suite.
```
AndWac, line 92: The message is clear: If you want to let this test being run, you'll need to set the XTEST_GPGME_SRCDIR variable to refer to the GPGME sources, eg. add the following line to the Makefile:
COMPONENT_TEST_ENV += $(COMPONENT_DIR)/../gpgme/gpgme-1.13.1
I am not sure whether this works, though. You maybe need to experiment with the settings.

components/sysutils/gnupg/gnupg.p5m
balrog-nona: Yes, that must've been the case. When I run it properly, the line disappears. I removed it from both files. Thanks.

components/sysutils/gnupg/test/results-64.master
balrog-nona: Thanks, I'll try it and play with it.

AndWac: Ups, should be
COMPONENT_TEST_ENV += XTEST_GPGME_SRCDIR=$(COMPONENT_DIR)/../gpgme/gpgme-1.13.1

AndWac: But IMO it's not that important to let this test case succeed. Skipping is ok for me.

AndWac: Pro-tip: look for similar things in our sources, eg.
cd .../components
find . -name Makefile | xargs grep COMPONENT_TEST_ENV
Looking for similar solutions is always a good idea ;-)

balrog-nona: > Ups, should be
> COMPONENT_TEST_ENV += XTEST_GPGME_SRCDIR=$(COMPONENT_DIR)/../gpgme/gpgme-1.13.1

Thank you for showing me how to assign two variables at the same time! Yet, it didn't work, now the message says: 
```
SKIP: Environment variable 'XTEST_GPGME_BUILDDIR' not set.  Please point it to a recent GPGME build tree to run the GPGME test suite.
```
It looks like it could go on for a long time.

balrog-nona: > But IMO it's not that important to let this test case succeed. Skipping is ok for me.

Perfect, then I think it's ready.

balrog-nona: > Pro-tip: look for similar things in our sources, eg.
> cd .../components
> find . -name Makefile | xargs grep COMPONENT_TEST_ENV
> Looking for similar solutions is always a good idea ;-)


Thank you, normally I would do that (this is how I got COMPONENT_TEST_TRANSFORMS :) but in this case I didn't know what was wrong. And I didn't notice that gpgme is another component...

AndWac: Ok, new trick: run
gmake print-COMPONENT_TEST_ENV
and see what it spits out :)

AndWac: Plus, you may need to run
gmake unpack
for the gpgme component in order make this work.

AndWac (approved):

PR #6420, opened 2021-01-14
userland-mangler: avoid broken packages

alarcher:


tools/userland-mangler
```diff
@@ -263,9 +263,14 @@ def mangle_elf(manifest, action, src, dest):
                                 destdir = os.path.dirname(dest)
                                 if not os.path.exists(destdir):
                                         os.makedirs(destdir)
-                                # Create a copy to mangle if it doesn't exist yet.
-                                if os.path.isfile(dest) == False:
-                                        shutil.copy2(src, dest)
+                                # Create a copy to mangle
+                                # Earlier the code would check that the destination file doe not exist
+                                # yet, however internal library versioning can be different while the
```
AndWac, line 268: doe -> does

tools/userland-mangler
```diff
@@ -263,9 +263,14 @@ def mangle_elf(manifest, action, src, dest):
...
+                                # filename remains the same.
+                                # When publishing from a non-clean prototype directory older libraries
+                                # which may be ABI incompatible would then be republished in the new
+                                # package instead of the new version.
+                                shutil.copy2(src, dest)
```
AndWac, line 274: The code looks reasonable. But I am neither familiar with the userland-mangler nor a vivid python developer.
Maybe we should try to create some test cases to check it.

Toasterson:
LGTM. I don't know about the process details but from what you told me the code looks as it should.

PR #6422, opened 2021-01-14
userland: add rule to print depend require actions

alarcher:


make-rules/ips.mk
```diff
@@ -151,6 +151,23 @@ ifneq ($(wildcard $(HISTORY)),)
...
 endif
 
+define ips-print-depend-require-rule
+$(shell cat $(1) $(WS_TOP)/transforms/print-depend-require |\
+	$(PKGMOGRIFY) $(PKG_OPTIONS) /dev/fd/0 |\
+	sed -e '/^$$/d' -e '/^#.*$$/d' | sort -u)
+endef
+
+define ips-print--depend-require-versioned-rule
+$(foreach v,$($(1)V_VALUES),\
```
AndWac, line 161: Is there a typo here: ...print--depend... (double hyphen)?
In the foreach loop it is written with a single hyphen.

alarcher: You are right, fixed.

PR #6434, opened 2021-01-17
WIP: meson: bump to 0.56.1

alarcher:
@AndWac Need to check if old patches still apply... At least I know one of them is incorrect and breaks expected behaviour. Neither solaris-userland nor omniosorg seem to patch any of the bugs we had observed and addressed.


components/developer/meson/patches/013-environment.patch
```diff
@@ -0,0 +1,30 @@
...
+--- a/mesonbuild/environment.py
++++ b/mesonbuild/environment.py
+@@ -813,7 +813,7 @@ class Environment:
+         """
+         # Arguments to output compiler pre-processor defines to stdout
+         # gcc, g++, and gfortran all support these arguments
+-        args = compiler + ['-E', '-dM', '-']
++        args = compiler + os.environ.get('CFLAGS', '').split() + ['-E', '-dM', '-']
+         p, output, error = Popen_safe(args, write='', stdin=subprocess.PIPE)
+         if p.returncode != 0:
```
AndWac, line 29: The new patch can't be applied with the remaining one.
It needs to be reworked. I recommend to use our typical diff -u old new style and remove the PATCH_LEVEL=0 from the Makefile.
Regarding the other patches: I have seen similar situtations for other packages. Alas we miss thorough test cases and our patches usually don't have comments. I really wish for comments even in patches in order to give context and some explanations. This would be helpful in situations like this.

AndWac: Sorry, I have to correct me: You simply need to remove PATCH_LEVEL=0 from the Makefile in order to make the patch work. I mixed this with my own unfinished work on meson 0.56

AndWac:
After removing PATCH_LEVEL=0 from Makefile, I have successfully created a new sample-manifest and with that a new meson.p5m.
Alas when I run gmake publish or gmake REQUIRED_PACKAGES it will end with an error like this:
Could not find the file for __main__ imported in /oibuild/oi-userland/components/developer/meson/build/prototype/i386/usr/lib/python3.9/vendor-packages/mesonbuild/rewriter.py
Could not find the file for ast imported in /oibuild/oi-userland/components/developer/meson/build/prototype/i386/usr/lib/python3.9/vendor-packages/mesonbuild/rewriter.py
gmake: *** [/oibuild/oi-userland/make-rules/ips.mk:461: /oibuild/oi-userland/components/developer/meson/build/manifest-i386-meson.depend] Error 1

AndWac:
superceeded by #6554

PR #6454, opened 2021-01-20
userland-component: refactoring rules and update

alarcher:
Later auto-update of Makefiles...

So far just some refactoring rules, just 'gmake update' in the component's directory.
Probably the code is very ugly and incomplete.

### DIFF
--- a/tools/bass/makefiles.py
+++ b/tools/bass/makefiles.py
@@ -22,10 +22,12 @@
 # This is all very naive and will hurt pythonists' eyes.
 #
 
+import json
 import os
 import re
 import subprocess
 import warnings
+from urllib.request import urlopen
 
 from .component import Component
 
@@ -281,6 +283,15 @@ class Makefile(object):
 
         return result
 
+    def print_value(self, name):
+        return self.run('print-value-'+name)[0]
+
+    def build_style(self):
+        return self.variables['BUILD_STYLE'].value()
+
+    def build_bits(self):
+        return self.variables['BUILD_BITS'].value()
+
     def has_variable(self, variable):
         return variable in self.variables
 
@@ -331,6 +342,9 @@ class Makefile(object):
                 contents.extend(self.contents[line:])
                 self.update(contents)
 
+    def set_archive_hash(self, checksum):
+        self.set_variable('COMPONENT_ARCHIVE_HASH', "sha256:"+str(checksum))
+
     def variable_assignment(self, variable):
         return self.variables[variable].variable_assignment(variable)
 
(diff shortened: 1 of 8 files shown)


tools/bass/makefiles.py
```diff
@@ -0,0 +1,239 @@
...
+        self.includes = []
+        self.variables = {}
+        self.targets = {}
+        makefile = os.path.join(path, 'Makefile')
+        with open(makefile, 'r') as f:
+            self.contents = f.readlines()
+        self.update()
+
+    def update(self):
+        # Construct list of keywords
```
AndWac, line 157: General comments on programming with high level languages:
- Procedures / methods should be short and have an appropriate name.
- Whenever you need a comment to explain some lines of code you should think about refactoring the code into a separate method with a sound name and thus don't need an explanation.
- Another general idea: don't mix high level and low level language constructs. Mixing levels of abstraction is a bad sign.

=> The update() method is too long and is doing too many low level things. This is hard to understand and hard to maintain.

tools/userland-component
```diff
@@ -0,0 +1,185 @@
...
+from bass.makefiles import Item
+from bass.makefiles import Keywords
+from bass.makefiles import Makefile as MK
+
+# Refactoring rules
+# 000:  Use WS_* variables instead $(WS_TOP)/* 
+#       If $(WS_TOP)/make-rules is found in an include then replace with the
+#       variable $(WS_RULES). Do the same for other variables.
+def refactor000(mk):
+    for i in iter(mk.includes):
```
AndWac, line 46: Speaking of sound names: A better name of refactor000() would be refactor_use_WS_variables()

tools/userland-component
```diff
@@ -0,0 +1,185 @@
...
+            mk.contents[i.idx] = i.include_line()
+
+
+# 001:  Use common.mk
+#       If common.mk is not included then:
+#           1. infer the build system and set the BUILD_STYLE.
+#           2. set the BUILD_BITS from the existing targets.
+#           3. erase default target and keep the custom ones.
+def refactor001(mk):
+    kw = Keywords()
```
AndWac, line 63: refactor001() -> refactor_use_common_mk()

AndWac: And refactor some parts of its body into separate helper methods. Maybe some of them can be re-used by other refactor_* methods later.

tools/userland-component
alarcher: The naming reflects the order in which refactoring functions are supposed to be called, otherwise we may miss assumptions.

tools/bass/makefiles.py
alarcher: Right, I am fluent in C/C++ mainly and do not have use of high level languages :)
The update() method is a naive parsing I do not know how to do this can be simpler :)

alarcher: I updated the scripts to hide some of the details and polished a bit more.

PR #6465, opened 2021-01-29
Add latex package

cstes:
There still is a pkgmogrify error in the g,make publish step but the manifest could be OK.

This is basically building LaTeX through the use of the texlive package.

The meta-packages/texlive must be installed ifrst on the build server.


cstes:
pkgmogrify error

PR #6489, opened 2021-02-07
gtar: update to 1.33

balrog-nona:
It builds, installs and publishes fine. When installed locally, I tested it with some archives and it worked.

Regarding patches:

- hunk 1 of patch `02-remote-shell.patch` had to be redone because the old one didn't work in the new version. I hope the change is the same.
- I deleted patch `03-sys-limits.patch` because I think that now it's taken care of in the source code. And I found this discussion about it:
https://lists.gnu.org/archive/html/bug-gnulib/2019-01/msg00016.html
- there is a new patch because the test `extrac09.at` fails in the new version. The patch has been taken from here:
https://github.com/omniosorg/omnios-build/blob/master/build/gtar/patches/revert_check_return_value_from_xgetcwd.patch


components/archiver/gtar/patches/revert_check_return_value_from_xgetcwd.patch
```diff
@@ -0,0 +1,59 @@
+The patch was taken from omnios-build:
```
AndWac: Please prepend this patch with 03-. This generates an implicit order (OmniOS has explicit patch orders by a steering file).

balrog-nona: I did it and pushed it as a new commit. Thank you.

AndWac (approved):

PR #6499, opened 2021-02-13
Squeak5

cstes:
This is an update that builds Squeak-5 Cog Spur and Stack Spur in 32bit and 64bit.

The previous versions were 64bit only, I changed it now to also deliver a 32 bit version.

This works but I still think there is a problem with my component:

When I build it with gmake install and then gmake publish it works for me.

It works because it is doing : biuld-32 install-32 build-64 install-64 publish

But a direct "gmake publish" seems for me to use the wrong build order.

It seems to be doing : biuld-32 build-64 install-32 install-64 which is the wrong order

Maybe this is an issue with the make-rules framework, or maybe I have to add a fix for it in my component.

I'm not sure.

As said the workaround for me that works is doing : gmake install and gmake publish in 2 different steps apparently ...

Feedback welcome on how to fix this -- if a fix is required at all because possibly the publish framework
works with 2 different stages install and publish as well.


AndWac:
I have checked squeak5 and can confirm the problems.
This will not work on our build server as Jenkins directly runs "gmake publish".

cstes:
The OpenSmalltalk configure script needs perhaps a fix or a patch to make it work "as is".

For the momentI close this PR and I open another one with a solution-- for now.  As said, it may be that the OpenSmalltalk configure script requires a fix rather than the OI Makefile.

PR #6503, opened 2021-02-13
salt dependencies: part 2

balrog-nona:
Distro builds, installs and publishes fine. When installed locally, I tested it with 
```
>>> import distro
>>> distro.linux_distribution(full_distribution_name=False)
```
which gave me
`('sunos', '5.11', '')`

Msgpack also builds, installs and publishes fine. When installed locally, I tested it with 
```
>>> import msgpack
>>> msgpack.packb([1, 2, 3], use_bin_type=True)
'\x93\x01\x02\x03'
>>> msgpack.unpackb(_, raw=False)
[1, 2, 3]
>>> msgpack.unpackb(b'\x93\x01\x02\x03', use_list=False, raw=False)
(1, 2, 3)
```
which worked.

Tornado also buils, installs and publishes fine.

This PR must be merged after #6522 and before #6523


components/python/tornado/Makefile
```diff
@@ -54,4 +49,6 @@ REQUIRED_PACKAGES += library/python/python-mysql-27
 # Auto-generated dependencies
```
AndWac, line 49: The build dependencies look strange. Are you sure that only Python-2.7 versions are needed?

balrog-nona: I let it like that because Python 3.5 was already included before and the only thing I did was including more of Python 3. I thought there was a good reason for having Python-2.7 versions explicitly and didn't question it.

AndWac (approved):

PR #6507, opened 2021-02-14
Switch to openssl-1.1

AndWac:


AndWac:
These are just the first two packages. curl's tests are mostly failing (only 13 out of 1176 are ok).
mariadb-101 does not support openssl-1.1:
-- Performing Test OLD_OPENSSL_API
-- Performing Test OLD_OPENSSL_API - Failed
CMake Error at cmake/ssl.cmake:222 (MESSAGE):
  Cannot find appropriate system libraries for SSL.  Use WITH_SSL=bundled to
  enable SSL support
Call Stack (most recent call first):
  CMakeLists.txt:360 (MYSQL_CHECK_SSL)


-- Configuring incomplete, errors occurred!

AndWac:
mariadb-10.1 is out of support since November 2020. So we should deprecate it and set mariadb-10.3 as the default.

AndWac:
We need to deprecate php-7.0. It ran out of support in January 2019!
It depends on mariadb-10.1. Sooner or later we need to add php-7.4 or 8.0.

alarcher:
> mariadb-10.1 is out of support since November 2020. So we should deprecate it and set mariadb-10.3 as the default.

Yes!

alarcher:
> We need to deprecate php-7.0. It ran out of support in January 2019!

Yes!

> It depends on mariadb-10.1. Sooner or later we need to add php-7.4 or 8.0.

BTW I am taking care of Postgresql.

AndWac:
I think it would be best to separate the deprecations from the openssl switch.

alarcher:
> I think it would be best to separate the deprecations from the openssl switch.

Yes sure :)

AndWac:
I will close this and will start to merge simple packages immediately (simple packages adhere to pkgconfig and work without mediator settings). I will collect problematic packages in a new branch.

PR #6522, opened 2021-02-21
salt dependencies: part 1

balrog-nona:

Futures and msgpack-python are both obsoleted: 

- futures are now part of Python3's standard library: https://pypi.org/project/futures/
- msgpack-python is now msgpack: https://pypi.org/project/msgpack-python/

This PR must be merged before #6503


components/meta-packages/history/history
```diff
@@ -902,11 +902,16 @@ library/perl-5/xml-parser-5100@2.41,5.11-2015.0.1.0
 library/perl-5/xml-parser-584@2.36,5.11-5.12.0.0.0.14.0
 library/perl-5/xml-simple-5100@2.20,5.11-2014.0.1.1
 library/perl-5/xml-simple-584@2.18,5.11-5.12.0.0.0.14.0
+library/python/futures@3.2.0,5.11-2020.0.1.1
+library/python/futures-27@3.2.0,5.11-2020.0.1.1
 library/python/gnome-python@2.28.1,5.11-2018.0.0.4
```
AndWac: You cannot remove futures-27 because salt-27 depends on it.

components/meta-packages/history/history
```diff
@@ -902,11 +902,16 @@ library/perl-5/xml-parser-5100@2.41,5.11-2015.0.1.0
...
 library/python/gnome-python-26@2.28.1,5.11-2016.0.1.3
 library/python/gnome-python-27@2.28.1,5.11-2018.0.0.4
 library/python/instant@1.3.0,5.11-2020.0.1.1
 library/python/instant-27@1.3.0,5.11-2020.0.1.1
+library/python/msgpack-python@0.4.7,5.11-2020.0.1.5
+library/python/msgpack-python-27@0.4.7,5.11-2020.0.1.5
+library/python/msgpack-python-35@0.4.7,5.11-2020.0.1.5
 library/python/python-gnome-desktop-26@2.32.0,5.11-2016.0.1.6
```
AndWac: Here also: salt-27 depends on msgpack-python-27.

balrog-nona: You're right, I should obsolete salt-27 as well.

AndWac: When you only remove a single package and keep or update other packages within a single folder, it is better to create a local history file to remove a package, see eg. library/apr/history.
Furthermore, I am not sure whether this really works as intended. You have only added an entry in this history file. You also need to remove related files, especially its manifest file. In cases where they are generated by a script you usually will remove the python version from PYTHON_VERSIONS.

balrog-nona: I see. In that case, I'll obsolete salt-27 in #6691 because that is the stage of switching between Python 2.7 and Python 3.7.

AndWac: No, that won't work because of the dependencies here.
I haven't had such a situation yet. The salt package manifest is uncommon because it only creates a single package with the generic manifest. This should be an ordinary manifest instead.
We could try to make the obsoletion with this, I am just inexperienced. If it doesn't work as expected we need to find a solution on the fly.

AndWac: It might also work because only the global history is touched and not the package itself.
We should try to find a way to make salt an ordinary package (not the salt and salt-<PYVER> combination).

balrog-nona: I'm sorry, I've already moved the commit... Should I moved it back and have the obsoletion here in the meta-packages/history file?

AndWac: I really don't know how to deal with this situation. We might need a trick to handle. Something like:
1. Remove salt-27 dependencies on the packages we want to obsolete.
2. Obsolete salt-27 with a local history file and make a single package salt.
3. Obsolete the other packages
4. Update the other dependencies
5. Update salt and  add the dependencies back

AndWac: The salt package of 2. would need to be an almost empty package dummy.

balrog-nona: And is the fact of having salt and salt-PYV packages really such a problem? Because as far as I know, powerline behaves exactly the same and it's also a whole program itself.

What about do the update and transition of salt now and obsolete futures and msgpack-python afterwards when there is no salt-27 depending on them. This approach would only require dropping two commits now and have separate PR later.

AndWac: I am not sure whether this will work. Did you do some experiments?

Toasterson: This is a usual approach. As both python3 and python 2 have different directories they can be installed simultaneously. If we first update salt without removing old libraries and then in a next PR obsolete the libraries with a obsoleting package new installs will grab the new package and salt will be completely obsoleted. For the buildserver however we will need two PR, so we can have this run through in two steps. Package update merge both into one update.

Toasterson: In summary:
First remove deprecations from #6522
Then merge in #6503 then #6523
Then have a look at salt itself
Then deprecate packages not needed.

balrog-nona: @AndWac If you approve the above approach, I'll remove the deprecating commits from this PR and create separate PR for them.

AndWac: Yes, let's try that. Toasterson is our IPS wizzard ;-)

balrog-nona: Did it, thank you.

AndWac (approved):

PR #6526, opened 2021-02-23
remove libcouchbase

AndWac:
- libcouchbase-2.6.2 is not compatible with openssl-1.1 (newer versions exist)
- php-7_0-ext-couchbase is its only consument and will be removed
- merge after #6518


alarcher:
@AndWac Go for it.

AndWac:
> @AndWac <https://github.com/AndWac> Go for it.

Together with php-7_0 or separately?

alarcher:
> Together with php-7_0 or separately?

Together :)

PR #6527, opened 2021-02-23
switch to openssl-1.1 Part 1

AndWac:
More commits will be added


AndWac:
These packages need the openssl mediator version set to 1.1 before integration

PR #6535, opened 2021-02-27
Add golang-116

Toasterson:


components/developer/golang-116/Makefile
```diff
@@ -0,0 +1,128 @@
...
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright 2021 Till Wegmueller
+#
+
```
AndWac, line 25: Please modernize the Makefile according to doc/current-tasks.md. E.g. put
BUILD_BITS= 64
BUILD_STYLE= justmake
before the include
and then later remove the common tasks:
build:...
install:...
test:..

components/developer/golang-116/golang.p5m
```diff
@@ -22,4 +22,4 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
```
AndWac, line 24: I noticed that the manifest contains lots of static libraries. Are these really needed? We usually only deliver them if needed.

AndWac:
Running "gmake test" revealved some failing tests:
embed/internal/embedtest/embed_test.go:15:12: pattern c*.txt: cannot embed irregular file concurrency.txt
FAIL	embed/internal/embedtest [setup failed]

--- FAIL: TestReadOnlyWriteFile (0.00s)
    ioutil_test.go:90: Expected an error when writing to read-only file /tmp/TestReadOnlyWriteFile704425621/blurp.txt
FAIL
FAIL	io/ioutil	0.055s

--- FAIL: TestReadOnlyWriteFile (0.00s)
    read_test.go:87: Expected an error when writing to read-only file /tmp/TestReadOnlyWriteFile168340550/blurp.txt
--- FAIL: TestRemoveAllButReadOnlyAndPathError (0.00s)
    removeall_test.go:352: RemoveAll succeeded unexpectedly
FAIL
FAIL	os	0.476s

go test proxy running at GOPROXY=http://127.0.0.1:48010/mod
--- FAIL: TestScript (0.01s)
    --- FAIL: TestScript/mod_cache_rw (0.37s)
        script_test.go:213:
            # Regression test for golang.org/issue/31481. (0.000s)
            # golang.org/issue/31481: an explicit flag should make directories in the module
            # cache writable in order to work around the historical inability of 'rm -rf' to
            # forcibly remove files in unwritable directories. (0.352s)
            # After adding an extraneous file, 'go mod verify' should fail. (0.017s)
            # However, files within those directories should still be read-only to avoid
            # accidental mutations. (0.000s)
            > [!root] ! cp $WORK/extraneous.txt $GOPATH/pkg/mod/rsc.io/quote@v1.5.2/go.mod
            FAIL: testdata/script/mod_cache_rw.txt:16: unexpected command success

    --- FAIL: TestScript/build_nocache (2.35s)
        script_test.go:213:
            # As of Go 1.12, the module cache is required.
            # If none of the variables we use to locate GOCACHE are set, the cache is off
            # and we cannot build. (0.299s)
            # If GOCACHE is set but is not an absolute path, and we cannot build. (0.459s)
            # An explicit GOCACHE=off also disables builds. (0.227s)
            # If GOCACHE is set to an unwritable directory, we should diagnose it as such. (1.345s)
            > [windows] stop # Does not support unwritable directories.
            > [root] skip # Can write to unwritable directories.
            > mkdir $WORK/unwritable/home
            > chmod 0555 $WORK/unwritable/home
            > [!plan9] env HOME=$WORK/unwritable/home
            > [plan9] env home=$WORK/unwritable/home
            > env GOCACHE=$WORK/unwritable/home
            > ! go build -o triv triv.go
            FAIL: testdata/script/build_nocache.txt:35: unexpected command success

FAIL
FAIL	cmd/go	65.388s

Are these failures expected?

components/developer/golang-116/Makefile
Toasterson: I looked into current-tasks.md and could not find things that apply here.

Applying your suggestions, makes the build fail. I think this is a too heavily modified Makefile that we can modernise it.

components/developer/golang-116/golang.p5m
Toasterson: Yes this is golang specific. If we do not ship them the compile fails for anybody except root.

Toasterson:
I only get the test fail from embed and that is because of the symlink cloning. The golang builder is passing all tests in illumos https://build.golang.org/?repo=&branch=release-branch.go1.16# I usually use that as reference, as we have an official CI there.

components/developer/golang-116/Makefile
balrog-nona: > I looked into current-tasks.md and could not find things that apply here.
> 
> Applying your suggestions, makes the build fail. I think this is a too heavily modified Makefile that we can modernise it.

I think it could work - it worked for me even with older components' Makefiles. Here it could be more visible: 
https://github.com/OpenIndiana/oi-userland/pull/6543/files
In your case the BUILD_STYLE is justmake. And also, you can remove the lines 
```
```
and have `include $(WS_MAKE_RULES)/common.mk` instead.

AndWac: Nona is right. It also works on my machine.

Toasterson: Well if I knew to include `$(WS_MAKE_RULES)/common.mk` yes then it works :)

New makefile format applied.

AndWac (approved):

PR #6543, opened 2021-03-03
powerline: update to 2.8.2, add Python3.9

balrog-nona:
- The component changed its name and the newer versions are not available on pypi.org

- It builds, installs and publishes fine. The deleted patch made it to the upstream.
- New version is incompatible with the old one - before installing this one the old one has to be uninstalled. Therefore I added powerline packages into history file. There were three packages, so I added a line for every one of them. I hope I did it correctly.

- When installed locally, I tested it in bash, vim and tmux. 

For bash testing I added these lines into .profile:
```
powerline-daemon -q
POWERLINE_BASH_CONTINUATION=1
POWERLINE_BASH_SELECT=1
if [[ -z "$SSH_CONNECTION" || "$USER" != 'root' ]] && [[ "$TERM" != 'linux' ]]; then
        source /usr/share/powerline/bash/powerline.sh
fi
```

For vim testing I added these lines into .vimrc:
```
set laststatus=2
python3 import sys; sys.path.append("/usr/lib/python3.9/vendor-packages/")
python3 from powerline.vim import setup as powerline_setup
python3 powerline_setup()
python3 del powerline_setup
```
It looks like the second line wasn't necessary for some reason in the previous version. Therefore I thought I could write an email to oi-discuss to spare people the headache of vim settings.


components/meta-packages/history/history
```diff
@@ -999,6 +999,9 @@ network/dhcp/dhcpmgr/locale/zh_hk_tw@0.5.11-2013.0.0.0
 network/unison@2.27.57,5.11-0.170
 print/lp/filter/a2ps@4.14,5.11-2015.0.2.0 print/filter/a2ps@4.14,5.11-0.151.1.8.1
 print/mp@0.5.11,5.11-2018.0.0.0
+python/powerline@2.7-2020.0.1.1
+python/powerline-35@2.7-2020.0.1.1
+python/powerline-vim@2.7-2020.0.1.1
 redistributable@0.1,5.11-2015.0.2.0
```
AndWac, line 1005: The versions you listed exist already. You'll need to increment the COMPONENT_REVISION (and add the build version: ,5.11):
python/powerline@2.7,5.11-2020.0.1.2
python/powerline-35@2.7,5.11-2020.0.1.2
python/powerline-vim@2.7,5.11-2020.0.1.2

balrog-nona: Did it, thanks.

AndWac (approved):

PR #6544, opened 2021-03-03
Upgrade of pcs to python-37

jgmills:
As recommended, this PR is a direct update of the pcs package.

The software product consists of a python module and a private collection of ruby modules. The new version, 0.10.1, does support python-3. I used the same IPS package name, application/cluster/pcs, with a new version string. Although I started it in the old source directory, most of the files have now changed. The new package should replace the old one.

In particular, I've made many changes to Makefile. As well, most of the patches had to be changed to suit the new context. Many changes were also made to the manifest, pcs.p5m .

Ultimately, the build, install, and publish steps were all successful. I could not test the resulting package, as I do not have an OI cluster. However, it is complete, as far as I could tell.


AndWac (approved):
I have successfully build this package on my system.
I will wait some more days to give other people the chance to test this before I will merge it.

PR #6569, opened 2021-03-11
want nvidia-460

hoewweken:
Add Nvidia Driver Version 460.

This version doesn't ship 32 bit binaries.
I don't know if there are consumers for 32-bit libraries
nvidia_drv.so
libglx.so.1
libnvidia-cfg.so.1
libnvidia-gtk2.so.460.56
..and the 32 bit programs
nvidia-debugdump, nvidia-settings, ...

i would have version 410,440 ready as well, these do ship 32 bit binaries.
but i hesitate to make a PR to these versions if there's no real need to.

Test status is "works for me" on a GT 1030 card, lightdm, nvidia-settings etc...

### DIFF
--- a/components/meta-packages/xorg-video/xorg-video.p5m
+++ b/components/meta-packages/xorg-video/xorg-video.p5m
@@ -20,7 +20,7 @@ set name=info.classification value=org.opensolaris.category.2008:System/X11
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 depend fmri=pkg:/consolidation/X/X-incorporation type=require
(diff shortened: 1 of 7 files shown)


components/openindiana/nvidia-460/pkg5.fmris
```diff
@@ -0,0 +1 @@
+driver/graphics/nvidia-460
```
AndWac: You don't need this. It is only needed if multiple packages (fmris) will be delivered.

hoewweken: Removed file pkg5.fmris from PR.

AndWac:
Ok, I have checked it and it seems to work. Please squash the two commits into one before we merge it.

AndWac (approved):

PR #6631, opened 2021-03-29
Upgrade of rdiff-backup to 2.0.5

jgmills:
This PR upgrades the rdiff-backup product from 1.33 to 2.0.5 .  The old version supported python-2.7 .  The new one supports python versions ranging from 3.5 to 3.8 .  Of these, only 3.7 makes sense for OI.  Note that it does not support python-3.9 .  The next version of the product, not yet released, will drop 3.5 and add 3.9 .

The Makefile is changed to accomodate the new software product and the new python version.  It also builds only 64-bit software and copies python-independant files post-install to standard locations in the prototype directory.  One patch is omitted and another is added.  The deprecation.patch is omitted because it's no longer relevant.  The no-scm.patch is added because it eliminates the confusion that ensues when OI also uses git.

The manifests use a new mediator name: python3 .  Note that the package will only create symlinks in /usr/bin when python3=3.7 now.  Other changes to the manifests are to accomodate installed files.  The file history causes the old package, backup/rdiff-backup-27@1.3.3, to become obsolete.

I was able to build, install, and publish the new rdiff-backup packages successfully.  I did not test the backup function of the software.


components/sysutils/rdiff-backup/history
```diff
@@ -1 +1 @@
-backup/rdiff-backup-26@1.3.3,5.11-2016.0.1.2
+backup/rdiff-backup-27@1.3.3-2020.0.1.2
```
AndWac: Two things here:
1. You'll need to keep the old entry and just have to add a new one for 2.7
2. You'll need to add a full FMRI with incremented COMPONENT_REVISION. Thus, the new entry should be
backup/rdiff-backup-27@1.3.3,5.11-2020.0.1.3

jgmills:
Okay, I've updated the history file as you suggested, and did another gmake publish.  This time it created two obsolete packages.  As I'm sure you suspected, that was the first time I had encountered a history file.

components/sysutils/rdiff-backup/Makefile
```diff
@@ -19,36 +19,59 @@
...
+      usr/share/doc/rdiff-backup-$(COMPONENT_VERSION); \
+  done; \
+  $(CP) -p usr/lib/python$$V/vendor-packages/share/bash-completion/completions/rdiff-backup \
+    usr/share/bash-completion/completions; \
+  $(CP) -p usr/lib/python$$V/vendor-packages/share/man/man1/rdiff-backup-statistics.1 \
+    usr/share/man/man1; \
+  $(CP) -p usr/lib/python$$V/vendor-packages/share/man/man1/rdiff-backup.1 \
+    usr/share/man/man1; \
+done \
+)
```
AndWac, line 61: This works only for a single python version as I see it.
If we add 3.9 later the $(CP) commands will overwrite files from the former loop. Maybe we need a mediator?

jgmills:
No, it will work for multiple python versions.  The files are the same for any python version.  Overwriting does not matter.  They are independant of the python version, so a mediator would do nothing.  Those files only depend on the rdiff-backup version, which is constant for this package.  There's no need for a mediator.  The contents of the package never changes.

AndWac:
Ok, but then the outer loop is superfluous, isn't it?

jgmills:
No, one iteration of the loop is essential.  I can add a break statement at the end to ensure that there it only one iteration.

The for loop could be replaced by an asignment statement, but I could not find a POSIX and reliable way to assign only one element from the list, except for the for statement.

PR #6647, opened 2021-04-05
Upgrade of postgresql10 to 10.16 and python 3.7

jgmills:
This PR upgrades the postgresql10 version from 10.15 to 10.16 and upgrades the python version that it uses from 2.7 to 3.7 .  Changes to Makefile are to accomodate the version upgrades, and also to omit the 32-bit build.  As far as I can tell, the omission has no consequences.

The pg_config.patch is also omitted, as it seems to be no longer relevant.  All of the files produced by the 32-bit build are now omitted from the manifests.  Also in the manifests, some of the files have been renamed, as a result of the python version change.

The build, install, and publish steps were all successful.  The test step ran normally, indicating that one of 178 tests had failed.  I assume that this is the normal result.


pyhalov:
Hi,  Gary.
1) Don't drop 32-bit libraries. We need them. Now PGSQL_VERSION is 9.5, but is should be EOLed. So, 10 will become the default version and should deliver 32 bit libs.
2) Dropping plpython2 language is fine, but it's a breaking change, so  it should be announced in advance.

jgmills:
Is there a consumer for the postgresql 32-bit binaries?  In a quick look, I didn't see any.

Yes, 9.5 is unsupported.  9.6 soon will be unsupported.  I intended to recommend that 9.5 be dropped, and that 13 be added.

Dropping plpython2 and replacing it with plpython3 is a consequence of the python upgrade, and the loss of python-2.  I expect the upgrade to python-3 to be announced.

AndWac:
Alas we still seem to have several packages that support 32-Bit and depend on PostgreSQL:
1. libpqxx (albeit this itself has no consumer and could be dropped)
2. pgagent
3. pgadmin
4. pgbadger
5. pgtcl
6. asterisk
7. psycopg2
8. gammu
9. qt4
10. qt5
11. tntdb
12. apr-util
13. DBI-PostgreSQL

There will also be some more transitive dependencies, eg. for qt5.
So it looks like we should postpone killing 32-Bit support.

pyhalov:
I think we can kill 32-bit PostgreSQL libraries only with completely dropping support for 32-bit apps.

jgmills:
I reviewed that list of packages.  All of them use 32-bit builds or combined 32 and 64-bit builds.  All of them likely use the 32-bit postgresql libraries.  There is no doubt that we need to keep those libraries in postgresql for the time being.  You can expect an updated PR for postgresql10 in the next four or five days.

PR #6650, opened 2021-04-06
openldap: update to 2.4.58

AndWac:
gmake test run fine

### DIFF
--- a/components/network/openldap/Makefile
+++ b/components/network/openldap/Makefile
@@ -27,12 +27,12 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		openldap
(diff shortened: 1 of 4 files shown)


tsoome:
last time I did check openldap, there was an issue; openldap is obsoleting slapd.conf and SMF startup script was still using it...

AndWac:
What do you mean by "openldap is obsoleting slapd.conf"?

tsoome:
"Mind that slapd.conf is not used by the new openldap packages. The configuration is stored under the directory /etc/openldap/slapd.d" 

I was having fun for quite some time trying to find out why I can not authenticate with root ... till I found this note and removed slapd.conf and slapd was able to use slapd.d :D

PR #6660, opened 2021-04-09
rust: update 1.44.1

cgrzemba:
Please update Rust language. This update is a prerequisite for firefox newer than 68.0


AndWac (changes requested):
Just a minor nitpick: you omitted the changed sample-manifest.p5m file from the PR. Can you add it?

AndWac (changes requested):
Sorry, I missed that pkg5 needs to be added to the PR, too.
Also, when done please squash the (then three) commits into one with only one comment "rust: update to 1.44.1". The other commit logs are of no interest.

components/developer/rust/pkg5
```diff
@@ -16,4 +16,4 @@
         "developer/lang/rustc"
     ],
     "name": "rustc"
-}
\ No newline at end of file
+}
```
AndWac: You added developer/clang-90 to the required packages. That should also have an impact on pkg5. You can either run gmake "update-metadata" or "gmake publish" to get an updated pkg5 file.

cgrzemba: gmake update-metadata ...
That's was easy.

PR #6661, opened 2021-04-10
Upgrade of postgresql10 to 10.16 and python 3.7

jgmills:
This PR upgrades the postgresql10 version from 10.15 to 10.16 and upgrades the python version that it uses from 2.7 to 3.7 .  Changes to Makefile are to accomodate the version upgrades.  Both 32-bit and 64-bit binaries are contained in the IPS packages.  32-bit libraries are needed for some consumers.

The pg_config.patch is also omitted, as it seems to be no longer relevant.  In the manifests, some of the files have been renamed, as a result of the python version change.

The build, install, and publish steps were all successful.  The test step ran normally, indicating that one of 178 tests had failed.  I assume that this is the normal result.


components/database/postgresql-10/Makefile
```diff
@@ -78,7 +79,7 @@ CONFIGURE_OPTIONS.64 += --libdir=/usr/postgres/$(COMPONENT_MJR_VERSION)/lib/$(MA
 
 CONFIGURE_OPTIONS+=$(CONFIGURE_OPTIONS.$(BITS))
 
-CONFIGURE_ENV	+= PYTHON=$(PYTHON)
+CONFIGURE_ENV	+= PYTHON=/usr/bin/python3.7
```
AndWac: You can keep the original
CONFIGURE_ENV += PYTHON=$(PYTHON) when you explicitly set the python version to use by setting
PYTHON_VERSION=		3.7
some lines above.

jgmills:
Yes, that is an improvement.  I've updated the PR with that change to Makefile .  Everything else was unchanged, including the publish step.

AndWac (approved):

PR #6670, opened 2021-04-15
ccache: add ccache-update-symlinks service and script

jimklimov:
Finally got tired of maintaining the symlinks I need manually, so proposing an SMF service to update those according to current compatible compilers in PATH.

I hope I've got right the triggering of this service when actual compilers are installed or updated, but did not test yet what happens if they are installed but this service is absent (e.g. ccache package not installed or updated). Comments welcome :)
UPDATE: Triggering moved to PR #6671 so not a concern in this one.

Note: while there is a newer ccache version released since our last bump, updating to it is not in scope of this PR.

### DIFF
--- a/components/developer/ccache/files/ccache-update-symlinks.sh
+++ b/components/developer/ccache/files/ccache-update-symlinks.sh
@@ -1,0 +1,125 @@
+#!/bin/sh
+
+# ccache-update-symlinks - keep ccache symlinks in sync with installed
+# versions of compatible compilers
+# Copyright (C) 2021 by Jim Klimov
+
+# This script can be used standalone, or as an SMF method script
+[ -s /lib/svc/share/smf_include.sh ] && . /lib/svc/share/smf_include.sh
+
+SMF_FMRI_BASE=""
+if [ -n "$SMF_FMRI" ]; then
+    SMF_FMRI_BASE="`echo "$SMF_FMRI" | sed 's|:[^:]*$||'`"
+fi
+
+getproparg() {
+    [ -n "$SMF_FMRI" ] || return
+    val="`svcprop -p "$1" "$SMF_FMRI" 2>/dev/null`" || val="`svcprop -p "$1" "$SMF_FMRI_BASE" 2>/dev/null`"
+    [ -n "$val" ] && echo "$val"
+}
+
+# Caller may `export DEBUG=yes` to trace the decisions in the script
+# Caller may `export DRYRUN=echo` to avoid actual link manipulations
+
+# This is where symlinks named identically to compiler binaries are present.
+# These symlinks point to the ccache binary however, and it knows how to
+# wrap the real thing if called (gcc, clang, maybe icc... and compatibles).
+# Users who want build stuff with a speedup add this to front of their PATH,
+# and their PATH must later contain the location of the real tool's binary.
+# While it is not generally a problem for common /usr/bin it may be for e.g.
+# gcc-4.4.4-il or other not-exposed compilers.
+# Also note that even if the caller has this location in their PATH, they
+# can turn off effects of ccache without changing runtime PATHs just by
+# `export CCACHE_DISABLE=1` before a build.
+LINKDIR="/usr/lib/ccache"
+SYMLINK="../../bin/ccache"
+
+# These are the filenames we manage symlinks for, either exact or suffixed
+# by a dash and number, and optionally add values from SMF instance:
+TOOLS="
+    gcc g++ gcpp
+    clang clang++ clang-cpp
+    c++ cc cpp
+    i386-pc-solaris2.11-c++ i386-pc-solaris2.11-cpp i386-pc-solaris2.11-g++ i386-pc-solaris2.11-gcc
+    `getproparg ccache-update-symlinks/TOOLS_ADD`
+"
+# Rearrange for easier parsing below
+TOOLS="`echo $TOOLS | tr ' ' '\n' | sort -n | uniq`"
+
+PATH_ADD="`getproparg ccache-update-symlinks/PATH_ADD`"
+if [ -n "$PATH_ADD" ]; then
+    PATH="$PATH:$PATH_ADD"
+fi
+
+cd "$LINKDIR" || exit
+
+# TODO: Option in the SMF service to enable additions to PATH for gcc-4.4.4-il
+# (/opt/gcc/4.4.4/bin/...) or other special compilers? It would help everyone
+# if that particular system's admin rather symlinked the custom compiler to
+# common /usr/bin/ instead.
+# Set value from SMF instance if present there
+[ -n "${ALLOW_DELETE-}" ] || ALLOW_DELETE="`getproparg ccache-update-symlinks/ALLOW_DELETE`"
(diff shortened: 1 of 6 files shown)


jimklimov:
On a separate note, current `gmake sample-manifest` updated to a 2020 Copyright... should be 2021 for some time now ;)

jimklimov:
Took a few iterations to wrinkle out the bugs :) But works for me now.

A few points to consider:
* Should we also expose `illumos-gcc` by moving it out into `/usr/bin` limelight (via symlinks)? I do that locally, but maybe it should be done in the package?
* Should there be a special PATH pre-set in the script or service? Currently it is SMF default `/usr/sbin:/usr/bin` which differs from even a common user profile like `/usr/gnu/bin:/usr/bin:/usr/sbin:/sbin` or what non-root people have in their setups.
* Should there be a kill-switch for removing "cruft" symlinks, which users can enable consciously to keep their systems in sync, but otherwise this script would only keep registering new compilers?

Regarding the latter point: The service script, as made now, ensures that the symlinks under /usr/lib/ccache/ refer to names of compilers and tools that it can find in PATH and removes the cruft (e.g. my earlier incorrect manual attempts to link a `gcc10` instead of `gcc-10`, shot and forgotten). In particular, this helps build-configuration tools to avoid detecting compilers by just a name of an executable seen in PATH when there is nothing behind it (package removed, etc.).

As a corollary, if someone had a web of symlinks set up to cover their local tool names that are not in the inspected list of tools, or not in the default PATH when the service runs (which might be adjusted in the SMF instance or script directly), those links would be logged and lost as "cruft". 

I can see this can be a hiccup for people using/testing non-packaged compilers (locally built, SFE, illumos-gcc as it is currently), e.g. with a custom PATH during their build so at that moment, unlike when the service script runs, the filename does make sense. This probably impacts cases where the non-default `CC` and `CXX` values are passed into the build anyway, so maybe the aggressive cleanup has clear benefits only when someone removes plain `gcc` or `g++` from PATH... I guess I'll add that switch :)

jimklimov:
Question to SMF experts: would this chicken-and-egg FMRI be a problem? e.g. on first install? (want to trigger rescans whenever ccache package is updated):

````
 file files/ccache-update-symlinks.xml path=lib/svc/manifest/application/ccache-update-symlinks.xml restart_fmri=svc:/system/manifest-import:default
-file files/ccache-update-symlinks.sh path=lib/svc/method/ccache-update-symlinks.sh mode=0555
+file files/ccache-update-symlinks.sh path=lib/svc/method/ccache-update-symlinks.sh mode=0555 restart_fmri=svc:/application/ccache-update-symlinks:default
````

**Comment by jimklimov** [comment]
So, now the failsafe is there (no removals by default) and optional properties for the service are handled and documented.

While iterating the package in my system, had to run `svcadm refresh svc:/system/ccache-update-symlinks` time and again... should not be a problem for usual end-users though. Hope (but did not try yet) that the proposed changes for packaged compilers would help trigger this service to make symlinks appear when the package is installed/updated, and hope even more that this would not botch compiler installations on systems where this new ccache service is absent.

**Comment by Toasterson** [comment]
Not a bad Hack to be Honest. I was thinking about a way to do this with the Packaging system alone, so can have a system state rather than defining these things at runtime. 

I would rather see the compilers deliver these links themselves. We can facet the links so we can remove them. And we won't need a dependency to ccache as those are just support links.

I think a package can deliver a facet. So we could have the links faceted and the ccache directory deliver the facet. But I don't know if it is possible. Worst case we have some dead links.

**Comment by jimklimov** [comment]
Thanks, I thought about that too but somehow it did not feel right to bog down compiler packages with symlinks for the benefit of a third-party optionally present helper tool :) And the original symlinks hardcoded into ccache.p5m were not particularly scalable either, as we have a dozen compilers packaged by now.

Also, the approach here allows users to more easily manage addition of custom compilers (say, you build gcc-12 to see how it works for OI, and wrap it into ccache by refreshing the service). Or similarly for adding other package sources like `localhostoih` (though for that likely `PATH_ADD` in the service instance would need to adapt).

On the other hand, symlinks in packages would be more deterministic for what the OS delivers (WRT the deletions that are a vague point here, but at least are not the default now :) )

**Comment by jimklimov** [comment]
Since I am not sure currently about the implications of changes to the compiler packages, to trigger this new service (e.g. what happens to them if ccache is not installed?), I've separated those commits into #6671 so those concerns do not block evaluation and merge of this PR :) At the very least, developers can refresh the service or reboot the system/container to same effect, and get new symlinks.

**Comment by AndWac** [comment]
I will postpone this after the coming release. After the release has landed I am open to merge this PR.

**Comment by AndWac** [comment]
Can you please squash the commits and remove irrelevant commit messages, please? I will merge the PR when this happened.

File: components/developer/ccache/files/ccache-update-symlinks.sh [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+        P="`echo "$P" | sed 's|^"\(.*\)"$|\1|'`"
+        [ "$P" = "$LINKDIR" ] && continue
+        for F in `ls -1d "$P/$T"* 2>/dev/null` ; do
+            [ -f "$F" ] && [ -x "$F" ] || continue
+            # F represents an existing filename in PATH component P
+            # that starts with tool name T
+            B="`basename "$F"`"
+            case "$B" in
+                ccache*) ;; # Quiet skip
+                "$T"|"$T"-{0,1,2,3,4,5,6,7,8,9}|"$T"-[0123456789]|"$T"-{0,1,2,3,4,5,6,7,8,9}*|"$T"-[0123456789]*)
```
pyhalov: Why  do we have both `"$T"-{0,1,2,3,4,5,6,7,8,9}` and `"$T"-[0123456789]`?

jimklimov:
Can't really say, might be different experience with different shells,
parts of the script were tried on various systems.

pyhalov:
Perhaps, simplify it, so that it works in our /bin/sh and next person looking at it doesn't wonder on these lines?

components/developer/ccache/files/ccache-update-symlinks.sh
jimklimov: Seems that the bracket expansions are more standard than curly brace lists, which are mostly a bash/ksh thing, so I'll drop that variant.

jimklimov: Double-checked, a test with:
````
:; ls -1 /usr/bin | while read L ; do case "$L" in gcc-[0123456789]*) echo "=== $L";; esac; done

=== gcc-10
=== gcc-4.4.4
=== gcc-4.9.4
=== gcc-6
=== gcc-7
````
...did match all numbered gcc versions including single-digit (7), double-digit(10) and dotted (4.9.4) with OI bash, dash, ksh and zsh.

jimklimov:
Gentle bump: squashed and simplified as specified :)

pyhalov:
Thanks. Overall looks fine. Should ALLOW_DELETE be true?

jimklimov:
As detailed in XML manifest comment for the option, it is better to err on the safe side and not remove whatever custom setup users might have before this service. Those inclined would need some path option setup/verification and activate the deleter.

PR #6673, opened 2021-04-16
Upgrade of postgresql12 to 12.6 and python 3.7

jgmills:
This PR upgrades the postgresql12 version from 12.5 to 12.6 and upgrades the python version that it uses from 2.7 to 3.7 .  Changes to Makefile are to accomodate the version upgrades.  Both 32-bit and 64-bit binaries are contained in the IPS packages.  32-bit libraries are needed for some consumers.

The pg_config.patch is also omitted, as it seems to be no longer relevant.  In the manifests, some of the files have been renamed, as a result of the python version change.

The build, install, and publish steps were all successful.  The test step ran normally, indicating that one of 194 tests had failed.  I assume that this is the normal result.


AndWac (approved):

AndWac:
You'll also need to update and add this PR the test/results-64.master file by copying the build/test/amd64/results-64.snapshot file into test/results-64.master

jgmills:
Thanks for letting me know about the one missing updated file.  I've added it and squashed the two commits.  The PR should be complete now.

PR #6676, opened 2021-04-18
squeak-4: add Cuis VectorEnginePlugin as a patch

cstes:
This adds the Cuis plugin developed by Juan Vuletich (author of Cuis Smalltalk)

The plugin is compiled both in 32bit and 64bit .

There is a transform rule which I hope is fixing the pkglint error of 64bit object in 32bit path,
but for safety I set again

PKGLINT=/bin/true

it is niot clear to me whether the transform fixes the pkglint error

The new Cuis plugin speeds up SVG scalable vector graphics in Cuis and the 64bit version is installed in /usr/lib/amd64


AndWac:
I think it's a mistake (created by our dependency tool) to have packages that will be delivered by a Makefile contents are themselves occuring as REQUIRED_PACKAGES.
Here it is
REQUIRED_PACKAGES += runtime/squeak-4-display-X11
REQUIRED_PACKAGES += runtime/squeak-4-nodisplay
(which are from a former PR). Nevertheless I think you should comment them out.

AndWac:
Please run "gmake update-metadata" and add the updated pkg5 file, too.

AndWac:
Oh sorry, I just realized that the pkg5 file contents is correct. You didn't include it yet so the contents is still without these two dependencies.

PR #6677, opened 2021-04-21
Cups: fixup man page conflicts with illumos lp

Toasterson:


AndWac:
There are more conflicts than just the man pages. The binaries conflict also, eg. usr/bin/lp

jgmills:
CUPS client commands should be installed in /usr/lib/cups/bin .  The perl script /usr/sbin/print-service describes how that location is used.  The previous manifest would have the same information.

Toasterson:
Which is why the new manifest still has the binaries under usr/lib/cups ;) except lpinfo which was in sbin IIRC. I can also put that under usr/lib/cups

```
file usr/bin/cancel path=usr/lib/cups/bin/cancel mode=0555
file usr/bin/lp path=usr/lib/cups/bin/lp mode=0555
file usr/bin/lpr path=usr/lib/cups/bin/lpr mode=0555
file usr/bin/lpstat path=usr/lib/cups/bin/lpstat mode=0555
file usr/bin/lpq path=usr/lib/cups/bin/lpq mode=0555
file usr/bin/lprm path=usr/lib/cups/bin/lprm mode=0555
file usr/sbin/cupsaccept path=usr/lib/cups/bin/cupsaccept mode=0555
file usr/sbin/lpc path=usr/lib/cups/bin/lpc mode=0555
file usr/sbin/lpmove path=usr/lib/cups/bin/lpmove mode=0555
file usr/sbin/lpadmin path=usr/lib/cups/bin/lpadmin mode=0555
file usr/sbin/lpinfo path=usr/sbin/lpinfo
```

AndWac:
You both are right. I shouldn't do things like this after a hard working day.

You also need to set COMPONENT_REVISION in the Makefile.

Toasterson:
> You both are right. I shouldn't do things like this after a hard working day.
> 
> You also need to set COMPONENT_REVISION in the Makefile.

Done and squashed.

AndWac (approved):

PR #6678, opened 2021-04-22
Upgrade of postgresql95 to 9.5.25 and python 3.7

jgmills:
This PR upgrades the postgresql95 version from 9.5.24 to 9.5.25 and upgrades the python version that it uses from 2.7 to 3.7 .  This postgres version is still unsupported.  The python option is now only available for the 64-bit version of postgres because the new python version is only available in 64-bits.  Changes to Makefile are to accomodate the version upgrades.  Both 32-bit and 64-bit binaries are contained in the IPS packages.  32-bit libraries are needed for some consumers.

The pg_config.patch is also omitted, as it seems to be no longer relevant.  In the manifests, some of the files have been renamed or deleted, as a result of the python version change.

The build, install, and publish steps were all successful.  The test step ran normally, indicating that one of 157 tests had failed.  I assume that this is the normal result.


components/database/postgresql-95/Makefile
```diff
@@ -11,22 +11,27 @@
...
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.bz2
-COMPONENT_ARCHIVE_HASH=	sha256:065cfd3db9f5aca84e794e73e71a797c984b2e728e760f4f4226a9162a99c22a
+COMPONENT_ARCHIVE_HASH=	sha256:7628c55eb23768a2c799c018988d8f2ab48ee3d80f5e11259938f7a935f0d603
 COMPONENT_ARCHIVE_URL=	http://ftp.postgresql.org/pub/source/v$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL=  https://www.postgresql.org
 
+# Set python version used by this product
+PYTHON_VERSION=3.7
+
 TEST_TARGET=$(TEST_32_and_64)
```
AndWac: You don't need to set TEST_TARGET explicitly here. The default derived from BUILD_BITS should be sufficient here.

jgmills:
Okay, I deleted that line.  As you suggested, everything worked as before, including gmake test.

AndWac (approved):

PR #6683, opened 2021-04-23
illumos-gate: disable SMB printing

AndWac:
Since CUPS 2.3b4 the use of private structures of IPP is prohibited.
Our SMB printing is broken since then.


Toasterson (approved):
LGTM. It's the same Change @citrus-it suggested in IRC

PR #6688, opened 2021-04-24
http-parser: assert in test is wrong on 32-bit sparc

tsoome:
Due to algnment (padding), we have sizeof 32 even as we have 32-bit pointers.


AndWac:
Could please add some commenting lines on top of the patch. This would make a future maintainers life easier if something changes.

AndWac:
I meant to put the comment on top of the patch, but this way is also ok.

AndWac (approved):

PR #6693, opened 2021-04-24
fcron: add 3.3.0 package

cstes:
Fcron is the "Frequency Cron" by Thibault Godouet .  Category job scheduler software.
Use case for this package : scripts that should be ran with a given frequency. 
The initial commit required fcron user/group to be created manually, but the fcron manifest is doing that automatically now.
The author (Thibault Godouet) seems prepared and ready to accept any (reasonable) patches if required.

### DIFF
--- a/components/sysutils/fcron/Makefile
+++ b/components/sysutils/fcron/Makefile
@@ -1,0 +1,69 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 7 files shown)


cstes:
Hi,  although I realise that fcron is not "perfect" I still think it would be a nice addition to OpenIndiana.   Also this is of interest to both fcron and will be appreciated by Thibault Godouet (and by myself who packaged fcron) but also it is of interest to OpenIndiana to support portable small reusable UNIX services and tools.   Note that FreeBSD has fcron in its ports collection https://svnweb.freebsd.org/ports/head/sysutils/fcron/ ...   Of course this utility is useful in the first place for scheduling backups (for example with a script that is ran with a given frequency instead of at a given day).

PR #6699, opened 2021-04-26
Changing the default postgres version to 11

jgmills:
This PR is the first step in breaking the logjam of postgres versions.  By a modification of shared-macros.mk, it changes the default version from 9.5 to 11.  As well, it changes the location of the binary directories to match reality.  In this step, the version 9.5 package  still exists, leaving no need to rebuild packages that depend on it.

If the package must be rebuilt for other reasons, developers can use the new default if it works.  Otherwise, they can use the original default version by adding:

    PG_VERSION = 9.5

to the component Makefile.  This is a temporary solution.

The next step is to obsolete 9.5, perhaps simplifying shared-macros.mk at the same time.  As well, the version-independant and pan-version postgres files can be moved to their own component directory.  Either then or later, 9.6 can be obsoleted and 13 packaged.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -18,7 +18,7 @@
 #
 # CDDL HEADER END
 #
-# Copyright 2017 Gary Mills
+# Copyright 2017,2021 Gary Mills
 # Copyright (c) 2010, 2016, Oracle and/or its affiliates. All rights reserved.
 #
 
@@ -814,8 +814,8 @@ PKG_MACROS +=   PERL_ARCH=$(PERL_ARCH)
 PKG_MACROS +=   PERL_VERSION=$(PERL_VERSION)
 
 # Config magic for Postgres/EnterpriseDB/...
-# Default DB version is the oldest one, for hopefully best built complatibility
-PG_VERSION ?=   9.5
+# Default DB version.  Can be overridden in component Makefile.
+PG_VERSION ?=   11
 PG_IMPLEM ?=    postgres
 PG_VERNUM =     $(subst .,,$(PG_VERSION))
 # For dependencies, including REQUIRED_PACKAGES if needed
@@ -827,8 +827,14 @@ REQUIRED_PACKAGES_SUBST+= PG_DEVELOPER_PKG
 REQUIRED_PACKAGES_SUBST+= PG_LIBRARY_PKG
 
 PG_HOME =       $(USRDIR)/$(PG_IMPLEM)/$(PG_VERSION)
+# Binary directories match reality now
+ifeq ($(strip $(PG_VERSION)),9.5)
 PG_BINDIR.32 =  $(PG_HOME)/bin
 PG_BINDIR.64 =  $(PG_HOME)/bin/$(MACH64)
+else
+PG_BINDIR.32 =  $(PG_HOME)/bin/$(MACH32)
+PG_BINDIR.64 =  $(PG_HOME)/bin
+endif
 PG_BINDIR =     $(PG_BINDIR.$(BITS))
 PG_INCDIR =     $(PG_HOME)/include
 PG_MANDIR =     $(PG_HOME)/man


make-rules/shared-macros.mk
```diff
@@ -814,8 +814,9 @@ PKG_MACROS +=   PERL_ARCH=$(PERL_ARCH)
 PKG_MACROS +=   PERL_VERSION=$(PERL_VERSION)
 
 # Config magic for Postgres/EnterpriseDB/...
-# Default DB version is the oldest one, for hopefully best built complatibility
-PG_VERSION ?=   9.5
+# Default DB version is one of: 9.5 9.6 10 11 12
+# 9.5 is unsupported
+PG_VERSION ?=   11
```
AndWac, line 818: I am not happy with the new comment. That 9.5 is unsupported is not a relevant information for the future, but just for the PR.
Like before, the comment should explain why we chose 11 and nothing else. Even the concrete candidates will change over time.

jgmills:
By writing "9.5 is unsupported", I intended to mean that it was not a suitable default.  I can omit that line, if you prefer.

The line "Default DB version is one of: 9.5 9.6 10 11 12" was meant to specify the currently acceptable values of PG_VERSION .  I suppose that the word "Default" should be omitted.  As well, the possible values will change with time.  Perhaps the whole line should be omitted?

Choosing a suitable default is a difficult process.  It requires predicting what will happen in the future, and how quickly it happens.  The versions supported by the postgres developers are: 9.6 10 11 12 13 .  I arbitrarily chose one in the middle.  How do I express that in a Makefile comment?

make-rules/shared-macros.mk
pyhalov: The safe is the minimum supported version. If it's not so and by chance application/library uses libpq.so.5 from PG 10, application (at least, in theory) can break.

jgmills:
The minimum supported version is not a good choice for the postgres default, because that will always be the next version to become unsupported.  A version that will be supported for several years is a better default.  The only alternative is frequent changes of the default version.  We can't do that because we don't have people to do the work.

I already posted in my initial comment, a solution for applications or libraries that require a specific postgres library.  The alternative is never to change the default.  We already know that we must change it.  We do have to test applications or libraries to ensure that they work with a new postgres default.

pyhalov:
This is the sane choice in current layout. It's better than finding out reasons of strange segfaults...
Currently 
a) /usr/lib/libpq.so.5 can come from any PostgreSQL. 
b) If RPATH is not embedded in libpq-dependent library/app, it's compiled  with PG_VERSION, but uses /usr/lib/libpq.so.5.
If current libpq.so.5 comes from earlier version, this can lead to crash (e.g., symbol not found or has wrong semantics).
I see several possible sollutions:
1) current approach, when application is always compiled against earliest PG.
2) libpq.so.5 can be made non-mediated, shipped by 'true one' PostgreSQL and all libraries/apps, depending on PG, should be linked against this PG version.
3) some combination - libpq.so.5 is deliveres by the earliest PG version, we don't care about RPATH unless it's strictly necessary (postgresql extensions).

But just compiling everything against 'true one' PG version doesn't seem safe - what happens if you compile some complex library (like qt) against PG 11, RPATH is lost, user has PG 10 libs installed and compiles something against QT5? "Symbol not found" is the most evident and easiest issue which can happen. I don't think anyone proved that other issues are not possible.

jgmills:
Thanks for telling me about /usr/lib/libpq.so.5 .  I had assumed that each version of postgres was self-contained.  I see now that that's not quite the case.  Here's what I discovered:

/usr/lib/libpq.so.5.8 is a fallback shared object for applications or libraries that don't set the postgres version in the runtime path.  It provides one consistent version of postgres, using the mediator value to determine the version.

/usr/lib/libpq.so.5.8 (and /usr/lib/libpq.so.5) are symlinks.  The targets depend on the mediator value, but will be one of: /usr/postgres/*/lib/libpq.so.5.8 .  The target is unpredictable at build time.  This does not matter as long as the shared object API is the same.  I'm assuming that the postgres developers will change the SO name if the API changes.  This design ensures that those applications or libraries will use a specific version of postgres and no other.

So, we seem to be okay with the current design.  The default can be set to any version without causing any problems.  Applications or libraries that set a runtime path will use the postgres version embedded in that path.  Those that don't will follow the symlink in /usr/lib, and derive the version from the mediator value.

AndWac:
I am not sure how to proceed here. Did anybody run any tests to make sure everything works as expected?

jgmills:
I'm intending to change the comments to make them more general.  However, I have no way to do any testing of the Makefile include file.  It seems correct to me.

pyhalov:
Once again - minimum supported version is the only sane default.

jgmills:
It's not possible to be that sane or that safe.  Besides, absolute safety is not the only criterion in choosing a default.  You also want it to be stable over time, for example.  In the case of postgres, the user-selected mediator value is more important than the default, because that's what will be used to determine the version to be run.

As well, users don't care about the default version.  If somebody wants to use postgres on OI, they will install the latest supported version, not the oldest supported version.

As I said before, any version makes a good default.

jgmills:
I've updated the comments that you found objectionable.  I hope they are better now.

AndWac:
Did anyone of you run some tests to proof your claims? I cannot decide who of you is right. Shall we ask users in the openindiana mailing list?

jgmills:
We differ only in the behavior of applications that require a postgres database but don't care which version of postgres they use.  I don't know the names of any such applications.  If I did, I could do some testing.  These applications use the mediated symlink /usr/lib/libpq.so.5.8 that points to one postgres library of the same name.  Each postgres version has it's own copy of libpq.so.5.8 and runs independantly of the others.  The mediator is chosen by the user, although it defaults to version 9.5 now.  Users are free to install any version of postgres that is available in OI, usually causing a change in the mediator value.

I don't think either of us have run tests.  As well, users generally are not aware of the postgres default.  Only developers might know about the default.  However, when I posted a question to the oi-dev mailing list, I only heard from postgres users.  You probably won't be succesful in finding testers on the mailing lists.

pyhalov:
This is an over-simplified use case.
```
sudo pkg install database/postgres-12/developer database/postgres-95/developer
gcc -I /usr/postgres/12 test.c -o test1 -lpq
./test1 'host=192.168.1.2'
Result size is 2720b
oid            datname        datdba         encoding       datcollate     datctype       datistemplate  datallowconn   datconnlimit   datlastsysoid  datfrozenxid   datminmxid     dattablespace  datacl         

14016          postgres       10             6              ru_RU.UTF-8    ru_RU.UTF-8    f              t              -1             14015          579            1              1663                          
16384          leoric         10             6              ru_RU.UTF-8    ru_RU.UTF-8    f              t              -1             14015          579            1              1663                          
1              template1      10             6              ru_RU.UTF-8    ru_RU.UTF-8    t              t              -1             14015          579            1              1663           {=c/leoric,leoric=CTc/leoric}
14015          template0      10             6              ru_RU.UTF-8    ru_RU.UTF-8    t              f              -1             14015          579            1              1663           {=c/leoric,leoric=CTc/leoric}
sudo pkg set-mediator -V 9.5 postgres
./test1 'host=192.168.1.2'
ld.so.1: test1: fatal: relocation error: file test1: symbol PQresultMemorySize: referenced symbol not found
Killed
```

[test.c.gz](https://github.com/OpenIndiana/oi-userland/files/6461472/test.c.gz)

What I want to say is simple - compile applications with minimal available PG version. This is over-simplified example as build system would ignore dependency on /usr/lib/libpq.so.5, but would search for libpq symbols, but if such dependency is indirect (for example, via QT), build system doesn't catch it and we have the same issue. So, I consider that building apps using minimal available libpq version is the safest approach.

pyhalov:
The steps to reproduce:
- Use some libpq features, not available in minimal PG verssion,
- Lose RUNPATH

jgmills:
The cause of the "symbol not found" error in the example cited is that the library/qt5 package is built with the default postgres version of 9.5, but the missing symbol is only available in postgres version 12 .  The test program only works when the mediator is set to 12, but fails when the mediator is set to 9.5 .  This error could be seen as a problem with the default.  It could be corrected by rebuilding qt5 with "PG_VERSION = 12" near the top of the Makefile, assuming that qt5 really requires this version of postgres.

Since QT is partially a compiler, it should set the runtime path for binaries, as other compilers on OI do.  QT appears not to be one of the applications that works with any version of postgres.  The example cited is not relevant to this discussion.

pyhalov:
What I say is that in non-controlled build environment user programs easily loose RPATH, and this can't be reliably fixed.

jgmills:
Of course, users building their own software can do anything.  This behavior is not limited to postgres.  It can can affect any OI package.  That's why we build software and package it for OI.  We don't want to prevent users building software.  We do want to avoid leaving traps for unwary users.  /usr/lib/libpq.so is a trap, because it can point to any version of postgres, even one that does not work for the user's binary.  Maybe we should omit those symlinks entirely?

pyhalov:
Not sure. Omitting them now will likely break things,

Toasterson:
@pyhalov Lovely to see you in discussions again :)

As far as the compatibility question goes, I think your example has a bug, it does not compile with 9.5. If you use a feature that only exists since a version after 9.5 it won't work. Backwards compatibility means exactly that this has to fail! And it fails safely. In your example you are testing for Forward Compatibility. A in theory awesome concept, but I doubt anybody ever implemented it. 

In any case. Your input proves that we need a modern postgresql version as default. If everybody uses function introduced in a later version that we have we are not doing our job properly. We need to provide a modern default. Otherwise, people have to port to our system, and unless we have become a giant like RedHat overnight, that won't happen.

```bash
pfexec pkg install database/postgres-12/developer database/postgres-95/developer
pfexec pkg set-mediator -V 9.5 postgres
gcc -I /usr/postgres/9.5/include test.c -o test1 -lpq
```
```text
test.c: In function 'main':
test.c:99:37: warning: implicit declaration of function 'PQresultMemorySize'; did you mean 'PQresultErrorField'? [-Wimplicit-function-declaration]
     printf("Result size is %ldb\n", PQresultMemorySize(res));
                                     ^~~~~~~~~~~~~~~~~~
                                     PQresultErrorField
PQresultMemorySize                  /var/tmp//ccYqaW5A.o
ld: fatal: symbol referencing errors. No output written to test1
```

pyhalov:
It's not a bug, it's an example of program which requires particular libpq version. Anyway, what I intended to say - compile applications with minimum requirements to libpq, in such way they will not break in runtime if mediator is set to another version. So, if minimal libpq version we ship is PG 9.5, we should compile all apps against it.

Toasterson:
Hi Alexander,

Ok, I get the point sorry was a bit confused by the sample and lack of coffee.

To summarize:
If Apps assume 11 as feature set but 9.5 is chosen via mediator, then the apps break. If Apps by default assume 9.5 but 11 is chosen it works. But we set the mediator to switch automatically to the latest version installed unless specifically chosen. 

Does that make sense?

pyhalov:
It seems this makes sense, by default applications use the most feature-rich library. However, if one manually 'downgrade' mediator, after compiling some apps, this can cause issues.

AndWac:
I will close this as it might create troubles in some situations.

PR #6720, opened 2021-05-06
ejabberd: bump to 21.04 

grueni:
See https://github.com/OpenIndiana/oi-userland/pull/6719


AndWac:
You'll also need to increment COMPONENT_REVISION in the Makefile otherwise it's not guaranteed that this update will be installed.

components/web/ejabberd/Makefile
```diff
@@ -21,30 +21,31 @@
...
+
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= ejabberd
 COMPONENT_FMRI= web/server/$(COMPONENT_NAME)
-COMPONENT_VERSION= 17.11
-COMPONENT_REVISION= 1
+# 21.04 will fail with "ValueError: Zero padded number" during mogrify
+COMPONENT_VERSION= 21.4
 COMPONENT_PROJECT_URL=	http://www.ejabberd.im/
```
AndWac, line 35: In this case we have IPS_COMPONENT_VERSION, too. So you can use
COMPONENT_VERSION= 21.04
IPS_COMPONENT_VERSION=21.0.4
(We usually model leading zeroes like this).

grueni: Thanks, works as it should

components/web/ejabberd/files/svc-xmpp-ejabberd
```diff
@@ -22,7 +22,7 @@
 
 #
 # Copyright (c) 2009, 2011, Oracle and/or its affiliates. All rights reserved.
-# Copyright 2017 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
+# Copyright 20212121212121212121212121212121212121212121rueni). All rights reserved.
 #
```
AndWac, line 26: Is this strange copyright message intended?

grueni: Not really

AndWac:
Please squash the commits before I merge this PR. While squashing please remove commit messages that are irrelevant.

PR #6723, opened 2021-05-07
cog-spur, stack-spur : update to 5.0.2957

cstes:
Rename squeak-5 to opensmalltalk/stack-spur
Rename squeak-5c to opensmalltalk/cog-spur

Update to opensmalltalk oscog-eem.2957  (the version 5.0.2957 corresponds to the VMMaker oscog-eem.2957)

### DIFF
--- a/components/runtime/squeak5/Makefile
+++ b/components/runtime/squeak5/Makefile
@@ -14,12 +14,11 @@
 
 
 #
-# Package Makefile for the "OpenSmalltalk Cog Spur" Squeak Smalltalk V5 system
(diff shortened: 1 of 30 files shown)


AndWac:
One problem with how you'd done the renaming seems to be that with every future release of the new packages similar old packages will also be created and published. That is not good as we already have a problem with too much meta data.

AndWac:
Obsoletion and moving installations to newer packages shall be done by means of history files. The preferred location of history files is within the package's folder. The global history file should only be used when the package's folder will also be removed, eg. when the package will completely removed.

cstes:
I'll try to do this renaming with a history file ...

PR #6749, opened 2021-05-14
CouchDB: initial commit of version 3.1.1

grueni:


components/database/couchdb-31/couchdb-common.p5m
```diff
@@ -0,0 +1,26 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
+#
+
+set name=pkg.fmri value=pkg:/database/$(COMPONENT_NAME)-common-3@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
```
pyhalov: Are there any reasons to avoid calling package just couchdb-common (instead of couchdb-common-3)? I mean, if/when couchdb 4 is released, should it have separate user?

grueni: couchdb-common is obsoloted.
It should be resurrected but I have no idea how to manage that.

Just remove 
database/couchdb-common@2.1.1,5.11-2020.0.1.3
from components/meta-packages/history/history?

And rename database/$(COMPONENT_NAME)-common-3 of course.

pyhalov: Yes, remove from history and ensure that new version > old one

grueni:
Do we need a COMPONENT-REVISION?

AndWac:
No, because we don't have 3.1.1 released yet. It would only be necessary to distinguish a new build without a version increment, eg. when fixing internal problems.

grueni:
I asked because couchdb-common will have the same name if PR https://github.com/OpenIndiana/oi-userland/pull/6754 will be merged.

AndWac:
Yes, but that's ok because you will create a couchdb-common with an increased version number. The COMPONENT_REVISION is just an artificial part of a version number that is under our control. We need to increase it every time we change something internally.
I probably wouldn't have created a separate PR for it but just add the change of the global history within this PR because they belong together. But I am not experienced with this yet.

AndWac:
Please squash the commits.

components/database/couchdb-31/couchdb.p5m
```diff
@@ -0,0 +1,1256 @@
...
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend type=require fmri=database/couchdb-common-3
```
pyhalov: Package name was changed

AndWac (approved):

PR #6752, opened 2021-05-15
Drop PostgreSQL 9.5

pyhalov:


pyhalov:
This is a simple alternative to https://github.com/OpenIndiana/oi-userland/pull/6699

So far build is still going here...

PR #6753, opened 2021-05-15
squeak: move packages and classification to smalltalk

cstes:
Hopefully this commit is acceptable:

- move to runtime/smalltalk/*
- set COMPONENT_CLASSIFICATION to Development/Smalltalk

I've added history files in the component dirs as requested qnd kept the old directory names.

Advantage is it is now convenient to pkg uninstall runtime/smalltalk/* with a wildcard


components/runtime/squeak4/history
```diff
@@ -0,0 +1,3 @@
+runtime/squeak-4@4.19.6-2020.0.1.4 runtime/smalltalk/squeak
+runtime/squeak-4-nodisplay@4.19.6-2020.0.1.4 runtime/smalltalk/squeak-nodisplay
```
AndWac: All your FMRIs are missing ",5.11-" inside, eg.
runtime/squeak-4@4.19.6-2020.0.1.4 runtime/smalltalk/squeak should be
runtime/squeak-4@4.19.6,5.11-2020.0.1.4 runtime/smalltalk/squeak

This is the \<release\> part of an FMRI and usually omitted when being shown.

cstes:
I have updated the history files.   I am sorry I didn't understand those files before but the are input files for the history.mk make rules which deal with them ...

PR #6757, opened 2021-05-17
mbuffer: bump to version 20210328

grueni:


grueni:
gmake REQUIRED_PACKAGES finds only system/library as dependency but pkg5 contains 6 packages. Is this normal behaviour?
Found my error: after gmake REQUIRED_PACKAGES and editing Makefile you must use gmake publish to create a correct pkg5.

components/sysutils/mbuffer/Makefile
```diff
@@ -13,32 +13,38 @@
...
-include $(WS_MAKE_RULES)/configure.mk
-include $(WS_MAKE_RULES)/ips.mk
+COMPONENT_PRE_CONFIGURE_ACTION =	( $(CLONEY) $(SOURCE_DIR) $(@D) )
+
+CONFIGURE_OPTIONS +=	--sysconfdir=/etc
+
+include $(WS_MAKE_RULES)/common.mk
 
-build:		$(BUILD_32_and_64)
+test:		$(NO_TESTS)
```
AndWac, line 45: Make this
TEST_TARGET= $(NO_TESTS)
and move it above the include common.mk

components/sysutils/mbuffer/Makefile
```diff
@@ -13,32 +13,38 @@
...
 # Copyright 2016, Jim Klimov
+# Copyright 2021 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
 #
+
+BUILD_BITS= 64
+
 include ../../../make-rules/shared-macros.mk
 
+GCC_VERSION= 10
+
```
AndWac, line 24: I wonder if we should make gcc 10 our new default instead of manually set it all over.

grueni: If is assured that the compilation with gcc-10 works without problems it can removed from the Makefile.

AndWac:
Alternatively you can run "gmake update-metadata". This will only update pkg5.

PR #6768, opened 2021-05-19
binutils: update to 2.36.1

AndWac:


pyhalov:
Check illumos-gate and compilers builds with new binutils.

AndWac:
I have build and installed gcc-7 and gcc-10 successfully with the new version of binutils.
These new compilers then successfully built illumos-gate, gcc-7, gcc-10, nodejs-16, nodejs-14, and some other packages.

PR #6776, opened 2021-05-22
libp11: fix conflicting manifest actions

AndWac:
Without these changes libp11 cannot be installed and the installation on the build server doesn't look correct as the pkcs11.so files are missing from the engines folders.

### DIFF
--- a/components/library/libp11/Makefile
+++ b/components/library/libp11/Makefile
@@ -17,6 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libp11
 COMPONENT_VERSION=	0.4.11
+COMPONENT_REVISION=	1
 COMPONENT_ARCHIVE=	$(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.gz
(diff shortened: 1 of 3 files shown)


cgrzemba:
Looks good and works for me.

cgrzemba:
Do we need a separate build of libp11 for openssl1.1 then?

AndWac:
I don't know. The engine files are only delivered by openssl-1.0 and libp11 is only available for this version. I don't expect it to work with openssl-1.1 but haven't tried yet.

AndWac:
This PR has broken updates on system where it is installed.
The reason seems to be that openssl-1.0.2 delivers broken symlinks:
╰─➤  ls -l /lib/openssl
total 2
lrwxrwxrwx   1 root     root          28 Mar 11 18:51 default -> ../../../usr/openssl/1.0/lib/
lrwxrwxrwx   1 root     root          36 Mar 11 18:51 engines -> ../../../usr/openssl/1.0/lib/engines/

These symlinks go up one folder too far. That doesn't work when root is temporarily mounted during pkg update and thus it fails.

PR #6809, opened 2021-06-04
squeak,cog-spur: split ssl plugin in its own pkg

cstes:
installation of cog-spur-ssl  gives a basic minimal runtime with the SSL plugin but without any other plugins

### DIFF
--- a/components/runtime/squeak4/Makefile
+++ b/components/runtime/squeak4/Makefile
@@ -49,7 +49,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		squeak
 COMPONENT_VERSION=	4.19.6
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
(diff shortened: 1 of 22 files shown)


AndWac:
What is the purpose of this split? I don't expect too many squeak users on OI.

cstes:
This makes it  possible to use Smalltalk servers that use only the ST kernel and the SSL plugin.

For example a student at Potsdam university has written a Squeak package that acts as a server ("bot") for the Telegram messaging app.

This is basically a server that uses HTTPS (SSL).

By installing OI without GUI and installing cog-spur-ssl the runtime is present to run this Telegram bot in a hardened minimal setup without any of the GNOME/MATE packages.

So IPS allows an install in a minimal setup.

The same thing is true for the cog-spur-nodisplay package but that package cannot run the Squeqk bot for Telegram because HTTPS/SSL is required.

PR #6822, opened 2021-06-11
Upgrade of gtk-doc to 1.33.2 and python 3.7

jgmills:
Gtk-doc is a documentation tool that is primarily used to document the API of libraries from specially-formatted comment blocks.  It is usually used to develop GTK+ and GNOME libraries.  The product itself consists of a series of Python scripts and a private python module.  Gtk-doc has a circular dependancy that affects the OI publish operation and the built-in tests.  This PR upgrades the gtk-doc version from 1.25 to 1.33.2 and upgrades the python version that it uses from 2.7 to 3.7 .

The Makefile is updated for the new software version and the new python version.  It's also rewritten in the new style.  The Makefile disables the built-in tests, as they do not work unless the package is first installed on the build system.

The manifest gtk-doc.p5m now contains a transform to suppress unresolved reference errors during the publish operation.  It also contains some new or changed files.  These files are also present in the manifests/sample-manifest.p5m file.

The build, install, and publish operations were all successful.  I tried the built-in tests (gmake test) but they hang because of an error in the test framework.  The "gmake REQUIRED_PACKAGES" operation only added a comment to the Makefile.  I did install the newly-published package on the build system and did some manual tests of the python scripts.  These were only the "--version" and "--help" options, but they verify that the scripts ran correctly and that they found the private python module.


components/developer/gtk-doc/Makefile
```diff
@@ -9,44 +9,50 @@
...
-install:	$(INSTALL_32)
+# Build configure script
+COMPONENT_PREP_ACTION += (cd $(@D); autoreconf -fiv)
 
-test:		$(NO_TESTS)
+CONFIGURE_OPTIONS+=	--bindir=$(USRBINDIR)
+CONFIGURE_OPTIONS+=	--sbindir=$(USRSBINDIR)
+CONFIGURE_OPTIONS+=	--sysconfdir=$(ETCDIR)
+CONFIGURE_ENV+=		PYTHON=$(PYTHON)
+CONFIGURE_ENV+=		PERL=$(PERL)
```
AndWac, line 51: COMPONENT_TEST_MASTER = $(COMPONENT_TEST_RESULTS_DIR)/results-all.master
COMPONENT_TEST_ARGS =   -k -i
COMPONENT_TEST_TRANSFORMS += \
        '-n ' \
        '-e "/result_count/d" ' \
        '-e "/TOTAL:/p" ' \
        '-e "/SKIP:/p" ' \
        '-e "/PASS:/p" ' \
        '-e "/FAIL:/p" ' \
        '-e "/ERROR:/p" '

AndWac: Adding this (from solaris-userland) make the gmake test semi successful: 3 tests pass while 11 fail

jgmills:
I tried your suggestion, but it did not work for me.  The test operation only got this far:

`(cd /dpool/export/home/mills/Downloads/code/oi-userland-gh/components/developer/gtk-doc/build/amd64 ; \
        /usr/bin/env  \
        /usr/gnu/bin/make \
        -k -i check) \
        &> /dpool/export/home/mills/Downloads/code/oi-userland-gh/components/developer/gtk-doc/build/test/amd64/test-64-results`

At that point, it hung.  Here's part of the process tree:

`        20091  gmake test
          20148  /bin/bash -c (cd /dpool/export/home/mills/Downloads/code/oi-us
            20149  /bin/bash -c (cd /dpool/export/home/mills/Downloads/code/oi-
              20150  /usr/gnu/bin/make -k -i check
                20151  /bin/bash -c fail=; \ if (target_option=k; case ${target
                  20203  /bin/bash -c fail=; \ if (target_option=k; case ${targ
                    20204  /usr/gnu/bin/make check
                      20205  /bin/bash -c fail=; \ if (target_option=k; case ${
                        24461  /bin/bash -c fail=; \ if (target_option=k; case 
                          24462  /usr/gnu/bin/make check-am
                            24463  /usr/gnu/bin/make check-TESTS
                              24469  /bin/bash -c set +e; bases='tools.sh.log c
                                24478  /usr/gnu/bin/make test-suite.log TEST_LO
                                  24479  /bin/bash /dpool/export/home/mills/Dow
                                    24483  /bin/sh ./tools.sh
                                      24485  sh -n`

Part of the tools.sh script looks like this:

`# test shell scripts
for file in gtkdocize; do
  sh -n `which $file`
  if test $? != 0 ; then failed=`expr $failed + 1`; fi
  tested=`expr $tested + 1`
done`

There are three things wrong with this script segment, but on OI /bin/sh is a symlink to ksh93, not bash.  I found these messages in build/amd64/tests/tools.sh.log:

`Running suite(s): gtk-doc-tools
which: no gtkdocize in
(/dpool/export/home/mills/Downloads/code/oi-userland-gh/components/developer/gtk-doc/build/amd64:/dpool/export/home/mills/Downloads/code/oi-userland-gh/components/developer/gtk-doc/gtk-doc-1.33.2/tests::/usr/gnu/bin:/usr/bin/amd64:/usr/bin:/usr/sbin/amd64:/usr/sbin:/usr/perl5/5.22/bin)`

The error is correct.  There is no gtkdocize in any of those directories.  There would be if I had installed the package, but I did not.  The first thing wrong is that "sh -n" does not perform a test.  The second thing is a failure to account for a null result from the command enclosed in backticks.  Enclosing them in double quotes fixes that one.  Finally, "whence" is only a built-in for bash.  On OI, /usr/bin/whence is a csh script.

So, NO_TESTS is probably the best thing for this Makefile.

PR #6824, opened 2021-06-12
infra: add jenkins PR buildscripts

Toasterson:
This adds the tested scripts for the New Jenkins to automatically build PR's

I can install jenkins into the Organisation and after this is merged we have our branch and PR autobuilder which publishes to 
https://pkg-test.openindiana.aurora-opencloud.org/PR-$PRNUMBER

URL can be adapted in the reverse proxy config under https://github.com/Toasterson/openindiana-reverse-proxy-config/

@olbohlen I changed the Git line to one that shows all files different between current branch or PR and oi/hipster and not just the one in the last commit. Feel free to review.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,37 @@
+// OpenIndiana Userland Jenkinsfile, (c) 2021 Olaf Bohlen <olbohlen@eenfach.de>
+
+// ensure the jenkins user can modify SMF services, e.g. run:
+// usermod -A solaris.smf.* -P 'Service Operator' jenkins
+
+pipeline {
+    options {
+        // set a timeout of 600 minutes for this pipeline
+        timeout(time: 600, unit: 'MINUTES')
+    }
+    agent {
+      node {
+        label 'buildserver'
+      }
+    }
+    stages {
+        stage('setup stage') {
+            steps {
+		sh './tools/jenkinshelper.ksh -s setup'
+            }
+
+        }
+
+        stage('build changed components') {
+            steps {
+                sh './tools/jenkinshelper.ksh -s build_changed'
+            }
+        }
+
+	stage('create initial pkgdepotd.conf if not exists') {
+	    steps {
+		sh './tools/jenkinshelper.ksh -s prepare_pkgdepotd'
+	    }
+	}
+    }
+}
+
--- a/tools/jenkinshelper.ksh
+++ b/tools/jenkinshelper.ksh
@@ -1,0 +1,140 @@
+#!/bin/ksh
+# 2021-04-07 Olaf Bohlen <olbohlen@eenfach.de>
+# instead of putting all this into the Jenkinsfile I decided to put the actual code in this script
+
+# global config
+HTTPCONF="/etc/apache2/2.4/conf.d/pkgdepotd.conf"
+
+# just run prepare once to initially set up the environment
+# this must be run as root
+prepare() {
+    echo "jenkinshelper: preparing..."
+    pkg install web/server/apache-24
+    mkdir -p /etc/apache2/2.4/conf.d && chown root:sys /etc/apache2/2.4/conf.d
+    grep "#ProxyPassMatch /default/(.*)\$ http://127.0.0.1:10000/\$1 nocanon max=200" "${HTTPCONF}" >/dev/null 2>&1
+    if [ $? -gt 0 ]; then
+	echo "jenkinshelper: Initializing a new apache config at ${HTTPCONF}"
+	echo "#ProxyPassMatch /default/(.*)\$ http://127.0.0.1:10000/\$1 nocanon max=200" >"${HTTPCONF}"
+    else
+	echo "jenkinshelper: Preserving an existing ${HTTPCONF}"
+    fi
+
+    cat >/etc/apache2/2.4/conf.d/00-proxy.conf <<EOF
+LoadModule proxy_module libexec/mod_proxy.so
+LoadModule proxy_connect_module libexec/mod_proxy_connect.so
+LoadModule proxy_ftp_module libexec/mod_proxy_ftp.so
+LoadModule proxy_http_module libexec/mod_proxy_http.so
+LoadModule proxy_ajp_module libexec/mod_proxy_ajp.so
+
+ProxyRequests Off
+ProxyVia Off
+ProxyPreserveHost on
+RequestHeader unset Origin
+AllowEncodedSlashes NoDecode
+EOF
+    
+    chown root:builders "${HTTPCONF}"
+    chmod 664 "${HTTPCONF}"
+    svcadm enable svc:/network/http:apache24
+}
+
+# setup oi-userland
+stage_setup() {
+    if [ ! -f /etc/apache2/2.4/conf.d/00-proxy.conf ]; then
+	echo "jenkinshelper: aborting, please run \"jenkinshelper -p\" initially as root on this jenkins instance once"
+	exit 1
+    fi
+    echo "jenkinshelper: running gmake setup"
+    gmake setup
+}
+
+# scan the git log for changed components
+# we try to be smart and assume that all updates will always touch
+# the components Makefile also (to update COMPONENT_REVISION, etc)
+stage_build_changed() { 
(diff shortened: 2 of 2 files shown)


jenkinshelper.ksh
```diff
@@ -0,0 +1,139 @@
+#!/bin/ksh
```
AndWac: Did you consider to move this file to the tools folder?

jenkinshelper.ksh
```diff
@@ -0,0 +1,139 @@
...
+# 2021-04-07 Olaf Bohlen <olbohlen@eenfach.de>
+# instead of putting all this into the Jenkinsfile I decided to put the actual code in this script
+
+# global config
+HTTPCONF="/etc/apache2/2.4/conf.d/pkgdepotd.conf"
+
+# just run prepare once to initially set up the environment
```
jimklimov: Seems this part needs to be run as root, makes sense to document as such

jenkinshelper.ksh
```diff
@@ -0,0 +1,139 @@
...
+LoadModule proxy_ajp_module libexec/mod_proxy_ajp.so
+
+ProxyRequests Off
+ProxyVia Off
+ProxyPreserveHost on
+RequestHeader unset Origin
+AllowEncodedSlashes NoDecode
+EOF
+    
+    chown root:builders "${HTTPCONF}"
```
jimklimov: This group does not exist initially, right?

On a related note, I had a somewhat suspended work (that can move forward since a week ago, that pkg(5) PR contributing to that got merged), to add a standard account in our userland mappings, `abuild:399` (GID and UID) to match "automated builder" originally of OBS, but also our dayjob CI farm uses that to run Jenkins builds (in an account separate from the controller's `jenkins` account) since the two CI-related systems toss the ball a lot in our case.

So I either had a PR or planned to have one, to have such user defined in jenkins-common and to deliver finally the jenkins-swarm as a package to get automated self-registering builders running as local `abuild` in their zones or wherever.

So TL;DR: it may make sense to cooperate here and define a new special account once in our packaging, and in a more standardized manner at that :-)

UPDATE: That was about PR #6827 that I have now posted :)

jenkinshelper.ksh
```diff
@@ -0,0 +1,139 @@
...
+# prepare the pkg.depotd server instance, if an instance already
+# exists for this branch, we will use that - otherwise create a
+# fresh one
+stage_prepare_pkgdepotd()
+{
+    # we need the platform for the path to the repo
+    platform=$(uname -p)
+
+    # check if we already have this branch set up as a pkg.depotd:
+    if ! svcs pkg/server:${BRANCH_NAME} >/dev/null 2>&1; then
```
jimklimov: Oh, the horrors: I did find an un-quoted variable expansion! (Maybe ksh is safer in this regard -- what it considers a token after expansion, if the var value had a space -- but better safe than sorry) :)

jenkinshelper.ksh
```diff
@@ -0,0 +1,139 @@
...
+
+	# get highest port from ${HTTPCONF} to figure out the next free one
+	nextport=$(( $(nawk -F '[/:]' '{ print $7 }' < "${HTTPCONF}" | sort -n | tail -1) + 1 ))
+	
+	# set tup a new pkg/server instance
```
jimklimov: ty ypo :)

jenkinshelper.ksh
```diff
@@ -0,0 +1,139 @@
...
+	s)         stage="stage_${OPTARG}";;
+	p)         stage="prepare";;
+        h)         usage;;
+    esac
+done
+shift `expr ${OPTIND} - 1`
+
+# run requested stage
+${stage}
+rt=$?
```
jimklimov: Do we expect to `set -e` in this script eventually (so it fails ASAP on a line with non-zero result code) and yet handle caught errors? This couple of lines may be better rewritten as

````
rt=0
${stage} || rt=$?
````
so no line has a non-zero result and you have a chance to have and handle the non-zero `rt`.

Also, may be safer to sub-shell `( ${stage} )` to protect oneself from untamed evolution, so the parent script is not corrupted by whatever it runs or `exec`s in the stage :)

jimklimov (changes requested):
Quite decent, thanks :)

Had a few coding-style related comments, and did not finger-point at the mix of 4-space indents vs TABs for double-indents, but that better be made consistent also, whichever way feels better but some *one* way.

Also, since I dabbled with automated builds and setups for that, commented about a possible shared effort (and use-case => raison d'etre) for the static `abuild` GID/UID definition that automated build processes can run as, and which is not exactly a concern of any one build system like Jenkins or OBS or buildbot or whatever, same as a common `webservd` is not just for apache or nginx or...

jenkinshelper.ksh
Toasterson: Ah, no forgot bout that. Moving it there. The Jenkinsfile needs to be in the repo root though.

jenkinshelper.ksh
Toasterson: This is specific to the jenkins setup, which started before the PR about abuild landed. builders is the group that all users that login to the build zone to build and run their work have. This could be made related to abuild group but it's definetly future infra work. We should start doing tasks for these.

jenkinshelper.ksh
Toasterson: Thanks

jenkinshelper.ksh
Toasterson: fixed

jenkinshelper.ksh
Toasterson: Hmm your shell foo is probably better than mine. This is from @olbohlen don't know if he knows some more details about this.

jenkinshelper.ksh
Toasterson: Ok done, let me know where to explicitly document it if you have a good place to put it.

Toasterson:
@jimklimov the mix of spacing is probably from copy/pasting the file around. If you have a cli that makes it consistent let me know otherwise it's a bit to much work.

jimklimov:
Regarding TAB vs SPACE indents, I favor Midnight Commander and its Editor, so would just replace (F4/ESC+4) the <Ctrl+Q>+TAB as eight spaces. Or `\t` in regex substitution mode. And keep watch to only do this to indents (just in case there are valid tab chars inside the text somewhere).

Probably can do similar with
````
sed 's,\t,        ,g' -i filename
````
(or complicate that for avoiding tabs inside text), but depending on shell and/or layers of escaping from Jenkins groovy or other wrapper script languages to produce shell script, passing tabs into sed can be complicated -- too often I have to use something like
````
TABCHAR="$(printf '\t')"; sed "s,${TABCHAR},        ,g" -i filename
````

Finally, in OI we have `meld` or `gedit` and can TAB/SHIFT+TAB to indent in a chosen manner :)

jenkinshelper.ksh
jimklimov: Ah, might as well rename that there in the Jenkins setup =p

One semi-useful point is that since this PR and recipe lands into common oi-userland, it might as well be usable on developers' machines with (a zone for) local Jenkins building the same. And so using bits delivered by OI packaging.

Toasterson: Yes but I don't want to depend on abuild right now. Let's change this when we add abuild to jenkins dependencies.

Toasterson:
Apart from abuild anything else you would like to see here? Otherwise, I'll merge this and we can continue with the abuild in another task.

jimklimov (approved):
Apart from abuild suggestion, yes this LGTMs (can it be a verb?) now. Note that I did have reservations about "abuild-common" being a dependency for jenkins per se, there are pro's and con's to both approaches. In any case, that package (after PR merge) may be addressed here and documented as an explicit requirement for any local setup of Jenkins to build oi-userland.

Toasterson:
@jimklimov cool. Can you open a discussion in the Discussion tab here on Github about that? Then we can start co-ordinating and gather the changes we will need.

jimklimov:
Added discussions #6829 and #6831 for ideas raised in this PR and in my related recent others.
I thought they would make sense as "cards" in https://github.com/OpenIndiana/oi-userland/projects/9 but it seems only issues or PRs can be referenced there. At least, can make notes and type the name and link there :)

PR #6828, opened 2021-06-12
Introduce jenkins-swarm-client package 

jimklimov:
Note: relies on PR #6827 to deliver the `abuild-common` package to consistently create the `abuild` UID and GID accounts.

Currently posting as a "Draft PR" because:
1) Relies on a feature in another PR not yet merged (includes a copy of its current commit though)
2) Initial posting was a while ago and converted from scripts I used in another project, so may need more testing in practice (and/or code ports) that I soon hope to have in that project's re-engineered CI. Others are welcome to test and give feedback though :)
3) Includes a commit to test the `check-version.sh` script to bump from initial 3.25 to 3.27; this may be squashed later to merge the PR.


components/developer/jenkins-swarm-client/files/jenkins-swarm-client.sample-conf
```diff
@@ -0,0 +1,147 @@
...
+
+# No default => no option, if not specified
+CUSTOM_WORKERDESC_FILE="/opt/jenkins-swarm-client/etc/client.description"
+
+# Override some of the settings used for the swarm client
+CUSTOM_NUMEXECUTORS_FILE="/opt/jenkins-swarm-client/etc/client.numexecutors"
+DEFAULT_NUMEXECUTORS=1
+CUSTOM_WORKDIR_FILE="/opt/jenkins-swarm-client/etc/client.workdir"
+DEFAULT_WORKDIR="/tmp/jenkins-swarm-client"
+if [ -d /dev/shm ] && [ -w /dev/shm ] ; then
```
Toasterson, line 25: This is a linux specific directory. Do we need this?

components/developer/jenkins-swarm-client/files/jenkins-swarm-client.sh
```diff
@@ -0,0 +1,148 @@
...
+shift
+
+# Default setting: Allow to scale RAM consumption
+JAVA_OPTS="-Xms64m -Xmx512m"
+
+# Allow to investigate why an agent crashed, which rarely happens, to report
+# details into  https://issues.jenkins-ci.org/browse/JENKINS-60673
+JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
+
+# This script can be used standalone, or as an SMF method script
```
Toasterson, line 39: I think a "dual mode" like this is not such a good idea. If we are using SMF it should call a simple wrapper in /usr/bin to handle application specific startup and handle the SMF part in the method file. Keep things where they belong.

Toasterson (changes requested):
This PR duplicates the abuild-common related commits, please cleanup.

In general I like tha application and would love to see it for our building in zones setup. Could you work with the assumption of using it for that and add extensions for that? So that we can have a host VM that spins up a new Zone from golden snapshot and builds packages in there.

components/developer/jenkins-swarm-client/files/jenkins-swarm-client.sample-conf
jimklimov: In OI alone, probably not. The script originated as a multiplatform thing for various builders, and it is easier to maintain that way to set up swarm workers on different OSes with the same code.

jimklimov:
Regarding "host VM that spins up a new Zone from golden snapshot and builds packages in there", did you mean something like https://github.com/jimklimov/illumos-zonemgr I started a few fosdems ago but sadly never fully finished (and I think eventually @xen0l's approach to builds in zones landed... or was PRed at least)?

In my case, the skeleton is there and works to manage the zone lifecycles, and needs putting some flesh on the bones to actually preinstall desired packages, perhaps detect which golden image zone (per package set) is the best starting point to clone and adapt, and stuff.

Also it was made deliberately not as a zone brand, to be conceptually reusable for any type of zone where we might want golden images and clones of those (not necessarily builds). Can be lx, can be bhyve, why not. But that choice made it more complicated to interact for package preinstalls and inspections :)

components/developer/jenkins-swarm-client/files/jenkins-swarm-client.sample-conf
Toasterson: Well it would probably be better not to have too many complex configs. Otherwise it gets too confusing for new people. Let's keep configs we deliver via the OS group tailored to that OS. Everybody can have their own config afterwards.

Toasterson:
pre-intalls should be done by the build scripts IMHO. They have code to install the required packages and the metadata. So no need for that. Just the base image with the snapshoted starting point with the swarm agent installed..

components/developer/jenkins-swarm-client/files/jenkins-swarm-client.sample-conf
jimklimov: Hm, right. I first thought the question was about the script (which better be reusable). Sample config file indeed can be tailored.

components/developer/jenkins-swarm-client/files/jenkins-swarm-client.sh
jimklimov: Repeating the misplaced comment form another conversation: this script originated from, and hopefully will remain usable as, a script for setting up the Jenkins Swarm Client on a large variety of OSes. So I am not in favor of tailoring it to any single OS and/or service management framework (that is, in a way that reduces its usability "as is" in other cases).

jimklimov:
Regarding commits for abuild, they are one of the reasons this is so far a draft PR and not a "real" one :)
Until that PR #6827 for abuild-common is merged, this branch can be checked out "as is" and built locally to test the PR for swarm client.

Regarding preinstalls, the scripts in illumos-zonemgr purported to have arguments to make that simple for CI systems to use (e.g. pass a list of needed deps, and zonemgr makes the magic happen). If that gets externalized, at least by one currently available CI system (oi-userland scripts) so much the better, I think that project can already help "as is" then. And might absorb the way to install desired packages later =D

PR #6849, opened 2021-06-26
Upgrade of dtrace/toolkit to python 3.7

jgmills:
The entire purpose of this PR is removal of the dependancy of developer/dtrace/toolkit on python-2 .  The DTraceToolkit itself is a collection of tools written using DTrace for the Solaris 10[tm] OS by Sun Microsystems[tm]. Most of these scripts will also work on illumos and OI.  There is no software upgrade in this PR.  The version remains at 0.99 .  The python version, used by the software, is upgraded from 2.7 to 3.7 .

The only significant change to Makefile is to increment COMPONENT_REVISION .  The patch patches/shebang.patch is revised to omit the python changes.  A new patch patches/python3.patch incorporates all changes for the two python scripts.  The only significant change to the manifest dtracetoolkit.p5m is the change to the python dependancy.

The build, install, and publish operations were all successful.  There are no built-in tests.  However, I did try some of the most useful dtrace scripts after I had installed the package.  The scripts all seemed to work normally.

### DIFF
--- a/components/developer/dtracetoolkit/manifests/sample-manifest.p5m
+++ b/components/developer/dtracetoolkit/manifests/sample-manifest.p5m
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2016 <contributor>
+# Copyright 2021 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
@@ -273,7 +273,9 @@ file path=opt/DTT/Code/Perl/hello_strict.pl
 file path=opt/DTT/Code/Perl/hello_strict.pl.~1~
 file path=opt/DTT/Code/Php/func_abc.php
 file path=opt/DTT/Code/Python/func_abc.py
+file path=opt/DTT/Code/Python/func_abc.py.~1~
 file path=opt/DTT/Code/Python/func_slow.py
+file path=opt/DTT/Code/Python/func_slow.py.~1~
 file path=opt/DTT/Code/Readme
 file path=opt/DTT/Code/Ruby/func_abc.rb
 file path=opt/DTT/Code/Ruby/func_abc.rb.~1~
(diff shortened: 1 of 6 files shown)


components/developer/dtracetoolkit/manifests/sample-manifest.p5m
```diff
@@ -273,7 +273,9 @@ file path=opt/DTT/Code/Perl/hello_strict.pl
 file path=opt/DTT/Code/Perl/hello_strict.pl.~1~
 file path=opt/DTT/Code/Php/func_abc.php
 file path=opt/DTT/Code/Python/func_abc.py
+file path=opt/DTT/Code/Python/func_abc.py.~1~
```
Toasterson, line 276: This should not be installed. Can we make sure it is not landing on the End user computers?

components/developer/dtracetoolkit/manifests/sample-manifest.p5m
```diff
@@ -273,7 +273,9 @@ file path=opt/DTT/Code/Perl/hello_strict.pl
...
 file path=opt/DTT/Code/Python/func_slow.py
+file path=opt/DTT/Code/Python/func_slow.py.~1~
```
Toasterson, line 278: Also a backup file

Toasterson (commented):
It looks like we install some backup files at least in the Prototype area.
Can we make sure those backup files from patch process are not installed in the enduser systems?

I am not blocking the merge, so if another developer wants to merge go ahead.

components/developer/dtracetoolkit/manifests/sample-manifest.p5m
jgmills: The patch backup files all have a suffix of  "~1~".  A "gmake install" installs them to the prototype area, but a "gmake" publish ignores them.  This has always been the case and is still the case.  So, these files never land on  a user's desktop.  Your concerns are unfounded.

Toasterson: Ah, good!

PR #6879, opened 2021-07-07
uthash: add version 2.3.0

torculus:
Adding [uthash](https://github.com/troydhanson/uthash), a library of C macros for hash tables and more


components/library/uthash/Makefile
```diff
@@ -0,0 +1,34 @@
...
+COMPONENT_CLASSIFICATION=Development/C
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=https://github.com/troydhanson/uthash/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:e10382ab75518bad8319eb922ad04f907cb20cccb451a3aa980c9d005e661acc  
+COMPONENT_LICENSE=BSD
+COMPONENT_LICENSE_FILE=	uthash.license
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
+include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 32: With the new style (BUILD_BITS et al.) only include $(WS_MAKE_RULES)/common.mk is necessary. Please remove prep.mk and ips.mk

components/library/uthash/Makefile
```diff
@@ -0,0 +1,34 @@
...
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 Benjamin S. Osenbach
+#
+
+BUILD_BITS=32_and_64 # for binaries or 32_and_64 for libraries
+BUILD_STYLE=justmake
```
AndWac, line 16: BUILD_BITS=NO_ARCH should be enough (or 32 or 64) because you'll only copy the header files and they are identical for both variants.

AndWac:
One more question: Why do you want to add it as a package? Is there a consumer for it?

torculus:
Okay, I can make those changes and re-submit

I'm packaging uthash in the hopes of compiling [picom](https://github.com/yshui/picom), which requires it as a dependency. However, it is also a dependency for several other [projects](https://archlinux.org/packages/community/any/uthash/)

components/library/uthash/patches/01-add-Makefile.patch
```diff
@@ -0,0 +1,24 @@
...
++	@cd src/ && cp -p utarray.h $(DESTDIR)$(PREFIX)/include
++	@cd src/ && cp -p uthash.h $(DESTDIR)$(PREFIX)/include
++	@cd src/ && cp -p utlist.h $(DESTDIR)$(PREFIX)/include
++	@cd src/ && cp -p utringbuffer.h $(DESTDIR)$(PREFIX)/include
++	@cd src/ && cp -p utstack.h $(DESTDIR)$(PREFIX)/include
++	@cd src/ && cp -p utstring.h $(DESTDIR)$(PREFIX)/include
++
++uninstall:
++	@rm -rf $(DESTDIR)$(PREFIX)/include/utarray.h
++	@rm -rf $(DESTDIR)$(PREFIX)/include/uthash.h
```
AndWac: I think we should move the files into its own folder, eg. include/uthash/.
This is what Solaris did for many libraries.

torculus: Would other programs be able to find the headers then? I was under the impression that `#include <something.h>` searches for `something.h` in `/usr/include/`, not in any subdirectories

torculus:
After some more thought, I decided that you're right. It will be easy to patch the file in picom if it can't find uthash

The manifest now reflects that each header is installed to `/usr/include/uthash`

components/library/uthash/patches/01-add-Makefile.patch
AndWac: Either by changing it to #include <uthash/something.h> or by adding the path to the include-search-path to gcc, eg. by adding  -I /usr/include/uthash

PR #6882, opened 2021-07-08
fetchmail: update to 6.4.19, remove Python

balrog-nona:
This new version builds, installs and publishes fine. When installed locally, I tested it with this `.fetchmailrc` file:
```
set logfile /export/home/user_name/fetchmail.log
poll imap.seznam.cz protocol IMAP
    user "mail_address" is user_name here
    password "my_passwd"
    keep
    ssl
```
and I ran commands from man page: https://www.fetchmail.info/fetchmail-man.html#4
```
env LC_ALL=C fetchmail -V -v --nodetach --nosyslog
env LC_ALL=C fetchmail -vvv  --nodetach --nosyslog
```
and it worked - it fetched my mail using imap or pop3. I wasn't able to test it with gmail because it looks like gmail has another layer of security and it would require some setting on the gmail's side.

Building fetchmail with `fetchmailconf.py`  was very problematic. Therefore, I blocked it and this way avoids depending on Python. Fedora did similar thing: https://src.fedoraproject.org/rpms/fetchmail/blob/rawhide/f/fetchmail.spec#_49

### DIFF
--- a/components/mail/fetchmail/Makefile
+++ b/components/mail/fetchmail/Makefile
@@ -22,54 +22,36 @@
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
 # Copyright 2018, Michal Nowak
+# Copyright 2021, Nona Hansel
 #
 
-PREFERRED_BITS=64
-
+BUILD_BITS= 64
+USE_OPENSSL11= yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		fetchmail
-COMPONENT_VERSION=	6.3.26
-COMPONENT_REVISION=	3
+COMPONENT_VERSION=	6.4.19
+COMPONENT_DESCRIPTION=	Full-featured, robust, well-documented remote-mail retrieval and forwarding utility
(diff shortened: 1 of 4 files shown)


components/mail/fetchmail/Makefile
```diff
@@ -22,54 +22,36 @@
...
+BUILD_BITS= 64
+USE_OPENSSL11= yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		fetchmail
-COMPONENT_VERSION=	6.3.26
-COMPONENT_REVISION=	3
+COMPONENT_VERSION=	6.4.19
+COMPONENT_DESCRIPTION=	'Full-featured, robust, well-documented remote-mail retrieval and forwarding utility'
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
```
AndWac, line 35: Don't put hyphens around the description text. In the manifest the text will be sourrounded by "".

balrog-nona: Removed them, thanks.

PR #6893, opened 2021-07-15
borgbackup: add openssl-11, add Python3.9

balrog-nona:
I added two things:

- Python3.9 because I got warning about EOL during building

- openssl-11. Even though I believe it uses openssl-11, for some reason `gmake REQUIRED_PACKAGES` doesn't find it so I added it manually. 

When installed locally, I tested it with
```
borg init --encryption=repokey /path/to/repo
borg create /path/to/repo::Monday ~/src ~/Documents
```
and it worked.


AndWac:
I doubt that configuring openssl-1.1 like this really works.
If I check the crypto library:
╰─➤  ldd low_level.cpython-39.so
	libcrypto.so.1.0.0 =>	 /lib/64/libcrypto.so.1.0.0
	libc.so.1 =>	 /lib/64/libc.so.1
	libsocket.so.1 =>	 /lib/64/libsocket.so.1
	libnsl.so.1 =>	 /lib/64/libnsl.so.1
	libdl.so.1 =>	 /lib/64/libdl.so.1
	libmp.so.2 =>	 /lib/64/libmp.so.2
	libmd.so.1 =>	 /lib/64/libmd.so.1
	libm.so.2 =>	 /lib/64/libm.so.2

you can see that the library links against openssl-1.0. This would need the mediator set to openssl-1.1 which isn't possible because it will fail other packages not prepared for it.

balrog-nona:
I'll close this PR. I didn't know how to test it properly, thanks for showing me.

AndWac:
Maybe switching to Python-3.9 would be enough for now.
We have a big problem with packages like borgbackup. I converted all (?) packages from openssl-1.0 to 1.1 where it possible without setting the openssl-mediator. We still have a lot of packages that would need to be re-published with the meditor set to 1.1. Alas we also have some package that aren't compatible with openssl-1.1 and also depend on the correct mediator settings.
I am not sure how to proceed:
- either change the openssl mediator to 1.1 and re-publish appropriate packages (and obsolete problematic ones (or let them rot)).
- or try to remove the mediator dependency from the remaining, problematic packages. In fact, we have pkgconfig files that should be used by the packages. Alas only a few follow the settings they provide.

balrog-nona:
I opened a new PR for adding Python 3.9.
As for incorporating openssl-1.1, unfortunately, I don't have sufficient knowledge to give any advice about how to fix the situation.

PR #6894, opened 2021-07-15
mlocate: use group id 98 instead of already used gid 95.

klausz65:
while I was a producing a new ISO for SPARC, distro constructor told be GID 95 is already used by group slocate, which belongs to the illumos-gate, therefore GID for package mlocate  needs to be amended. The next free GID 98 was choosen for this porpose.


AndWac:
You'll need to add/increment the COMPONENT_REVISION in the Makefile.

AndWac (approved):

PR #6897, opened 2021-07-16
openconnect: remove dependency libp11

cgrzemba:
With update openssl to version 1.1 we disable all extra engines like solaris-userland it does. 
So we drop also the engine libp11. The only package which has an IPS dependency is openconnect. For openconnect it is a weak dependency, so if it is not avail it will not used and should not raise any complains. 
Beside openssl also gnutls could use libp11, but there are no package dependnecies.


AndWac:
You'll need to split this PR into two - you cannot retire a package and rebuild a dependant one in one step. The retirement is in place after it has successfully been merged.

PR #6904, opened 2021-07-18
bison: update to 3.7.6

AndWac:


AndWac:
I am not sure whether the .gnulib files should be in bison-runtime (or if we should merge the two manifests into one).

PR #6908, opened 2021-07-19
binutils: update to 2.37

AndWac:


AndWac:
Integration postponed until https://illumos.topicbox.com/groups/developer/T801d5f232ba7f2af/review-13958-kptitrampolines-s-fails-to-assemble-with-binutils-2-37

PR #6921, opened 2021-07-21
borgbackup: add Python 3.9

balrog-nona:
When installed locally, I tested it with
```
borg init --encryption=repokey /path/to/repo
borg create /path/to/repo::Monday ~/src ~/Documents
```
and it worked.


pyhalov:
Hi. At least borg-35 package should be marked obsolete.

balrog-nona:
> Hi. At least borg-35 package should be marked obsolete.

I created `history` file, thanks.

alarcher:
@pyhalov Should there also be a dependency type=optional to the version deprecating 3.5 to avoid clashes w.r.t /usr/bin/borg

pyhalov:
In fact yes, but it doesn't matter much as the main way of getting updates on OI seems to be just `pkg update`, and it will both fetch  new and remove packages.

components/sysutils/borgbackup/history
```diff
@@ -0,0 +1 @@
+backup/borg-35@1.1.17-2020.0.1.1
```
AndWac: This should be "backup/borg-35@1.1.17,5.11-2020.0.1.1"
5.11 is the build version, typically left out when the FMRI is being printed (and separated by a "," from the component version). According to the docs the build version is the version of the OS the component is being built for.

balrog-nona: Sorry for that. I took FMRI from `pkg info borg-35` which now I see doesn't contain the whole version.

AndWac: Never mind, I made the same mistake in the past and it took some time to understand it.

PR #6925, opened 2021-07-22
pyasn1: update to 0.4.8 & add Python 3.7, 3.9 & obsolete 2.7 versions

AndWac:


pyhalov:
library/python/python-twisted-27 still depends on pyasn

AndWac:
Yes, I am working on twisted (also removing 2.7 support) and its dependencies.
I wonder how far we can get with obsoleting python 2.7 and what needs to be done to switch our IPS to 3.9.

PR #6929, opened 2021-07-22
New python module lazy_object_proxy, version 1.6.0

jgmills:
This PR adds the lazy_object_proxy packages, version 1.6.0 .  The product is a fast and thorough lazy object proxy, used in situations where you want to call a function the first time the proxy object is used.  It is the second of two packages required by the astroid package, which is in turn required by recent versions of pylint.

The python module is known as lazy_object_proxy, but the archive file unpacks to python-lazy-object-proxy-1.6.0 .  This discrepancy is corrected in Makefile .  Packages are built for python 3.7 and 3.9 .  All files are new, including Makefile, the manifest lazy_object_proxy-PYVER.p5m, the two manifests manifests/generic-manifest.p5m and manifests/sample-manifest.p5m, and the two patches in the patches directory.  Both patches affect only the built-in tests.

Note that the manifest lazy_object_proxy-PYVER.p5m contains a file named cext.so .  This file is magically renamed to cext.cpython-37m.so or cext.cpython-39.so, as appropriate, during the publish operation.  The magic is accomplished by a built-in transform.  The build, install, and publish steps were all successful.  The built-in tests require that pytest be installed.  Typical results, showing 36 xfailed and 1504 passed tests, are included in the Makefile.

### DIFF
--- a/components/python/lazy_object_proxy/Makefile
+++ b/components/python/lazy_object_proxy/Makefile
@@ -1,0 +1,70 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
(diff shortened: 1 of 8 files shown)


AndWac:
General remark on patches: I'd like to have short explanations in front of every non-trivial patch.
The explanation should give some information about why it is necessary and only if it is not obvious how.
This will help in the future when sources have changed.
You can write arbitrary text in front of any patch. This is ignored by the patch utility.

PR #6936, opened 2021-07-24
Adding oniguruma regex library

Toasterson:
This is a build dependency for jq with test functionality

### DIFF
--- a/components/text/oniguruma/Makefile
+++ b/components/text/oniguruma/Makefile
@@ -1,0 +1,37 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 Till Wegmueller
+#
+
+BUILD_BITS=64 # for binaries or 32_and_64 for libraries
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=			oniguruma
+COMPONENT_VERSION=		6.9.7
+COMPONENT_SUMMARY=		oniguruma - regular expression library 
+COMPONENT_PROJECT_URL=		https://github.com/kkos/oniguruma
+COMPONENT_FMRI=			text/oniguruma
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_SRC=			onig-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=		$(COMPONENT_SRC).1.tar.gz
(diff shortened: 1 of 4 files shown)


components/text/oniguruma/Makefile
```diff
@@ -0,0 +1,36 @@
...
+COMPONENT_VERSION=		6.9.7
+COMPONENT_SUMMARY=		oniguruma - regular expression library 
+COMPONENT_PROJECT_URL=		https://github.com/kkos/oniguruma
+COMPONENT_FMRI=			text/oniguruma
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_SRC=			onig-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=		$(COMPONENT_SRC).1.tar.gz
+COMPONENT_ARCHIVE_URL=		https://github.com/kkos/$(COMPONENT_NAME)/releases/download/v$(COMPONENT_VERSION).1/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=		sha256:6444204b9c34e6eb6c0b23021ce89a0370dad2b2f5c00cd44c342753e0b204d9
+COMPONENT_LICENSE=		COPYING
```
AndWac: COMPONENT_LICENSE= oniguruma license
(This should be an appropriate name, like MIT, BSD, ... oniguruma seem have its own, BSD like)
COMPONENT_LICENSE_FILE = oniguruma.license
(This one is optional because it is set in shared_macros.mk to <comonent name>.license)

AndWac (changes requested):
Despite the license type ok

Toasterson:
Licenses fixed and removed static library files from delivery

components/text/oniguruma/Makefile
```diff
@@ -0,0 +1,37 @@
...
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 <contributor name>
```
alarcher: Contributor name missing.

components/text/oniguruma/Makefile
```diff
@@ -0,0 +1,37 @@
...
+#
+
+BUILD_BITS=32_and_64 # for binaries or 32_and_64 for libraries
+include ../../../make-rules/shared-macros.mk
```
AndWac, line 16: The only consumer is jq which is 64-bit only. Do we really want to add a new 32-bit variant? IMO we should try to phase out 32 bit where possible.

Toasterson: I have no real opinion on what I want here :shrug:. So I think we can go with yours. I'll send an update later today and make it 64bit only.

components/text/oniguruma/Makefile
Toasterson: Thanks!

AndWac (approved):

PR #6938, opened 2021-07-25
Gcc 102

klausz65:
Added -mcpu=ultrasparc to shared-makros: gcc_OPT.sparc.32 and gcc_OPT.sparc.64 FLAGS,
because v8 CPUs aren't supported since Solaris10.
Syncronize the SPARC default binary format for gcc-10.2.0 to be 64bit, as allready done for amd64.


alarcher:
Could you clean up the PR to avoid merge conflicts?

PR #6940, opened 2021-07-26
New python module astroid, version 2.6.2

jgmills:
This PR adds the astroid package, version 2.6.2 .  The product is a common base representation of python source code, used presently only as the library powering pylint's capabilities.  It is required by recent versions of pylint.  This package allows pylint to be upgraded to the current version.  The package is only built for python 3.9, the same as pylint will be.  All files are new, including Makefile, the manifests astroid-39.p5m and astroid.p5m, and manifests/sample-manifest.p5m .  There are no patches.

"gmake REQUIRED_PACKAGES" had no effect.  The build, install, and publish steps were all successful.  The built-in tests require that pytest be installed.  Typical results, showing 1 failed and 987 passed tests, are included in the Makefile.

### DIFF
--- a/components/python/astroid/Makefile
+++ b/components/python/astroid/Makefile
@@ -1,0 +1,65 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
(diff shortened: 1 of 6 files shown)


balrog-nona:
I believe `manifests/generic-manifest.p5m` is missing. Could you please add this file into this PR?

jgmills:
No, there is no file manifests/generic-manifest.p5m .  That file is only present when the Makefile uses more than one version of python.  This one only uses 3.9 .

balrog-nona:
> No, there is no file manifests/generic-manifest.p5m . That file is only present when the Makefile uses more than one version of python. This one only uses 3.9 .

My mistake. I thought it might be present because there is variable PYTHON_VERSIONS in Makefile. Because some time ago I asked about this file and I got this answer: "The build machinery iterates over all
version that are listed in PYTHON_VERSIONS (note the plural -S at its end)
using the <component name>-PYVER.p5m file to generate a manifest for
each python version." Here: https://github.com/OpenIndiana/oi-userland/discussions/6390

PR #6941, opened 2021-07-28
sudo: update to 1.9.7p2

balrog-nona:
It builds, installs and publishes fine. Tests also ran fine.
I set COMPONENT_REVISION to 1 because only SRC_PATCH_VERSION changed. I hope it's correct.

When installed locally, I tested it with `sudo cat /etc/sudoers` which worked and man page says the right version.


alarcher:
@balrog-nona It seems that COMPONENT_VERSION contains the patch number so it should not be necessary to bump COMPONENT_REVISION. Can you confirm that the new package version is 1.9.7.2?

balrog-nona:
..> @balrog-nona It seems that COMPONENT_VERSION contains the patch number so it should not be necessary to bump COMPONENT_REVISION. Can you confirm that the new package version is 1.9.7.2?

You're right, it is version 1.9.7.2. I'll remove COMPONENT_REVISION.

components/sysutils/sudo/Makefile
```diff
@@ -30,15 +30,15 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		sudo
 SRC_VERSION=		1.9.7
-SRC_PATCH_VERSION=	0
+SRC_PATCH_VERSION=	2
 SRC_DOT_PATCH_VERSION=	$(shell [ $(SRC_PATCH_VERSION) -gt 0 ] && echo .$(SRC_PATCH_VERSION))
```
AndWac, line 34: Additionally you can also supply HUMAN_VERSION here and in the manifest.

balrog-nona: Did it, thanks.

PR #6942, opened 2021-07-29
ccache: ccache-update-symlinks.sh: fix matching of shorter prog names (e.g. "gcc" is not "gcc-10")

jimklimov:
Also runtime needs `realpath` from unlisted dependency

### DIFF
--- a/components/developer/ccache/Makefile
+++ b/components/developer/ccache/Makefile
@@ -17,7 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		ccache
 COMPONENT_VERSION=	3.3.4
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_PROJECT_URL=	http://ccache.samba.org/
 COMPONENT_DOWNLOAD_URL=	http://samba.org/ftp/$(COMPONENT_NAME)/
 COMPONENT_DOCUMENTATION_URL =	http://ccache.samba.org/manual.html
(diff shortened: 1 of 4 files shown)


components/developer/ccache/files/ccache-update-symlinks.sh
```diff
@@ -109,16 +109,28 @@ if [ "$NAMES" = "$LINKS" ]; then
 fi
 
 # Link missing compilers to get wrapped by ccache for its consumers
+# Note: originally used `fgrep` but it misfired for substring matches
+# finding e.g. "gcc" or "gcc-4" in "gcc-4.9"; not a full-string match
 for N in $NAMES ; do
-    echo "$LINKS" | fgrep "$N" >/dev/null || \
```
pyhalov, line 113: will `grep -w` work?

jimklimov: Not quite, `gcc` will be found in `gcc-10` for example; might be not-found in `gcc10` but we do not use version suffixes like that (FreeBSD seems to do so though)

components/developer/ccache/Makefile
```diff
@@ -52,3 +52,4 @@ REQUIRED_PACKAGES += library/zlib
 REQUIRED_PACKAGES += SUNWcs
 REQUIRED_PACKAGES += system/library
 REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += file/gnu-coreutils
```
AndWac, line 55: The new required packages makes an updated pkg5 file necessary.

jimklimov: Strange, thought I added that, sorry

PR #6955, opened 2021-08-02
ekiga: retire component

cgrzemba:
Unfortunately Ekiga is not under development any more since 2014 and was removed from Linux distributions already also. 
The follower Empathy is also "stable". There was recommendations to use software https://jami.net for audio- and video- conferencing.


components/meta-packages/history/history
```diff
@@ -455,6 +455,7 @@ SUNWzsh@4.3.9,5.11-0.133 shell/zsh@4.3.9-0.133
 babel_install@0.1,5.11-2015.0.2.0
 backup/areca@7.1,5.11-2016.0.1.0
 backup/borg-34@1.1.10-2020.0.1.1
+communication/conferencing/ekiga@4.0.1-2020.0.1.6
 communication/infinoted@0.5.5,5.11-2017.0.0.2
```
AndWac, line 459: This should be
communication/conferencing/ekiga@4.0.1,5.11-2020.0.1.6

cgrzemba:
Thanks for the hint!

PR #6961, opened 2021-08-04
X11 libraries update

alarcher:
Tested on my workstation (Nvidia).


torculus:
I'm pretty sure these commits need to be squashed into one before they can be committed

alarcher:
> I'm pretty sure these commits need to be squashed into one before they can be committed

Any specific reason why you would squash them?

AndWac (approved):

PR #6969, opened 2021-08-06
automat: bump to 20.2.0 and drop -27

alarcher:


components/python/automat/manifests/sample-manifest.p5m
```diff
@@ -25,6 +25,8 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 file path=usr/bin/automat-visualize
 file path=usr/bin/automat-visualize-2.7
 file path=usr/bin/automat-visualize-3.5
```
AndWac, line 26: Something looks wrong here. You have removed Python-2.7 support so I would expect all related files to be removed from the sample-manifest, too.

components/python/automat/test/results-3.5-32.master
```diff
@@ -0,0 +1,5 @@
+_test/test_core.py .....                                                 [  8%]
+_test/test_discover.py .......................                           [ 45%]
+_test/test_methodical.py .................                               [ 73%]
```
AndWac, line 3: running the tests here gives
-_test/test_discover.py .......................                           [ 45%]
+_test/test_discover.py sssssssssssssssssssssss                           [ 45%]

components/python/automat/manifests/sample-manifest.p5m
alarcher: Probably I forgot to clobber... damn I hate that we do not clean the prototype before each install.

components/python/automat/test/results-3.5-32.master
alarcher: Who knows which other python modules these tests use... :S

components/python/automat/manifests/sample-manifest.p5m
alarcher: FIxed.

PR #6977, opened 2021-08-08
change '$gmp_alias' into @gmp_alias@ in gcc/Makefile.in

klausz65:
This bug can lead to problems on virtualized build invironments


alarcher:
If this has been tested only on SPARC the component should be also rebuilt on x86 and the testsuite run.

AndWac (approved):
I run "gmake test" twice on my build server. One without the patch and another with this patch. The results were identical.

PR #6982, opened 2021-08-10
Extend hypothesis to python 37 and 39, dropping 27

jgmills:
This PR extends the python versions used by the hypothesis module to 37 and 39, but drops python 27.  Hypothesis is an advanced testing library for python, required by some python test modules, and used only by developers.  The version of hypothesis remains the same at 4.57.1 .

There is a newer version (6.14.5) that I tried to use, but I was unable to do so, for several reasons.  The new version has a long list of dependancies, most of which are not available in OI.  As well, "gmake test" of the new version failed with this error message:

  AttributeError: module 'datetime' has no attribute 'date'

I was forced to revert to the original version (4.57.1).

Hypothesis is required by the isort module, which is in turn required by recent versions of pylint.  Adding isort allows pylint to be upgraded.

For this PR, only a few files needed to be changed.  Makefile is changed to accomodate the addition and deletion of python versions.  The file history was changed to obsolete the python 27 version.  Note that there's no need to add ",5.11-" to FMRIs in history.  Finally, the file hypothesis-PYVER.p5m was changed to omit a dependency that was only needed for python 27 .

The build, install, and publish steps were all successful.  The source archive has no built-in tests.  I was able to do some external testing by running "gmake test" for the isort module.  Once the hypothesis module was installed, this step no longer issued error messages that mentioned hypothesis.


AndWac:
manifest/sample-manifest.p5m needs to be updated, too.

jgmills:
I assumed that sample-manifest.p5m would not change.  I was wrong.  That file is now included in the PR.  The file generic-manifest.p5m did not change, except for the Copyright comment.

AndWac:
The sample-manifest has been changed because you added new python versions. In fact, it is almost irrelevant because we have the generic manifest. Nevertheless we still keep and update it even when the generic manifest is there.

PR #6994, opened 2021-08-15
liblouis: update to 3.13.0

AndWac:
Our liblouis is outdated


balrog-nona:
I built and installed it locally successfully. I got the same test result on my machine.

I tested it with `lou_translate --help` and 
`echo "Getting Liblouis to run and translate" | lou_translate unicode.dis,en-gb-g1.utb` which gave me `⠠⠛⠑⠞⠞⠊⠝⠛⠀⠠⠇⠊⠃⠇⠕⠥⠊⠎⠀⠞⠕⠀⠗⠥⠝⠀⠁⠝⠙⠀⠞⠗⠁⠝⠎⠇⠁⠞⠑` So to me it seems to work.

PR #7004, opened 2021-08-16
New python module isort, version 5.9.3

jgmills:
This PR adds the package isort, version 5.9.3 .  The product is a Python utility / library to sort imports alphabetically, used to quickly sort all imports.  It is one of the packages required by recent versions of pylint.  This package allows pylint to be upgraded to the current version.  The package is only built for python 3.9, the same as pylint will be.  All files are new, including Makefile, the manifests isort-39.p5m and isort.p5m and manifests/sample-manifest.p5m .  There are no patches.

"gmake REQUIRED_PACKAGES" had no effect.  The build, install, and publish steps were all successful.  The built-in tests require that pytest and hypothesis be installed.  Typical results, showing 13 failed, 502 passed tests, are included in the Makefile.

### DIFF
--- a/components/python/isort/Makefile
+++ b/components/python/isort/Makefile
@@ -1,0 +1,85 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright 2021 Gary Mills
+#
+
+BUILD_STYLE=	setup.py
+BUILD_BITS=	NO_ARCH
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		isort
+COMPONENT_VERSION=	5.9.3
+# COMPONENT_REVISION=	0
+COMPONENT_PROJECT_URL=	https://github.com/PyCQA/isort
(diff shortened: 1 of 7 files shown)


components/python/isort/Makefile
```diff
@@ -0,0 +1,84 @@
...
+#FAILED integration/test_setting_combinations.py::test_isort_is_idempotent - is...
+#FAILED integration/test_setting_combinations.py::test_isort_doesnt_lose_imports_or_comments
+#...
+#FAILED unit/test_ticketed_features.py::test_isort_literals_issue_1358 - isort....
+#FAILED unit/test_ticketed_features.py::test_sort_configurable_sort_issue_1732
+#ERROR benchmark/test_api.py::test_sort_file
+#ERROR benchmark/test_api.py::test_sort_file_in_place
+#ERROR unit/test_main.py::test_sort_imports_error_handling
+#= 13 failed, 502 passed, 1 skipped, 76 warnings, 3 errors in 329.14s (0:05:29) ==
+
```
AndWac, line 83: You'll need to add the test dependency on library/python/sortedcontainers-39

jgmills:
Ah, I overlooked sortedcontainers-39 .  Thanks for noticing.  I've just added that package to the comment.

components/python/isort/Makefile
```diff
@@ -0,0 +1,85 @@
...
+include $(WS_MAKE_RULES)/common.mk
+
+# Copy in a minimal setup.py
+COMPONENT_PRE_BUILD_ACTION = $(CP) OI/setup.py $(COMPONENT_SRC)
+
+# The tests are run using python 3.9 only and require that the
+# python-39 package is installed (does not have to be the default python).
+# The pytest-39, hypothesis-39, and sortedcontainers-39 packages must
+# also be installed
+
```
AndWac, line 55: We usually add REQUIRED_PACKAGES += statements for test dependencies with an appropriate comment (eg. # Manually added test dependencies) but I think such a comment is also ok.
What do you think?

jgmills: The problem is that pytest and a few other packages are not dependancies of the packages described by the P5M files;  Instead, they are dependancies of the test modules.  Adding them to the REQUIRED_PACKAGES list will have no effect, although it might provide a clue for the developer.  Comments are the best we can do just now.  The automatic package install always fails for me because I do the builds as an unprivileged user.

I suppose a different variable, such as TEST_REQUIRED_PACKAGES or REQUIRED_TEST_PACKAGES might be an improvement.  Then, the test procedure could at least ensure that packages from that list are installed.

AndWac: Yes, that sounds reasonable. We only need to find someone who will provide such an enhancement.

PR #7005, opened 2021-08-17
GCC-7 and GCC-10 strip -x

alarcher:


AndWac (approved):
This works fine for gcc-7 on my build system.

PR #7006, opened 2021-08-17
gcc: strip the DWARF (but not symbol tables) from the resulting binaries

richlowe:
This strips the resulting GCC binaries of DWARF (but not the symbol tables, which we want to keep).  Saving _lots_ of download.

I've checked that the 10 binaries seem good, and am currently building 5 (the oldest using gcc-component).  I haven't run the test suites since we're not changing behaviour.  It could really use someone with more horsepower and OI knowledge making sure I didn't screw anything up.

(figures from a compressed filesystem, so fairly similar to what would go over the wire during install/upgrade)
```
; du -sh {build/prototype/i386,}/usr/gcc/10/
247M	build/prototype/i386/usr/gcc/10/
970M	/usr/gcc/10/
```

### DIFF
--- a/make-rules/gcc-component.mk
+++ b/make-rules/gcc-component.mk
@@ -101,6 +101,8 @@ FCFLAGS= -O2
 COMMON_ENV=  LD_OPTIONS="-zignore -zcombreloc -i"
 COMMON_ENV+= LD_FOR_TARGET=/usr/bin/ld
 COMMON_ENV+= LD_FOR_HOST=/usr/bin/ld
+COMMON_ENV+= STRIP="/usr/bin/strip -x"
+COMMON_ENV+= STRIP_FOR_TARGET="/usr/bin/strip -x"
 COMMON_ENV+= LD=/usr/bin/ld
 
 CONFIGURE_ENV+= $(COMMON_ENV)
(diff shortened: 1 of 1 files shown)


richlowe:
```
; pkg info -r pkg://{openindiana.org,userland}/developer/gcc-10 | grep Size
          Size: 1.87 GB
          Size: 471.68 MB
```

alarcher:
@richlowe I checked earlier today but it seemed that STRIP and STRIP_FOR_TARGET were not passed to install-sh, only 'strip' shows. I did patch this script at #7005 to make sure it is not dependent on the environment. So far 7, 9, and 10 are rebuilt on my build server, 8 is ongoing.

richlowe:
They were passed ok on 10 (as evidenced by the amount it shrunk, but that .symtab survived).  I'm still waiting on 5

alarcher:
@richlowe OK then I will merge your commit first then the PR with the gcc rebuilds rebased on top of it.

richlowe:
Well, more thorough checking of the other versions and stuff would be good first!

alarcher:
@richlowe Sure the testsuite for 7, 9, and 10 is already running.

alarcher:
Together with checking that the symbol table is present in the main libs.

richlowe:
Oh, GCC 5 bombs because it looks like STRIP_FOR_TARGET gets unquoted at some point or something.  Yuck

Toasterson:
If you guys bump all makefiles for all GCC releases in this PR, then Jenkins will build all of them (not the testsuite) so we can see the results and grab binaries from the temporary repo for testing.

AndWac:
We might want to drop gcc-5, gcc-6 and gcc-8.

alarcher:
@Toasterson I started queueing the updates ay #7005 yesterday as testsuites complete.

PR #7015, opened 2021-08-22
argh: add python-3.7 and 3.9 versions

AndWac:
barman-37 needs it


alarcher:
@AndWac Could you add -39 as well or is something missing?

AndWac:
The argh project is old and seems to be unmaintained. Officially even
3.7 is not supported. I was just reluctant to add 3.9 because of this.
I will update the PR.

Am 8/22/21 um 3:11 PM schrieb Aurélien Larcher:

PR #7016, opened 2021-08-22
New python module mccabe, version 0.6.1

jgmills:
This PR adds the package mccabe, version 0.6.1 .  The product is the McCabe complexity checker, used to detect over-complex code.  It is the last of the packages required by recent versions of pylint.  This package allows pylint to be upgraded to the current version.  The package is only built for python 3.9, the same as pylint will be.  All files are new, including Makefile, the manifests mccabe-39.p5m and mccabe.p5m, and manifests/sample-manifest.p5m .  There are no patches.

"gmake REQUIRED_PACKAGES" had no effect.  The build, install, and publish steps were all successful.  The built-in tests require that pytest and hypothesis be installed.  Typical results, showing 14 passed tests with no failures, are included in the Makefile.

### DIFF
--- a/components/python/mccabe/Makefile
+++ b/components/python/mccabe/Makefile
@@ -1,0 +1,65 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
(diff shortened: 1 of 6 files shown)


pyhalov:
Sorry for the noice, while updating pylint, consider  if we need support for Python 2.7 checking (and does new versions support it).

jgmills:
Regarding pylint, I'm following Aurlien Larcher's list of packages that still depend on python-27.  He states that python-27 will be removed from OI.  One of the packages in the list is pylint-27.  It is a python script installed in /usr/bin .

Strictly speaking, pylint only checks the source version for which it was built.  More generally, the 2.7 version checks any python-2 source.  The 3.9 version checks any python-3 source.  The new version of pylint only supports 3.5 to 3.9 for building.  The old version only supports 2.7 for building.

Consequently, the new version will not check python 2.7 files.  This may seem to be a problem, but OI never needs to do this.  Pylint is not specified in any OI Makefile.  As well, python 2.7 is disappearing.  Any development on OI will only be for python-3.  We will need the new pylint for that.

PR #7027, opened 2021-08-31
Upgraded to ntfs-3g_ntfsprogs 2021.8.22

jpandre:
This fixes vulnerability threats caused by maliciously tampered NTFS
partitions
See : https://sourceforge.net/p/ntfs-3g/mailman/message/37343528/

### DIFF
--- a/components/library/libfuse/patches/mount_util.c.patch
+++ b/components/library/libfuse/patches/mount_util.c.patch
@@ -1,5 +1,5 @@
---- libfuse/mount_util.c.orig	2017-12-06 22:17:00.200753628 +0000
-+++ libfuse/mount_util.c	2017-12-06 22:16:13.152587853 +0000
+--- mount_util.c.ref	2010-06-15 16:46:54.000000000 +0200
++++ mount_util.c	2016-05-03 21:36:52.782854200 +0200
 @@ -58,6 +58,7 @@
          return -1;
      }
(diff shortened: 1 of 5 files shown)


jpandre:
Wanted to add I kept "BUILD_BITS=32" in the Makefile, which was inserted by Alexander Pyhalov on Apr 14, 2020 for some reason unknown to me.

pyhalov:
ntfsprogs were initially 32-bit. Perhaps, it's fine to build them 64 and 32-bit (as the deliver library). If library is unused currenrly, perhaps, they can be built 64-bit only. Anyway, switching to 64-bit ntfsprogs needs testing.

jpandre:
Being all in user space, it does not really matter whether 32 or 64-bit, and I have tested both. I personally use 64-bit mode.

AndWac:
Can you give me a hint when you are done with updating the PR?

jpandre:
Hi Andreas,

I'm done now.

Actually, it was not my intention to merge libfuse into the ntfs-3g update.
However these are related, as OpenIndiana ships ntfs-3g with the external (unsupported) libfuse.

Sorry for the confusion.

Jean-Pierre

Andreas Wacknitz wrote on 9/1/21 5:49 PM:

AndWac:
Your patches against libfuse won't work as expected. You'll need to separate them from this PR and also increment libfuse's COMPONENT_REVISION.

AndWac:
Furtermore I recommend to work on a separate branch. Otherwise you'll sooner or later get in trouble.

PR #6971, opened 2021-08-06
cheroot: bump to 8.5.2 and drop -27

alarcher:

PR #7036, opened 2021-09-02
cheroot: update to 8.5.2 and add python-3.7 and 3.9 versions

AndWac:


AndWac:
This PR is based on #6971 and replaces it.

PR #7039, opened 2021-09-02
Generic packages should depend at least on one versioned package

pyhalov:


AndWac (approved):

Toasterson:
@pyhalov Can you quickly document here how it is setup right now and how the dependencies look right now? My sleepy brain keeps insisting, that we could run into a dependency cycle if we are not careful. But I can't remember How it is setup currently. This is to fix the missing python deps thing right?

AndWac:
Ok, wer macht‘s?

pyhalov:
> @pyhalov Can you quickly document here how it is setup right now and how the dependencies look right now? My sleepy brain keeps insisting, that we could run into a dependency cycle if we are not careful. But I can't remember How it is setup currently. This is to fix the missing python deps thing right?

Currently from GENFRAG manifests common packages are generated, which conditionally depend on -27, -35 packages and so on. If python-27 is installed, -27 package is installed and so on. The issue is that when PYTHON_VERSIONS is restricted, corresponding python can be missing and so installing  (or updating) common package will not lead to installing one of the -27, -35,... packages. I try to fix this specifying that common package depends on at least one of the python-specific packages.

Toasterson:
Ah, yeah that sounds reasonable. Just out of curiosity how does the output of the manifests change exactly?

Otherwise LGTM

Toasterson (approved):

pyhalov:
> Ah, yeah that sounds reasonable. Just out of curiosity how does the output of the manifests change exactly?
> 
> Otherwise LGTM

For example, now `backup/borg` gets `depend fmri=backup/borg-39@1.1.17,5.11-2020.0.1.1 type=require-any` dependency in addition to `depend fmri=backup/borg-39@1.1.17,5.11-2020.0.1.1 predicate=runtime/python-39 type=conditional`, and `kafka-python` - `depend fmri=library/python/kafka-python-27@2.0.1,5.11-2020.0.1.0 fmri=library/python/kafka-python-35@2.0.1,5.11-2020.0.1.0 type=require-any` in addition to 
```
depend fmri=library/python/kafka-python-27@2.0.1,5.11-2020.0.1.0 predicate=runtime/python-27 type=conditional
depend fmri=library/python/kafka-python-35@2.0.1,5.11-2020.0.1.0 predicate=runtime/python-35 type=conditional
```

AndWac:
I wonder if and how we can document this properly.

pyhalov:
Likely now PYVER components should be rebuilt.

Toasterson:
I agree we should rebuild the PYVER packages. 

Lovely changes in general good to see them.

PR #7083, opened 2021-09-15
Add missing mdl dependencies

pyhalov:


pyhalov:
I know nothing about ruby, but these changes seems to be enough to fix our mdl.  Without this it fails with lots of missing dependencies.

components/ruby/mdl/Makefile
```diff
@@ -15,6 +15,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		mdl
 COMPONENT_VERSION=	0.4.0
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	Markdown lint tool
```
AndWac, line 19: I think you'll need to put this in a separate PR.

pyhalov: Why?

AndWac: Because tt requires the newly introduced packages:

depend type=require fmri=library/ruby/kramdown
depend type=require fmri=library/ruby/kramdown-parser-gfm
depend type=require fmri=library/ruby/mixlib-cli
depend type=require fmri=library/ruby/mixlib-config

pyhalov: No, as it's runtime dependency, not build-time.

AndWac (approved):

PR #7091, opened 2021-09-17
Introduce infrastructure for several Ruby versions

pyhalov:


pyhalov:
Wanted to work on vagrant prerequirements, but so far it seems I've managed only to prepare Ruby build infrastructure to support multiple packages built from one source (like Python one). I've tested that it works for simple cases, but of course it could be incomplete...

PR #7113, opened 2021-09-24
gcc-10: add andy's patch for thread-local errno to sync up with -il1

richlowe:
This adds andy's patch and syncs up with -il1, but doesn't change how we do the build to take from the illumos tarball.
I also didn't sync up the test results because it appears that they're locale sensitive and I didn't want to screw things up for jenkins
(but perhaps not doing it screws them up too?)

### DIFF
--- a/components/developer/gcc-10/Makefile
+++ b/components/developer/gcc-10/Makefile
@@ -18,7 +18,7 @@ BUILD_BITS=64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_VERSION= 10.3.0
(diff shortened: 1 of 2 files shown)


Toasterson:
Thanks for the assist. You need to bump COMPONENT_REVISION for it to build properly though. Can you do that?

Toasterson:
Some notes to test

https://github.com/illumos/gcc/pull/44#issuecomment-913049551201

Toasterson (approved):

AndWac (approved):

PR #7122, opened 2021-09-26
freerdp: update to 2.4.0

mvwieringen:


components/desktop/freerdp/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2020 <contributor>
+# Copyright 2021 <contributor>
 #
```
AndWac: Please remove the sample-manifest from the PR. It's only needed if something has changed. Change of the copyright year is not sufficient.

AndWac (approved):

PR #7125, opened 2021-09-26
wireshark: update to 3.2.16

mvwieringen:
Tried this before and then it go next to no traction but I think it makes sense to upgrade to a somewhat more modern version of wireshark. This uses qt5 instead of gtk for its interface as wireshark has moved on. Some stuff comes from a previous version of Solaris userland hence the updated copyrights in the Makefile. wireshark 3.4 is the next target but that has some extra dependencies and last time I couldn't get it to compile. This compiles and runs ok I can capture data and its decoding packets. 

I think this might help others but am fine to just keep this to myself and don't upstream any further packages.


AndWac (approved):

AndWac:
Can you add the test results in you future work? It should be easy to add the needed things for our test machinery.

mvwieringen:
> Can you add the test results in you future work? It should be easy to add the needed things for our test machinery.

Could you elaborate exactly what you are after ? I may push 2 update pull requests to bring things to 3.4.8 which seems
to compile correctly nowadays (were not last time I tried but seems Oracle upstreamed most changes.)

AndWac:
Am 9/27/21 um 11:01 AM schrieb Marco van Wieringen:
>
>     Can you add the test results in you future work? It should be easy
>     to add the needed things for our test machinery.
>
You can read about it in oi-userland/doc/testing.txt

mvwieringen:
Ok clear but the upstream Oracle userland has no tests so I have to dive in and see how and if there is some test framework for wireshark (probably based on tshark). For now this was low hanging fruit bringing things a few years upfront and somewhat closer to the upstream versions. For now this is for the moment where I can spend my time.

PR #7128, opened 2021-09-27
wireshark: update to 3.4.8

mvwieringen:
Didn't include updated header files as we don't distribute them. Upgrade from 3.2.16 is limited but 3.4.8 is the new stable version and 3.2.16 was the previous stable.


AndWac:
On my build machine library/opus is installed (as is on our official build server) and will be used by wireshark.
This leads to the following error during gmake publish:
/oibuild/oi-userland/components/sysutils/wireshark/build/manifest-i386-wireshark-common.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libopus.so.0 \
        pkg.debug.depend.reason=usr/lib/amd64/wireshark/plugins/3.4/codecs/opus_dec.so \
        pkg.debug.depend.type=elf \
gmake: *** [/oibuild/oi-userland/make-rules/ips.mk:502: /oibuild/oi-userland/components/sysutils/wireshark/build/.resolved-i386] Error 1

So, please either add REQUIRED_PACKAGES += library/opus or explicitly disable its use in the Makefile.

mvwieringen:
Strange have that also on my fresh build zone. And publishing was fine there last week and installed also fine on my prod server install and running fine there. Let me see what extra is needed in manifests etc will take some time to fully rebuild everything.

mvwieringen:
Ok rebased and pushed.

AndWac (approved):

PR #7173, opened 2021-10-14
Updating ca-certificates to 3.71

Toasterson:
This updates ca-certificates to 3.71

This also removes the openssl-0.9.8 compat links. If we need these we can also update the openssl-.0.9.8 file to remove the now removed certs.

If we need to regenerate the 0.9.8 hashes then we will need to document how to actually do that.


Toasterson:
@richlowe pinging you CC

components/openindiana/ca-certificates/ca-certificates.p5m
```diff
@@ -195,40 +186,30 @@ link path=etc/openssl/certs/064e0aa9.0 \
...
-link path=etc/openssl/certs/157753a5.0 \
-    target=../../certs/CA/AddTrust_External_Root.pem
 link path=etc/openssl/certs/1636090b.0 \
     target=../../certs/CA/Hellenic_Academic_and_Research_Institutions_RootCA_2011.pem
 link path=etc/openssl/certs/18856ac4.0 \
     target=../../certs/CA/SecureSign_RootCA11.pem
 link path=etc/openssl/certs/1d3472b9.0 \
     target=../../certs/CA/GlobalSign_ECC_Root_CA_-_R5.pem
-link path=etc/openssl/certs/1e08bfd1.0 \
+link path=etc/openssl/certs/1e08bfd$(IPS_COMPONENT_VERSION) \
```
richlowe: See, this is the crud the userland transforms did to you that you probably want to remove.

richlowe (commented):
Just the thing I mentioned on IRC I'm not sure how OI usually deals with, except it seemed like it hadn't been allowed into the manifest before.

@jclulow did you want to look too?

components/openindiana/ca-certificates/ca-certificates.p5m
```diff
@@ -41,14 +41,13 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 # We want to call svc:/system/ca-certificates everytime we modify /etc/certs/CA somehow
 <transform file path=etc/certs/CA/.* -> add refresh_fmri "svc:/system/ca-certificates:default">
 
-<include openssl-0.9.8-compat-links.inc>
```
richlowe, line 44: Oh, and if doing this.  Remove the include file too?

Toasterson: Definitely, but I am not going to remove it now if we need it. I want to wait for the discussion to settle on "We don't need it" before I remove it. Otherwise I need got foo to restore it.

components/openindiana/ca-certificates/ca-certificates.p5m
Toasterson: Yeah I noticed. I think it's a too general replacement script we do in the generation of the manifest. We want to replace certain variables but it's a bit odd replacing this one.

components/openindiana/ca-certificates/Makefile
```diff
@@ -12,21 +12,22 @@
...
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= ca-certificates
 COMPONENT_VERSION_MAJOR=3
-COMPONENT_VERSION_MINOR=45
+COMPONENT_VERSION_MINOR=71
 COMPONENT_VERSION=	$(COMPONENT_VERSION_MAJOR).$(COMPONENT_VERSION_MINOR)
 IPS_COMPONENT_VERSION=	1.0
-COMPONENT_REVISION=	7
+COMPONENT_REVISION=	8
```
AndWac: Why do we need IPS_COMPONENT_VERSION stick at 1.0? If we could use the ordinary COMPONENT_VERSION the problem with arbitrarily replacements of 1.0 in the manifest would vanish. And using the real version information of the origin would also be more appropriate.

Toasterson: :shrug: I did not want to replace it as I have no way to test for eventual breakages. But I agree with you. I can go ahead and setup that change.

Toasterson: Done

components/openindiana/ca-certificates/ca-certificates.p5m
Toasterson: The last change also resolves this as we don't have 1.0 as IPS_COMPONENT_VERSION anymore.

Toasterson:
I am going to test https://www.illumos.org/issues/11625 in a Vagrantbox and then we can merge @olbohlen

PR #7224, opened 2021-10-23
chezmoi: add version 2.7.3

torculus:
Adding `chezmoi`, a utility to securely synchronize configuration files (aka "dotfiles") across multiple OSes

The main developer has worked specifically to support illumos (see https://github.com/twpayne/chezmoi/releases/tag/v2.7.3)

### DIFF
--- a/components/sysutils/chezmoi/chezmoi.p5m
+++ b/components/sysutils/chezmoi/chezmoi.p5m
@@ -1,0 +1,25 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 Benjamin S. Osenbach
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/chezmoi
--- a/components/sysutils/chezmoi/manifests/sample-manifest.p5m
+++ b/components/sysutils/chezmoi/manifests/sample-manifest.p5m
@@ -1,0 +1,25 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 <contributor>
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/chezmoi
--- a/components/sysutils/chezmoi/patches/01-only-build-illumos.patch
+++ b/components/sysutils/chezmoi/patches/01-only-build-illumos.patch
@@ -1,0 +1,25 @@
+#Chezmoi by default uses git to find the version it is built from. This does not work as we download a tarball copy from Github and build inside oi-userland, thus we remove the version embedding here
+--- chezmoi-2.7.3/Makefile.orig	2021-10-22 15:34:17.000000000 +0000
++++ chezmoi-2.7.3/Makefile	2021-10-29 09:59:04.758384704 +0000
+@@ -2,17 +2,15 @@
+ GOLANGCI_LINT_VERSION=$(shell grep GOLANGCI_LINT_VERSION: .github/workflows/main.yml | awk '{ print $$2 }')
+ 
+ .PHONY: default
+-default: run build test lint format
++default: build
+ 
+ .PHONY: install
+ install:
+-	go install -ldflags "-X main.version=$(shell git describe --abbrev=0 --tags) \
+-		-X main.commit=$(shell git rev-parse HEAD) \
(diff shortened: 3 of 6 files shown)


torculus:
@Toasterson , here's hoping you can be a reviewer

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
```diff
@@ -0,0 +1,21 @@
...
+-default: run build test lint format
++default: run
+ 
+ .PHONY: install
+ install:
+-	go install -ldflags "-X main.version=$(shell git describe --abbrev=0 --tags) \
+-		-X main.commit=$(shell git rev-parse HEAD) \
+-		-X main.date=$(shell date -u +%Y-%m-%dT%H:%M:%SZ) \
+-		-X main.builtBy=source"
++	GOOS=illumos GOARCH=amd64 ${GO} build -o /usr/bin
```
torculus: This piece needs to be changed before the pull request can be merged, since the package manifest does not list `/usr/bin` as an install location (and the documentation should also probably be put somewhere). However, every time I tried `go install`, it placed the files in my go working directory instead of `/usr/bin`. So I'm not really sure how to fix it on my own

components/sysutils/chezmoi/chezmoi.p5m
```diff
@@ -0,0 +1,24 @@
...
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
```
Toasterson, line 24: There must be a file action here otherwise the install fails.

Toasterson: Or rather it did not install properly into the proto area.

components/sysutils/chezmoi/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,24 @@
...
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
```
Toasterson, line 24: Same as the other manifest. If no file action is generated the package installs nothing

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
Toasterson: Yeah this looks like the source of the problem we are having with the manifests. For oi-userland you must not use /usr/bin hardcoded as GOBIN. Rather in the Makefile of the component or a golang.mk set `GOBIN=$PROTO_DIR/usr/bin` This makefile should pass the environment variables along. This patch must go away. Then go install should work. (Cannot test right now but from knowledge it's sound)

Toasterson: Also re run `gmake sample-manifest` after changing this.

Toasterson (changes requested):
This needs changes. Right now it does not install chezmoi but rather an empty package. The main thing will be to set GOBIN to the appropriate directory under `$PROTO_DIR` so go install installs there.

torculus:
So, I have set
```
COMPONENT_BUILD_ENV += GOOS="illumos"
COMPONENT_BUILD_ENV += GOPATH="$(SOURCE_DIR)/gopath"
COMPONENT_BUILD_ENV += GOBIN="$(PROTO_DIR)/usr/bin"

COMPONENT_INSTALL_ENV += GOPATH="$(SOURCE_DIR)/gopath"
COMPONENT_INSTALL_ENV += GOBIN="$(PROTO_DIR)/usr/bin"
```
in my Makefile, and `go` is not recognizing any of them (I've checked with `echo $GOBIN`, etc.). There also is no such thing as `golang.mk` in the make-rules directory

components/sysutils/chezmoi/chezmoi.p5m
torculus: Okay, I have resolved the issues and managed to build and install a working package. Please see my latest commit

Toasterson: Awesome!

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
```diff
@@ -0,0 +1,24 @@
...
+-default: run build test lint format
++default: build
+ 
+ .PHONY: install
+ install:
+-	go install -ldflags "-X main.version=$(shell git describe --abbrev=0 --tags) \
+-		-X main.commit=$(shell git rev-parse HEAD) \
+-		-X main.date=$(shell date -u +%Y-%m-%dT%H:%M:%SZ) \
+-		-X main.builtBy=source"
++	go install
```
Toasterson, line 17: Is there a reason we skip backing in the metadata into the binary?

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
```diff
@@ -0,0 +1,24 @@
...
+ 
+ .PHONY: build
+-build: build-darwin build-freebsd build-linux build-windows
++build:
++	GOARCH=amd64 ${GO} build -o /dev/null .
```
Toasterson, line 22: Can this not be done inside the Makefile of the component?

Toasterson (changes requested):
If we can not get away without a patch file, please write as a comment to the .patch file why we need it so in later updates we know if we can ditch it or not.

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
torculus: Yes. For some reason, including the metadata results in an error that the binary is "not in dotted-tri format". Just using `go install` without metadata resolves this issue

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
torculus: So in the component Makefile, I put
```
COMPONENT_BUILD_ACTION=(go build)
COMPONENT_INSTALL_ACTION=(go install)
```
but then it says `/bin/bash: line 1: ((: go build: syntax error in expression (error token is "build")`

torculus:
> If we can not get away without a patch file, please write as a comment to the .patch file why we need it so in later updates we know if we can ditch it or not.

I cannot seem to get away without it, so I have added a comment to the top of the .patch file

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
```diff
@@ -0,0 +1,25 @@
+#This patch file exists because the source code's Makefile does not build for illumos by default
```
Toasterson: Update the text as following.

`Chezmoi by default uses git to find the version it is built from. This does not work as we download a tarball copy from Github and build inside oi-userland, thus we remove the version embedding here`

Toasterson (commented):
Ok, last change to the patch then we can merge.

Toasterson (approved):

torculus:
Okay, the comment has been changed, and my branch has been rebased, squashed, and is ready for merge

components/sysutils/chezmoi/patches/01-only-build-illumos.patch
torculus: Complete

Toasterson (approved):

Toasterson:
@torculus Merged. Thank you for your work

PR #7232, opened 2021-10-28
gstreamer1: update to 1.18.5

timmooney:
Update gstreamer1 from 1.16.2 to 1.18.5 (current latest).

Changes include converting from autotools to meson for the build system and new test transforms to deal with the change to the test output.

The ABI tracker seems to have quit working, as there's no ABI compatibility info available for comparing 1.16.2 to 1.18.5.  I've been able to use rhythmbox 3.4.4 without a rebuild without any issues.

Updates for all the gstreamer1 plugins will follow.


AndWac:
This doesn't build here:
[318/434] Compiling C object tests/check/gst_gstutils.p/gst_gstutils.c.o
FAILED: tests/check/gst_gstutils.p/gst_gstutils.c.o 
/usr/gcc/7/bin/gcc -Itests/check/gst_gstutils.p -Itests/check -I../../gstreamer-1.18.5/tests/check -I. -I../../gstreamer-1.18.5 -Ilibs -I../../gstreamer-1.18.5/libs -Igst -Ilibs/gst/controller -I/usr/lib/libffi-3.2.1/include -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -I/usr/include/pcre -I/usr/include/gio-unix-2.0 -I/usr/X11/include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -O3 -fvisibility=hidden -fno-strict-aliasing -DG_DISABLE_CAST_CHECKS -Wmissing-declarations -Wmissing-prototypes -Wredundant-decls -Wundef -Wwrite-strings -Wformat -Wformat-nonliteral -Wformat-security -Wold-style-definition -Winit-self -Wmissing-include-dirs -Waddress -Waggregate-return -Wno-multichar -Wdeclaration-after-statement -Wvla -Wpointer-arith -m32 -O2 -m32 -pthread -DFFI_NO_RAW_API -DHAVE_CONFIG_H -UG_DISABLE_ASSERT -UG_DISABLE_CAST_CHECKS '-DGST_CHECK_TEST_ENVIRONMENT_BEACON="GST_STATE_IGNORE_ELEMENTS"' '-DTESTFILE="/oibuild/oi-userland/components/multimedia/gstreamer1/gstreamer-1.18.5/tests/check/meson.build"' -DGST_DISABLE_DEPRECATED -MD -MQ tests/check/gst_gstutils.p/gst_gstutils.c.o -MF tests/check/gst_gstutils.p/gst_gstutils.c.o.d -o tests/check/gst_gstutils.p/gst_gstutils.c.o -c ../../gstreamer-1.18.5/tests/check/gst/gstutils.c
../../gstreamer-1.18.5/tests/check/gst/gstutils.c:770:10: fatal error: gmp.h: No such file or directory
 #include <gmp.h>
          ^~~~~~~
compilation terminated.

timmooney:
Interesting.  I found an unlisted dependency for one of the plugins and added it to the build dependencies, but I must have had library/gmp installed for some other dependency.

Do you want me to add library/gmp as a manual build dependency and redo my git push for this?

AndWac:
It wouldn't be enough to manually add the build dependency. The error indicates that you are also missing an appropriate include path, eg. CFLAGS += -I/usr/include/gmp because the compiler is searching in the standard paths and doesn't finde gmp.h.
If it would have found the gmp.h file we would get an error at the ips dependency check in a later step.

timmooney:
Does meson's setup say:

```
Library gmp found: YES
```

?

So that's one of the tests that's failing.  Interestingly, I get this message from the linker for that:

```
[318/434] Linking target tests/check/gst_gsturi
[319/434] Linking target tests/check/gst_gstutils
ld: warning: symbol 'times' has differing types:
    (file tests/check/gst_gstutils.p/gst_gstutils.c.o type=OBJT; file /lib/libc.so type=FUNC);
    tests/check/gst_gstutils.p/gst_gstutils.c.o definition taken
ld: warning: symbol 'times' has differing types:
    (file tests/check/gst_gstutils.p/gst_gstutils.c.o type=OBJT; file /lib/libc.so type=FUNC);
```

I don't know why you're seeing different results than me.  This is a very recently installed build-box.  I haven't installed anything other than the dependencies that come by way of 'gmake setup', 'gmake env-prep', and the packages themselves.

This box **was** installed from the Live image, though, and it has a full GUI.  I needed that to be able to test gstreamer and things like rhythmbox, etc.

AndWac:
Our meta data is lacking. Thus, not all packages have all required packages right yet.
My build system is older and has already build all available packages. So, it has almost all available packages installed and obviously gstreamer1 detects gmp if it's already installed. You can install it manually and that probably will give you similar results like mine. My system should be similar enough to out official build server that is being used for years and thus has also many packages installed.

timmooney:
The line I included shows that meson **found** gmp on my system.   That's presumably why I didn't have the build problem; it used pkgconfig to automatically add the necessary CFLAGS/LDFLAGS etc.

Did meson find gmp on yours?

timmooney:
Does the behavior change for you if you re-add the

```
unexport SHELLOPTS
```

to the Makefile?   I removed that from gstreamer1's Makefile, but that was mostly unintentional.   I had intended to put it back before pushing my commit, but I see that didn't happen.

AndWac:
unexport SHELLOPTS is only needed to run tests successfully

AndWac:
Library gmp found: YES

AndWac:
Simply change the Makefile:
PATH = $(PATH.gnu)                                                                                                                   
                                                                                                                                     
CFLAGS += -I/usr/X11/include                                                                                                         
CFLAGS += -I/usr/include/gmp                                                                                                                                                                       
                                                                                                                                     
I have only added the 2nd CFLAGS line. You might have copied or moved gmp.h into /usr/include and forgot about it.

timmooney:
>  You might have copied or moved gmp.h into /usr/include and forgot about it.

Ummm, no.  I would **never** do that and I sure as heck wouldn't have forgotten about it if I had done so!

Andreas, my build box is brand new, as in created in the last 10 days.  Isn't it possible that the problem is some difference on your build box?  Perhaps something wrong with you gmp.pc file, if meson is finding 'gmp.pc' but not adding '-I/usr/include/gmp' to CFLAGS automatically once it has done so?  Or you have an old, extra copy of gmp.pc somewhere?

AndWac:
Our gmp.h is at /usr/include/gmp/gmp.h. Your build seems to find it without explicitly telling where to search for it (the code uses \#include <gmp.h>). How so? It's not being found on my system and exhibits the error and I expect that our build server will behave similarly. Thus, I expect that we'll need the additional CFLAGS+= -I/usr/include/gmp. We have some other packages with similar problems because Linux distros usually put header files directly under /usr/include.

timmooney:
> Our gmp.h is at /usr/include/gmp/gmp.h. 

Yes.

> Your build seems to find it without explicitly telling where to search for it (the code uses #include <gmp.h>).

No, it doesn't.  Look at your failed compilation output.  The failure is coming at line 770 in gst_gstutils.c.  Looking at the code there:


```
#ifdef HAVE_GSL
#ifdef HAVE_GMP

#include <gsl/gsl_rng.h>
#include <gmp.h>

```

I have GMP installed, **but** I don't have GSL installed.  Like I said, this is a new build box.   The only stuff I have installed is the stuff that 'gmake setup' pulled in and that was listed as a dependency for gstreamer1.

Your older build box probably does have GSL installed.

That section of code that's failing for you isn't even being included for me, so gst_gstutils.c builds just fine for me.  It's not failing because of something being "wrong" with my build box, it's failing because of additional packages you have installed, that I do not.

> How so? It's not being found on my system and exhibits the error and I expect that our build server will behave similarly. 

If that's how the build system behaves, then yes, the same issue will probably happen there.

It seems like if the build system has "everything under the sun" installed for every package build, then "gmake setup" should probably install a lot more stuff as part of preparing a new build system for a developer?  If the goal is to have reproducible builds, which I definitely would want, then having a build system set up very similar to the system that actually generates the official packages seems like the best approach.

> Thus, I expect that we'll need the additional CFLAGS+= -I/usr/include/gmp. We have some other packages with similar problems because Linux distros usually put header files directly under /usr/include.

Now that we know why you're seeing the failure and I am not, it seems like two work-arounds are needed:

1. The CFLAGS change that you have proposed should be added.
2. both library/gmp and developer/gsl need to be manually added to the REQUIRED_PACKAGES.  They will never be auto-detected because they're only needed for the test suite, not the installed software, and we need to guarantee that **both** of them are installed on a build box,

Do you agree?

AndWac:
Am 10/29/21 um 10:20 PM schrieb Tim Mooney:
>
>     Our gmp.h is at /usr/include/gmp/gmp.h.
>
> Yes.
>
>     Your build seems to find it without explicitly telling where to
>     search for it (the code uses #include <gmp.h>).
>
> No, it doesn't. Look at your failed compilation output. The failure is
> coming at line 770 in gst_gstutils.c. Looking at the code there:
>
> |#ifdef HAVE_GSL #ifdef HAVE_GMP #include <gsl/gsl_rng.h> #include
> <gmp.h> |
>
> I have GMP installed, *but* I don't have GSL installed. Like I said,
> this is a new build box. The only stuff I have installed is the stuff
> that 'gmake setup' pulled in and that was listed as a dependency for
> gstreamer1.
>
> Your older build box probably does have GSL installed.
>
> That section of code that's failing for you isn't even being included
> for me, so gst_gstutils.c builds just fine for me. It's not failing
> because of something being "wrong" with my build box, it's failing
> because of additional packages you have installed, that I do not.
>
>     How so? It's not being found on my system and exhibits the error
>     and I expect that our build server will behave similarly.
>
> If that's how the build system behaves, then yes, the same issue will
> probably happen there.
>
> It seems like if the build system has "everything under the sun"
> installed for every package build, then "gmake setup" should probably
> install a lot more stuff as part of preparing a new build system for a
> developer? If the goal is to have reproducible builds, which I
> definitely would want, then having a build system set up very similar
> to the system that actually generates the official packages seems like
> the best approach.
>
>     Thus, I expect that we'll need the additional CFLAGS+=
>     -I/usr/include/gmp. We have some other packages with similar
>     problems because Linux distros usually put header files directly
>     under /usr/include.
>
> Now that we know why you're seeing the failure and I am not, it seems
> like two work-arounds are needed:
>
>  1. The CFLAGS change that you have proposed should be added.
>  2. both library/gmp and developer/gsl need to be manually added to
>     the REQUIRED_PACKAGES. They will never be auto-detected because
>     they're only needed for the test suite, not the installed
>     software, and we need to guarantee that *both* of them are
>     installed on a build box,
>
> Do you agree?
>
Yes, you are right. Sorry that I didn't spend enough time to look as
close as you did. I have checked the official build server: it has
developer/gsl installed like my machine.

The build system installs every required package automatically and as it
has been used to build (probably) all packages (and many that already
have been obsoleted). So it is quite likely that it has more packages
installed than "younger" build systems. Alas our build system has some
weaknesses in its dependency detection  and many packages still have
incomplete meta data. Packages required only for tests usually won't be
found by the dependency resolver. I guess many of those test
dependencies might have been installed manually in the past.

So, you are right with both of your work arounds.

timmooney:
It's OK, my first assumption was wrong too: I thought that because it found gmp.pc it was "doing the right thing" and adding -I/usr/include/gmp to CFLAGS automatically and that's why my build was working.  The truth was more complicated, for sure.  I was just a little put off when you suggested that the discrepancy was because I had done something stupid like modifying header files on my build system.

I'll get both changes made and force-push my branch again, after I've had a chance to verify the changes don't introduce anything that breaks the test comparisons.  It will probably be a few hours yet, perhaps tomorrow.

timmooney:
Just pushed an update that adds the following changes over my previous commit:

1. add `CFLAGS += -I/usr/include/gmp` plus a comment about it
2. re-instate `unexport SHELLOPTS`, since I hadn't intended to remove it.
3. add `library/gmp` and `developer/gsl` as `REQUIRED_PACKAGES` with a comment about them being for the test suite only.

Even with the changes to the test suite, the comparison still succeeds, so no changes needed to `results-32.master` or `results-64.master`

AndWac:
Can you please remove the contents of the "old" folder? The files don't seem to be necessary anymore.

timmooney:
The "old" folder and the previous patches have been removed.

AndWac (approved):

PR #7233, opened 2021-10-28
gst-plugins-base1: update to 1.18.5

timmooney:
Update gst-plugins-base1 from 1.16.2 to 1.18.5 (current latest).

The big changes were converting from autotools to meson and updated test transforms to deal with the new test format.  There are a few tests that seem to fail non-deterministically, so I completely filter them from the test results.

I also had to specify several OpenGL-related parameters to minimize the inclusion of Qt5, as the qmake build steps require more work to respect the multi-ABI builds.  I judge it's not worth including Qt5 as a dependency here anyway.  It's very minimal functionality.

This also switches gst-plugins-base1 from depending upon libjpeg6 to now depending upon libjpeg8-turbo.


AndWac:
Please remove the "old" folder and its contents here, too.

timmooney:
`old` folder has been removed.  I didn't rebuild or republish, since nothing that's part of the component generation was touched.

PR #7234, opened 2021-10-28
gst-plugins-good1: update to 1.18.5

timmooney:
Update gst-plugins-good1 from 1.16.2 to 1.18.5

Same changes as mentioned with previous PRs: switch from autotools to meson and update the test transforms.

This picks up a new dependency (taglib).


components/multimedia/gst-plugins-good1/manifests/sample-manifest.p5m
```diff
@@ -136,6 +139,7 @@ file path=usr/lib/gstreamer-1.0/libgstsmpte.so
 file path=usr/lib/gstreamer-1.0/libgstsoup.so
 file path=usr/lib/gstreamer-1.0/libgstspectrum.so
 file path=usr/lib/gstreamer-1.0/libgstspeex.so
+file path=usr/lib/gstreamer-1.0/libgsttaglib.so
 file path=usr/lib/gstreamer-1.0/libgsttwolame.so
 file path=usr/lib/gstreamer-1.0/libgstudp.so
 file path=usr/lib/gstreamer-1.0/libgstvideo4linux2.so
```
AndWac, line 146: When I run gmake sample-manifest it will have two additional libs:
file path=usr/lib/gstreamer-1.0/libgstwavpack.so
file path=usr/lib/$(MACH64)/gstreamer-1.0/libgstwavpack.so

components/multimedia/gst-plugins-good1/Makefile
```diff
@@ -89,6 +103,7 @@ REQUIRED_PACKAGES += library/aalib
 REQUIRED_PACKAGES += library/audio/gstreamer1
```
AndWac, line 104: REQUIRED_PACKAGES += audio/wavpack

audio/wavpack is installed on our build server

timmooney:
Just pushed an update with

1. installed wavpack
2. verified rebuild
3. updated `results-32.master` and `results-64.master` since the # of tests have now changed
4. updated `manifests/sample-manifest.p5m` with the new actions
5. updated `gst-plugins-good1.p5m` with the new actions
6. re-ran `gmake REQUIRED_PACKAGES` and updated the `Makefile` with the new requirement
7. re-ran `gmake publish` and picked up the addition to `pkg5`
8. dropped the `old` directory

I believe this should be everything needed for the additional dependency you identified.

AndWac (approved):

PR #7235, opened 2021-10-28
gst-plugins-bad1: update to 1.18.5

timmooney:
Update gst-plugins-bad1 from 1.16.2 to 1.18.5.

Mostly the same updates as gstreamer1 and gst-plugins-base1 and gst-plugins-good1: switch from autotools to meson and update the test transforms and results.

Since this previously depended upon openssl, I switched it to depend upon `library/security/openssl-11` now.

The upstream dropped the `vdpau` plugin so `library/graphics/libvdpau` is no longer a dependency.

Also, OIs libdrm is too old (gst-plugins-bad1 now requires version 2.4.98, we're currently using 2.4.75) so the KMS module doesn't currently build.  I noted the change in the Makefile and in `gst-plugins-bad1.p5m`.


components/encumbered/gst-plugins-bad1/Makefile
```diff
@@ -99,7 +109,6 @@ REQUIRED_PACKAGES += library/audio/openal
 REQUIRED_PACKAGES += library/desktop/cairo
```
AndWac, line 110: REQUIRED_PACKAGES += library/audio/soundtouch

The soundtouch package is installed on the build server.
This will bring
file path=usr/lib/$(MACH64)/gstreamer-1.0/libgstsoundtouch.so
file path=usr/lib/gstreamer-1.0/libgstsoundtouch.so
to the manifests

timmooney:
I'll add soundtouch and update my commit in a few hours.

timmooney:
Updated commit has the following changes over the previous one:

1. installed `library/audio/soundtouch`
2. rebuilt with the new dependency in place
3. re-ran the test suite.  No new tests and no changes needed for the existing results files
4. ran `gmake sample-manifest` to pick up the new actions
5. updated `gst-plugins-bad1.p5m` with the new actions
6. ran `gmake REQUIRED_PACKAGES` to pick up the new dependency
7.  published and picked up the change to `pkg5`
8. dropped the `old` directory

I believe this takes care of the changes needed for the additional soundtouch dependency Andreas pointed out

AndWac (approved):

PR #7236, opened 2021-10-28
gst-plugins-ugly1: update to 1.18.5

timmooney:
Final component update in the batch for gstreamer1: gst-plugins-ugly1

Same main changes as the previous ones: switch to meson and update the test transforms.

This ended up being the easiest of all the updates.  No installed paths or dependencies changed, and all the package tests are still passing.

### DIFF
--- a/components/encumbered/gst-plugins-ugly1/Makefile
+++ b/components/encumbered/gst-plugins-ugly1/Makefile
@@ -12,34 +12,32 @@
 #
 # Copyright 2016-2018 Aurelien Larcher.  All rights reserved.
 # Copyright 2019 Michal Nowak
+# Copyright 2021 Tim Mooney.  All rights reserved.
 #
 
 BUILD_BITS= 32_and_64
+BUILD_STYLE=	meson
 
 include ../../../make-rules/shared-macros.mk
 
-COMPONENT_NAME= gst-plugins-ugly1
-COMPONENT_VERSION= 1.16.2
-COMPONENT_SUMMARY= GNOME streaming media framework plugins
+COMPONENT_NAME=		gst-plugins-ugly1
+COMPONENT_VERSION=	1.18.5
+COMPONENT_SUMMARY=	GNOME streaming media framework plugins
 COMPONENT_CLASSIFICATION= System/Multimedia Libraries
-COMPONENT_FMRI= library/audio/gstreamer1/plugin/ugly
-COMPONENT_SRC_NAME= gst-plugins-ugly
-COMPONENT_SRC= $(COMPONENT_SRC_NAME)-$(COMPONENT_VERSION)
-COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.xz
+COMPONENT_FMRI=		library/audio/gstreamer1/plugin/ugly
+COMPONENT_SRC_NAME=	gst-plugins-ugly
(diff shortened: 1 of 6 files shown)


AndWac:
What about library/audio/gstreamer1/plugin/libav? It's also in the encumbered area.

timmooney:
You're right, I noticed that after I finished up with gst-plugins-ugly1.

I started looking at gst-libav1 last night.  It requires several libav* libraries, but the REQUIRED_PACKAGES don't list them, so I don't know quite what was done in the past to get them.  Were they installed separately, or is the real dependency on `ffmpeg`, which uses them too?  I suspect it's ffmpeg, so I started looking at that last night.

I would like to update ffmpeg from our current 3.2.x to the latest 4.4.1, but although it builds quite easily, I have several questions about our package for it:

1. our current build includes OSS for both input and output devices, but it doesn't include pulseaudio.  Why not?  Is that intentional?    Is it a problem if I try enable pulseaudio for input and output too?
2. Our current 3.2.14 build of ffmpeg selects options that generates a binary that is "nonfree and unredistributable" according to the output from ffmpeg's configure.  Is that intentional?  Is it OK for us to package and distribute something that says it's "unredistributable"?
3. Where I spent the most time last night, and still don't understand the issue, is the test suite.  ffmpeg has an extensive test suite but I think our previous comparison results may have been for just the basic suite.  I need to spend some more time on that, but having answers to the first two questions I outlined would be very useful.

timmooney:
pushed a change to drop the `old` directory from this component too.  I did not rebuild or re-publish, since nothing else changed.

AndWac:
Alas I don't have definitive answers to your questions. We lack a lot of meta data, and as many original contributors left or are busy we cannot expect answers. I will try to answer your questions from my perspective:
1. I think that the omission is not itentionally. I guess that the inclusion of pulseaudio was driven by the fact that firefox needs it. So maybe the ffmpeg package predates pulseaudio in OI.
2. The ffmpeg package is most probably in encumbered (and the corresponding repository) because of these problems. I think we should check the concrete reasons and if the newer version adds problems.
3. The test suite is the least problematic thing. It a feature works sufficient well it's ok for me if tests fail. Of course we should try our best to bring the tests in an acceptable state but it should not block the merge of a package if that is not possible with a reasonable amount of work. We have many other packages with problematic tests or no tests at all.

components/encumbered/gst-plugins-ugly1/Makefile
```diff
@@ -12,67 +12,70 @@
...
-	'-e "/XPASS/p" ' \
-	'-e "/ERROR/p" '
+	'-e "s/[ 	]*[0-9][0-9]*\.[0-9][0-9]s//" ' \
+	'-e "s/[ 	]*exit status [0-9][0-9]*//" ' \
+	'-e "/^[ 	]*[0-9][0-9]*\/[0-9][0-9]* /p" ' \
+	'-e "/^Summary of/p" ' \
+	'-e "/^Ok:/,/^Timeout:/p" ' \
+	'-e "/^Full log written.*/,/^ninja: build stopped.*/d" '
+
+test: $(TEST_32_and_64)
```
AndWac: You shouldn't need to set the test target as it's the default when BUILD_BITS= 32_and_64

timmooney:
Updated commit that drops 'test' from the Makefile.

PR #7240, opened 2021-10-29
ca-certificates: remove expired DST Root CA X3

cstes:
Could be interesting as a temporary update until December 2021

https://bugzilla.mozilla.org/show_bug.cgi?id=1733560
https://bugzilla.mozilla.org/show_bug.cgi?id=1733003

The problem for Squeak Smalltalk is that the certificate of https://squeak.org is signed by DST Root CA X3 which expired in September 2021.

The article https://www.openssl.org/blog/blog/2021/09/13/LetsEncryptRootCertExpire/  blog explains that either openssl should be updated to 1.0.2.zb or some other workaround implemented because openssl 1.0.2 (unlike openssl 1.1 ?) prefers the "OpenSSL 1.0.2 which always prefers the untrusted chain"  instead of the trusted chain X1.

### DIFF
--- a/components/openindiana/ca-certificates/Makefile
+++ b/components/openindiana/ca-certificates/Makefile
@@ -13,6 +13,7 @@
 # Copyright 2016 Alexander Pyhalov
 # Copyright 2019 Michal Nowak
 # Copyright 2021 Till Wegmueller
+# Copyright 2021 David Stes
 #
 
 include ../../../make-rules/shared-macros.mk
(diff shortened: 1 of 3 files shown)


cstes:
See also the warning by Mozilla 

https://blog.mozilla.org/security/2021/05/10/beware-of-applications-misusing-root-stores/ 

Unclear what the conclusion could be: the following remark can be interpreted that this PR is ok because it filters out an expired certificate at the OpenIndiana manifest side.

Mozilla wrote:
_That said, this does seem like an instance of https://blog.mozilla.org/security/2021/05/10/beware-of-applications-misusing-root-stores/ where if the downstream consumer is incompatible with expired roots (e.g. due to path building), they should be filtering those out themselves. It’s totally reasonable to bump this forward, but it seems longer-term, this is an issue in how OpenIndiana consumes the store._

PR #7257, opened 2021-11-07
thunderbird: update 91.3.0

cgrzemba:
- add and comment some patches
	- remove mediator


components/mail/thunderbird/Makefile
```diff
@@ -27,16 +27,20 @@
...
 include ../../../make-rules/shared-macros.mk
 
+# unfortunatly I see -m32 in CFLAGS despite that BUILD_BITS = 64 is set
+BITS = 64
+
 COMPONENT_NAME =	thunderbird
-COMPONENT_VERSION =	78.13.0
+COMPONENT_VERSION =	91.3.0
+COMPONENT_REVISION = 1
 COMPONENT_SUMMARY = Mozilla Thunderbird Email Application
```
AndWac, line 38: COMPONENT_REVISION=1 is wrong

components/mail/thunderbird/Makefile
```diff
@@ -83,9 +90,12 @@ MOZCONFIG_DEBUG =	/dev/null
 MOZCONFIG_NON_DEBUG =	$(MOZCONFIG)
 endif
 
+COMPILER = gcc
+
```
AndWac: COMPILER = gcc is the default in shared-macros.mk and should not be necessary here.

components/mail/thunderbird/thunderbird.p5m
```diff
@@ -32,16 +32,9 @@ license thunderbird.license license="MPL v2.0"
 file files/thunderbird.desktop path=usr/share/applications/thunderbird.desktop
 file usr/lib/$(MACH64)/thunderbird/chrome/icons/default/default48.png path=usr/share/pixmaps/thunderbird-icon.png
 
-link path=usr/bin/thunderbird target=../lib/amd64/thunderbird/thunderbird mediator=thunderbird mediator-version=78
-
```
AndWac, line 36: You have removed too much - the link action needs to be kept, it's only the mediator attribute that has to be removed. We still need the link.

cgrzemba: have the link readded.

3eka: Version should be 91, right?

cgrzemba: we remove the mediator, because installation of different versions is not intended

PR #7259, opened 2021-11-08
new component dav1d, a fast, open AV1 decoder

timmooney:
This is a new component, a multimedia library 'dav1d' for the AV1 format.

The [Wikipedia article on AV1](https://en.wikipedia.org/wiki/AV1) has good information on both its openness and its adoption by many popular streaming services.

This is set to become a dependency for my next PR (update ffmpeg to 4.4.1) and for the rebuild of VLC required by the ffmpeg
update.

The AV1 codec is very open and this library is BSD 2-clause, so there should be no need for this to be in "encumbered".

I asked on the oi-dev mailing list a few days ago for direction about what belongs in codec/ and what belongs elsewhere and didn't get any responses, so I went ahead and put this in codec/dav1d  .  If you have objections to codec/dav1d, that's fine, I'll take direction for where you would prefer it.

### DIFF
--- a/components/encumbered/dav1d/Makefile
+++ b/components/encumbered/dav1d/Makefile
@@ -1,0 +1,57 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 (c) Tim Mooney. All rights reserved.
+#
+BUILD_BITS=		64_and_32
+BUILD_STYLE=		meson
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		dav1d
+COMPONENT_VERSION=	0.9.2
+COMPONENT_FMRI=		codec/dav1d
+COMPONENT_SUMMARY=	A fast, cross-platform AV1 decoder
+COMPONENT_CLASSIFICATION=System/Multimedia Libraries
+COMPONENT_PROJECT_URL=	https://code.videolan.org/videolan/dav1d
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL= \
+	$(COMPONENT_PROJECT_URL)/-/archive/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:59a5fc9cc5d8ea780ad71ede6d589ed33fb5179d87780dcf80a00ee854952935
+COMPONENT_LICENSE=	BSD 2-Clause Simplified License
+
+include $(WS_MAKE_RULES)/encumbered.mk
+include $(WS_MAKE_RULES)/common.mk
+
+CONFIGURE_OPTIONS += -Denable_asm=true
+
+unexport SHELLOPTS
+# (first) delete the timing information from any line
+# delete the "exit status N", it can cause issues too
+# print lines that start with " N/NNN"
+# print everything between "^Ok:" and "^Timeout:"
+# delete the ninja log-related stuff at the end
+COMPONENT_TEST_TRANSFORMS += \
+	'-n ' \
+	'-e "s/[ 	]*[0-9][0-9]*\.[0-9][0-9]s//" ' \
+	'-e "s/[ 	]*exit status [0-9][0-9]*//" ' \
+	'-e "/^[ 	]*[0-9][0-9]*\/[0-9][0-9]* /p" ' \
+	'-e "/^Summary of/p" ' \
+	'-e "/^Ok:/,/^Timeout:/p" ' \
+	'-e "/^Full log written.*/,/^ninja: build stopped.*/d" '
+
+# build requirement
+REQUIRED_PACKAGES += developer/assembler/nasm
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/math
--- a/components/encumbered/dav1d/dav1d.license
+++ b/components/encumbered/dav1d/dav1d.license
@@ -1,0 +1,23 @@
+Copyright © 2018-2019, VideoLAN and dav1d authors
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this
+   list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice,
+   this list of conditions and the following disclaimer in the documentation
+   and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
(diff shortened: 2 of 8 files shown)


AndWac:
Are you sure that this package doesn't need to be in encumbered? Did you check the patent situation? Similar packages are in encumbered...

timmooney:
The wikipedia article I linked has extensive information on the patent situation for AV1.   IANAL, but this seems to be one of the most open codecs in many years.

AndWac:
I prefer to be safe and move it to the encumbered repository. As I see it, all its consumers are already there. What I understand from the articles I read about it there are different parties involved with different interests and opinions.

components/multimedia/dav1d/Makefile
```diff
@@ -0,0 +1,57 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 (c) Tim Mooney. All rights reserved.
+#
+BUILD_BITS=		32_and_64
+BUILD_STYLE=		meson
```
AndWac: Change BUILD_BITS to
BUILD_BITS= 64_and_32
This will bring the 64-bit program(s( into /usr/bin and 32-bit ones into /usr/$(MACH32)/bin.
Then you can filter out the 32-binaries in the manifest file.
Background: We only want 64-bit binaries and try to get rid of 32-bit binaries wherever possible.
Maybe it's even simpler here as we don't have any existing consumers and we can also change ffmpeg to make it 64-bit only. We provide 32-bit libraries where necessary for backwards compatibility.

components/multimedia/dav1d/Makefile
```diff
@@ -0,0 +1,57 @@
...
+COMPONENT_PROJECT_URL=	https://code.videolan.org/videolan/dav1d
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL= \
+	$(COMPONENT_PROJECT_URL)/-/archive/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:59a5fc9cc5d8ea780ad71ede6d589ed33fb5179d87780dcf80a00ee854952935
+COMPONENT_LICENSE=	BSD 2-Clause Simplified License
+COMPONENT_LICENSE_FILE=	$(COMPONENT_NAME).license
+
```
AndWac: You don't need to set COMPONENT_LICENSE_FILE. It's the default in shared-macros.mk

components/multimedia/dav1d/Makefile
timmooney: I'm not opposed to switching ffmpeg to 64 bit only, and two of its dependencies (minidlna and vlc) appear to already be 64 bit only, but gst-libav1 would also need to switch to be 64 bit only.  I'm also not opposed to that, but it does leave us with the weird situation where most of gstreamer1 is 32 bit and 64 bit, but the libav stuff would be 64 bit only.

components/multimedia/dav1d/Makefile
timmooney: Fixed.  Removed `COMPONENT_LICENSE_FILE`

components/multimedia/dav1d/Makefile
timmooney: Partially fixed.  Switched it to `64_and_32`, rebuilt the `sample-manifest` and updated the `dav1d.p5m` file, skipping the one 32 bit binary.

Once you've weighed in on whether it's OK to have `gst-libav1` switch to be 64 bit only and be unlike the other `gstreamer1` components then I'll know if any additional changes are needed or the current commit addresses all concerns.

components/encumbered/dav1d/Makefile
```diff
@@ -0,0 +1,57 @@
...
+	$(COMPONENT_PROJECT_URL)/-/archive/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:59a5fc9cc5d8ea780ad71ede6d589ed33fb5179d87780dcf80a00ee854952935
+COMPONENT_LICENSE=	BSD 2-Clause Simplified License
+
+include $(WS_MAKE_RULES)/encumbered.mk
+include $(WS_MAKE_RULES)/common.mk
+
+CONFIGURE_OPTIONS += -Denable-asm=true
+
```
AndWac, line 36: Build fails:
meson.build:25:0: ERROR: Unknown options: "enable-asm"

You have misspelled the option (correct: enable_asm), but its default is true anyways.

timmooney: Thanks for pointing out the incorrect option name, I'll fix that.  However for me, it's only a warning and does not stop the build.  Do you have some additional "meson" flags in your environment that convert warnings to errors? 

Here's a small snippet from a build log where it warns, and then continues with the `setup` process:

```
The Meson build system
Version: 0.59.3
Source dir: /export/home/mooney/oi-userland/components/encumbered/dav1d/dav1d-0.9.2
Build dir: /export/home/mooney/oi-userland/components/encumbered/dav1d/build/i86
Build type: native build
WARNING: Unknown options: "enable-asm"
The value of new options can be set with:
meson setup <builddir> --reconfigure -Dnew_option=new_value ...
Project name: dav1d
Project version: 0.9.2
C compiler for the host machine: /usr/gcc/7/bin/gcc (gcc 7.5.0 "gcc (OpenIndiana 7.5.0-il-0) 7.5.0")
C linker for the host machine: /usr/gcc/7/bin/gcc ld.solaris 5.11-1.1771 (illumos)
Host machine cpu family: x86
Host machine cpu: i686
Run-time dependency threads found: YES
Checking for function "clock_gettime" : YES 
Library dl found: YES
Checking for function "dlsym" with dependency -ldl: YES 
```

The option name will be fixed in my next commit.

AndWac:
Am 09.11.21 um 08:05 schrieb Tim Mooney:
>
>
>
> In components/encumbered/dav1d/Makefile
> <https://github.com/OpenIndiana/oi-userland/pull/7259#discussion_r745335866>:
>
> > +COMPONENT_SUMMARY=	A fast, cross-platform AV1 decoder
> +COMPONENT_CLASSIFICATION=System/Multimedia Libraries
> +COMPONENT_PROJECT_URL=	https://code.videolan.org/videolan/dav1d
> +COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
> +COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
> +COMPONENT_ARCHIVE_URL= \
> +	$(COMPONENT_PROJECT_URL)/-/archive/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
> +COMPONENT_ARCHIVE_HASH=	\
> +	sha256:59a5fc9cc5d8ea780ad71ede6d589ed33fb5179d87780dcf80a00ee854952935
> +COMPONENT_LICENSE=	BSD 2-Clause Simplified License
> +
> +include $(WS_MAKE_RULES)/encumbered.mk
> +include $(WS_MAKE_RULES)/common.mk
> +
> +CONFIGURE_OPTIONS += -Denable-asm=true
> +
>
> Thanks for pointing out the incorrect option name, I'll fix that.
> However for me, it's only a warning and does not stop the build. Do
> you have some additional "meson" flags in your environment that
> convert warnings to errors?
>
I have an up-to-date build system with meson-0.60.1. Our official build
server updates itself after each successful jenkins invocation. It has
one frozen package, though: osnet-incorporation. This prevents for
automatic illumos-gate updates because they could render the machine
unusable and we have no control over it. Of course we have to check
certain userland packages thoroughly (eg. openssh). In case of an
inaccessible system we need to contact Krystal support because we don't
have console access to the hypervisor.

I recommend everyone to update their machines regularly, especially
before creating PR's.

timmooney:
Thanks Andreas.  I made sure I was fully up to date before I started with gstreamer1 and I've updated once since then, but I will make it a habit to update more frequently.

PR #7263, opened 2021-11-08
update gst-libav(1) to 1.18.5

timmooney:
Update gstreamer1's gst-libav to 1.18.5, to match the other gstreamer1 updates.

- switch from `configure` to `meson`
- the old package was missing metadata for the `video/ffmpeg` dependency (for all the libav* libraries), but `gmake REQUIRED_PACKAGES` picked that up for this update.
- none of the compression libraries (`compress/bzip2`, `compress/xz`, `library/zlib`) show up as direct dependencies any more.  I'm not sure if that's because the old package never really depended upon them or because of changes related to meson, but they are no longer included.
- fix the `COPYING.LIB` file, replacing it with the current `COPYING` file from the distribution.


timmooney:
Sorry, I accidentally had some extra stuff in the original commit, I believe I have amended it and my last force push should be the intended commit.

PR #7269, opened 2021-11-09
chezmoi: update to v2.7.5

torculus:
Updating `chezmoi`, a utility to synchronize configuration files (aka "dotfiles")

This update also trims down my old patch file so that more metadata can successfully be included. It should make the component easier to maintain in the future

### DIFF
--- a/components/sysutils/chezmoi/Makefile
+++ b/components/sysutils/chezmoi/Makefile
@@ -16,7 +16,7 @@ BUILD_STYLE=justmake
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=chezmoi
-COMPONENT_VERSION=2.7.3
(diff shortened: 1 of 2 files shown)


Toasterson (approved):
Nice work, I will have a testrun once I am out of the Plane again.  But I do not expect it to run into any trouble.

For the future patch, could we make an upstream issue about this? The Makefile is a bit to strongly bound to their CI to be useful for packaging in general. It should not assume to always be running in a git repo and not cross-build by default

torculus:
@Toasterson, should we just wait on the next point release from upstream before merging this? That will probably incorporate the changes in the Makefile that make the patch unnecessary

Toasterson:
We can also merge this and wait on the next release to remove the patch. I am for that.

Toasterson (approved):
LGTM

PR #7289, opened 2021-11-15
add editor/nano to the minimal install

timmooney:
Add `editor/nano` to the minimal system install.

This was discussed on the mailing list in early January of 2021.


AndWac:
You also need to increment the COMPONENT_REVISION!

timmooney:
Thanks Andreas.  My updated push increments COMPONENT_REVISION for `install-types`

PR #7293, opened 2021-11-18
Dnscrypt proxy rebuild

timmooney:
rebuild for libsodium update

- bump `COMPONENT_REVISION`
- set `BUILD_BITS=64`
- drop the 3 separate make-rules includes and replace with the include for `common.mk`
- publish target picked up the addition of `shell/ksh93`

### DIFF
--- a/components/network/dnscrypt-proxy/Makefile
+++ b/components/network/dnscrypt-proxy/Makefile
@@ -13,12 +13,12 @@
 # Copyright 2016 Alexander Pyhalov
 # Copyright 2017 Andreas Wacknitz
 #
-
+BUILD_BITS=	64
 include ../../../make-rules/shared-macros.mk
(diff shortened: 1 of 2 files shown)


AndWac:
This PR has commits from 2 commits: dnscrypt & libzmq.
You'll need to fix this as it collides with PR #7292

timmooney:
Sorry for the mixup Andreas.  I think I have this branch fixed up now.

I'm still not super good with git, especially when I'm actively juggling multiple branches.  I'm definitely getting more comfortable, but still have a ways to go.  I appreciate the coaching (and the patience).

AndWac:
No problem. You can make your life easier next time when you collect all simple update commits into a single PR.
Technically it's only necessary to split dependant updates.

PR #7301, opened 2021-11-19
squeak: add CameraPlugin, cog-spur: upgrade to 5.0.3101

cstes:
CameraPlugin for webcam in Smalltalk

http://www.squeaksource.com/WebCamMorph
http://www.squeaksource.com/CameraPlugin.html

part of old MIT Scratch eToys


AndWac:
As I wrote after your last PR: please split your PR into 3 separates as our build server fails when building all three related packages at once. It then needs manual intervention.

cstes:
Can you please accept it this time.  There was a misunderstanding.

Last time I submitted a PR containing 3 commits for 3 packages.  You said then indeed that you need 1 single commit.

So this time I submitted a PR with 1 single commit.  However , I understand now from your feedback that the problem is that there are 3 packages at the IPS level involved.

So next time I will submit 3 PR with each containing a single commit.

PR #7302, opened 2021-11-19
library/pcre2: Update to 10.39

richlowe:
This updates pcre2 to 10.39, which adds a feature (`PCRE2_MATCH_INVALID_UTF`) git 2.34 expects where pcre can ignore invalid utf-8.  This is the root cause of 'git grep' not working almost anywhere anymore since the 2.34 update.

I say I update it badly, because this introduces a test failure in pcre2 which seems to be because of some shell magic that isn't working out how they expect (I have not updated the baseline, so if you build/test you'll see the failure).

```
echo "---------------------------- Test 132 -----------------------------" >>testtrygrep
(cd $srcdir; $valgrind $vjs $pcre2grep -m1 -A3 '^match'; echo '---'; head -1) <$srcdir/testdata/grepinput >>testtrygrep 2>&1
echo "RC=$?" >>testtrygrep
```

This shows the first line of `grepinput` from the `head -1` rather than what the test expects.

Installing this package and then rebuilding git fixes for eg `git grep` in the oi-userland source root.


richlowe:
Also I was surprised pcre2 didn't seem to be constrained by `userland-incorporation`, but I have no idea if that's on purpose, nor how to fix it

Toasterson:
The thing with userland-incorporation seems odd. I would say that was a build problem from last time. Let's see what happens this time round.

PR #7303, opened 2021-11-20
Rebuilding git after pcre2 update

Toasterson:


Toasterson:
@richlowe

AndWac:
If rebuilding git was necessary after pcre2 update I would expect it to be necessary for all dependant packages:
antivirus/clamav-core
application/cluster/pacemaker
developer/versioning/git
library/accessibility/brltty
library/desktop/vte-291
library/ncurses
network/chat/weechat
terminal/tmux
web/wget

PR #7304, opened 2021-11-20
stunnel: update to 5.60 and switch to openssl-11

timmooney:
Update stunnel to current latest and switch to openssl-11

- ran `gmake update` to get the standard Makefile changes from that
- switch from `32_and_64` to `64_and_32`
- specify `COMPONENT_SUMMARY`, `COMPONENT_FMRI` and `COMPONENT_LICENSE` in the Makefile and update `stunnel.p5m` to reference them, instead of specifying them directly in that file
- replaced the reference to perl 5.16 with `PATH=$(PERL5BINDIR):$PATH`
- new version defaults to allowing config fiile inclusion from `/etc/stunnel/conf.d` so create that and include the dir in the `stunnel.p5m`
- added (currently local, may be upstreamed) patch to get the test suite working and changed the `test` target from `$(NO_TESTS)` to the default.  Include `test/results-32.master` and `test/results-64.master`
- added the `diagnostic/nmap-cli` dependency, required for the test suite
- updated the patch for `authpriv` syslog for the new version
- dropped the `stunnel-4.56-32_64.patch` patch for the interposer; that change was previously incorporated almost verbatim into upstream
- included a pre-release patch from upstream related to a deprecated `CRYPTO_set_id_callback` openssl call.  It won't be needed after this release, the fix will be part of 5.61.
- removed the `stunnel-4.56-CRYPTO_num_locks.patch`, that code has completely changed and locking is handled in a much different way for OpenSSL 1.1
- updated the stock `stunnel.conf` file to be the sample file that's generated for UNIX systems during the build.  Our previous default config file was for Windows and had windows-style line endings.  We install the generated sample file in two places, the config directory and the share directory (as an example).  Also included a patch `04-stunnel.conf-sample.in.patch` to make very minor customizations for OI related to cert CAs.
- dropped all of the `dir path=etc`, `dir path=usr/lib`, `dir path=usr/bin` etc. lines from `stunnel.p5m` for any directories that are shared system directories.
- commented out and noted that we're not including the `usr/bin/$(MACH32)` binaries

I haven't tested the service.  If you know anyone that uses the stunnel service on OI, it would be great to have them test the new version.


timmooney:
Forgot to mention it but I'm also working on updating `socat` and switching that to `openssl-11`, if possible, but I'm waiting for a new release that's coming from upstream that is supposed to fix some portability issues.  It might be a while for `socat`, since I don't know how soon that release will be.

components/sysutils/stunnel/Makefile
```diff
@@ -21,31 +21,32 @@
 
 #
 # Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
+# copyright (c) 2021, Tim Mooney.  All rights reserved.
 #
 
+BUILD_STYLE=configure
+BUILD_BITS=64_and_32
+USE_OPENSSL11=          yes
```
AndWac, line 28: We don't seem to have dependant packages. Thus, we can drop 32-bit completely.

timmooney:
Switched to `BUILD_BITS=64` and adjusted test results, sample-manifest, and `stunnel.p5m` for only building 64 bit.

components/sysutils/stunnel/Makefile
```diff
@@ -21,31 +21,32 @@
 
 #
 # Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
+# copyright (c) 2021, Tim Mooney.  All rights reserved.
 #
 
+BUILD_STYLE=            configure
+BUILD_BITS=             64
```
AndWac, line 27: BUILD_STYLE= configure is the default so you don't need to set it here.

timmooney: Dropped `BUILD_STYLE`.   It may have been inserted automatically by `gmake update`

PR #7314, opened 2021-11-22
chezmoi: update to 2.8.0

torculus:


torculus:
@Toasterson, I'd appreciate your help with this. You were part of the [discussion](https://github.com/twpayne/chezmoi/pull/1617) on upstream to change the makefile, and now the install target has changed from `go install` to `install -m 755 chezmoi "${DESTDIR}${PREFIX}/bin"`. The problem is that `gmake` either can't find or create the `prototype` directory in the build folder

The exact error message is `install: cannot create regular file '/root/oi-userland/components/sysutils/chezmoi/build/prototype/i386/usr/local/bin': No such file or directory`

Is this something better handled by opening an issue in upstream, or by adjusting the component Makefile?

Toasterson:
You need to pass PREFIX=/usr in the COMPONENT_BUILD_ENV thats a requirement with autotools too.

torculus:
Figured it out, `install` is not POSIX-compatible. I'll have it ready soon

torculus:
Okay, so I'm going to mark this as ready for review since I got it to build, but can't figure out how to do so without a (small) patch file. It turns out that `install -m 755 chezmoi ${DESTDIR}${PROTOTYPE}/bin"` in the upstream makefile does NOT create `${DESTDIR}${PROTOTYPE}/bin` with illumos' version of `install`, you need `install -d` to do that. Requesting a change in upstream might break compatibility across the other *nix systems it's built on

torculus:
Solved it! We can now build the component without a patch!

Toasterson (approved):
Ah, thought we may actually need GNU paths set. but yeah that solves it :)

PR #7319, opened 2021-11-24
WIP: new perl runtime 5.34

timmooney:
new perl version 5.34.0 (current latest stable).

builds (64 bit only), tests, and publishes, but can't be committed until I have all the perl module rebuilds ready.

The module rebuilds aren't quite as simple as bumping `COMPONENT_REVISION` and doing a `gmake publish`, since there will be a new package generated for each module for the perl-534 runtime.  There's also some cleanup and Makefile modernization I'll do for each module, and it's possible that I will be forced to update some modules (I'm preferring to do rebuilds of existing module versions where possible;  I would then come back and start tackling module updates at a later time).

There are multiple different approaches I could take with the module rebuilds, I'm looking for input on how you would like me to handle them.

### Option 1:

1. a pull request for the changes to `make-rules/makemaker.mk` and `make-ruiles/shared-macros.mk`
2. a separate pull request for each perl module

### Option 2:
One giant pull request for both the macro changes and all the modules, but the commits are separate within the pull request.  There would be one commit for the two files in `make-rules`, one commit for `Archive-Zip`, one commit for `DBI`, etc.

### Option 3:
One giant pull request with one giant (squashed) commit, covering the macros and every one of the modules.

### Option 4:
Something else I'm not thinking of?


timmooney:
@AndWac low priority question, but any preference on which option you would like me to pursue for the perl module rebuilds?

If I know which option you prefer from the ones I've outlined, I may be able to make some progress on this over the next few weeks.

AndWac:
I don't know enough about perl and how changes would affect areas of OI to give a qualified advice.
I am happy to support either way of integrating the updates. My preferred option would be where changes can be integrated in smaller portions and tested separately.

components/perl/perl-534/Makefile
```diff
@@ -0,0 +1,110 @@
...
+# Note this perl is 64-bit only.
+#
+
+# Override PERL_VERSION because when building a new version
+# of perl it probably isn't installed so shared-macros.mk
+# can't call it.
+
+COMPONENT_NAME=         perl
+COMPONENT_VERSION=      5.34.0
+COMPONENT_SUMMARY=      Perl 5.30
```
pyhalov: 5.34?

pyhalov:
Modules will not be built until PERL_VERSIONS is updated. Perl is usually done in several pieces - first interpreter, then modules and PERL_VERSIONS. Check mediators to ensure that new perl is not becoming default until it's tested.
Option 2 seems sane if there's no dependencies between modules.

components/perl/perl-534/Makefile
timmooney: Nice catch @pyhalov !  I'll fix that tomorrow when I have some focused work time.

It's a leftover from when I originally started the perl update, just before the pandemic started.  I was originally building 5.30.0.

Nice to see you're still checking in on OI once in a while.  I hope things are great with you!

klausz65: If perl architecture names do change with this new perl version, please also verify their consitency in:
oi-userland/tools/python/pkglint/userland.py
Thanks

timmooney:
Updated the commit with a few minor changes:

1. fix the `COMPONENT_SUMMARY` to include the correct version (thanks @pyhalov !)
2. add a small patch from @klausz65 that fixes a warning in the `Cwd` module.
3. Remove the pkglint override I had in the `perl-534.p5m` file to disable the "64 bit files in wrong location" lint.  That (like the incorrect `COMPONENT_SUMMARY`) was a holdover from an earlier set of diffs.  The pkglint override was no longer needed.

timmooney:
Regarding my earlier question about what approach to take for the module rebuilds, and especially related to @AndWac 's comments about allowing change integration in small portions, I propose using the following procedure:

### Call this "Option 5":

1. After this PR for perl 5.34.0 is merged, there's a second PR that updates `makemaker.mk` with the necessary .configured/.built/.installed/.tested etc. rules to add support for perl 5.34 and updates `shared-macros.mk` to define `PERL.5.34`, `POD2MAN.5.34`, and `PERL5BINDIR.5.34` for the new perl, **but `PERL_VERSIONS` and `PERL_64_ONLY_VERSIONS` is not yet updated to include 5.34**.   Basically, all the macro infrastructure is updated so oi-userland can build modules for it, but it isn't (yet) part of the default set of `PERL_VERSIONS`
2. I then start doing module rebuilds, with each module a separate PR, but for each module I rebuild, I set `PERL_VERSIONS` **in that module** to start including 5.34 (as well as the other 2 perl versions) in the list.  That way, we can go as slow as needed with module rebuilds: we don't get the 5.34 version of a module until that module is rebuilt, but the other two versions of perl are unaffected.
3. Once **all** existing modules have been rebuilt and perl 5.34 has its own version of each module, I then submit a "cleanup" PR that enables `5.34` in `PERL_VERSIONS` in `shared-macros.mk` and removes the `PERL_VERSIONS` setting from the Makefile of each of the modules.

After those steps are complete, we can look at the steps necessary to obsolete the 5.22 and 5.24 versions, in particular what impact the new perl might have on `illumos-gate`.

I think this should be the safest process for adding 5.34 and slowly building up its packaged modules, with the least danger to existing modules and versions.

Comments?  Concerns?

AndWac:
This sounds like a good option to follow.

timmooney:
> This sounds like a good option to follow.

Excellent!  I'll proceed following that plan.  Thanks Andreas!

PR #7324, opened 2021-11-26
new component gtksourceview4

timmooney:
new component gtksourceview4, can be installed at the same time as gtksourceview3 without conflict.

This component is required by pluma 1.26.0 and presumably also mate-applets 1.26.0 (both components previously required gtksourceview3).

This component **requires** PR #7323 first; that's the only reason I updated vala, as this wouldn't build with the older version of vala we previously had.

I enabled the optional 'glade_catalogs' option, so that this package installs its glade XML file for use with glade.  I don't know if there's a catalog rebuild that's triggered automatically by some transform for the glade catalogs, but that might be required.

Tests all require X11 so they can't be run in an automated fasion, but I ran them and included the "all pass" output for manual comparison as this package is updated in the future.


timmooney:
Sorry, I hadn't squashed commits with my first push, that's taken care of now.

AndWac:
Tests don't run automatically so you can add your test results as an indicator for what is expected and we can check against other build environments.
As vala is needed for the build you should add a manual REQUIRED_PACKAGES += entry for it in the Makefile.

timmooney:
Pushed a minor change that adds `developer/vala` to the required (build) packages.

AndWac:
Sorry for being not precisely enough: you should add test/results-32.master and test/results-64,master (or test/results-all.master) and COMPONENT_TEST_TRANSFORMS because it can be used. You can simply add a comment to the Makefile stating that X11 is needed for the tests.
"gmake test" is only being run manually. I am doing it when I test a PR before I merge it. And of course also if I prepare a package by myself.

timmooney:
Added the transforms and the `results-32.master` and `results-64.master` .

PR #7332, opened 2021-11-27
Upgrade of fio to 3.28 with python 3.9

jgmills:
Fio is a flexible I/O tester.  This PR upgrades fio from 2.15 to 3.28 and the python version used by fio from 2.7 to 3.9 .

The changes are mainly to accomodate the fio and python version upgrades.  The Makefile also includes a new REQUIRED_PACKAGES list and is converted the new style.

Most of the patches were removed as they were no longer necessary.  One patch was rebuilt to suit the new fio source.  One patch was completely new.

The manifest file fio.p5m was edited to match sample-manifest.p5m .  As well, the python script fio-histo-log-pctiles.py was removed because it requires a python module not available for python3.9 .  Finally, it was necessary to bypass dependency checking for fio2gnuplot because it misinterpreted a valid dependency on the six python module.

The build, install, and publish steps were all successful.  There are no built-in tests.  However, I was able to test the fio executable after the package was installed.  "fio --version" and "fio --help" produced the expected results.  As well, when I gave it an actual jobfile so that fio performed an actual I/O test, it worked nicely.


AndWac:
system/test/zfstest depends on it. Have you tested it after the fio upgrade?

jgmills:
I have not tested it with zfstest, but I have tested it outside of that framework, on a ZFS filesystem.  I just looked at the README file for zfs-tests .  The tests require three scratch disks.  I don't have any system with three scratch disks.  As far as I can tell, I'm not able to run the zfs tests.

PR #7352, opened 2021-12-01
mysql-common: separate from other packages

AndWac:


components/database/mysql-common/Makefile
```diff
@@ -0,0 +1,53 @@
...
+#
+# Copyright 2021 Jim Klimov
+# Copyright 2021 Andreas Wacknitz
+#
+include ../../../make-rules/shared-macros.mk
+
+# Beside the p5m file, there is really nothing to this component which just
+# delivers a user and group actions as a common ground for database servers of mysql heritage.
+COMPONENT_NAME=				mysql-common
+COMPONENT_VERSION=			1.0
```
Toasterson, line 21: I think we should bump this just to be safe.

make-rules/ips-buildinfo.mk
```diff
@@ -31,7 +31,7 @@
 #
 # Release major number: 2014, 2015, etc.
 #
-RELEASE_MAJOR ?= 2020
+RELEASE_MAJOR ?= 2021
```
Toasterson, line 34: Should we not do this in a seperate PR?

Toasterson (changes requested):
Some small points on the things we talked like the current version having COMPONENT_REVISION=1 and that we should seperate make-rules changes.

make-rules/ips-buildinfo.mk
AndWac: Last time it happened here:
commit c2b8df7766849ddb8f75305e9075721dd400d754
Author: Alexander Pyhalov <apyhalov@gmail.com>
Date:   Tue Jan 14 09:09:45 2020 +0300

    Remove Python2 illumos-gate modules

:100644 100644 8ab8f07f7 c6e71ef85 M    components/meta-packages/history/history
:100644 100644 62f8b7d17 ad7ef8cef M    components/openindiana/illumos-gate/Makefile
:100644 100644 9b1641a60 1f0c43e54 M    make-rules/ips-buildinfo.mk

Toasterson: Hmmmm, welp. We should maybe do this properly when the Year increments aswell.... Well let's leave it in then and do it correctly in 2022

components/database/mysql-common/Makefile
Toasterson: @AndWac This we still need to do. With incorporations we nail the version so branch_version helps, but unless we bump `COMPONENT_VERSION` Old versions could still be installed.

Toasterson (changes requested):
See comment in Makefile we need to bump `COMPONENT_VERSION`

Toasterson:
From out docs: [IPS Dev Guide Chapter 3 section about version precedence](https://github.com/OpenIndiana/pkg5/blob/oi/doc/dev-guide/chpt3.txt#L135)
```
The package versions are ordered using left-to-right precedence; thus
the timestamp is the least significant part of the version space; the
number immediately after the ‘``@``’ is the most significant.
```

AndWac:
mysql-common is different from other packages as it doesn't really use COMPONENT_VERSION from the Makefile, instead it uses a fix version numer 0.5.11 in the manifest:
set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@0.5.11,$(BUILD_VERSION)
So, the only relevant/changing part here is BUILD_VERSION which is 5.11-2021.0.0.0 with the change of the RELEASE_MAJOR.

Toasterson (approved):
Ah, we had another discussion about COMPONENT_REVISION, but then that is not changing from the existing package. LGTM then.

PR #7370, opened 2021-12-07
libfuse : Fix the condition for not setting a timestamp in utimensat(2)

jpandre:


jpandre:
Oops. Patch missing... will fix

PR #7378, opened 2021-12-10
Fix emacs build to be 64bit, delete insecure movemail from

klausz65:
Delete insecure movemail from emacs distribution, by utilizing movemail from GNU mailutils.
Add patch: 08-allow-composition-of-pure-ASCII-strings-in-the-mode-line.patch


AndWac:
Please also update the sample-manifest

AndWac (approved):

PR #7380, opened 2021-12-10
Fix building of top 64 bit only binary.

klausz65:
enable configure option --disable-dualarch which disables building 32 and 64 bit executabels


AndWac:
sample-manifest and manifest also need to be updated

AndWac (approved):

PR #7382, opened 2021-12-11
relocate gmp.h into /usr/include from /usr/include/gmp.

klausz65:
Many opensource packages expect gmp.h to be located in /usr/include rather then /usr/include/gmp.
These packages depend on library/gmp:
desktop/remote-desktop/rdesktop
developer/gnu-cobol
editor/gnu-emacs/gnu-emacs-gtk
editor/gnu-emacs/gnu-emacs-no-x11
editor/gnu-emacs/gnu-emacs-x11
file/gnu-coreutils
library/gnutls-3
library/guile
library/mpc
library/mpfr
library/nettle
text/gawk
web/aria2
All of them don't need a recompile because of this change, however emacs would benefit from a recompile,
because gmp has been enabled in the Makefile but was never detected during configure/build time. All
the depending packages have already been re-packaged on SPARC without any problems.

### DIFF
--- a/components/library/gmp/Makefile
+++ b/components/library/gmp/Makefile
@@ -59,7 +59,7 @@ LDFLAGS += $(LD_Z_REDLOCSYM) $(LD_Z_RESCAN_NOW)
 MPN32_i386 = x86/pentium x86 generic
 MPN64_i386 = x86_64/pentium4 x86_64 generic
 MPN32_sparc = sparc32/v9 sparc32 generic
-MPN64_sparc = sparc64 generic
+MPN64_sparc = sparc64/ultrasparc34 sparc64/ultrasparc1234 sparc64 generic
 MPN_32 = $(MPN32_$(MACH))
 MPN_64 = $(MPN64_$(MACH))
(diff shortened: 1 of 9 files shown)


alarcher (commented):
I must say I do not agree with this change.

AndWac:
Can you elaborate why you don't agree? I have checked similar packages, eg. mpfr in solaris-userland. Oracle seem to have stopped putting include files for every package into its own folders several years ago. What are possible reason to stick with the burden?

alarcher:
You are basically creating a merge request to address a non-problem (the burden of passing one flag to cpp) while it would have been easier to wait the next update of gmp to do it if you really want, then while claiming that you only move files around for gmp, some variables for SPARC are actually changed in gmp+mpfr and other actions in guile without stating it.
I just do not think this is OK to proceed like this.

alarcher:
Also a lot of software probably relies on pkgconfig files so the argument is a bit overstated....

klausz65:
Okay, I cancel this PR, and split it off to a few more PR's

PR #7387, opened 2021-12-13
WIP : Driver and utilities to support exFAT file systems

jpandre:
Note : I should add a readme and a few header files, but I could
not find the proper way to do that.


AndWac:
For problems like yours I suggest to look at other packages. Eg. web/nginx brings additional files in a folder named "files" and adds them to the manifest like this:
file files/nginx-startup.sh path=/lib/svc/method/http-nginx mode=555
Here files/nginx-startup.sh is an additional file that is in oi-userland. This will be installed as file /lib/svc/method/http-nginx by the manifest.

PR #7389, opened 2021-12-14
Driver and utilities to support exFAT file systems

jpandre:
Retrying, sorry for the inconvenience

### DIFF
--- a/components/sysutils/exfat/Makefile
+++ b/components/sysutils/exfat/Makefile
@@ -1,0 +1,43 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 Jean-Pierre André. All rights reserved.
+#
+
+BUILD_BITS=64
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME= exfat
+IPS_COMPONENT_VERSION= 1.3.0
+COMPONENT_VERSION= $(IPS_COMPONENT_VERSION)
+COMPONENT_SUMMARY= exfat - Utilities that provide access to exFAT
+COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE= v$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH= sha256:689bcb4a639acd2d45e6fa0ff455f7f18edb2421d4f4f42909943775adc0e375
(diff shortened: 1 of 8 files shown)


components/sysutils/exfat/patches/libexfat-platform.h.patch
```diff
@@ -0,0 +1,19 @@
+--- tries/libexfat/platform.h.orig	2021-12-13 10:39:39.520746794 +0000
```
Toasterson, line 5: Can you make a comment here to describe the reason why this patch exists so we can make decisions if we want to keep it or not in the future?

components/sysutils/exfat/patches/use-shared-object.patch
```diff
@@ -0,0 +1,62 @@
+--- tries/dump/Makefile.am.orig	2018-09-15 07:03:24.000000000 +0000
```
Toasterson, line 5: Can you make a comment here to describe the reason why this patch exists so we can make decisions if we want to keep it or not in the future?

components/sysutils/exfat/patches/configure.ac.patch
```diff
@@ -0,0 +1,22 @@
+--- tries/configure.ac.orig	2018-09-15 07:03:24.000000000 +0000
```
Toasterson, line 5: Can you make a comment here to describe the reason why this patch exists so we can make decisions if we want to keep it or not in the future?

components/sysutils/exfat/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+TEST_TARGET= $(NO_TESTS)
+include $(WS_MAKE_RULES)/common.mk
+
+COMPONENT_PREP_ACTION = ( cd $(@D) && autoreconf -f -i )
+
+PATH=$(PATH.gnu)
+
+CFLAGS += -D__LITTLE_ENDIAN=1234
+CFLAGS += -D__BYTE_ORDER=1234
```
Toasterson: Can you make a comment here to describe the reason why this two lines exists so we can make decisions if we want to keep it or not in the future?

Toasterson (commented):
Can you add the requested comments? Otherwise LGTM

components/sysutils/exfat/patches/libexfat-platform.h.patch
jpandre: Add macroes to describe the endianness of the platform used, and to define how bytes are to be swapped accordingly.

components/sysutils/exfat/patches/use-shared-object.patch
jpandre: Use the code common to driver and utilities from a library defined as a shared object instead of static linking it into each of them.

components/sysutils/exfat/Makefile
jpandre: Good point. These lines are not needed anymore.

(Will update tomorrow)

Toasterson:
Can you add the comments for the patches to the file?

Toasterson (approved):

Toasterson:
@jpandre We had during the Maintainer meeting a discussion about the License Situation. And as far as we can tell the Patent is only opened to linux. So we can't have it in the main repository. If you really want this inside of OI we could put it into hipster-encumbered

PR #7393, opened 2021-12-15
add gcc-11.2.0

AndWac:


components/developer/gcc-11/manifests/sample-manifest.p5m.erste_version
```diff
@@ -0,0 +1,1704 @@
+#
+# This file and its contents are supplied under the terms of the
```
Toasterson: Leftover file?

Toasterson (approved):
I went through this and several libraries (gmp, mpfr, boost) and all build fine and are usable.

Anything you want me to test in particular? Otherwise LGTM

AndWac:
There is one alarming area you can see in the test results for g++:
        === g++ Summary for unix/-m64 ===
# of unexpected failures    6554 

I thing we need to investigate about the reasons of the big number of unexpected failures. With almost the same patches OmniOs has only a fraction of this number.

PR #7394, opened 2021-12-15
move gmp header files to /usr/include and add MPN64_sparc options.

klausz65:
This is to follow the coding standards of Oracle Solaris 11.4, solaris-userland has switched re-locating gmp.h to /usr/include about 4 years ago now. Dependent other software packages will be changed in separate PR(s).


Toasterson (commented):
Same as in #7395

PR #7399, opened 2021-12-16
Add nzbget 21.1

erwinlem:
Am I supposed to add the sample manifest too? That seems a bit weird but all the other packages have this as well.

### DIFF
--- a/components/network/nzbget/Makefile
+++ b/components/network/nzbget/Makefile
@@ -1,0 +1,45 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 erwinlem
+#
+
+BUILD_BITS=64
+USE_OPENSSL11=		yes
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		nzbget
+COMPONENT_VERSION=	21.1
+COMPONENT_SUMMARY=	Binary newsreader supporting NZB files
+COMPONENT_PROJECT_URL=	https://nzbget.net
+COMPONENT_FMRI=		network/nzbget
+COMPONENT_CLASSIFICATION=
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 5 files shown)


AndWac:
Yes, but it's not weird because you can use it to detect changes when the package gets updated.

components/network/nzbget/nzbget.p5m
```diff
@@ -0,0 +1,66 @@
...
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 <contributor>
+#
```
AndWac, line 14: Replace <contributor> with your information

components/network/nzbget/nzbget.p5m
```diff
@@ -0,0 +1,66 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file usr/share/nzbget/nzbget.conf path=etc/nzbget.conf 
+file files/nzbget.xml path=lib/svc/manifest/network/nzbget.xml restart_fmri=svc:/system/manifest-import:default
```
AndWac, line 26: for config files you probably want preserve=true

components/network/nzbget/Makefile
```diff
@@ -0,0 +1,43 @@
...
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2021 erwinlem
+#
+
+BUILD_BITS=64
+include ../../../make-rules/shared-macros.mk
```
AndWac, line 17: Please add
USE_OPENSSL11=yes
here and re-run gmake REQUIRED_PACKAGES. This should update the dependencies to openssl-1.1. Otherwise the deprecated openssl-1.0 will be used.

components/network/nzbget/pkg5
```diff
@@ -0,0 +1,17 @@
+{
+    "dependencies": [
+        "SUNWcs",
+        "library/libxml2",
+        "library/ncurses",
+        "library/security/openssl",
+        "library/zlib",
```
AndWac, line 7: pkg5 needs to be updated, too.

AndWac (approved):

PR #7406, opened 2021-12-18
Bump webkitgtk to 2.32.3

klausz65:


components/library/webkitgtk/Makefile
```diff
@@ -37,8 +37,10 @@ COMPONENT_LICENSE_FILE=	webkitgtk.license
 
 # Build single threaded (parallel build fails on some systems):
 # COMPONENT_BUILD_ARGS=	-j1
+COMPONENT_PREP_ACTION= $(GPATCH) --strip=0 --backup < patches/mapfile.$(MACH)
```
pyhalov: You can make something like https://github.com/oracle/solaris-userland/blob/cc6fd8770c5234162b2d2c3ee4be9f7d362309e8/components/desktop/webkitgtk4/Makefile#L59 . Look also on corresponding changes in make-rules/prep-patch.mk .

klausz65: See #PR 7412

AndWac (approved):

PR #7412, opened 2021-12-19
add macro ADDITIONAL_PATCHES to prep-patch.mk

klausz65:
This is needed to be able to add $MACH specific patches to component Makefiles e.g. different mapfiles


make-rules/prep-patch.mk
```diff
@@ -39,7 +39,7 @@ GPATCH_FLAGS =	--strip=$(PATCH_LEVEL) $(GPATCH_BACKUP)
 # PATCH_DIR can be overridden to move patches to a different location
 # PATCH_PATTERN can be overridden to adjust the patch naming scheme that the
 #     build recognizes.
-# EXTRA_PATCHES{version} can be defined in the component Makefile to include
+# ADDITIONAL_PATCHES can be defined in the component Makefile to include
```
pyhalov, line 42: If EXTRA_PATCHES is no-op for now, please, remove it from oi-userland Makefiles. Otherwise looks good.

PR #7426, opened 2021-12-23
Update gnome-connection-manager to current GitHub HEAD - d2f66e3 @ 20210420

jimklimov:
Not sure when a GCM release would be issued, but some development after 1.2.1 did seemingly fix an issue I had with https://github.com/kuthulux/gnome-connection-manager/issues/5 that made `git log` and `git diff` use unpleasant, and seems to be related to `ksh93` hangs that nobody else could reproduce.

On the downside, that Git HEAD suffers from https://github.com/kuthulux/gnome-connection-manager/issues/64 misinterpreting double-clicks in Midnight Commander... So maybe there would be another PR (hopefully soon - maybe another commit in this one) to take advantage of some solution to that issue too.

### DIFF
--- a/components/desktop/gnome2/gnome-connection-manager/Makefile
+++ b/components/desktop/gnome2/gnome-connection-manager/Makefile
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright (c) 2015-2020 Jim Klimov
+# Copyright (c) 2015-2021 Jim Klimov
 #
 
 include ../../../../make-rules/shared-macros.mk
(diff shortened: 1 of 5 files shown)


jimklimov:
Well, hoping for a better fix to the mentioned issue in future upstream revisions, but as of today the current state of this PR is satisfactory to me as a major user of both GCM and MC :)

AndWac:
Can you squash the commits?

jimklimov:
Probably, left them apart to git revert the patchfile easily later.

Or you may Squash&Merge...

UPDATE: Then again, in the future a plain `git rm` and `git commit` is more likely ;) Pushed...

PR #7428, opened 2021-12-23
phoronix-test-suite: add version 10.6.1

torculus:
Adding [`phoronix-test-suite`](https://github.com/phoronix-test-suite/phoronix-test-suite), a popular benchmarking tool that might get OpenIndiana / illumos some more attention

I'm not sure how best to remove the `systemd` files. Some help would be appreciated there


AndWac:
Did you run the test suite and checked its results?

torculus:
I ran it, but then the holidays came and upstream released a new version. I will rebase and resubmit soon

Also, can you prevent installation of certain files just by removing them from the manifest? We don't need the `systemd` files on illumos

torculus:
I ran the suite and the test completed without error (see below)
phoronix_finished

You can also see the results at https://openbenchmarking.org/result/2201039-NE-TESTS069642

Everything seems set to go, but I would still appreciate guidance on how to remove the `systemd` and `.service` files from the manifest. The component is pretty huge, even if it is simple to build

AndWac:
When I run "gmake sample-manifest" I'll get two differences. This has to be solved, because when this happens on the build server publishing will fail. It might be a missing dependency (something installed on your machine that's not on mine).
╰─➤  git diff sample-manifest.p5m                                                                                                                                                                                                 1 ↵
diff --git a/components/sysutils/phoronix-test-suite/manifests/sample-manifest.p5m b/components/sysutils/phoronix-test-suite/manifests/sample-manifest.p5m
index 7f344bcec..c1dbbac96 100644
--- a/components/sysutils/phoronix-test-suite/manifests/sample-manifest.p5m
+++ b/components/sysutils/phoronix-test-suite/manifests/sample-manifest.p5m
@@ -58,9 +58,7 @@ file path=usr/share/doc/phoronix-test-suite/stubs/downloads.html
 file path=usr/share/doc/phoronix-test-suite/stubs/readme-basics.txt
 file path=usr/share/doc/phoronix-test-suite/test-profile-creation.md
 file path=usr/share/icons/hicolor/48x48/apps/phoronix-test-suite.png
-file path=usr/share/icons/hicolor/64x64/mimetypes/application-x-openbenchmarking.png
 file path=usr/share/man/man1/phoronix-test-suite.1
-file path=usr/share/mime/packages/openbenchmarking-mime.xml
 file path=usr/share/phoronix-test-suite/deploy/deb-package/build-package-deb.php
 file path=usr/share/phoronix-test-suite/deploy/deployments.md
 file path=usr/share/phoronix-test-suite/deploy/docker/ubuntu-pts-docker-build.sh

alarcher:
@torculus Thanks for your work. May I suggest to find out how to get all the system information to make benchmark results relevant? Processor is marked as unknown. In case nothing is implemented you can use for instance 'kstat -m cpu_info -i 0 -s brand'

AndWac:
Am 03.01.22 um 23:07 schrieb Benjamin S. Osenbach:
>
> I ran the suite and the test completed without error (see below)
> phoronix_finished
> <https://user-images.githubusercontent.com/20175597/147985552-b5a88971-25f6-4da7-a2fe-f1eb57f2dcaa.png>
>
> You can also see the results at
> https://openbenchmarking.org/result/2201039-NE-TESTS069642
>
> Everything seems set to go, but I would still appreciate guidance on
> how to remove the |systemd| and |.service| files from the manifest.
> The component is pretty
>
Removing files from a manifest is easy. Straight way is to place a
comment sign (#) in the first column of the file action. Preferred way
is to use a transform,
eg. look for existing drops with running

find . -name "*.p5m" -exec grep drop\> {} \; -print

inside the oi-userland repository folder.

Message ID:
***@***.***>Andreas

torculus:
> @torculus Thanks for your work. May I suggest to find out how to get all the system information to make benchmark results relevant? Processor is marked as unknown. In case nothing is implemented you can use for instance 'kstat -m cpu_info -i 0 -s brand'

@alarcher, I did some digging and it turns out that the issue is coming from two sources:
1. Upstream uses `dmi_info` (see [here](https://github.com/phoronix-test-suite/phoronix-test-suite/blob/4f18296a1862fe54a4c58701a1f5ec6bd62a4d94/pts-core/objects/phodevi/parsers/phodevi_solaris_parser.php#L33)), and OpenIndiana has an incomplete implementation (see [this message thread](https://www.mail-archive.com/oi-dev@openindiana.org/msg06711.html))
2. Even when I use OpenIndiana's `smbios`, my old HP laptop (yes, I am running a server OS on an old laptop) doesn't support the bios information

All of that to say that my CPU non-reporting is probably specific to my machine

torculus:
> When I run "gmake sample-manifest" I'll get two differences. This has to be solved, because when this happens on the build server publishing will fail. It might be a missing dependency (something installed on your machine that's not on mine).

@AndWac, I modified the manifest and checked it against both the sample-manifest and the manifest generated in the build directory. Those files are not marked as missing on my machine. `gmake REQUIRED_PACKAGES` reports that you need `ksh93` and `php-73/php-cli` as the only two dependencies. I did also change the Makefile halfway through this pull request to upgrade to the newest release. Would you mind checking again to see if the problem is still there?

torculus:
@alarcher, @AndWac, I see that PHP 8.1 is pending in the pull requests. Since PHP server is the only dependency for phoronix-test-suite, we can wait until that is merged, and then I'll rebase to see if that fixes some of the issues here

PR #7430, opened 2021-12-24
Upgrade of pygobject to 3.42.0 with python 3.7 and 3.9

jgmills:
Pygobject is the Python bindings for GObject.  This PR upgrades the software version to 3.42.0, using python versions 3.7 and 3.9 .  Note that this version does not support python 2.7 or 3.5, requiring that the old version of pygobject be retained.  The new version is therefore introduced in a new component directory: python/pygobject-34 .  This version supplies packages library/python/pygobject-3-37 and library/python/pygobject-3-39 .

There are two existing component directories for pygobject.  python/pygobject-3 contains a version that supplies library/python/pygobject-3-27 and library/python/pygobject-3-35 .  It is changed to suit the new component directory.  python/pygobject contains an even older version that supplies the package library/python/pygobject-27 .  This component is unchanged.  Both of these component directories can be deleted when the corresponding python versions are removed from OI.

The new component directory contains a Makefile that uses the setup.py build style.  The built-in tests are disabled because they require the gobject-introspection test files, which are not packaged for OI.  They will be packaged when the gobject-introspection package is upgraded to the current version.  This directory also contains pygobject-34-GENFRAG.p5m and pygobject-34-PYVER.p5m, needed to populate three new packages.  Note that the pygobject-34-PYVER.p5m package uses mediated symlinks, with the python3 mediator.

In the python/pygobject-3 component directory, several changes were required to Makefile.  In addition, two new manifests were needed to describe the python 2.7 and 3.5 packages.

The build, install, and publish steps were all successful.  As mentioned above, there were no built-in tests performed.  As well, no external tests were performed.  The subsequent upgrade of python/terminator will be an external test.  I did do a "pkg install" of library/python/pygobject-3-39 .  It behaved normally.


components/python/pygobject-34/Makefile
```diff
@@ -0,0 +1,66 @@
+#
```
AndWac: Why did you choose pygobject-34 for packages that provide versions for python-3.7 and 3.9?
My recommendation is to use pygobject-legacy for the old python versions and pygobject (or pygobject-3) for the latest python versions.

components/python/pygobject-34/pygobject-34-GENFRAG.p5m
```diff
@@ -10,13 +10,16 @@
...
-file path=usr/lib/$(MACH64)/pkgconfig/pygobject-3.0.pc
+# Manually added since python2 version is built in ../pygobject-3
+depend type=conditional fmri=$(COMPONENT_FMRI)-27 \
+    predicate=runtime/python-27
 
-depend type=require \
-    fmri=library/python/pycairo@1.10.0
+# Manually added since python3 version is also built in ../pygobject-3
+depend type=conditional fmri=$(COMPONENT_FMRI)-35 \
+    predicate=runtime/python-35
```
AndWac: These two dependencies look wrong. The python runtime dependencies belong to the corresponding manifest files. And the dependencies for python runtimes 3.7 and 3.9 seem to miss.

AndWac:
I have done some checks.
1. We don't seem to have an package that depends on pygobject-3-27. Thus, it's a good idea to obsolete this package now.
2. The only packages that depend on pygobject-27 are
     diagnostic/nmap-gui
     image/editor/gimp
Maybe it's possible to rebuild these with the corresponding new pygobject-3 versions (or maybe even better to rebuild them with python-3.7). If that's possible we can also drop pygobject and only provide pygobject-3 versions.
What do you think?

pyhalov:
Be careful to test carefully second option. I vaguely remember that python-gobject dependency can be 'hidden' (build-time only?).

components/python/pygobject-34/Makefile
jgmills: The "-34" suffix is derived from version 3.40.0 of pygobject.  It is not related to python-34 .

I assumed that it was not possible to rename existing component directories, without introducing complications.  Was I wrong?

I'm willing to do the two renames but, how do I do that, after having done the "git commit"?
I'd prefer to rename the old directory to pygobject-3-legacy and the new one to pygobject-3, instead.

components/python/pygobject-34/pygobject-34-GENFRAG.p5m
jgmills: They seem correct to me, for a conditional dependency.  The condition is that the package runtime/python-35 be installed.  When I do "pkg contents -m library/python/pygobject-3-35 | grep depend", I get "depend fmri=library/python/pygobject-3 type=require".  I have the package runtime/python-35 installed on my build server.

jgmills:
I prefer to retain the python-27 and python-35 libraries until all applications using python-27 or python-35 are upgraded.  Some of these packages may be rebuilt, but I can't predict when this will happen or who will do it.

components/python/pygobject-34/Makefile
AndWac: After you have changed your local git history you can simply use a force push (eg. "git push -f <branch name> origin") and Github will automatically update your PR.

components/python/pygobject-34/pygobject-34-GENFRAG.p5m
AndWac: But you are forcing runtime dependencies for python 2.7 and 3.5 in your pygobject-34 packages where you only provide package versions for python 3.7 and 3.9. The require dependencies should be were the packages are delivered. So, pygobjet-3-35 should require python 3.5, pygobject-3-37 should require python 3.7 and so on. You can look in other python packages, eg. setuptools, how it is done in the PYVER manifest:

\# force a dependency on the Python $(PYVER) runtime
depend type=require fmri=__TBD pkg.debug.depend.file=python$(PYVER) \
    pkg.debug.depend.path=usr/bin

components/python/pygobject-34/Makefile
jgmills: That was not my question.  I wanted to know what do I do locally before the push?  I've never done this before.  I want to to it right.

Obviously I have to be in the pygobject branch.  Should I just do "git mv" twice?  Or, should I undo the commit first?  After that, I'd do "git mv" twice, all the "git add/rm" commands, and the commit again.

Of course, I'd squash the commits before doing the forced push.

components/python/pygobject-34/pygobject-34-GENFRAG.p5m
jgmills: Ah, I see the problem.  The conditional dependencies are correct, but an unrelated dependency is missing.  I compared the pygobject-3-35 and pygobject-3-37 packages, generated from different manifests.  Both pygobject-3-35 and -27 packages are missing an explicit dependency on runtime/python-35 or -27, respectively.  I will correct that omission in the next push.

jgmills:
With this push, I've added the missing dependencies.  I've also renamed two of the component directories, pygobject-3 to pygobject-3-legacy and pygobject-34 to pygobject-3 .  I've also renamed two of the files to eliminate "-34" from the names, and changed two files to eliminate "-34" from comments.  After many commits and squashed, this updated PR is the result.

components/python/pygobject-3-legacy/pygobject-3-27.p5m
```diff
@@ -0,0 +1,69 @@
...
+set name=info.classification value="org.opensolaris.category.2008:Development/Python"
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license COPYING license="LGPL v2.1"
+
+depend type=require fmri=runtime/python-27
+depend type=require fmri=library/python/pygobject-3
+depend type=require fmri=library/desktop/gtk3@3.14.15,5.11-2016.0.0.2
```
AndWac: I think that depending on pygobject-3 is wrong here. We have it the other way around: pygobject-3 depends on specialized packages like pygobject-3-35 because pygobject-3 is only a meta package. If you install pygobject-3 you'll get pygobject packages for all python versions. If you install a pygobject-3 package for a certain python version, eg. pygobject-3-35 you don't necessarily want all other python versions of this package, too.

AndWac: I am not sure whether it's a good idea to depend on exact versions. I recommend to use
depend type=require fmri=library/desktop/gtk3
instead.

jgmills: Getting packages for all python versions is normal behavior for packages published via: *GENFRAG.p5m *PYVER.p5m .  This is our model, even though they are published from separate component directories.

However, since python-27 and -35 will soon be removed, the next push will change it to a conditional dependency.  The condition will be the presence of the appropriate python runtime package.  This change is already done for the reverse dependency.

jgmills: I noticed that version string in the original manifest.  I don't know why it depended on an obsolete version, but I decided not to alter it.  Certainly, omitting the version string is better.  That change will appear in the next push.

AndWac:
> Be careful to test carefully second option. I vaguely remember that python-gobject dependency can be 'hidden' (build-time only?).

I have already found a package that doesn't build anymore with the current state of affairs: pyatspi.
checking for PYGOBJ... no
configure: error: Package requirements (pygobject-3.0 >= 2.90.1) were not met:

No package 'pygobject-3.0' found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables PYGOBJ_CFLAGS
and PYGOBJ_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.

jgmills:
That's the error message you get if you are building with python-39 or -39 but don't have the pygobject-3-37 or -39 packages installed.  It's also the error you get if pkgconfig is looking in the wrong place for the PC file.  Usually, it's looking in the old 64-bit place but not in the standard place.  Check the config.log file to see the current setting of PKG_CONFIG_PATH.  If it's only the amd64 path, the error is likely in the build system.  For a simple fix, add this to the Makefile:

PKG_CONFIG_PATH= $(PKG_CONFIG_PATH.64):$(PKG_CONFIG_PATH.32)

It should be inserted after "shared-macros.mk" but before "common.mk".

AndWac:
Yes, that fixed the build problem, thanks.

components/python/pygobject-3-legacy/pygobject-3-27.p5m
```diff
@@ -0,0 +1,70 @@
...
+depend type=require fmri=runtime/python-27
+depend type=conditional fmri=library/python/pygobject-3 \
+    predicate=runtime/python-27
+depend type=require fmri=library/desktop/gtk3
+
+link path=usr/include/pygobject-3.0/pygobject.h target=pygobject.h-2.7 \
+	mediator=python3 mediator-version=2.7
+link path=usr/lib/$(MACH64)/pkgconfig/pygobject-3.0.pc target=pygobject-3.0.pc-2.7 \
+	mediator=python3 mediator-version=2.7
+
```
AndWac, line 38: You are making python-2.7 a python-3 version in OpenIndiana with these two links and its mediator settings.
We don't actually seem to have a package that depends on pygobject-3-27 and retaining this kind of compability seems to be not worth the problems this will probably create.

jgmills:
You wrote "Yes, that fixed the build problem, thanks."

It's best to fix this problem in shared-macros.mk .  Otherwise all python applications will need that fix.  The original line in shared-macros.mk is this:

PKG_CONFIG_PATH = $(OPENSSL_PKG_CONFIG_PATH):$(PKG_CONFIG_PATH.$(BITS))

It should be changed to become this:

PKG_CONFIG_PATH = $(OPENSSL_PKG_CONFIG_PATH):$(PKG_CONFIG_PATH.$(BITS)):$(PKG_CONFIG_PATH.32)

All that change to the PC path variable does is to tell pkgconfig to look in the standard location last.  Then you won't need that line, although setting it in Makefile too won't cause any problems.

components/python/pygobject-3-legacy/pygobject-3-27.p5m
jgmills: I wanted to use an existing mediator and also realized that the "python3" mediator could never assume the value "2.7".  Thus, the symlinks would never be created.  The only alternative left now is to delete those lines.  This change means that the package would still be published, but that dependent packages could not be built.  I'll delete those lines in the next push.

PR #7436, opened 2021-12-25
boost: update to 1.78.0

AndWac:


alarcher:
The libraries fiber and stacktrace are broken in this update. The boost buildsystem does not throw error if libraries fail to build so the hint is that there is an issue if some disappear from the sample manifest.

AndWac:
Oh, I am sad to read this. Are you able to help us out here?


Am 19.01.22 um 16:24 schrieb Aurélien Larcher:

alarcher:
No worries, I have just set up a new build zone and started looking into the issue :)

alarcher:
I should have written a note about the boost build system, sorry I did not think about it.

AndWac:
That's great! Welcome back! I am quite happy to hear from you again.

In the meantime I am also happy about being able to get at least a few
more people to create PR's :)

Am 19.01.22 um 20:43 schrieb Aurélien Larcher:

AndWac:
Haha, I think that's one of the smaller problems. I am still learning
the Python eco system and the dependency hell it has created over the
last years.
I am not even near at the point that I am satisfies. Our virtualenv is
broken in many ways so as many other packages. I would like to throw out
32 bit and all python versions despite one. And of course we seem to
have problems with libraries in some packages (see #13824).


Am 19.01.22 um 20:44 schrieb Aurélien Larcher:

PR #7441, opened 2021-12-25
squeak:  patch to use pkg:/image/library/libjpeg6-ijg

cstes:
The Squeak source code contains a (builtin) plugin for reading JPEG

Unfortunately this is done by a copy of the libjpeg6-ijg library copied into the source code of Squeak

There is work going on to link against system provide libjpeg libraries for Debian/Ubuntu.

For OpenIndiana we can succesfully do the same thing using the image/library/libjpeg6-ijg    
it does not work with libjpeg8-turbo

This is being escalated to the Squeak developer(s).

The current patches change the dependencies so that squeak now depend on OpenIndiana provided libjpeg instead of using a builtin libjpeg (statically linked into the Squeakvm)

### DIFF
--- a/components/runtime/smalltalk/squeak/Makefile
+++ b/components/runtime/smalltalk/squeak/Makefile
@@ -42,6 +42,12 @@
 
 BUILD_BITS=32_and_64
 
+# the goal is to use libjpeg8-turbo but that links, but does not display
+# working jpeg images; needs to be fixed upstream by the Squeak team
+# currently use libjpeg6-ijg which is better than using the copy of
+# those libjpeg files in the squeak source tree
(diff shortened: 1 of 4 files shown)


cstes:
This patch adds/changes

REQUIRED_PACKAGES += image/library/libjpeg6
REQUIRED_PACKAGES += image/library/libjpeg6-ijg

cstes:
Longtem goal is 

https://www.illumos.org/issues/7391

"All components should use libjpeg8-turbo as JPEG implementation"

We are trying to do so but  are already happy that squeak linked against the system provided libjpeg6-ijg works, which is better than using a statically linked libjpeg into squeakvm (as it was before) because that's not visible at the IPS dependency level

PR #7445, opened 2021-12-26
libjpeg9-ijg : version 9.4.0 of libjpeg

cstes:
This version does not work with Squeak but still it may be useful ...

Attempt can be made to check whether Squeak can be made compatible with libjpeg 9.4.0


cstes:
libjpeg9-ijg works fine with Squeak

temorarily closing this PR although it may make sense to add libjpeg9-ijg

PR #7450, opened 2021-12-27
libmatemixer: update to 1.26.0

timmooney:
This is the 2nd in a series of MATE-related updates, all part of the 1.26.0 update.

For dependency reasons, I tackled the packages in several "groups".

libmatemixer is part of "Group 1". The packages in that group can be built in any order.

Very little changed in libmatemixer.  You should be able to apply this without needing to rush later updates/rebuilds.

### DIFF
--- a/components/desktop/mate/libmatemixer/pkg5
+++ b/components/desktop/mate/libmatemixer/pkg5
@@ -1,8 +1,11 @@
 {
     "dependencies": [
         "SUNWcs",
+        "developer/build/autoconf-archive",
         "library/audio/pulseaudio",
+        "library/desktop/mate/mate-common",
         "library/glib2",
+        "shell/ksh93",
         "system/header/header-audio",
         "system/library"
     ],
--- a/components/desktop/mate/libmatemixer/Makefile
+++ b/components/desktop/mate/libmatemixer/Makefile
@@ -12,22 +12,23 @@
 # Copyright 2016 Till Wegmueller
 # Copyright 2016 Ken Mays
 # Copyright 2020 Michal Nowak
+# Copyright (c) 2021 Tim Mooney. All rights reserved
 #
 
-BUILD_BITS=		32_and_64
+BUILD_BITS=		64_and_32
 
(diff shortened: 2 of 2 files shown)


components/desktop/mate/libmatemixer/pkg5
```diff
@@ -3,6 +3,7 @@
         "SUNWcs",
         "library/audio/pulseaudio",
         "library/glib2",
+        "shell/ksh93",
         "system/header/header-audio",
```
AndWac, line 9: If you add new REQUIRED_PACKAGES in the Makefile corresponding entries in pkg5 are expected.

AndWac: As soon as you have fixed pkg5 I will start integrating the mate PR's. I don't think that we will find more testers here and it runs stable enough for me on my desktop.

mvwieringen: Over today I cherry-pick the patches and build them one by one and eventually upgraded my server(s) and from the first basic testing the thing I use daily work. The one running my SunRays was the most important and there I have not seen anything that works to a degree worse then the previous version. So for me also a +1 as to integrating this.

AndWac: Good to know that we have someone who is really interested in Mate tests ;-)

timmooney:
Sorry for the delay, I had missed your earlier comment on this one.

The updated push has just one change:
1. include `pkg5` from after `publish`, so the updated requirements are present

components/desktop/mate/libmatemixer/pkg5
timmooney: Thanks @mvwieringen for the feedback, it's very useful.

I am concerned that there may be a memory corruption or uninitialized value somewhere in `caja` that's causing the one issue that @AndWac noted, but until I can reproduce it, tracking down the problem is difficult.  If it's very rare and no one else triggers it, I guess that's OK, but if not I'm hoping someone can provider a reproducer.

PR #7478, opened 2021-12-27
revert copyright owner back to Andrzej Szeszo

klausz65:


Toasterson:
Merging as discussed via mail.

PR #7496, opened 2021-12-30
adding perl module Tk to perl-522 and perl-524

fritzkink:


timmooney:
So there's a couple of things that likely need to be adjusted before this could be accepted.

When you're working on a single component (package), the workflow they've adopted for OI userland is that you should have 1 commit that contains all your changes.  The "working with oi-userland" doc has a section near the end about how to work incrementally in your local git branch, probably making several different commits, but then before you push your work for this PR, it recommends _squashing_  all of your related commits into a single commit.

Here, you have 4 commits, and some of them appear to be completely unrelated to Tk.  There are reasons why you might intentionally touch multiple components in one PR, but in this case it appears to be accidental.  That means that the unrelated commits shouldn't be squashed down into a single commit, they should instead be dropped from this commit.  If they are things that you think should be changed in OI, then they ideally would be on a different branch and you would open a separate PR for them.

Now, I know of ways to remove unintended commits from a branch (because I've done the same thing), but I don't know what the **best** way is here.  I'm hoping that Andreas or one of the other people can provide some guidance on what the recommendation is to get rid of the 3 commits that aren't related to Tk.

Once we're down to just one commit and it's all related to perl Tk, then the focused review can start.

As an example, I can see that `COMPONENT_REVISION = 1` is specified in the `Makefile`.  That's not needed for a brand new component, so that line should be removed from the `Makefile`.  You use `COMPONENT_REVISION` when you need to rebuild an existing package.  It basically signifies that the package has been updated, not that there's a newer version of the software itself.

Anyway, the first thing that needs to be worked out is getting things down to be just related to Tk.

balrog-nona:
I know how I would proceed if I were to get rid of some commits: I would create a brand new branch and cherry-pick relevant commits to this branch. After that, I would create a new PR submitting the new branch. I would do this to avoid any commit removal because these procedures makes me nervous. So I would do this:

1. on the branch where I have all the commits, I would `git log` to see the commits with their hash which is the long sequence of number and letters (or maybe it is hexadecimal number? sorry for my ignorance) after the word commit, for example ff5a5696227e0794bddfc6dd89b401f13b9a78ea
2. I would create a brand new branch from oi/hipster which should be up to date (this can be done by `git fetch upstream` and after that `git rebase upstream/oi/hipster`)
creating new branch: `git checkout -b new_branch`
4. on this new_branch, I would `git cherry-pick hash_of_relevant_commit` which makes a copy of that commit and apply it in new_branch. You can cherry-pick as many commits as you need.

The advantage of this procedure is that it leaves this branch (meaning the old one) intact, just in case.

If the PR is about adding a component, it should contain one commit. In case that the change is currently part of several commits, these commits should be squashed after they're cherry-picked to new_branch. If you need help with that, I can provide some guidance so feel free to write me.

fritzkink:
Thanks a lot for your comments. I'll close this one and start with a new one hopefully better then.

PR #7517, opened 2022-01-03
added new Perl module Tk for Perl versions 5.22 and 5.24

fritzkink:


timmooney:
Hi Fritz!

This PR is much, much better, and is very close.  There are a couple of minor changes that are required, and there's a couple of stylistic things we can discuss.

First, the sample manifest.  Presumably at some point during the process you ran `gmake sample-manifest` and that generated `manifests/sample-manifest.p5m`.  You likely used that `sample-manifest.p5m` as the template to create `Tk-PERLVER.p5m`.

Whenever the `sample-manifest.p5m` changes, including the first time it's generated for a completely new package, it should be included in the commit.  Unlike the `Makefile` and the actual `.p5m` file that's used to generate the package, you **should not** change the copyright in the `sample-manifest.p5m`.  Just leave that file as a "template", but it needs to be part of the commit.

You can add it as a new commit in your local working copy, and we'll _squash_ that commit later on.  So to take care of this file, you would do something like:

    cd your_oi_checkout_dir/components/perl/Tk/
    git checkout whatever_your_Tk_branch_name_is
    git add manifests/sample-manifest.p5m
    git commit -m "include the sample-manifests.p5m"

Next, your `Makefile` references a `$(COMPONENT_LICENSE_FILE)` named `Tk.license`.  Presumably that's a copy of whatever license file is included in the `Tk` source distribution.  Whatever it is, it too needs to be part of the commit.  Again, another separate local commit that we'll squash into a single commit in a later step:

    # still in your oi-userland/components/perl/Tk
    # still on whatever your working branch is called
    git add Tk.license
    git commit -m "include Tk.license"

Finally, since this is a new commit, it's typical to include the `pkg5` metadata file that's generated as part of a successful `gmake publish`.  You don't need to make any changes to it, just include it in your commit:

    # still in your oi-userland/components/perl/Tk
    # still on whatever your working branch is called
    git add pkg5
    git commit -m "include pkg5 metadata file"

At this point you've fixed up the 3 missing files and you have them in your local working copy.  If you followed my example, you did each of them as a separate commit, which means you have 3 commits **plus** the original commit message, for a total of 4 commits in your branch.  Before you re-push your updates back to github, we want you to _squash_ those four commits into a single commit.  This is done via a `git rebase` operation.

You can use `git log` to verify how many commits you have on your branch, but assuming it's 4, you would do

    git rebase -i HEAD~4

That starts an interactive rebase with the 4 commits "after" HEAD.  It will put you in an editor and you will see something like:

    pick abcdef  my original commit message
    pick bafged include the sample-manifests.p5m
    pick fafeeee include Tk.license
    pick efefgab include pkg5 metadata file

To squash the 3 later commits into the first one, you change it to say:

    pick abcdef  my original commit message
    squash bafged include the sample-manifests.p5m
    squash fafeeee include Tk.license
    squash efefgab include pkg5 metadata file

(you can use just "s" instead of "squash", as explained in the comments).  Once you've done that, have your editor save the buffer/changes.

You will then be taken to a second editor screen, with your commit messages.  You've now squashed all the commits into one, but it's giving you a chance to adjust the commit messages.  I typically just delete all the additional messages, leaving just the original commit message, and save.

Once you've done that, the `git rebase` will be complete.  If you do

    git log

now you'll only see one commit since HEAD on your branch, and if you do

    git diff <whatever_the_long_commit_id_is_for_head>

you'll see all your changes in the one commit.

At that point, you just have one thing left to do: push your new consolidated commit "over the top" of the original commit you pushed to github.  That just means doing the same push to github you did before, but this time with the `-f` flag, to force it.  It will probably be something like

    git push -f origin your_tk_branch_name_here

Once that's done, when you look at this PR, you'll see that the original commit has been replaced by your updated commit.

If you can do those steps, you'll have fixed up the most important stuff for the PR.  Assuming things go well with those adjustments, I'll have some additional minor suggestions, that will be easy to make and will follow the same basic workflow.

Let me know if you have any questions or you run into any issues with the procedure as outlined.

components/perl/Tk/Makefile
```diff
@@ -0,0 +1,63 @@
...
+
+#
+# Copyright (c) 2021, Friedrich Kink. All rights reserved.
+#
+
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		Tk
+COMPONENT_VERSION=	804.036
```
AndWac: I am not happy with the package's name: First, it should be lower case. Second, it is too generic. If I am correct, this is not the tk package but just perl bindings for it. We already hava a package named "tk-8" which is the "real" thing. So, this should be named differently, eg. tk-perl (or whatever is most applicable).

components/perl/Tk/Makefile
```diff
@@ -0,0 +1,63 @@
...
+IPS_COMPONENT_VERSION=	5.11
+COMPONENT_REVISION=	1
```
AndWac: The IPS_COMPONENT_VERSION should be 804.0.36 but not 5.11. This version number is reserved for packages tightly coupled with illumos gate.
Also please add
HUMAN_VERSION= $(COMPONENT_VERSION) and also add an appropriate entry in the manifest. This will add information will that can be used outside IPS (eg, by humans or repology).

fritzkink:
Thank you very much for the feedback. I implemented all suggested comments. Still have to learn a lot.

AndWac:
You seem to have removed the Makefile.

AndWac:
Please squash your commits.

fritzkink:
restart with my pull requests

PR #7531, opened 2022-01-08
library/perl-5/lwp-mediatypes: perl-5.34 rebuild

timmooney:
rebuild lwp-mediatypes for all perl versions, including 5.34

No dependency on other perl modules and PR #7512 has already been merged, so this can be done in any order relative to the other module rebuilds.

All the same updates and modernizations as other recent perl module rebuilds

`Makefile`:
1. `BUILD_STYLE=makemaker`, `BUILD_BITS=32_and_64`, and include `common.mk`
2. bump `COMPONENT_REVISION`
3. add `COMPONENT_FMRI`, `COMPONENT_SUMMARY`, `COMPONENT_CLASSIFICATION`, and `COMPONENT_LICENSE` based upon values from the manifest
4. update the URLs for https and current locations/maintainer
5. specify `PERL_VERSIONS` and `PERL_64_ONLY_VERSIONS` so this includes 5.34 in the rebuild
6. drop `COMPONENT_TEST_TARGETS=test` and add `COMPONENT_TEST_MASTER` with one `results-all.master` and suitable `COMPONENT_TEST_TRANSFORMS`
7. drop `build/install/test`
8. update `REQUIRED_PACKAGES`

`LWP-MediaTypes-PERLVER.p5m`:
1. reference `$(COMPONENT_FMRI)`, now specified in the `Makefile`
5. add a runtime dependency on the perl runtime used for each rebuild of this module
6. add a dependency upon `library/perl-5/lwp-mediatypes`

### DIFF
--- a/components/perl/LWP-MediaTypes/LWP-MediaTypes-PERLVER.p5m
+++ b/components/perl/LWP-MediaTypes/LWP-MediaTypes-PERLVER.p5m
@@ -11,18 +11,28 @@
 
 #
 # Copyright 2013 Alexander Pyhalov.  All rights reserved.
+# Copyright (c) 2022 Tim Mooney.  All rights reserved.
 #
 
-set name=pkg.fmri value=pkg:/library/perl-5/lwp-mediatypes-$(PLV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
-set name=pkg.summary value="LWP::MediaTypes - guess media type for a file or a URL"
-set name=info.classification value=org.opensolaris.category.2008:Development/Perl
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)-$(PLV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
-license LWP-MediaTypes.license license="Artistic"
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 <transform file path=usr.*/man/.+ -> default mangler.man.stability committed>
+
+# force a dependency on the Perl runtime
+depend fmri=__TBD pkg.debug.depend.file=perl \
+	pkg.debug.depend.path=usr/perl5/$(PERLVER)/bin type=require
+
+# force a dependency on the non-PLV version of this module
+depend type=require \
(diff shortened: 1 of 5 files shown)


components/perl/LWP-MediaTypes/LWP-MediaTypes-PERLVER.p5m
```diff
@@ -11,19 +11,29 @@
...
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 <transform file path=usr.*/man/.+ -> default mangler.man.stability committed>
 
+# force a dependency on the Perl runtime
+depend fmri=__TBD pkg.debug.depend.file=perl \
+	pkg.debug.depend.path=usr/perl5/$(PERLVER)/bin type=require
+
+# force a dependency on the non-PLV version of this module
+depend type=require \
```
AndWac, line 33: I am not sure whether this is correct. I know that we already have other Perl and Python packages that are having similar dependencies. I am not yet convinced that this is a thing we want to have.
In my opinion the non-PLV (like non-PYV) version is only a convenience package, eg. install it an you'll get all packages for all Perl (Python) versions. But if a user installs a package for a certain Perl (Python) version he just wants or needs this version and not all packages for all supported Perl (Python) versions.
Are there any reasons to have this "backwards" dependency?

timmooney: Hey Andreas.  I think that's a really good question.

I'm doing it because (as you noted) a few others had it, and also because either Alexander, Michal, or Aurélien (I forget which one) suggested to me years ago that it was something that we wanted.  At that time I was just beginning any packaging in OI, so I just assumed it was something that would become clear to me over time.   It has never become clear to me, though, what the reasons are for it.

So, I'm doing it without understanding why it was "a good idea", basically because I trust that they have spent more time thinking about it than I have.

I suppose we could look at solaris-userland and see if they're still doing that.  Ultimately, doesn't matter to me whether we include the depenency or not.  If it were up to just me, and no one more experienced in OI had told me to include it, I definitely would not include it, for exactly the reason you mention:  if I just want perl-5.34 and some modules installed, it's really annoying that it effectively "forces" me to also install 5.22 and 5.24 and their modules too.

I haven't been able to make either of the last couple meetups (hopefully I'll make some in the future), but maybe it's something you want to discuss at one of those (or ask on oi-discuss) and see if everyone is in favor of just dropping that dependency.  I'm certainly OK with getting rid of it.

Once I have our current set of module versions rebuilt as part of the perl 5.34 integration, I will eventually go back and touch each of these modules again, to update them to more recent versions (if appropriate).  If we want to make a change with that dependency, I could do it then.  That gives time for discussion, if appropriate, about whether we want it or not.

AndWac: Solaris-userland doesn't have this dependency but it has only Perl version 5.32 at the moment.

AndWac: I have merged the PR's but we need to discuss this further. I just don't want to hinder your work further.

PR #7544, opened 2022-01-09
pgloader: update to 3.6.3

AndWac:
According to https://github.com/dimitri/pgloader/issues/843 the problem that prohibited an update has been fixed.


pyhalov:
Have you checked that mentioned in  https://github.com/dimitri/pgloader/issues/843 example works?

PR #7546, opened 2022-01-09
meld: update to 3.21.0, add Python 3.9

balrog-nona:
It builds, installs and publishes fine. Both tests passed.
One patch was deleted since the file is no longer part of source code.
When installed locally, it looked to me it worked and I didn't get any error message even when I started it from command line.


components/developer/meld/Makefile
```diff
@@ -35,42 +38,50 @@ and edit them in place (diffs update dynamically). You can compare two or \
...
 
-# Build dependencies
-REQUIRED_PACKAGES += text/rarian
+# Runtime dependencies
+REQUIRED_PACKAGES += library/desktop/gtk3
+REQUIRED_PACKAGES += library/glib2
+REQUIRED_PACKAGES += library/desktop/gtksourceview4
+REQUIRED_PACKAGES += library/python/pygobject-3
+REQUIRED_PACKAGES += library/python/pycairo
+REQUIRED_PACKAGES += text/itstool
```
AndWac, line 83: Runtime dependencies need to be defined in the manifest. And I would expect that meld should depend on python-3.9 versions of dependencies, eg. pygobject-3 is a meta package which depends on all available python versions of it. Depending on a python meta packages would bring a lot superfluous package versions.

AndWac: Thus, I would expect that meld depends on library/python/pygobject-3-39 because it only uses python-3.9.

components/developer/meld/meld.p5m
```diff
@@ -37,80 +36,197 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
...
-file path=usr/share/gnome/help/meld/C/meld.xml
-file path=usr/share/gnome/help/meld/es/meld.xml
-file path=usr/share/gnome/help/meld/fr/meld.xml
-file path=usr/share/icons/HighContrast/scalable/apps/meld.svg
-file path=usr/share/icons/hicolor/16x16/apps/meld.png
-file path=usr/share/icons/hicolor/22x22/apps/meld.png
-file path=usr/share/icons/hicolor/32x32/apps/meld.png
-file path=usr/share/icons/hicolor/48x48/apps/meld.png
-file path=usr/share/icons/hicolor/scalable/apps/meld.svg
+# Bypass Tk dependencies, since they're only used to display an error dialog
```
AndWac, line 46: I think you should add the runtime dependencies here like I did for ansible. Ansible's runtime dependencies (it uses python-3.7 atm) look like this:
depend type=require fmri=library/python/cryptography-37
depend type=require fmri=library/python/jinja2-37
depend type=require fmri=library/python/paramiko-37
depend type=require fmri=library/python/pyasn1-37
depend type=require fmri=library/python/pyyaml-37
depend type=require fmri=library/python/six-37

Note: Depending on a python meta-package doesn't necessarily bring the package version that is needed. It just installs all available versions but the version for the actual python version used doesn't need to be amoung them.
This was the case for some packages ansible depend on and can only be found by testing or using the application.

AndWac (approved):

AndWac:
I have run some simple tests with it and it worked for me.

components/developer/meld/Makefile
balrog-nona: Did it, thanks.

components/developer/meld/meld.p5m
balrog-nona: Did it, thanks. I skipped `library/python/pygobject-3-39` because it's automatically detected as dependency and pkglint gave me an error of duplicate dependency.

balrog-nona:
> I have run some simple tests with it and it worked for me.

It works in my GUI without any error message so I believe it's ready for merge.

PR #7547, opened 2022-01-09
image/rrdtool updated to lates version 1.7.2 incl. bindings for perl versions 5.22 and 5.34 and lua

fritzkink:


AndWac:
You'll need to squash your commits and get rid of unimportant commit messages.

AndWac:
You now have a commit from your prior work in this PR. Your git history is bad.

fritzkink:
But now it should be ok. Still struggling with git but getting better.

Am 09.01.2022 um 17:26 schrieb Andreas Wacknitz:

PR #7552, opened 2022-01-11
developer/openjade: fix for newer autoconf

erwinlem:
See also https://bugs.gentoo.org/765034


AndWac:
You'll also need to touch the Makefile because our jenkins relies on changes there in order to publish a package.

PR #7554, opened 2022-01-11
illumos-gate: remove obsolete fix for lc_core.h

AndWac:
I am not sure whether this is the right thing but in our proto area there is no more lc_core.h file beneath root_i386.
But there still is an lc_core,h beneath root_i386-nd.

### DIFF
--- a/components/openindiana/illumos-gate/Makefile
+++ b/components/openindiana/illumos-gate/Makefile
@@ -122,21 +122,6 @@ $(BUILD_DIR)/$(MACH)/.overlay: $(BUILD_DIR)/$(MACH)/.built
 	$(CP) $(BUILD_DIR)/$(MACH)/overlay/boot/splashimage.xpm \
 	    $(BUILD_DIR)/$(MACH)/overlay/boot/solaris.xpm || true
 
-	# Fix closed lc_core.h header
-	# Patch was taken from here https://www.illumos.org/issues/3853
-
-	if ! grep 199711L \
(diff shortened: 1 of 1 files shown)


pyhalov:
Of course, it's closed. I'm not sure if this particular fix was working, but check /usr/include/sys/lc_core.h. It should guard `struct tm` definition as in https://www.illumos.org/attachments/937 (this should match https://github.com/illumos/illumos-gate/blob/9ecd05bdc59e4a1091c51ce68cce2028d5ba6fd1/usr/src/head/iso/time_iso.h). Otherwise, there can be issues while compiling C++ programs, using both headers IIRC.

AndWac:
Am 11.01.22 um 21:30 schrieb Alexander Pyhalov:
>
> Of course, it's closed. I'm not sure if this particular fix was
> working, but check /usr/include/sys/lc_core.h. It should guard |struct
> tm| definition as in https://www.illumos.org/attachments/937 (this
> should match
> https://github.com/illumos/illumos-gate/blob/9ecd05bdc59e4a1091c51ce68cce2028d5ba6fd1/usr/src/head/iso/time_iso.h).
> Otherwise, there can be issues while compiling C++ programs, using
> both headers IIRC.
>
We don't have /usr/include/sys/lc_core.h anymore. According to yuripv on
IRC it's related to

pyhalov:
OK, then it can be safe. Check if current clang headers include sys/localedef.h (or perhaps need recompilation to avoid using it).

PR #7559, opened 2022-01-12
add Clang-11 to oi-userland

klausz65:


AndWac:
Running "gmake publish" gives an error:
ERROR pkglint.manifest005.2       duplicate depend actions in pkg:/developer/clang-11@11.0.0,5.11-2022.0.0.0 runtime/clang-11

alarcher:
I am surprised by the small number of patches, have you run the testsuite and compiled some code (C++ usually poses more issues)?

Toasterson:
There where some tests with gfx-drm and others so far it works. We probably discover somethings in the future but omnios does not have many more patches either.

klausz65:
OmniOS also has no clang-11, therefore only introduce clang-13

PR #7573, opened 2022-01-14
library/perl-5/mysql-dbi: rebuild for perl 5.34

timmooney:
rebuild dbi-mysql to add perl 5.34 support

Special notes for this package:
1. no test suite, since nearly all of it requires a database
2. I did not modify any of the existing runtime dependencies at the end of the `dbi-mysql-PERLVER.p5m` file.

`Makefile`:
2. bump `COMPONENT_REVISION` to 5
3. add `COMPONENT_FMRI`, `COMPONENT_SUMMARY`, `COMPONENT_CLASSIFICATION`, `COMPONENT_LICENSE_FILE` and `COMPONENT_LICENSE` with values from the manifest
6. drop `build/install`.  I kept `test` and added a note about the test suite
7. `gmake REQUIRED_PACKAGES`

`dbi-mysql-PERLVER.p5m`:
5. add the runtime dependency on the same version of perl
6. add the commented-out stuff for the `non-PLV` version

### DIFF
--- a/components/perl/DBI-MySQL/Makefile
+++ b/components/perl/DBI-MySQL/Makefile
@@ -22,44 +22,57 @@
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2013, Alexander Pyhalov. All rights reserved.
+# Copyright (c) 2022 Tim Mooney.  All rights reserved.
 #
 
+BUILD_STYLE=makemaker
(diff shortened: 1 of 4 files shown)


timmooney:
Oh, one more thing I forgot to mention: our default Maria DB version is still 10.1, though that ended support from upstream in 2020.  At some point, we should probably switch our default version to be 10.3.  I haven't looked at what's involved in doing that.

PR #7575, opened 2022-01-14
library/perl-5/file-listing: rebuild for perl 5.34

timmooney:
rebuild `file-listing` to add perl 5.34 support

Of note:
1. added the missing runtime dependency on `http-date` from the same perl version
2. since it's needed at runtime and we're running the test suite, also added `library/perl-5/http-date` as a build-time dependency to the `Makefile`

`Makefile`:
2. bump `COMPONENT_REVISION` to 2
3. add `COMPONENT_FMRI`, `COMPONENT_SUMMARY`, `COMPONENT_CLASSIFICATION`, and `COMPONENT_LICENSE` with values from the manifest
6. drop `build/install/test`
7. add `COMPONENT_TEST_MASTER` and specify `results-all.master`
8. add standard `COMPONENT_TEST_TRANSFORMS`
9. add missing `library/perl-5/http-date` dependency
10. `gmake REQUIRED_PACKAGES`

`File-Listing-PERLVER.p5m`:
5. add the runtime dependency on the same version of perl
6. add the commented-out stuff for the `non-PLV` version
7. add a runtime dependency on `http-date` for the same perl version

### DIFF
--- a/components/perl/File-Listing/Makefile
+++ b/components/perl/File-Listing/Makefile
@@ -22,30 +22,55 @@
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2014, Alexander Pyhalov. All rights reserved.
+# Copyright (c) 2022 Tim Mooney.  All rights reserved.
 #
 
+BUILD_STYLE=makemaker
+BUILD_BITS=32_and_64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		File-Listing
 COMPONENT_VERSION=	6.04
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 IPS_COMPONENT_VERSION=	6.4
+COMPONENT_FMRI=		library/perl-5/file-listing
+COMPONENT_SUMMARY=	File::Listing - parse directory listing
+COMPONENT_CLASSIFICATION=Development/Perl
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
-COMPONENT_PROJECT_URL=	http://search.cpan.org/~gaas/File-Listing/
+COMPONENT_PROJECT_URL=	https://metacpan.org/pod/File::Listing
 COMPONENT_ARCHIVE_HASH=	\
     sha256:1e0050fcd6789a2179ec0db282bf1e90fb92be35d1171588bd9c47d52d959cf5
-COMPONENT_ARCHIVE_URL=	http://search.cpan.org/CPAN/authors/id/G/GA/GAAS/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_URL=	https://cpan.metacpan.org/authors/id/P/PL/PLICEASE/$(COMPONENT_ARCHIVE)
+COMPONENT_LICENSE=	Artistic
 
-include $(WS_TOP)/make-rules/prep.mk
-include $(WS_TOP)/make-rules/makemaker.mk
-include $(WS_TOP)/make-rules/ips.mk
+# until all perl modules have been rebuilt with 5.34, each module sets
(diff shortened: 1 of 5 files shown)


components/perl/File-Listing/Makefile
```diff
@@ -22,30 +22,53 @@
...
+# filter out timing information and anything else that varies between builds
+#
+COMPONENT_TEST_TRANSFORMS += \
+	'-e "1,/test_harness/d"' \
+	'-e "s/,  *[0-9]* wallclock.*//"'
 
-install:	$(INSTALL_32_and_64)
+# because of the runtime dependency on http-date, it must be installed
+# at build time to run the tests
+REQUIRED_PACKAGES += library/perl-5/http-date
```
AndWac: I think the correct way is to have dependencies on real packages not the meta package:
REQUIRED_PACKAGES += library/perl-5/http-date-522
REQUIRED_PACKAGES += library/perl-5/http-date-524
REQUIRED_PACKAGES += library/perl-5/http-date-534

timmooney: In the `Makefile`, if you use the meta-package, it results in `gmake prep` automatically pulling in the meta package, which also means that the versioned package for each version of perl we're currently building is also installed.  The advantage to that is build dependencies won't need to be adjusted every time we add or retire a perl version.  It's effectively a "short cut" for "http-date for all perl versions".

As we've discussed, I've stopped using that syntax in the manifest, instead leaving a commented out block about it.  I think you and I are in agreement about not liking having .e.g. `library/perl-5/authen-pam-534` depend upon `library/perl-5/authen-pam`.  The _opposite_ dependency, though, where `library/perl-5/authen-pam` requires `library/perl-5/authen-pam-534` (and `library/perl-5/authen-pam-522` and `library/perl-5/authen-pam-524`) does make sense to have, and it's this dependency relationship that I'm using in the `Makefile`.

Hopefully that makes sense and is clear.

If you feel strongly that we shouldn't be referencing the meta package in build dependencies in the `Makefile`, I'm willing to switch to using the versioned packages in the `Makefile`, but then I would question what purpose the meta-package even serves?

AndWac: I think that meta packages are for convenience purposes of humans only.
If you use meta packages for dependency management you are prone to errors, eg. you won't detect if somebody removes package versions from a meta package. I detected a real problem for the ansible package in the past. Ansible was dependent only on meta packages (python in this case). When we changed ansible's python version to 3.7 lots of dependent packages weren't available for python 3.7 and this stayed undetected by the build and runtime dependency checks.

timmooney: Ok, I'll switch to using versioned dependencies in the `Makefile` for build dependencies.

If I go back and fix the couple of modules where I've already used the meta-package for the build dependency and submit a new PR, do I need to increase `COMPONENT_REVISION` for those packages, or is changing just the build dependency "safe" and we don't need to trigger a package rebuild?

AndWac: Every build needs a change in the package's version information. The version information has different parts:
COMPONENT_VERSION (or more precisely IPS_COMPONENT_VERSION) +
RELEASE_MAJOR
RELEASE_MINOR
COMPONENT_REVISION
(look at ips-buildinfo.mk for their definitions).
These defintions are hierarchical, that means if a definition higher up changes you can reset the definitions below.
Beginning of 2022 I changed the definition of RELEASE_MAJOR to 2022. That makes it possible to reset RELEASE_MINOR and COMPONENT_REVISION to 0, but only one time for each package.
After that you either need a new IPS_COMPONENT_VERSION or a new COMPONENT_REVISION (RELEASE_MAJOR and RELEASE_MINOR can be changed but this happens only for certain circumstances).

So, yes you'll need to increment the COMPONENT_REVISION for every package that have already been touched this year, and you can reset COMPONENT_REVISION for every package you change the COMPONENT_VERSION or that haven't been touched recently.

timmooney:
I pushed a change that switches the `Makefile` to using the versioned packages for the build dependencies, rather than using the meta-package.

PR #7576, opened 2022-01-14
library/perl-5/www-robotrules: rebuild for 5.34

timmooney:
rebuild `www-robotrules` to add support for perl 5.34

Of note:
1. add the missing runtime dependency on `library/perl-5/uri-$(PLV)` to the manifest
2. add the `library/perl-5/uri` (non-versioned) to the build dependencies in the `Makefile`

`Makefile`:
2. bump `COMPONENT_REVISION` to 2
3. add `COMPONENT_FMRI`, `COMPONENT_SUMMARY`, `COMPONENT_CLASSIFICATION`, and `COMPONENT_LICENSE` with values from the manifest
6. drop `build/install/test`
7. add `COMPONENT_TEST_MASTER` and specify `results-all.master`
8. add standard `COMPONENT_TEST_TRANSFORMS`
9. add the `library/perl-5/uri` build dependency so we can run the test suite
10. `gmake REQUIRED_PACKAGES`

`WWW-RobotRules-PERLVER.p5m`:
5. add the missing runtime dependency on `library/perl-5/uri-${PLV)`
6. add the runtime dependency on the same version of perl
7. add the commented-out stuff for the `non-PLV` version

### DIFF
--- a/components/perl/WWW-RobotRules/Makefile
+++ b/components/perl/WWW-RobotRules/Makefile
@@ -22,29 +22,54 @@
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2014, Alexander Pyhalov. All rights reserved.
+# Copyright (c) 2022 Tim Mooney.  All rights reserved.
 #
 
+BUILD_STYLE=makemaker
+BUILD_BITS=32_and_64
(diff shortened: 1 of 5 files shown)


timmooney:
I updated the commit to use the versioned dependencies for URI in the Makefile, instead of the meta-package

PR #7579, opened 2022-01-14
Add pidgin-save-conv-order (plugin for Pidgin IM client)

jimklimov:
Something I missed on OI while migrating my Pidgin setup from another system. Hordes of chat-logs are still copying, so I would test this PR in practice a bit later.


jimklimov:
Seems to work - at least, pidgin recognizes the plugin via installed locally built package, and allows to enable it, and does not crash. Not fully sure (memory fades) how to practically test it though - which order of what it should have retained, compared to not having it. Behaviorally seems like on my other platform so probably working :)

AndWac (approved):

PR #7580, opened 2022-01-14
geany-plugins: rebuild with pygobject-3-39

balrog-nona:
It builds, installs and publishes fine. When installed locally, I used it in geany and it looked to me it worked but please note that I am not a geany user.

### DIFF
--- a/components/editor/geany-plugins/Makefile
+++ b/components/editor/geany-plugins/Makefile
@@ -11,6 +11,7 @@
 
 #
 # Copyright 2016-2017 Predrag Zečević.  All rights reserved.
+# Copyright 2022 Nona Hansel
 #
 
 BUILD_BITS=			32
@@ -18,6 +19,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		geany-plugins
 COMPONENT_VERSION=	1.38
+COMPONENT_REVISION=	1
 COMPONENT_CLASSIFICATION=	System/Text Tools
 COMPONENT_FMRI=	editor/geany/plugins
 COMPONENT_SUMMARY= 	Plugins for Geany editor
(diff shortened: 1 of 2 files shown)


components/editor/geany-plugins/Makefile
```diff
@@ -60,7 +62,7 @@ CONFIGURE_OPTIONS+= --disable-lipsum
 REQUIRED_PACKAGES += developer/vala
 REQUIRED_PACKAGES += gnome/config/gconf
 REQUIRED_PACKAGES += library/desktop/gtkspell
-REQUIRED_PACKAGES += library/python/pygobject
+REQUIRED_PACKAGES += library/python/pygobject-3-39
```
AndWac, line 72: The default python version is 2.7 and it's not sufficient to add a dependency on a python-3.9 package version of pygobject to let the plugins use python-3.9.
It needs also
PYTHON_VERSION= 3.9
and
CONFIGURE_ENV+= PYTHON=$(PYTHON.3.9.64)
Furthermore, you'll need to patch all .py files that start with
#!/usr/bin/env python
to use the correct python version:
#!/usr/bin/python3.9

Also make sure that the plugins are really developed for python3 (#!/usr/bin/env python looks more like python2 is being intended).

balrog-nona: > The default python version is 2.7 and it's not sufficient to add a dependency on a python-3.9 package version of pygobject to let the plugins use python-3.9. It needs also PYTHON_VERSION= 3.9 and CONFIGURE_ENV+= PYTHON=$(PYTHON.3.9.64) Furthermore, you'll need to patch all .py files that start with #!/usr/bin/env python to use the correct python version: #!/usr/bin/python3.9
> 
> Also make sure that the plugins are really developed for python3 (#!/usr/bin/env python looks more like python2 is being intended).

Thank you! I incorporated everything you suggested. I'm glad I know now how to do it in the future.

As for patching all .py files that start with #!/usr/bin/env python. It turned out there was just one such a file: `geany-plugins-1.38/geanypy/geany/console.py` but the whole geanypy plugin had to be disabled because it requires Python2.7. You were right.

When I tested it now, I tried plugins `auto-close`, `auto-mark` and they worked well. Yet, I had a problem with `Spell check` - I get warning into terminal that Stored language ('en_US') could not be loaded. When I try to uncheck this plugin in plugin menu, geany closes without warning. But I get the same warning and behavior even with current build of plugins. It may be something wrong on my machine but I didn't figure out how to fix it.

Any additional testing by a geany user will be most welcomed.

PR #7595, opened 2022-01-15
Clamav - updated to latest version 0.104.2

fritzkink:


AndWac:
This PR contains also an unrelated old commit. You'll need to fix that.

AndWac (approved):

PR #7620, opened 2022-01-17
build-essential: Replace illumos-gcc by gcc-10

mtelka:
illumos-gcc is not needed for building illumos-gate, but gcc-10 is.

### DIFF
--- a/components/meta-packages/build-essential/Makefile
+++ b/components/meta-packages/build-essential/Makefile
@@ -17,6 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		build-essential
 COMPONENT_VERSION=	1.0
+COMPONENT_REVISION=	1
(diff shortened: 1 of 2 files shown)


components/meta-packages/build-essential/Makefile
```diff
@@ -17,6 +17,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		build-essential
 COMPONENT_VERSION=	1.0
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	A meta-package that installs common development tools such as gcc
```
AndWac, line 21: No need to change the COMPONENT_REVISION as RELEASE_MAJOR has been changed to 2022 recently.

AndWac: Nevertheless we need a change in the Makefile in order to get jenkins buiding it.

PR #7623, opened 2022-01-17
onigurama - add new package (needed for php81)

fritzkink:


AndWac:
We already have text/oniguruma

PR #7638, opened 2022-01-19
Upgrade of python-dbus to 1.2.18 with python 3.5, 3.7, and 3.9

jgmills:
Python-dbus is a set of python bindings for D-Bus.  This PR upgrades the software from 1.2.16 to 1.2.18 and adds packages for python 3.7 and 3.9 .

As part of the upgrade, the build style was changed from configure to setup.py, as the setup.py style does iterate over python versions.  Unfortunately, this style neglected to install some files.  These are now installed by the Makefile.

The test target now works, showing 38 passed tests and 4 warnings.

The Makefile installs pkgconfig files in /usr/lib/pkgconfig .  For python 3.5 only, it also installs them in /usr/lib/amd64/pkgconfig .

The new patch corrects errors in one python file.

The build, install, and publish steps were successful, as were the built-in tests and the package install.

### DIFF
--- a/components/python/dbus-python/Makefile
+++ b/components/python/dbus-python/Makefile
@@ -11,21 +11,23 @@
 #
 # Copyright 2017 Alexander Pyhalov
 # Copyright 2019 Michal Nowak
+# Copyright 2022 Gary Mills
 #
 
 BUILD_BITS=		64
+BUILD_STYLE=		setup.py
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		dbus-python
-COMPONENT_VERSION=	1.2.16
-COMPONENT_PROJECT_URL=	http://www.freedesktop.org/wiki/Software/DBusBindings/
+COMPONENT_VERSION=	1.2.18
+COMPONENT_PROJECT_URL=	https://www.freedesktop.org/wiki/Software/DBusBindings/
 COMPONENT_SUMMARY=	Python bindings for D-Bus
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
 COMPONENT_ARCHIVE_HASH=	\
-	sha256:11238f1d86c995d8aed2e22f04a1e3779f0d70e587caffeab4857f3c662ed5a4
-COMPONENT_ARCHIVE_URL=	http://dbus.freedesktop.org/releases/dbus-python/$(COMPONENT_ARCHIVE)
+	sha256:92bdd1e68b45596c833307a5ff4b217ee6929a1502f5341bae28fd120acf7260
+COMPONENT_ARCHIVE_URL=	https://dbus.freedesktop.org/releases/dbus-python/$(COMPONENT_ARCHIVE)
 COMPONENT_FMRI=		library/python/python-dbus
 COMPONENT_CLASSIFICATTION= Development/Python
 COMPONENT_LICENSE=	MIT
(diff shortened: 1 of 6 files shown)


components/python/dbus-python/Makefile
```diff
@@ -33,20 +35,71 @@ COMPONENT_LICENSE_FILE=	COPYING
...
+#======================== 38 passed, 4 warnings in 0.27s =====================
+
+# Added run-time dependencies
 REQUIRED_PACKAGES += runtime/python-35
+REQUIRED_PACKAGES += runtime/python-37
+REQUIRED_PACKAGES += runtime/python-39
+
+# Added test dependencies for all python versions
+REQUIRED_PACKAGES += library/python/pytest
+REQUIRED_PACKAGES += library/python/py
```
AndWac: The reference to meta packages is wrong here. We need to be explicit here, as we do above for the runtime versions.
Otherwise missing dependencies may stay undetected or additional packages may be installed (eg. for pyhon 2.7 here). What you achieve with the use of meta packages here is that all available packages will be install but a meta package does not guarantee for a version of a certain python version. I consider meta packages as a convenience for humans.

components/python/dbus-python/Makefile
```diff
@@ -33,20 +35,71 @@ COMPONENT_LICENSE_FILE=	COPYING
 
 PATH=$(PATH.gnu)
 
-PYTHON_VERSION = 3.5
-PYTHON_VERSIONS = 3.5
+PYTHON_VERSIONS = 3.5 3.7 3.9
```
AndWac: This can be expressed as
PYTHON_VERSIONS= $(PYTHON3_VERSIONS)

components/python/dbus-python/Makefile
```diff
@@ -11,20 +11,22 @@
...
 BUILD_BITS=		64
+BUILD_STYLE=		setup.py
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		dbus-python
-COMPONENT_VERSION=	1.2.16
+COMPONENT_VERSION=	1.2.18
 COMPONENT_PROJECT_URL=	http://www.freedesktop.org/wiki/Software/DBusBindings/
 COMPONENT_SUMMARY=	Python bindings for D-Bus
```
AndWac, line 25: Please use https where possible.

components/python/dbus-python/Makefile
```diff
@@ -33,20 +35,71 @@ COMPONENT_LICENSE_FILE=	COPYING
...
+
+# Added run-time dependencies
 REQUIRED_PACKAGES += runtime/python-35
+REQUIRED_PACKAGES += runtime/python-37
+REQUIRED_PACKAGES += runtime/python-39
+
+# Added test dependencies for all python versions
+REQUIRED_PACKAGES += library/python/pytest
+REQUIRED_PACKAGES += library/python/py
+REQUIRED_PACKAGES += library/python/pluggy
```
AndWac: You don't need to add py or pluggy to the test dependencies as they are runtime dependencies of pytest.
(I have learned and fixed that recently.)

alarcher: Probably we should introduce TEST_REQUIRED_PACKAGES that install only deps if tests are invoked. However this is not going to work for all packages as some of them build tests only if the testing framework is detected.

AndWac:
Am 20.01.22 um 12:28 schrieb Aurélien Larcher:
>
>
>
> In components/python/dbus-python/Makefile
> <https://github.com/OpenIndiana/oi-userland/pull/7638#discussion_r788672529>:
>
> >  REQUIRED_PACKAGES += runtime/python-35
> +REQUIRED_PACKAGES += runtime/python-37
> +REQUIRED_PACKAGES += runtime/python-39
> +
> +# Added test dependencies for all python versions
> +REQUIRED_PACKAGES += library/python/pytest
> +REQUIRED_PACKAGES += library/python/py
> +REQUIRED_PACKAGES += library/python/pluggy
>
> Probably we should introduce TEST_REQUIRED_PACKAGES that install only
> deps if tests are invoked. However this is not going to work for all
> packages as some of them build tests only if the testing framework is
> detected.
>
I am still learning about python's eco system and I am not sure whether
I am right, but I think we need something that interpretes setup.py,
setup.cfg and probably PKG-INFO files of python packages (similar to
pipdeptree) and generates dependencies (eg. by calculating the
transitive closure) for it.

The more I work on python packages the more I wish we could drop all
python versions but one. But that would probably also mean to drop
Hipsters rolling release model and switch to an ordinary release model
where we can have baselines per release.

alarcher:
> I am still learning about python's eco system and I am not sure whether I am right, but I think we need something that interpretes setup.py, setup.cfg and probably PKG-INFO files of python packages (similar to pipdeptree) and generates dependencies (eg. by calculating the transitive closure) for it. 

Our build system should learn to generate dependencies from the python metadata and the pypi website, I agree.

alarcher:
> > I am still learning about python's eco system and I am not sure whether I am right, but I think we need something that interpretes setup.py, setup.cfg and probably PKG-INFO files of python packages (similar to pipdeptree) and generates dependencies (eg. by calculating the transitive closure) for it.
> 
> Our build system should learn to generate dependencies from the python metadata and the pypi website, I agree.

And I actually introduced pipdeptree to help with that.

components/python/dbus-python/Makefile
jgmills: I've changed that macro.

components/python/dbus-python/Makefile
jgmills: I've changed both of those, and verified that they worked.

components/python/dbus-python/Makefile
jgmills: I've removed those two dependencies.

components/python/dbus-python/Makefile
jgmills: I've changed those dependencies to specify the three version-dependent versions.

PR #7640, opened 2022-01-20
runtime/perl-534/modules/sun-solaris: new FMRI for sun-solaris for perl 5.34

timmooney:
~~The changes here are far more complicated than any of the other perl module rebuilds done for perl 5.34.  This should be tagged "needs review", and it should be very carefully considered before merging.~~

This PR **should not** be merged.  As discussed in the comments below, the copy of `sun-solaris` that's in `oi-userland` is effectively a decoy.  It's not what's used to build `runtime/perl-522/module/sun-solaris`.  That module is built as part of our build process of `illumos-gate`, so the appropriate place to change what version of perl is involved is in `oi-userland/components/openindiana/illumos-gate`, specifically by adjusting what environment is set via the `Makefile`.

I'll be closing this PR, but the comments from Alexander may be useful reference for the actual PR #7733

### DIFF
--- a/components/perl/sun_solaris/Makefile
+++ b/components/perl/sun_solaris/Makefile
@@ -11,26 +11,45 @@
 
 #
 # Copyright 2013 Alexander Pyhalov.  All rights reserved.
+# Copyright 2022 (c) Tim Mooney.  All rights reserved
 #
 
+BUILD_STYLE=makemaker
+BUILD_BITS=32_and_64
 include ../../../make-rules/shared-macros.mk
 
-COMPONENT_NAME= sun-solaris
-
-COMPONENT_SUMMARY=Perl Sun::Solaris Modules
-COMPONENT_SRC= src
-
-PERL_VERSIONS = 5.16
+COMPONENT_NAME=		sun-solaris
 # Illumos component, no explicit version
-COMPONENT_VERSION= 5.11
+COMPONENT_VERSION=	0.5.11
+COMPONENT_REVISION=	1
+# keep FMRI in the manifest for now
+#COMPONENT_FMRI=
+COMPONENT_SUMMARY=	Perl Sun::Solaris Modules
+COMPONENT_CLASSIFICATION=Development/Perl
+COMPONENT_SRC=		src
+COMPONENT_LICENSE=	Artistic
+COMPONENT_LICENSE_FILE=	sun-solaris.license
 
-include ../../../make-rules/prep.mk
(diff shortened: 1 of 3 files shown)


pyhalov:
The source of truth for these parts is illumos-gate. Perhaps, correct option here is just to teach gate build perl modules for several perl versions. On other side, it seems nothing besides scripts in base system (and intrd) requires them.

timmooney:
> The source of truth for these parts is illumos-gate.

Thanks Alexander!  I knew you would know the answer to some of these mysteries.  I really appreciate you taking the time to reply here.

With your hint, I found at least the manifest for it in `illumos-gate`:

[illumos-gate runtime/perl-###/module/sun-solaris](https://github.com/illumos/illumos-gate/blob/master/usr/src/pkg/manifests/runtime-perl-module-sun-solaris.p5m)

So why do we have a component for it in `oi-userland` then?   Even if we keep (a copy of) the source code for the various modules (`Exacct`, `Ucred`, `Kstat`, etc.), wouldn't we be better off deleting our `Makefile` and manifest, so they don't cause confusion for future contributors?

Also, how do I go about getting the component from `illumos-gate` to update to our perl 5.34 then?  Is the first step to update the `illumos.sh` script, or do I need to start elsewhere?  I still have your email from several years ago from `oi-discuss` where you outlined many of the (unrelated) steps to getting a new perl into OI, but some of those steps were waiting on patches and things that (I'm guessing) have now almost certainly been incorporated.

pyhalov:
illumos-gate builds perl modules only for one version, specified as PERL_VERSION in illumos.sh. It doesn't have rules to build and deliver several versions. So if you want to deliver these modules for some other perl version, you should either teach illumos-gate to build them or do it in some external way, for example, as done here.
Updating illumos-gate perl, IIRC, was a chicken-and-egg problem at least some time ago. You should build illumos-gate with new perl. It will build perl modules, but will likely fail on later stages. But you'll be able to copy these module to build system. After performing this operation, illumos-gate can be built as usual.

timmooney:
My personal opinion is that we only need it for one version of perl, not all of them.  If it were up to just me, I don't think it's needed to teach illumos-gate how to build for all perl versions.

I just want to be able to get to the point where we can switch it to depend upon perl 5.34, and then obsolete perl 5.22 and perl 5.24.

Thanks for the info about the chicken-and-egg problem with illumos gate.  I've never built all of illumos-gate before, but I may give that a try this weekend to see what happens.

PR #7641, opened 2022-01-20
bcc: fix build with gcc-10

alarcher:
The install build: target should be kept as a workaround to the build system.


Toasterson (approved):
LGTM

AndWac (approved):

PR #7655, opened 2022-01-24
top-down builds occasionally fail to install dependencies when gmake is called with -j(2 or more)

dbh201:
The build system fails on a top-level `gmake publish -j(2 or more)` when multiple jobthreads attempt to install dependencies at the same time.
Fixed by adding a wait loop which tests `pkg` for return code 7 (busy).


alarcher:
Nice :)

AndWac:
Please squash your commits.

PR #7657, opened 2022-01-24
shared-macros.mk: fixed overwriting of certain envvars

dbh201:
The Makefiles for many components seem to expect macro expansion to occur on assignment, for example:

- Setting `PYTHON_VERSION` to "3.9" should cause `PYTHON` to evaluate to `/usr/bin/python3.9`
- Setting `GCC_VERSION` to "10" should cause `GCC_ROOT` to evaluate to `/usr/gcc/10` 

Unfortunately, since macro expansion only occurs at the exact point `make-rules/shared-macros.mk` is included, these configuration variables are often ignored. In addition,  `make-rules/shared-macros.mk` also **overwrites** `GCC_VERSION`, `PYTHON_VERSION`, and others. This causes build errors.

The included commit resolves _part_  of the issue by conditionally assigning default values for some configuration variables. Makefiles will need to be updated to have configuration variables assigned _before_ including `make-rules/shared-macros.mk`.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -159,8 +159,8 @@ else
 MK_BITS=$(strip $(BUILD_BITS))
 endif
 
-PYTHON_VERSION =	2.7
-PYTHON_VERSIONS =	2.7
+PYTHON_VERSION?=	2.7
+PYTHON_VERSIONS?=	2.7
 
 PYTHON2_VERSIONS = 2.7
 PYTHON2_VERSION = 2.7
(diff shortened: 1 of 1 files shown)


dbh201:
Other config variables may need to be conditional. I'll update as my oi-userland build progresses.

Toasterson:
Hmmmm, tht might explain a lot of build errors. Good catch thanks.

alarcher:
Not really, what matters is where you set the variables.

AndWac:
As Aurélien already mentioned: This helps only for when you have set the variables before you have included common.mk.
We typically set them later. Can you provide examples where you have been hit by this?

dbh201:
Looking at the git history here, I had build failures on mongodb-44, node16, couchdb-31, and meson. Not ruling out a PEBKAC here; let me try cleaning them and building again. 
Just to be clear: for Makefiles with GCC_VERSION, add it between `shared-macros.mk` and `common.mk`?

AndWac:
Yes, as these versions are only set in shared-macros.mk.

dbh201:
Builds are working. I'll keep an eye on where the versions are set next time, thanks guys.

PR #7659, opened 2022-01-24
mongodb-44: ccache fix

dbh201:
(see #7648 #7649) Patch allowing scons to pass necessary ccache envvars to spawned processes.


AndWac:
Please squash your commits and get rid of the superfluous 2nd commit message.

AndWac:
Your branch has problems. The commit now contains several files from other packages.

PR #7662, opened 2022-01-25
gthumb: move to 64-bit, fix dependencies and gcc vector detection

alarcher:


alarcher:
@AndWac Some testing needed before merging.

alarcher:
Tested working.

PR #7674, opened 2022-01-29
php-81 - current version of php (based on openssl 1.1)

fritzkink:


fritzkink:
wrong approach as already merged

PR #7675, opened 2022-01-29
php-81 fix sed to cover all conditions to build correctly with openss…

fritzkink:
…l 1.1


fritzkink:
forgot to rebase against upstream

PR #7681, opened 2022-01-31
 php-812 - current version of php (based on openssl 1.1) includes fix…

fritzkink:
… for mod_php81.so crash 
former php 8.1.1 package has a wrong php81.conf and mod_php81.so crashes during apache startup


AndWac:
You have put these files into a new folder php-812, leaving the older version 8.1.1 back in its original folder php-8_1.
That's problematic. You should re-use the php-8_1 folder for the update.

fritzkink:
remove pull request to create a new one in php-8_1 folder

fritzkink:
remove pull request to create a new one in php-8_1 folder

PR #7682, opened 2022-01-31
 php-812 - current version of php (based on openssl 1.1) includes fix…

fritzkink:
… for mod_php81.so crash
former php 8.1.1 package has a wrong php81.conf and mod_php81.so crashes during apache startup


fritzkink:
to problematic

PR #7692, opened 2022-02-01
Make GNU-M4 64bit only

klausz65:


AndWac:
Doesn't the sample-manifest also change?

PR #7710, opened 2022-02-04
Update pkglint userland checker, the old presents false positives for run…

cgrzemba:
Our old pkglint userland checker complains about runpaths which are correct. So we should update this tool.
Thanks to Oracle!

e.g.
ERROR userland.action001.3        bad RUNPATH, 'opt/SUNWsamfs/bin/archive' includes '/opt/SUNWsamfs/lib'
ERROR userland.action001.3        bad RUNPATH, 'opt/SUNWsamfs/bin/notify' includes '/opt/SUNWsamfs/lib'
ERROR userland.action001.3        bad RUNPATH, 'opt/SUNWsamfs/bin/request' includes '/opt/SUNWsamfs/lib'

### DIFF
--- a/tools/pkglintrc
+++ b/tools/pkglintrc
@@ -25,7 +25,7 @@
 pkglint.ext.content = pkglint.userland
 
 # Used by the publisher_in_fmri method in .../tools/python/pkglint/userland.py
-userland.manifest.allowed_pubs = userland userland-localizable
+userland.manifest002.allowed_pubs = userland userland-localizable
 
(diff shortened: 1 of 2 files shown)


cgrzemba:
I will do some additional checks.

cgrzemba:
There is some more work to do.
This version will not accept runpaths like /usr/gcc/.. and /usr/mariadb/..., which are correct.
We can allow this per default. The most complete solution would be to allow this paths, if there resolved dependencies for this.

PR #7714, opened 2022-02-05
opendkim - DKIM (Domain Keys Identified Mail) sender authentication s…

fritzkink:
…ystem v2.10.3

### DIFF
--- a/components/mail/opendkim/Makefile
+++ b/components/mail/opendkim/Makefile
@@ -1,0 +1,107 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Friedrich Kink
+#
+
+BUILD_BITS=		64 # for binaries or 32_and_64 for libraries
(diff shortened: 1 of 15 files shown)


AndWac:
This package depends on OpenSSL-1.0. Did you try to build it with OpenSSL-1.1?

components/mail/opendkim/Makefile
```diff
@@ -0,0 +1,104 @@
...
+REQUIRED_PACKAGES += library/jansson
+REQUIRED_PACKAGES += library/libmemcached
+REQUIRED_PACKAGES += library/libmilter
+REQUIRED_PACKAGES += library/opendbx
+REQUIRED_PACKAGES += library/openldap
+REQUIRED_PACKAGES += library/security/openssl
+REQUIRED_PACKAGES += library/tre
+REQUIRED_PACKAGES += runtime/lua
+REQUIRED_PACKAGES += runtime/perl-534
+REQUIRED_PACKAGES += shell/ksh93
```
AndWac, line 105: When I run "gmake REQUIRED_PACKAGES" on my system it'll replaces runtime/perl-534 with runtime/perl-524.

fritzkink:
Your comment about openssl 1.1 made me rethink about it ;-). Initially I thought this package is to old for openssl 1.1 because it was released before openssl 1.1 was started. But I found some patches which made the build with openssl 1.1 successful.

components/mail/opendkim/Makefile
fritzkink: with new setup it is the same on my system.

PR #7715, opened 2022-02-05
graphviz: update to 2.50.0 and drop 32-bit

mtelka:


components/image/graphviz/Makefile
```diff
@@ -23,92 +23,106 @@
...
 				automake --add-missing)
 
-# Can't find headers without this.
-COMPONENT_PRE_CONFIGURE_ACTION =        ($(CLONEY) $(SOURCE_DIR) $(@D))
-
-CONFIGURE_ENV +=	CPPFLAGS="$(CPPFLAGS)"
+# configure
+CONFIGURE_ENV +=	LIBPOSTFIX="/$(MACH64)"
+CONFIGURE_ENV +=	JAVA_INCLUDES="-I$(JAVA_HOME)/include -I$(JAVA_HOME)/include/solaris"
 CONFIGURE_ENV +=	PERL=$(PERL)
```
AndWac, line 62: JAVA_INCLUDES indicates that there is an undetected build dependency on Java, eg. OpenJDK8.
You should add that to the REQUIRED_PACKAGES.

components/image/graphviz/Makefile
```diff
@@ -120,25 +134,17 @@ REQUIRED_PACKAGES += library/gd
...
 REQUIRED_PACKAGES += library/zlib
 REQUIRED_PACKAGES += print/filter/ghostscript
 REQUIRED_PACKAGES += runtime/lua
-REQUIRED_PACKAGES += runtime/perl-522
-REQUIRED_PACKAGES += runtime/ruby-23
-REQUIRED_PACKAGES += runtime/tk-8
+REQUIRED_PACKAGES += runtime/perl-534
+REQUIRED_PACKAGES += runtime/ruby-26
+REQUIRED_PACKAGES += shell/ksh93
 REQUIRED_PACKAGES += system/library
```
AndWac, line 148: Probably also tcl and tk are build dependencies, aren't they?

components/image/graphviz/Makefile
mtelka: Yes, openjdk8 was missing from REQUIRED_PACKAGES, so I added it.  OTOH, the explicit JAVA_INCLUDES here is because we need two include dirs and there is no other way (AFAIK) to do so.  Something similar was done earlier too.  I added a comment for JAVA_INCLUDES to make this more clear.

components/image/graphviz/Makefile
mtelka: Yes, tcl-8 was missing.  Added.  Thank you for the hint.

For tk I'm not sure.  I didn't find anything in the build log suggesting tk is really used during the build, but maybe I just didn't looked hard enough.

AndWac:
Thanks, can you squash your commits before I merge, please?

mtelka:
squashed

PR #7729, opened 2022-02-08
samqfs: add to openindiana

cgrzemba:
This adds SamQFS to Openindiana.
This based an the source which was released as OSS from Sun Microsystems in 2009 like OpenSolaris and Solaris Cluster Express.

For a short description of SamQFS see:
https://en.wikipedia.org/wiki/QFS

The source is buildable with ACSLS. The support for LTO8 is added.

### DIFF
--- a/components/openindiana/samqfs/Makefile
+++ b/components/openindiana/samqfs/Makefile
@@ -1,0 +1,55 @@
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Carsten Grzemba
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME= samqfs
+COMPONENT_VERSION= 5.0.1
+BRANCH_VERSION = 2022.0.0.0
+# COMPONENT_REVISION= 0
+COMPONENT_FMRI= system/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION= System/File System
+COMPONENT_PROJECT_URL= https://github.com/cgrzemba/$(COMPONENT_NAME)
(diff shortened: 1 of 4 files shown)


components/openindiana/samqfs/Makefile
```diff
@@ -0,0 +1,55 @@
...
+include $(WS_MAKE_RULES)/justmake.mk
+include $(WS_MAKE_RULES)/ips.mk
+
+COMPONENT_BUILD_ARGS = DEBUG=no MYSQL_VERSION=mariadb/10.3 SAMQFS_VERSION=$(COMPONENT_VERSION)-$(BRANCH_VERSION)
+
+COMPONENT_INSTALL_ARGS = install DEBUG=no DESTDIR=$(PROTO_DIR) SAMQFS_VERSION=$(COMPONENT_VERSION)-$(BRANCH_VERSION)
+
+build:          $(BUILD_32)
+
+install:          $(INSTALL_32)
```
Toasterson, line 42: Hmmm this would bind ourselves to 32bit things again. Can we have it in 64bit

Toasterson (approved):
Metadata looks good, but I have no clue what Hardware/Setup would be needed to test it. I assume you Tested it and are using it :) So I am ok with merging. Just would like it 64bit if it's possible.

components/openindiana/samqfs/Makefile
cgrzemba: This is a bigger task. The userland parts is still on 32bit, also on current Oracle HSM.
The kernel components are 64bit of course.

cgrzemba: For test, 
create 2 Zvols, 
a small for metadata samfsmm
a bigger for data /samfsmr 

in /etc/opt/SUNWsamfs/mcf add:

samfs1            10 ma samfs1
/dev/zvol/dsk/rpool/samfsmm 11 mm samfs1 -
/dev/zvol/dsk/rpool/samfsmr 12 mr samfs1 -

run for config check
    # sam-fsd 
for create filesystem
    # sammkfs samfs1
than mount sam file system
    # mount -T samfs samfs1 /mnt
and test.

cgrzemba: For test with tape library you can connect a quadstor VTL via iscsi, but this is a little bit longer story ...

PR #7742, opened 2022-02-11
imagemagick: update to 6.9.12-38

mtelka:
There are five packages with imagemagick in REQUIRED_PACKAGES: tango-icon-theme, rss-glx, awesome, inkscape, freeciv.  I do not know if they needs to be rebuilt or not after the imagemagick update, but to be safe I'll create new PR in few minutes to get them rebuild.


components/image/imagemagick/pkg5
```diff
@@ -20,6 +20,7 @@
         "library/openexr",
         "library/zlib",
         "print/filter/ghostscript/fonts/gnu-gs-fonts-std",
+        "shell/ksh93",
         "system/library",
         "system/library/freetype-2",
         "system/library/g++-7-runtime",
```
AndWac, line 25: pkg5 seem to need another update. You have remove library/fftw-3 from the REQUIRED_PACKAGES but it's still here.

mtelka: Sorry for that.  Fixed.

PR #7753, opened 2022-02-14
resection manual pages per IPD4

richlowe:
**THIS IS NOT REALLY A PULL REQUEST**

This contains the obvious things I can see to rename manual pages per IPD4 (which is largely to stop renaming them).
Section 1M, and subsections of 7 are obvious.  Sections 4 and 5 are troublesome because it's never clear which software never was correct in the first place (and so "fixing it" would make it wrong), and which was.  I only did the obvious pieces.

I have also not done section 7 in X11.  I think someone from OI re-syncing X11 with Oracle would be better, at this point.  I have no idea what state things are in, or what I would be making worse.  This is also, somewhat, true for almost all other software here.  It seems like in almost all cases updating it from upstream would be a good idea too.

Anyway.  This is not a pull request because I haven't tested any of it (nor, really, do I have the ability to).  But for OI people syncing up with illumos #14443 / IPD4 this might come in useful.


AndWac:
For every updated package we also need to bump the COMPONENT_REVISION. If the Makefile isn't changed jenkins won't build the package.

alarcher:
> For every updated package we also need to bump the COMPONENT_REVISION. If the Makefile isn't changed jenkins won't build the package.

Auto-bumping the revision can be achieved by executing gmake update in the component directory.
For very few exotic components the conversion of the Makefile may have border effects but this is very rare, usually the gain of time running such automated procedure outweights the few manual changes required afterwards.

AndWac:
I have created #7945 so you don't need to touch your commits. If you are ok with it you can remove the draft state and I will merge both PRs. I am fine with an incremental fix for man pages so we can deal with necessary changes for X11 later.

richlowe:
I don't think I really tested what I did at all, so I wouldn't merge it
as-is.  There's a better than usual chance it won't work.

AndWac:
Ah ok, so how shall we proceed with this?

richlowe:
So my perhaps unhelpful idea was that I'd push this since I had it, and anyone who was wanting to take out a piece of this could start from here rather than from scratch.  I'm not sure how helpful that is, but it seemed more realistic than me being able to build a complete oi-userland, and more helpful than doing nothing (the bit I could easily have been wrong about)

AndWac:
Yes, that sounds like a good idea. I wasn't sure whether you wanted to finish it and take all the credits for your work.

PR #7756, opened 2022-02-14
bacula: update to 11.0.5 and 64bit only

fritzkink:
now based on openssl 1.1
removed make_catalog_backup.pl because script included in original package
testers would be good ;-)


components/sysutils/bacula/Makefile
```diff
@@ -10,39 +10,34 @@
...
-COMPONENT_REVISION= 1
-COMPONENT_SUMMARY= Bacula (network backup solution)
-COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
-COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.gz
-COMPONENT_ARCHIVE_HASH= \
-  sha256:df0f8525ecdba380e74b537ca71adf3b7596593f02189b59ab0ebffc85ccc8de
-COMPONENT_ARCHIVE_URL= \
-  http://sourceforge.net/projects/bacula/files/bacula/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)/download
-COMPONENT_PROJECT_URL = http://www.bacula.org
+test: $(NO_TESTS)
```
AndWac: This doesn't work. Use TEST_TARGET= $(NO_TESTS) instead.

AndWac (approved):

components/sysutils/bacula/Makefile
fritzkink: did not check just took over but it is fixed now.

PR #7761, opened 2022-02-15
add victor-mono font

AndWac:


hoewweken:
Hello! This package is not available yet on pkg.openindiana.org. Should it ?

AndWac:
Am 17.02.22 um 10:51 schrieb hoewweken:
>
> Hello! This package is not available yet on pkg.openindiana.org.
> Should it ?
>
It should have. Thanks for the notification. I build it manually and
re-run jenkins. It should be available now.

PR #7764, opened 2022-02-16
Retirement of logilab-astng and logilab-common

jgmills:
The logilab-astng and logilab-common packages were formerly dependents of pylint.  They are now replaced in that role by the astroid package.  In detail, they are library/python/logilab-astng, library/python/logilab-astng-27, library/python/logilab-common, and library/python/logilab-common-27 .  These packages are made obsolete by this PR.

In addition, this PR removes their directories, which were components/python/logilab-astng and components/python/logilab-common .

Both of the history files, contained in these directories, are now merged into OI's global history file.  In doing this, I assumed that the order of lines in that history file did not matter.  I used the command "env LANG=C sort -n ..." as it seemed to produce the best approximation to the original order of the global history file.


components/meta-packages/history/history
```diff
@@ -941,10 +924,16 @@ library/python-2/gnome-python@2.28.1,5.11-2018.0.0.0
...
 library/python-2/instant-27@1.3.0,5.11-2020.0.1.0
 library/python-2/instant@1.3.0,5.11-2020.0.1.0
-library/python-2/libbe@0.5.11,5.11-2016.0.1.0
 library/python-2/libbe-27@0.5.11,5.11-2016.0.1.0
+library/python-2/libbe@0.5.11,5.11-2016.0.1.0
 library/python-2/libxml2-24@2.7.6,5.11-2015.0.2.0
 library/python-2/libxsl-24@1.1.26,5.11-2015.0.2.0
+library/python-2/logilab-astng-26@0.24.0,5.11-2016.0.1.0
+library/python-2/logilab-astng-27@0.24.0,5.11-2016.0.1.0 library/python/logilab-astng-27
+library/python-2/logilab-astng@0.24.0,5.11-2016.0.1.0 library/python/logilab-astng
```
AndWac: The 2nd column here is a target. It is being used to be replace the old package (1st column) with the new version (2nd column) if it is installed on a system. In case of logilab-astng and logilab-common it doesn't make sense anymore as you are obsoleting them now.
So please remove the 2nd column for the effected logilab-astng and logiliab-common entries.

jgmills: Okay, this updated PR should have the offending column removed now.

PR #7779, opened 2022-02-20
Update libuv to 1.43.0

shuryanc:
Update libuv 1.43.0 to prepare for porting megasync.


AndWac:
Please also run "gmake test" and update the test results accordingly.

components/library/libuv/Makefile
```diff
@@ -12,20 +12,20 @@
 # Copyright 2020, Michal Nowak
 #
 
-BUILD_BITS=		64
+BUILD_BITS=		32_and_64
```
AndWac: Why did you change this? The only actual consumer seems to be network/dns/bind which is 64-bit only.
Do you want to use libuv in 32-bit?
Furthermore, if you make changes like this, you'll also need to update manifest/sample-manifest.p5m by running "gmake sample-manifest" and check for its changes (compared to the old version). Then you either need to update the package's manifest (libuv.p5m in this case). Without the last step the addition of 32-bit files won't make it into the package.

shuryanc:
The application that I am going to port supports both 32 bit and 64 bit. But I am going to port 64 bit only. I changed to 32_and_64 just for in case other people may need it. If its not allowed, I'll change back to 64 bit. Let me know if I should go forward or keep it as 64 bit only so that I will do the corresponding changes.

AndWac:
If you don't need 32-bit then it's better to stay at 64-bit only here. We try to get rid of 32-bit wherever possible because it's costs more time and efforts than it brings.
You'll need to check whether libuv still works as expected before we can integrate your update. As there is only one consumer, bind, it shouldn't be too hard to test.

shuryanc:
I changed back to 64 only, built both libuv and bind. Then tested bind. Do I need to push the results-64.snapshot of bind testing with my change in libuv together?

AndWac:
I can't see your changes yet.

shuryanc:
Yes, because I tried to ask before I push. In the meanwhile, I am porting some other packages.

shuryanc:
Pushed libuv update.

AndWac:
We may want to include the patch from omnios or wait until it is merged upstream:
https://github.com/libuv/libuv/pull/3542

AndWac:
Superceded by #7898

PR #7785, opened 2022-02-21
Add PostgreSQL 14.2

mvwieringen:
This adds posgresql14 with the same ISM/DISM (e.g. (dynamic) Intimate Shared Memory) setup by default as what will land
in Omnios soon too. Cloned pg12 to pg14 build and repopulated the different p5m files from the changes found between pg12
and pg14 sample-manifest. Installed the package and started the database which works well and shows DISM as being used.

When this is integrated I will upstream a citus version of the latest version for pg14 which I also tested.


components/database/postgresql-14/Makefile
```diff
@@ -0,0 +1,130 @@
...
+	'-e "/failed/p"'
+
+
+REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += library/libedit
+REQUIRED_PACKAGES += library/libxml2
+REQUIRED_PACKAGES += library/libxslt
+REQUIRED_PACKAGES += library/lz4
+REQUIRED_PACKAGES += library/security/openssl-11
+REQUIRED_PACKAGES += library/zlib
```
AndWac, line 123: Did you change the openssl mediator on your system? When I try to publish here I'll get lots of dependency errors like this:
oi-userland/components/database/postgresql-14/build/manifest-i386-postgresql14-contrib.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libcrypto.so.1.0.0 \
        pkg.debug.depend.reason=usr/postgres/14/bin/pg_verifybackup \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=usr/postgres/14/lib/amd64'.

You'll probably need to tweak the CONFIGURE_OPTIONS to let it find openssl-1-1 as USE_OPENSSL11=yes doesn't seem to be enough.

mvwieringen: I didn't think about it but yes it seems I have done so ages ago.
e.g.

openssl             local     1.1     system     openssl

I do wonder why by today things are still not on 1.1 by default given that 3 is the new stable by now.

I already spend a whole day on this so not sure when and if I have more time to spend given that building packages
on OpenIndiana really is painfull compared to Omnios. I think the whole hand building of p5m files is getting somewhat
tiresome given the big changes in packages these days. Maybe postgresql is even worse given it has such a split set
of p5m files.

AndWac: Short answer: Not enough time/volunterrs
Long answer: We still have some packages that need to be changed to use openssl-1.1. Some of these packages need non-obvious tweaks, some need to be updated to support openssl-1.1 and some cannot be changed or updated because there is no update available.

mvwieringen: On the long answer you could opt to change to 1.1 default and fix the stuff that doesn't work which is probably a better approach (e.g. less work ) then putting the burden on all new stuff. And maybe you need to freeze and obsolete some stuff that doesn't work with openssl 1.1. I don't use postgresql on OpenIndiana only wanted to give people the same 
and more performance DISM solution. But I kind of get why no one bothered with pg13 or pg14.

alarcher: The issue is that consumers of the broken packages will likely not be happy... Often it is just a matter of importing a few patches. I have fixed myself some of these packages although I literally do not care about them.

mvwieringen: Changing the mediator doesn't break the old packages as they will never be rebuild anyway. It seems the configure from postgresql doesn't give you the option to specify a with value setting the INC or LIB dir. other option would be to compile against openssl-1.0. I'll give that a try now and update the deps then someone can fix things who has to much time to spend or sees a simple way to do it.

mvwieringen: Ok with openssl 1.0 this time build and published with mediator set to openssl 1.0

AndWac (approved):

mvwieringen:
Seems I now also pushed the lastest patch previous still had some 12 references.

AndWac:
Just a comment on your patches: It would be nice to have some explanations (eg. why needed and if non-obvious how done) at the top of the files. This help future maintainers to adapt when something changes.

mvwieringen:
> Just a comment on your patches: It would be nice to have some explanations (eg. why needed and if non-obvious how done) at the top of the files. This help future maintainers to adapt when something changes.

Yeah sorry added that will be good for Omnios too so will add it there too.

PR #7790, opened 2022-02-24
Code::Blocks 20.03 32bit

hoewweken:
Hello!
My first new package proposal for OI.
I have published, installed an run codeblocks with this, i don't see problems so  far.
Codeblocks is build against wxwidgets-2.8.
I didn't manage to build the 64 bit binary, so this is 32 bit only for now.

Any comments are welcome!


alarcher:
Is there a way to have something newer that does not depend on gtk2 and not 32-bit?

alarcher:
> Is there a way to have something newer that does not depend on gtk2 and not 32-bit?

IMO this should not be merged unless it uses wxwidgets-3 and provides 64-bit binaries.
Otherwise this brings another dependency to obsolete libraries that we want to get rid of.
@AndWac

hoewweken:
The problem with 64 bit is described here, and i don't know how to solve this "the right way "(TM)
How do we address things like this ?

https://forums.codeblocks.org/index.php?topic=13742.0

i tried with gcc-10 and gcc-11 as in wx-strings.h ther is this block with the comment "broken compiler" but all gcc variants behave the same way so i assume it  is _not_ a complier issue...

AndWac:
We try to get rid of 32-bit where possible. So I am reluctant to merge new applications that are 32-bit. I guess that creating a 64-bit version of it is a matter of fiddling with configure and linker options. Please ask either on IRC (#oi-dev) or mailing list for ideas to solve your problems. According to the changelog 20.03 should work with wxwidgets-3.

alarcher:
> The problem with 64 bit is described here, and i don't know how to solve this "the right way "(TM) How do we address things like this ?
> 
> https://forums.codeblocks.org/index.php?topic=13742.0
> 
> i tried with gcc-10 and gcc-11 as in wx-strings.h ther is this block with the comment "broken compiler" but all gcc variants behave the same way so i assume it is _not_ a complier issue...

This is not a compiler issue, you can fix it by either casting or removing the useless type specifier so that the compiler defaults to size_t.

See https://github.com/OpenIndiana/oi-userland/commit/edc115a172bfe44640465b415e9dca4a76c59463 for a similar issue.

PR #5272, opened 2019-08-27
Workaround '11625 zoneadm install fails: UnicodeEncodeError', update ca-certificates to 3.45

Mno-hime:
* Workaround for `11625 zoneadm install fails: UnicodeEncodeError: 'ascii' codec can't encode character '\u0151' in position 68: ordinal not in range(128)` by renaming two CAs using ASCII-only characters
* Update ca-certificates to 3.45

PR #7791, opened 2022-02-24
ca-certificates: update to 3.75, Switch to ASCII cert names

cstes:
This addresses illumos bug #14510

It also updates the script to process the certificates to the version in libcurl  (web/curl component)


AndWac (approved):

Toasterson (changes requested):
The certificate may be non UTF-8 Locale but we can't just remove it.

cstes:
The proposed change to drop / remove / delete this particular  Hungarian certificate is because the filename seems UTF-8 encoded with for example small latin letter o with double acute (the hungarian letter with something that looks a little bit like an umlaut but it is not an umlaut) for this hungarian entry.

I think it is correctly UTF-8 encoded.   It is not 7 bit ASCII but it is a valid UTF-8 encoding (I think).

On some other Linux distributions, it appears that this Hungarian certificate is included with the UTF-8 encoding as it is.

The decision to delete or drop the certificate on OpenIndiana is not arbitrary : it is because the criterium here is that the UTF-8 filename poses a problem because OpenIndiana seems to not fully support UTF-8 filenames in /etc.

Mozilla is releasing on a regular basis new updates to their NSS product and in the future we can either add the Hungarian cerrtificate again, in a next update of the  package,, where it is also possible that new certificates are added with different scripts (Turkish, Armenian, whatever).

OpenIndiana users that absolutely need this certificate can also add the Hungarian certificate themselves to the CA root store.

In my understanding []( https://www.illumos.org/issues/14510 ) is reporting that the presence of this Hungarian certificate poses a problem for OpenIndiana 's TAR command.      I think the title of that report should be : ca-certificates contains a UTF-8 path which is non-ASCII.

However as a test, I tried TAR and it seems to be able to deal with UTF-8 paths and filenames.   I think TAR works fine with UTF8 filenames.

There is a related bug report from 2018 where the ZONEADM command fails.   This is 

[](https://www.illumos.org/issues/11625)

In issue #11625  Michael Nowak has implemented a fix or solution which basically remaps the UTF-8 filename to an ASCII name and I think he did this manually by using a file utf8 path=ascii   entry in the manifest.

The git commit is in the bug report : [](https://github.com/OpenIndiana/oi-userland/pull/5272.)

Please either merge my pull request which drops the certficiate   or close / reject my pull request (that is no problem for me) and ask Michael Nowak to apply his solution again or please create a pull request which adopts such a solution in the same way as Michael Nowak did in the past.

I also observe that Linux distributions put UTF-8 filepaths in /etc   and if all applications (including ZONEADM) support the UTF-8 filepath than there is no need to filter or mangle this certificate filename.

tsoome:
> The proposed change to drop / remove / delete this particular Hungarian certificate is because the filename seems UTF-8 encoded with for example small latin letter o with double acute (the hungarian letter with something that looks a little bit like an umlaut but it is not an umlaut) for this hungarian entry.
> 
> I think it is correctly UTF-8 encoded. It is not 7 bit ASCII but it is a valid UTF-8 encoding (I think).
> 
> On some other Linux distributions, it appears that this Hungarian certificate is included with the UTF-8 encoding as it is.
> 
> The decision to delete or drop the certificate on OpenIndiana is not arbitrary : it is because the criterium here is that the UTF-8 filename poses a problem because OpenIndiana seems to not fully support UTF-8 filenames in /etc.
> 
> Mozilla is releasing on a regular basis new updates to their NSS product and in the future we can either add the Hungarian cerrtificate again, in a next update of the package,, where it is also possible that new certificates are added with different scripts (Turkish, Armenian, whatever).
> 
> OpenIndiana users that absolutely need this certificate can also add the Hungarian certificate themselves to the CA root store.
> 
> In my understanding is reporting that the presence of this Hungarian certificate poses a problem for OpenIndiana 's TAR command. I think the title of that report should be : ca-certificates contains a UTF-8 path which is non-ASCII.
> 
> However as a test, I tried TAR and it seems to be able to deal with UTF-8 paths and filenames. I think TAR works fine with UTF8 filenames.
> 
> There is a related bug report from 2018 where the ZONEADM command fails. This is
> 
> In issue #11625 Michael Nowak has implemented a fix or solution which basically remaps the UTF-8 filename to an ASCII name and I think he did this manually by using a file utf8 path=ascii entry in the manifest.
> 
> The git commit is in the bug report :
> 
> Please either merge my pull request which drops the certficiate or close / reject my pull request (that is no problem for me) and ask Michael Nowak to apply his solution again or please create a pull request which adopts such a solution in the same way as Michael Nowak did in the past.
> 
> I also observe that Linux distributions put UTF-8 filepaths in /etc and if all applications (including ZONEADM) support the UTF-8 filepath than there is no need to filter or mangle this certificate filename.

I see no problem with tar. If there is broken tool, that tool must be fixed. Therefore, fixing this issue should start by identifying what actually is broken.

Toasterson:
@richlowe reported the Problem with tar yes, but it's a Bug As far as I managed to follow the conversation. And it got fixed. Our tar from the commandline with -E does not cause problems. The thing is that the Hungarian cert is important in the chain and using non UTF-8 Characters. The main bug with zoneadm has been fixed in our ipk branding scripts which downgraded the locales to C instead of C.UTF-8 I fixed this a couple months ago. There is at this moment nothing blocking us from delivering this cert.

Toasterson (approved):
@cstes Thanks for the rename script now it's perfect. LGTM Ready to merge 

@richlowe @alanc @jclulow cc'ing you guys here if anything with this cert should come up again. With this we are back to ASCII cert filenames.

PR #7792, opened 2022-02-24
wxwidgets-3: disambiguate string.h

alarcher:
@hoewweken Could you try this?


hoewweken:
i have one successful 64 bit build now, gcc-7 64bit with wxwidgets-3 with your patch and
we need a patch to codeblocks as described here:
https://forums.codeblocks.org/index.php?topic=23687.0
.. and maybe more ...
After all my fiddling around i will setup a clean build environment to check your patch and what else needed for codeblocks.
Thanks!

AndWac (approved):

PR #7798, opened 2022-02-25
make wxPaintEvent public for building codeblocks

hoewweken:
i belive weneed this for compiling codeblocks, but maybe there is a more elegant solution for this ...


hoewweken:
Not needed for codeblocks svn source tree.

PR #7801, opened 2022-02-26
mysql_fdw: add extension for PostgreSQL 14, update to 2.7.0

mvwieringen:


components/database/postgresql-10-mysql_fdw/Makefile
```diff
@@ -43,9 +41,9 @@ TEST_TARGET=	$(NO_TESTS)
...
 
-COMPONENT_BUILD_ENV	+=	USE_PGXS=0
+COMPONENT_BUILD_ENV	+=	USE_PGXS=1
 COMPONENT_BUILD_ARGS	+=	MYSQL_LIBNAME=$(MYSQL_LIBDIR)/libmysqlclient.so
-COMPONENT_INSTALL_ENV	+=	USE_PGXS=0
+COMPONENT_INSTALL_ENV	+=	USE_PGXS=1
 
-REQUIRED_PACKAGES += $(PG_DEVELOPER_PKG)
+REQUIRED_PACKAGES += database/postgres-10/developer
 REQUIRED_PACKAGES += database/mariadb-101/library
```
AndWac, line 49: We probably should drop postgresql-96 before this.

mvwieringen: Yeah 96 is not longer supported. But it seems on the build server all devel versions of previous pg versions are installed because on my clean server all older versions fail to build to not being able to find pg_config for the right version.

AndWac: I am going to merge this when my tests have finished as I expect the discussions to go on for some time and I don't want to block this for its outcome.

mvwieringen: Removing pg 9.6 will be a challenge given to much depends on it. And the the question will be what next version I would
opt for pg14 hoping that all dependent packages work with that as pg10 is on for deprecation in November this year and
11 a year later and 12 a year later.

PR #7802, opened 2022-02-26
Add pg_repack 1.4.7 for PostgreSQL 14

mvwieringen:


components/database/postgresql-14-pg_repack/Makefile
```diff
@@ -0,0 +1,50 @@
...
+include $(WS_MAKE_RULES)/common.mk
+
+COMPONENT_BUILD_ENV	+=	USE_PGXS=0
+
+# Manually added build dependencies
+REQUIRED_PACKAGES += database/postgres-14/developer
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += $(PG_LIBRARY_PKG)
+REQUIRED_PACKAGES += system/library
```
AndWac, line 48: Is this really correct?

mvwieringen: No it should be on the pg14 sever package given its a plugin loaded into postmaster.
Then again all others are broken e.g. have same problem should I fix those too ?

mvwieringen: I get the feeling this was copied from the mysql_fdw package given the reference to USE_PGXS=0

mvwieringen: Ok pushed some fixes think this is the best we can do e.g. explicit depend on database itself in p5m files and have the 
build requirement on the version specific library package. It essentially needs libpq for pkgdepend to succeed.

PR #7803, opened 2022-02-26
enchant: Update to 2.3.2

ghost:
Update the enchant library to the latest version.
This is needed for #7735 
Please make sure to also merge my next PR,otherwise that will break a lot of applications.

PR #7804, opened 2022-02-26
Enchant rebuild

ghost:
Rebuild packages that depend on enchant after updating it to the latest version (#7803).
I tested them all and added patches on some (which expect a older version).


components/web/php/php-8_1/Makefile
```diff
@@ -20,6 +20,7 @@ include ../../../../make-rules/shared-macros.mk
 COMPONENT_NAME=         php
 COMPONENT_MJR_VERSION=	8.1
 COMPONENT_MNR_VERSION=	2
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=      PHP 8.1
```
AndWac, line 24: You'll need to fix that. We have already updated PHP-8.1.3

ghost:
I rebased my work on top of the current oi/hipster branch.
That should fix the version.

PR #7822, opened 2022-03-07
gmime3: Add package

ghost:
Add package for the gmime3 library.
I thought I'd need this for Pan,but currently it works better with gmime (version 2).
Will be required by future versions,however (and maybe other programs),so it can't hurt to have the package.

### DIFF
--- a/components/library/gmime3/Makefile
+++ b/components/library/gmime3/Makefile
@@ -1,0 +1,42 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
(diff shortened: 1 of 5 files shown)


alarcher:
This package is 32-bit and therefore should be either converted to 64 or 32_and_64. If there are no consumers then 64 is preferred.

ghost:
Thank you very much for the hint.
That was one of my first packages when I wasn't that experienced with the package system.
I made a new pull request to rebuild it with 64bit.

alarcher:
> Thank you very much for the hint. That was one of my first packages when I wasn't that experienced with the package system. I made a new pull request to rebuild it with 64bit.

Perfect! thank you :)

PR #7826, opened 2022-03-07
libreoffice: update to 7.2.5.2 64-bit

dbh201:
Tests passed:

**Base**
1. load folder of `.csv` files and display rows
2. Save as .odb file
3. Load .odb file and display rows

**Writer, Calc, Impress, Draw, Math**
1. Save simple file
2. Load file and display content


dbh201:
There was only one area of the code that wasn't portable (cpp<->uno bridge), and luckily all it needed were a few changes to adhere to the 64-bit ABI and some wider pointers. Most of the code is copy-pasted from the 32-bit version.

AndWac:
Can you also update sample-manifest.p5m, please.

PR #7829, opened 2022-03-07
gcc49: fix download URL for mpc

mtelka:


AndWac:
With every change you'll need to either change the COMPONENT_VERSION or COMPONENT_REVISION. Nevertheless I am not sure wether we should better drop all old gcc versions up to version 7. We have been accumulating a lot of cruft.

mtelka:
COMPONENT_REVISION updated.

PR #7878, opened 2022-03-14
Update bhyve UEFI firmware to stable202102; add VARS file

citrus-it:
This change updates the bhyve UEFI firmware to the stable 202102 version, as is used in FreeBSD and has been used in OmniOS for some time. This is quite a jump from the existing 2014 version and presents as UEFI 2.70 to the guest instead of 2.40.

The CSM firmware remains on the 2014 version since CSM is not working in the later version.

I've also added the template UEFI variables file (BHYVE_VARS.fd) which can be used with bhyve to support persistent UEFI variables alongside the bootrom. Support for this in gate arrived with [14372 bhyve upstream sync 2022 January](https://illumos.org/issues/14372), however this feature is not entirely reliable yet, pending [14489 bhyve should emulate imul](https://illumos.org/issues/14489) - reviews welcome there!

In order to build the later firmware, I needed to rebuild GNU binutils with plugin support, and that is a separate PR #7879. I ran the tests and there were no changes, the only difference appears to be one extra header file, and that building the new bhyve firmware works.


AndWac:
Thank you for your work.
If the firmware update depends on the binutils update you'll need to split into into separate PRs.
Update are not available before the jenkins job has completed.

citrus-it:
> If the firmware update depends on the binutils update you'll need to split into into separate PRs.

Done.

PR #7883, opened 2022-03-15
Add MediaInfoLib-21.09

shuryanc:
Add MediaInfoLib-21.09 before pushing Megasync package.

### DIFF
--- a/components/library/MediaInfoLib/Makefile
+++ b/components/library/MediaInfoLib/Makefile
@@ -1,0 +1,52 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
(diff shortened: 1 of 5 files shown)


components/library/MediaInfoLib/Makefile
```diff
@@ -0,0 +1,51 @@
...
+#
+
+BUILD_BITS=64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		MediaInfoLib
+COMPONENT_VERSION_DL= 21.09
+COMPONENT_VERSION=	21.9
+COMPONENT_SUMMARY=	MediaInfo(Lib) is a convenient unified display of the most relevant technical and tag data for video and audio files.
```
AndWac, line 23: You could also add HUMAN_VERSION=21.09 (and use it in the package manifest). This way pkg info will also show the correct version information. See sysutils/tmux for an example usage.

PR #7885, opened 2022-03-15
dnspython - additional package for pyhton 3.7 and 3.9 (required for l…

fritzkink:
…atest samba version)


alarcher:
pyhton -> python

PR #7906, opened 2022-03-19
dhcp: update to 4.4.3

balrog-nona:
It builds, installs and publishes fine. Sample-manifest didn't change.

I built it and tested it for gcc 7 as well as gcc 10. Now, it builds for gcc 10 without the need of patch file `005-fno.common.patch`. Upstream applied some changes for gcc 10 build: 
https://downloads.isc.org/isc/dhcp/4.4.3/dhcp-4.4.3-RELNOTES
https://gitlab.isc.org/isc-projects/dhcp/-/commit/6c7e61578b1b449272dbb40dd8b98d03dad8a57a

When installed locally, I tested it with minimal config file taken from here: https://gist.github.com/paul-ridgway/e419f7e52258db142326b691d0450be3
I was able to start the `svc:/network/dhcp/server:ipv4 service` after I added /etc/inet/dhcpd4.conf with definition of my local subnet:
`svcs -vp svc:/network/dhcp/server:ipv4` gave me:
```
STATE          NSTATE        STIME    CTID   FMRI
online         -             19:40:27    818 svc:/network/dhcp/server:ipv4
               19:40:27    17454 dhcpd
```


components/network/isc-dhcp/Makefile
```diff
@@ -20,16 +20,16 @@
...
 BUILD_BITS=		64
 
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		isc-dhcp
-COMPONENT_VERSION=	4.4.2
-COMPONENT_REVISION=	2
-IPS_COMPONENT_VERSION=	4.4.2.0.0
+COMPONENT_VERSION=	4.4.3
+IPS_COMPONENT_VERSION=	4.4.3.0.0
```
AndWac: I am not sure whether the use of IPS_COMPONENT_VERSION makes sense here. It is most probably a relic from times where COMPONENT_VERSION had some illegal characters. You should remove it.

balrog-nona: Removed, thanks.

PR #7930, opened 2022-03-23
Add developer/macro/cpp (binary from OpenSolaris)

mtelka:


AndWac:
We already have developer/macro/cpp from illumos-gate. Do you plan for a separation?

pyhalov:
IIRC, our cpp package is transfered 'as-is' from OpenSolaris.

mtelka:
> We already have developer/macro/cpp from illumos-gate.

I believe we don't.  It is as @pyhalov said.

components/closed/cpp/Makefile
```diff
@@ -0,0 +1,47 @@
...
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2012, Andrzej Szeszo
```
Toasterson, line 12: Is this a resurected Component or a wrong copyright notice?

Toasterson (commented):
I am unclear as to why we would want this component and not drop it, but it's a good cleanup nontheless. So LGTM. We can always drop it at a later stage.

Toasterson:
And I got Ninja'd while getting Coffee :)

components/closed/cpp/Makefile
mtelka: This file was born using this command: `cp ../mtsk/Makefile .`

mtelka:
> I am unclear as to why we would want this component and not drop it, but it's a good cleanup nontheless. So LGTM. We can always drop it at a later stage.

We want it because both illumos-gate (package developer/object-file) and oi-userland (build-essential and install-types/includes/minimal) requires it.  I'm not sure it is really used for anything, but dependency is still there.  Once the dependency is removed we could consider to remove (obsolete) this package too.

pyhalov:
illumos tools, like rpcgen and dtrace use it, so you can't just drop it. There were talks about replacement and one is provided in Joyent illumos-extra https://github.com/joyent/illumos-extra/tree/master/cpp , but it's not bug-to-bug compatible to Sun cpp. IIRC, when I tested it last time, it failed to parse some files. illumos-extra cpp was based on Schily Tools cpp - https://sourceforge.net/projects/schilytools/files/ . I don't know if somebody maintains them now, when Jorg has gone.

AndWac:
Alas this one doesn't seem to be compatible with our older revision cpp@5.11-2013.0.0.0.
At least it broke our illumos-gate build.

mtelka:
> Alas this one doesn't seem to be compatible with our older revision cpp@5.11-2013.0.0.0. At least it broke our illumos-gate build.

It looks like the file permissions are different in the new package.  Sorry for that.  I'll prepare new PR with the fix ASAP.

mtelka:
> > Alas this one doesn't seem to be compatible with our older revision cpp@5.11-2013.0.0.0. At least it broke our illumos-gate build.
> 
> It looks like the file permissions are different in the new package. Sorry for that. I'll prepare new PR with the fix ASAP.

Just for the record: fixed in #7943

PR #7939, opened 2022-03-23
illumos: Ship python 3.9 versions of illumos modules

richlowe:
This will allow, in a later change, smooth movement of package/pkg to python 3.9, since libbe will already be there for us.

I've tested this with a python3.9-using pkg of mine, in as much as letting it create boot environments etc.
@citrus-it did the work to add these flags, and should check I'm not messing them up


Toasterson (approved):
LGTM

PR #7941, opened 2022-03-23
Add system/library/c++/sunpro (binary from OpenSolaris)

mtelka:


mtelka:
I just double checked permissions in this package (to avoid #7943) and it looks reasonable.

PR #7942, opened 2022-03-23
Retirement of grails

jgmills:
OI grails is an ancient version (1.0.3).  The current version is 5.1.3 .  As far as I can tell, the existing version in OI has no consumers.  I attempted to upgrade grails to 5.1.3, but soon discovered that the installer (gradle) does reprehensible things.  The new grails is not amenable to being packaged.  The use of python in the existing version of grails is limited to one python-2 script.  It would be possible to convert this script to python-3, but it's not worth the effort for such an old version.

Grails is a web application framework.  Its package is library/java/grails .  This package is made obsolete by this PR.  Likewise, the directory components/library/grails is deleted by this PR.  The global history file is modified to obsolete the package.


Toasterson (approved):
I agree with Gary here. Grails is a framework that gets embedded into a War file during build process and a package would only work for development. As things go these tools are usually best installed by developers on the machines they need them and the devs have the skills to handle the tools. Packaging dev tools with the systems has fallen out of fashion and we should not continue doing it aswell.

PR #7946, opened 2022-03-24
libtdjson: Telegram TDLib 1.8.0

cstes:
the component is called td

but this seems a name that is too short and could be confusing as package name

therefore I used the name libtdjson which is the name of the library I need for a Smalltalk project which can use ths library and JSON support in Smalltalk to send Telegram messages


components/library/libtdjson/td.p5m
```diff
@@ -0,0 +1,67 @@
...
+#
+
+#
+# Copyright 2022 David Stes
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="org.opensolaris.category.2008:System/Libraries"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
```
AndWac: In the Makefile you have defined
COMPONENT_CLASSIFICATION=Web Services/Communications

components/library/libtdjson/Makefile
```diff
@@ -0,0 +1,63 @@
...
+COMPONENT_LICENSE_FILE= td.license
+
+include $(WS_MAKE_RULES)/common.mk
+CMAKE_OPTIONS +=	-DCMAKE_C_COMPILER=$(CC)
+CMAKE_OPTIONS +=	-DBIN_INSTALL_DIR=/usr/$(CMAKE_LIBDIR.$(BITS))
+CMAKE_OPTIONS +=	-DLIB_INSTALL_DIR=/usr/$(CMAKE_LIBDIR.$(BITS))
+CMAKE_OPTIONS +=	-DCMAKE_BUILD_TYPE=Release
+
+REQUIRED_PACKAGES += developer/gperf
+
```
AndWac: You also need to add
REQUIRED_PACKAGES +=developer/gcc-11
GCC_VERSION=11 is not sufficient to install it.

AndWac:
Now you'll need to update pkg5. Did you test this package? When I run "gmake test" the result is 1/1 tests failed.

cstes:
You may have noticed that I backported a few changes from 1.8.0 to 1.60.
The Squeak TelegramClient expects 1.6.0 libraries so if possible at all I'd prefer at this moment  TDLib 1.6.0.

I didn't run gmake test.

My test consisted of installing the library and trying out the Squeak TelegramClient which works in Squeak.
So in that sense, yes I tested the shared library.

I think the update of pkg5 is automatic when running "gmake publish".

Anyway, if I understand correclty you want 1) that I update pkg5 and 2) that I somehow disable the test targets for example by 

TEST_TARGET= $(NO_TESTS)


If I add TEST_TARGET= $(NO_TESTS) is that OK by you ?

By the way, given the fact that Telegram sources support illumos  (see CMake/illumos.cmake in their sources) and given the fact that they enable epoll for Illumos, I think they are effectively using Illumos or that somebody contributed fixes to them (not me).

Is there perhaps already somebody who contributed a IPS TDLib package because that would be easier ... and save work for me!

AndWac:
pkg5 is updated automatically by "gmake publish" but it needs to be updated in your commit, too.
If the automatic tests are expected to not work correctly it is better to document that fact in a short explanation in the Makefile.

components/library/libtdjson/Makefile
```diff
@@ -0,0 +1,74 @@
...
+# https://github.com/hpi-swa-teaching/TelegramClient
+
+BUILD_BITS=		32_and_64
+BUILD_STYLE=		cmake
+# USE_OPENSSL11=		yes
+include ../../../make-rules/shared-macros.mk
+GCC_VERSION= 		11
+
+# the FMRI is libtdjson , the project is tdlib, component is td
+
```
AndWac: Why this? Wouldn't it be better to set COMPONENT_NAME=tdlib and COMPONENT_FMRI=library/$(COMPONENT_NAME)?

cstes:
The name of td or tdlib does not seem compatible with the name scheme of most libraries in OpenIndiana.
Most library packages have a FMRI that start with lib.
The most important shared library in this package is libtdjson so I choose libtdjson as FMRI, because it seems more descriptive than for example libtd.  Also note that there seems to be already a libtdb  (tiny database) in OpenIndiana.
Given the fact that the sources are td-1.6.0 if you extract the sources, there is some choice to make.
If you want "td" as name of the package , I could try to do that , but td seems a very short name.
From telegram it seems logical "telegram database" but I doubt that someone who is just looking at IPS package names finds td sufficiently informative, the two letters td could be anything.

cstes:
As a test by the way, I use this libtdjson-1.6.0 with Squeak and the TelegramClient in Squeak and it works.   Also note that Squeak is currently using OpenSSL 1.0.2 so I am not building libtdjson with OpenSSL 1.1.
There are upstream requests from my part for Squeak to upgrade to OpenSSL 1.1 (also affects their Linux platforms by the way) and to TDLib 1.8.0.   It seems libtdjson 1.6.0 is slightly incompatbile with some fields in libtdjson 1.8.0.

AndWac:
The most important name is COMPONENT_FMRI and this already contains library/. So, in my eyes it isn't necessary to repeat lib. Just give it an appropriate name. You are right, td is very short. I would probably choose telegramdb but I have no plans to use it.

AndWac:
This doesn't look good:
/export/home/andreas/oi-userland/components/library/telegramdb/td-1.6.0/sqlite/sqlite/sqlite3.c: In function 'sqlite3Fts5IndexQuery':
/export/home/andreas/oi-userland/components/library/telegramdb/td-1.6.0/sqlite/sqlite/sqlite3.c:195003:5: warning: 'memcpy' specified bound 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
195003 |     memcpy(&buf.p[1], pToken, nToken);
       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

PR #7948, opened 2022-03-24
bhyve/firmware is shipping the wrong firmware file

citrus-it:
For the newer UEFI bootrom, there are three output files:
* BHYVE_CODE.fd
* BHYVE_VARS.fd
* BHYVE.fd

where the last one is a concatenation of the first two. For running bhyve in illumos, we need the first two (although VARS is optional) and the firmware package was mistakenly shipping the last two.

I've built the package with these changes and tested the ROM that ends up in the proto area.


citrus-it:
The file sizes look better too (128K smaller since the variables are no-longer bundled)
```
af@hipster:~/oi-userland/components/openindiana/bhyve-fw$ ls -l build/prototype//i386/usr/share/bhyve/firmware/
total 16263
lrwxrwxrwx   1 af       staff         20 Mar 24 20:48 BHYVE_CSM.fd -> BHYVE_RELEASE_CSM.fd
-rw-r--r--   1 af       staff    2097152 Mar 24 20:48 BHYVE_DEBUG_CSM.fd
-rw-r--r--   1 af       staff    1966080 Mar 24 20:54 BHYVE_DEBUG.fd
-rw-r--r--   1 af       staff    2097152 Mar 24 20:48 BHYVE_RELEASE_CSM.fd
-rw-r--r--   1 af       staff    1966080 Mar 24 20:54 BHYVE_RELEASE.fd
-rw-r--r--   1 af       staff     131072 Mar 24 20:54 BHYVE_VARS.fd
lrwxrwxrwx   1 af       staff         16 Mar 24 20:54 BHYVE.fd -> BHYVE_RELEASE.fd
```

Toasterson (approved):
LGTM

PR #7949, opened 2022-03-24
prep-unpack.mk: replace developer/java/jdk by developer/java/openjdk8

mtelka:
... because developer/java/jdk was renamed to developer/java/openjdk8 many years ago and we should not force developers to install renamed packages.


Toasterson (approved):
LGTM

PR #7950, opened 2022-03-24
ips.mk: the REQUIRED_PACKAGES target should update the list

mtelka:
This could save some manual work during package update/refresh.


Toasterson (approved):
I like the idea

PR #7974, opened 2022-03-31
fix pangox-compat build

AndWac:


AndWac:
I wonder whether anything is still using this unsupported library.

PR #7978, opened 2022-04-01
ca-certificates: update to nss-3.77

cstes:


AndWac:
Can you add HUMAN_VERSION="20220309" to the Makefile and use it in the manifest, please? On Repology you can see that ca-certificates are usually named after its release date on other distributions.

components/openindiana/ca-certificates/Makefile
```diff
@@ -20,20 +20,22 @@ include ../../../make-rules/shared-macros.mk
...
+		       sha256:825edf5a2fd35b788a23ff80face591f82919ae3ad2b8f77d424a450d618dedd
 COMPONENT_ARCHIVE_URL= \
 	https://ftp.mozilla.org/pub/security/nss/releases/NSS_$(COMPONENT_VERSION_MAJOR)_$(COMPONENT_VERSION_MINOR)_RTM/src/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL = https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS
 COMPONENT_FMRI=		crypto/ca-certificates
 COMPONENT_LICENSE=	MPLv2.0
 COMPONENT_LICENSE_FILE=	nss/COPYING
 
+HUMAN_VERSION="20220309"
+
```
AndWac, line 38: It's not sufficient to add HUMAN_VERSION to the Makefile. You also need to reference it from the manifest file:

Looking at fonts/gentium as an example, I think
HUMAN_VERSION=20220309
would be working, too (no need for quotation marks).

PR #7979, opened 2022-04-02
Add a new Build Style for Perl modules which use Module::Build or Module::Build::Tiny instead of MakeMaker

ghost:
Perl modules which use Module::Build or Module::Build::Tiny as build system can't be built on OpenIndiana currently because the makemaker.mk Build Style is hardcoded to Makefile.pl and $(GMAKE) while Module::Build(::Tiny) has Build.pl and ./Build.
This change adds a new Build Style, derived from makemaker.mk, which instead calls Build.pl and ./Build.
It's required for all packages listed below "bauen für" at: https://www.archlinux.de/packages/extra/x86_64/perl-module-build and https://www.archlinux.de/packages/community/x86_64/perl-module-build-tiny

### DIFF
--- a/make-rules/modulebuild.mk
+++ b/make-rules/modulebuild.mk
@@ -1,0 +1,80 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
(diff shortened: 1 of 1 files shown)


make-rules/modulebuild.mk
```diff
@@ -0,0 +1,175 @@
...
+# CDDL HEADER END
+#
+# Copyright (c) 2011, 2014, Oracle and/or its affiliates. All rights reserved.
+# Copyright 2022 Niklas Poslovski
+#
+
+# perl-5.22 is 32 bit (only), perl-5.24 and later are 64 bit (only)
+
+COMMON_PERL_ENV +=	MAKE=$(GMAKE)
+COMMON_PERL_ENV +=	PATH=$(dir $(CC)):$(SPRO_VROOT)/bin:$(PATH)
```
pyhalov: Is $(SPRO_VROOT)/bin still required?

make-rules/modulebuild.mk
```diff
@@ -0,0 +1,175 @@
+
```
pyhalov: Missiing "#"

pyhalov (commented):
I wonder if file can just include makemaker.mk and contain some overwrites?

ghost:
I made a new modulebuild.mk that only overwrites the needed variables as you suggested.
I tried it locally by building some packages that need Module::Build and it still works.

PR #7981, opened 2022-04-03
bacula - update to latest version 11.0.6

fritzkink:


components/sysutils/bacula/Makefile
```diff
@@ -68,7 +68,6 @@ REQUIRED_PACKAGES += database/mariadb-103
 REQUIRED_PACKAGES += database/mariadb-103/client
```
AndWac, line 70: You'll need to add database/postgres-12/developer as a manually added build dependency.

fritzkink: added

AndWac:
Now please update also pkg5.

PR #7992, opened 2022-04-05
openssh: update to 8.9p1

AndWac:


AndWac:
Note: Like OmniOS we dropped the GSS-API patch and thus, this new version doesn't support it anymore.

PR #8026, opened 2022-04-12
pari/gp: update to 2.13.4

cstes:


components/scientific/pari/manifests/sample-manifest.p5m
```diff
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2021 <contributor>
+# Copyright 2022 <contributor>
 #
```
AndWac: Don't update the sample-manifest when only the copyright year changes.

PR #8061, opened 2022-04-18
minimal_install: remove system/network/routing

cstes:
See Feature #8587

https://www.illumos.org/issues/8587

### DIFF
--- a/components/meta-packages/install-types/includes/minimal
+++ b/components/meta-packages/install-types/includes/minimal
@@ -195,7 +195,6 @@ depend type=require fmri=system/library/storage/scsi-plugins
 depend type=require fmri=system/man
 depend type=require fmri=system/network
 depend type=require fmri=system/network/nis
-depend type=require fmri=system/network/routing
 depend type=require fmri=system/prerequisite/gnu
 depend type=require fmri=system/scheduler/fss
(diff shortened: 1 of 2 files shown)


components/meta-packages/install-types/includes/minimal
```diff
@@ -195,7 +195,6 @@ depend type=require fmri=system/library/storage/scsi-plugins
 depend type=require fmri=system/man
 depend type=require fmri=system/network
 depend type=require fmri=system/network/nis
-depend type=require fmri=system/network/routing
 depend type=require fmri=system/prerequisite/gnu
```
AndWac, line 198: You simply removed it from minimal. Do we want routing in server_software instead?

cstes: I have no idea.   The idea behind removing it from the minimal group is , so that it is not always installed (as part of the minimal group).   The fact that it is in "minimal" seems not logical as there may be setups that do not need this software so it is not minimal in the sense of absolutely required (I think).

cstes: I have closed this PR but can reopen it later, perhaps I should first do more tests to build with distribution constructor images and see result of the change.   In any case I have opened bugs.illumos.org issues for these modifications and a request to split NPD neighbour ipv6 discovery from routing.

PR #8107, opened 2022-04-28
exfat file system via fuse

hoewweken:
Add extFAT support


jpandre:
Is this different from #7389 and #7521 ?

hoewweken:
Oh i missed that license issue, i'm sorry.

"_We had during the Maintainer meeting a discussion about the License Situation. And as far as we can tell the Patent is only opened to linux. So we can't have it in the main repository. If you really want this inside of OI we could put it into hipster-encumbered_"

PR #8136, opened 2022-05-01
vlc: update to 3.0.17.4

cstes:
This updates to 3.0.17.4 which is possible thanks to the previous update of libdvdnav and libdvdread to 6.0.1.

Note that there is a problem with the codec/dav1d but I think this is a upstream problem

Also see Debian Bug#1008609: vlc: FTBFS with dav1d 1.0.0
https://code.videolan.org/videolan/vlc/-/merge_requests/1611
https://code.videolan.org/videolan/vlc/-/merge_requests/1618

Anyway for my basic usage of VLC of trying to play some DVD's when I check the codec used for the DVD in my case, it is MPEG-1/2   and I have tested that my DVD play with both VLC 2.2.8 (best results) and with 3.0.16 and 3.0.17.4.

### DIFF
--- a/components/encumbered/vlc/Makefile
+++ b/components/encumbered/vlc/Makefile
@@ -11,6 +11,7 @@
 
 #
 # Copyright 2015 Aurelien Larcher
+# Copyright 2022 David Stes
 #
 
 BUILD_BITS=64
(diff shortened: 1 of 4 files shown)


alarcher:
There are at least some issues of videos stuttering because of VLC detecting capabilities of the GPU incorrectly. Some videos play fine with FFMPEG but not with VLC due to some parameters passed by VLC in their codec handling interface. I have experienced similar regressions with AVI/MKV.

AndWac:
Please rebase and adapt your PR after the merge of the newer qt5.

cstes:
will reopen later

PR #8139, opened 2022-05-02
Upgrade of docutils to 0.18.1 with python-3

jgmills:
Docutils is a python module and series of python scripts that processes documentation into useful formats.  This PR adds packages for python 3.5, 3.7, and 3.9, and obsoletes the python 2.7 module.  The software version is upgraded from 0.12 to 0.18.1 .  Note that docutils is generally a build dependency, and as such is not installed by ordinary users.  It's only needed by a very small number of packages.  These packages will not have to be rebuilt because of this dependency change.

The build, install, and publish steps were all successful, as was the package install.  Manual testing was successful: the "rst2man.py" script worked as expected.  With the "--help" option, it produced command and options help text.  It also did the conversion of an RST file to a man page file, the usual purpose of this script.


AndWac:
My question here is: is this a generic tool to create documentation or do we need variants for different python versions?
That is, is it a tool that happens to be implemented in python, like eg. ansible? If so, we can remove the -27 package (and don't publish -35, -37, and -39) and only have the docutils package with everything.

jgmills:
I'm not sure of the answer to that question because I'm not familiar with docutils or even python.  However, I do know that docutils consists of both libraries and python scripts.  The library, the largest part, resides in a version-specific directory.  The scripts, the smallest parts, reside in /usr/bin .  The scripts use the libraries.  The libraries can be used by themselves, although perhaps this was not done in OI.  I see that there was a -26 version of the libraries, but this got dropped when python 2.6 was dropped.  So, docutils is both a tool and a library.  We have to publish at least one library: OI might as well publish all of them.  After all, python 3.5 will one day be dropped.  Python 3.11 will one day be added.

AndWac:
Then we may skip at least 3.5 because it's not supported anymore and we just keep it because it's needed for IPS until we will be able to switch to 3.9 for it.

jgmills:
I agree that the python 3.5 version can be dropped.  I'll be submitting a PR in a day or two, with this change.

I'm concerned that many other OI packages depend on python 3.5 .  We need a complete list of such packages.  They will at least need to be rebuilt.

I'm also concerned that there has been no announcement that python 3.5 is being phased out, at least to my knowledge.  We need that announcement as well as a plan for carrying it out.

PR #8143, opened 2022-05-03
Fixed librecad for Qt5.15.3

shuryanc:
Fixed librecad for Qt5.15


AndWac:
Here you'll also need to change COMPONENT_REVISION=2 in the Makefile

PR #8144, opened 2022-05-03
Fixed vlc for Qt5.15.3

shuryanc:
Fixed vlc for Qt5.15.3


AndWac:
You'll need to disable dav1d as it makes vlc build fail as noted in #8136.
CONFIGURE_OPTIONS += --disable-dav1d
and update REQUIRED_PACKAGES and pkg5 file, too.

PR #8145, opened 2022-05-03
Fixed stellarium for Qt5.15.3

shuryanc:
Fixed stellarium for Qt5.15.3


AndWac:
When you touch a package you'll either need to change COMPONENT_VERSION or COMPONENT_REVISION. Otherwise the package won't be updated if already installed. Here you'll need to add COMPONENT_REVISION=1

PR #8146, opened 2022-05-03
Fixed megasync for Qt5.15.3

shuryanc:
Fixed megasync for Qt5.15.3


AndWac:
Again, you'll need to change COMPONENT_REVISION (set it to 1 if not yet there).

PR #8150, opened 2022-05-03
Fixed php-7_3 for icu70

shuryanc:
Fixed php-7_3 for icu70


AndWac:
Please update your fork before working on an update. We have obsoleted PHP-7.3 a while ago. We have PHP-8.1 instead.
You may also need to synchronize against OpenIndiana's oi/hipster several times until your PR has been merged because
others are also working on updates.

PR #8177, opened 2022-05-13
Added ECM-5.93.0

shuryanc:
Added ECM. Its needed before any kde stuff can be built.


AndWac:
Can you elaborate why you need this? If you plan to port KDE we need to discuss this before we start merging packages.

shuryanc:
I need to port krita

AndWac:
Ok, how far did you get already? You should finish the port locally before trying to integrate it. Without having it ported locally you might stop somewhere in the middle we will have lots of packages that are probably of no use for others.

shuryanc:
> Ok, how far did you get already? You should finish the port locally before trying to integrate it. Without having it ported locally you might stop somewhere in the middle we will have lots of packages that are probably of no use for others.

Already ported locally. Btw, I will also need it for kdenlive (video editing software)
圖片

AndWac:
Ok, that looks promising.

AndWac:
Some more remarks:
1. Use a tagged version and archive from here: https://github.com/KDE/extra-cmake-modules/tags
2. We need to get more structure in the packages if you continue to add more packages from KDE.  I suggest to follow what has been done for Mate. That is, move all packages belonging to KDE into a separate folder named "kde" below desktop.
3. I have added some definitions for QT like alaracher has proposed: https://github.com/OpenIndiana/oi-userland/pull/8181
   Please make use of them and if needed enhance or fix them.
4. If you find enough commonalities in the KDE related packages think about to extract them into a separate make-rule (.mk) file like alaracher did for x11 (see make-rules/x11.mk and x11 packages).

shuryanc:
> Some more remarks:
> 
>     1. Use a tagged version and archive from here: https://github.com/KDE/extra-cmake-modules/tags
> 
>     2. We need to get more structure in the packages if you continue to add more packages from KDE.  I suggest to follow what has been done for Mate. That is, move all packages belonging to KDE into a separate folder named "kde" below desktop.
> 
>     3. I have added some definitions for QT like alaracher has proposed: [provide qt5/6 macros #8181](https://github.com/OpenIndiana/oi-userland/pull/8181)
>        Please make use of them and if needed enhance or fix them.
> 
>     4. If you find enough commonalities in the KDE related packages think about to extract them into a separate make-rule (.mk) file like alaracher did for x11 (see make-rules/x11.mk and x11 packages).

This is very helpful! Will try to make them based on these instructions.

shuryanc:
I have a question for point 1. Based on prep-git.mk and my current setting as below, it will convert to the http url https://github.com/KDE/extra-cmake-modules/archive/refs/tags/v5.93.0.tar.gz. May I know if I can still use the below setting?

COMPONENT_PREP_GIT=yes
GIT_TAG=v$(COMPONENT_VERSION)
GIT_REPO=https://github.com/KDE/$(COMPONENT_NAME)
GIT_BRANCH=refs/tags/$(GIT_TAG)
COMPONENT_ARCHIVE_HASH=	sha256:6e53e70f293a11836353a5d81eac37d221d9525a85a28c02bb5e1060539c06c3

AndWac:
Remove everything git related and add COMPONENT_ARCHIVE_URL, eg.
COMPONENT_ARCHIVE_URL=https://github.com/KDE/$(COMPONENT_NAME)/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
Also remove COMPILER=gcc as this is our default.

components/developer/extra-cmake-modules/note
```diff
@@ -0,0 +1,5 @@
+Remove
```
AndWac: This file should be removed from the PR.

components/developer/extra-cmake-modules/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+BUILD_BITS=		64
+BUILD_STYLE=	cmake
+
+include ../../../make-rules/shared-macros.mk
+
+COMPILER=gcc
+COMPONENT_NAME=		extra-cmake-modules
+COMPONENT_VERSION=	5.93.0
+COMPONENT_FMRI=		developer/kde/ECM
```
Toasterson: Shortening this by ecm might be confising for people let's stick with extra-cmake-modules as the packages name.

PR #8180, opened 2022-05-13
Retirement of clang-90

jgmills:
OI clang-90 is an ancient version (9.0.1).  OI has a newer package available (runtime/clang-13) with version 13.0.1 .  Clang-90 also uses python 3.5, a python version soon to be obsolete in OI.  Since clang-13 is entirely adequate, the best solution for clang-90 is to obsolete it.

Clang-90 and clang-13 consist of the LLVM tools and the Clang compiler.  The clang-90 package is developer/clang-90 .  This package is made obsolete by this PR.  Likewise, the directory components/developer/clang-90 is deleted by this PR.  The global history file is modified to obsolete the package.


AndWac:
It is not sufficient to obsolete clang-90 only. We need a follow-up PR with all packages that are using it switched to clang-13.
pkg search -r -H -o pkg.name "depend:require:developer/clang-90" | sort -u                                                                                 developer/lang/rustc
library/qt5
library/qt6

So, you need to build rustc, qt5, and qt6 with clang-13, make sure they work as expected and create another PR with the changes before we can obsolete clang-90.

jgmills:
I can rebuild and test rust, but it's difficult and may not be possible.  In addition to the clang change, it uses python 2.7 for the build.  That has to be changed as well.  The whole build process is mostly undocumented.  All I can do is to try it and see.

I'll need some help with the two qt versions.  My build server does not have a graphic console, ruling out the testing of X11 applications.  A quick look at the qt? Makefiles suggests that the whole thing is simpler there, likely only requiring the clang change.  It may only require unintallling the old clang version and installing the new one, along with appropriate Makefile changes.

shuryanc:
I will handle qt6 upgrade to 6.3. Will change to clang13 at the same time.

AndWac:
I have updated qt5 and switched to clang-13.

hoewweken:
> I will handle qt6 upgrade to 6.3. Will change to clang13 at the same time.

Hello! i have read that QT6.2 is a LTS Release. Are we shure that we want/need 6.3? I have no clue :-)

shuryanc:
> > I will handle qt6 upgrade to 6.3. Will change to clang13 at the same time.
> 
> Hello! i have read that QT6.2 is a LTS Release. Are we shure that we want/need 6.3? I have no clue :-)

Nvm, run into issue in upgrade to 6.3 and cannot be fixed easily. Will stick with 6.2.

PR #8195, opened 2022-05-18
Add IO::Socket::SSL perl module

mtelka:


components/perl/IO-Socket-SSL/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+BUILD_STYLE = makemaker
+BUILD_BITS = 32_and_64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		IO-Socket-SSL
+COMPONENT_VERSION =		2.074
+IPS_COMPONENT_VERSION =		2.74
+COMPONENT_FMRI =		library/perl-5/io-socket-ssl
```
AndWac, line 25: Please make use of HUMAN_VERSION (here and in the manifest file). See archiver/lzop for an example use.

mtelka: Done.

components/perl/IO-Socket-SSL/Makefile
```diff
@@ -0,0 +1,66 @@
...
+
+BUILD_STYLE = makemaker
+BUILD_BITS = 32_and_64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		IO-Socket-SSL
+HUMAN_VERSION =			2.074
+COMPONENT_VERSION =		2.74
+COMPONENT_FMRI =		library/perl-5/io-socket-ssl
```
AndWac, line 25: Maybe COMPONENT_VERSION=2.0.74 would be better?

mtelka: Honestly, these perl developers use crazy numbering schemes.  For example this module have this history:
1.82 -> 1.83 -> 1.831 -> 1.84
then
1.94 -> 1.950 -> 1.951
and finally
1.999 -> 2.000 -> 2.001

What will we do when they go from 2.099 to 2.100?  Will we move from 2.0.99 to 2.1.0?  What about once they go from 2.079 to 2.0791?  They apparently do not understand what dots are for.

Maybe the best option is just to add dots between all digits after the first dot (like 2.0.7.4).  In a case there is a module with major version number greater than 9 (like 13.025) we would use 13.0.2.5.

BTW, most (maybe almost all) perl modules does this.

OTOH, in this particular case they apparently consider 074 as pure 74 (leading zero is insignificant). And in a case they release something like 2.0741 I'd just use 2.74.1.

AndWac: Ok, sounds reasonable.

PR #8200, opened 2022-05-19
Updated ply to 3.11

shuryanc:
Updated ply to 3.11


AndWac:
Have you checked that pkg still works with this version?
library/python/ply-35 has the following dependent packages:
developer/opensolaris/pkg5
package/pkg

shuryanc:
> 

May I know how to test? And do I need to go into some folders to build sth?
I am able to find the packages that depend on library/python/ply-35 but unable to locate the folder to build developer/opensolaris/pkg5 and package/pkg

AndWac:
These packages belong to IPS, the Image Packaging System. It's under openindiana/pkg.
IPS depends on some python packages we deliver, ply is one of them.
In order to test a python package that is needed by IPS you'll need to update/install it on your computer and test IPS, that is make sure that at least pkg is still working. As IPS is fundamental to OI changes need to tested thoroughly.
You can find oi-userland folders for a package when you seek for pkg5 files (which contain package meta data) and therein for the package, eg
find . -name pkg5 -exec grep package\\\/pkg {} \; -print

shuryanc:
Ok. Seems pkg is working fine after building ply3.11 and openindiana/pkg and installing ply3.11, developer/opensolaris/pkg5 and package/pkg. I have an issue that this Makefile has to keep COMPONENT_REVISION=5 or it won't be able to install due to this line in ply-PYVER.p5m:
depend type=require \
    fmri=library/python/ply@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)

The error is: No version matching 'conditional' dependency library/python/ply-35@3.11-2022.0.0.**5** can be installed

Is there anyway to reset the revision to 0?

AndWac:
The COMPONENT_REVISION is reset either by explicitly assigning 0 to it in the Makefile or by just removing the definition from the Makefile. This should be done when the COMPONENT_VERSION increases. Decreasing the COMPONENT_VERSION is not possible after a version has been published because the part of an FMRI that fully defines the version of a package needs to be strict monotone.
Something must be broken if you'll get this error. Either your package settings or your local repository must be broken. You can try to clean your local repository. pkgrepo is capable of removing FMRI's from a repository. The fastest way to clean a repository is to 
rm -rf i386/repo && gmake setup
from oi-userland's top folder. This resets the userland repository (encumbered has its own folder). The drawback is that you may need to rebuild all your formerly locally published packages again.

AndWac:
Ok, I also tested it on my build system for some days.

PR #8205, opened 2022-05-19
Updated pyparsing to 3.0.9

shuryanc:
Updated pyparsing to 3.0.9


AndWac:
What do you want to achieve? This PR looks only half-baked.
The old pyparsing Makefile (and its helpers pyparsing-PYV.p5m and pyparsing-GENFRAG.p5m) had created four packages:
1. python/pyparsing
2. python/pyparsing-35
3. python/pyparsing-37
4. python/pyparsing-39

With your changes we would end with 3 dangling packages. And we wouldn't be able to re-publish them because you have removed the machinery that creates the pyparsing-$(PYV) variants of the packages.
Furthermore, these packages (pyparsing-$(PYV)) are being used by other packages, eg:
pkg search -r -H -o pkg.name "depend:require:library/python/pyparsing-35" | sort -u
will tell you which package(s) depend on pyparsing-35.

shuryanc:
sip6 will need newer pyparsing, which will be needed by pyqt5 (Which will be used by some kde framework libraries). But latest pyparsing requires Python >=3.6.8. Maybe I should create an extra pyparsing such as pyparsing3?
https://pypi.org/project/pyparsing/

AndWac:
We still need python-3.5 as IPS depends on it. Work is underway to replace it with python-3.9 but there is still some work to do. Parts of it are from illumos-gate.
What you need to do for pyparsing has already been done for some other python packages. You need to move those packages that cannot be updated (because they aren't supported by the update) into a separate folder, eg. pyparsing-legacy (or in this case pyparsing-35).  The move can be done by copying the original folder contents into a newly created folder (before the changes for the update happen) and change there the PYTHON_VERSIONS to only 3.5.
Increment its COMPONENT_REVISION, update its sample-manifest.p5m and generic-manifest.p5m and then add all files into a separate commit.
Then you can make your changes to the pyparsing folder, updating pyparsing for python 3.7 and 3.9.
Note 1: You still need pyparsing-PYVER.p5m in both folders. I am not sure about pyparsing-GENFRAG.p5m, which only contains an old dependency. It might be ok to just keep one copy of it.
Note 2: With the right flag enabled (GENERATE_GENERIC_TRANSFORMS=1) generic-manifest files will be generated where applicable (another requirement is <package>-PYVER.p5m exists, even if empty). This generic-manifest.p5m file can be used for the <package>-PYVER.p5m file with only minimal changes.
If (and only if in my view) a <package>-GENFRAG.p5m exists (and has some contents) we also need to make the <package>-PYVER packages also depend on the base package. Here it would be that pyparsing-35, pyparsing-37, and pyparsing-39 depend on pyparsing. That's because <package>-GENFRAG.p5m contains actions that is common to all packages and have been refactored into a separate file.
Note 3: We have some python packages where the formerly mentioned dependency to the base package is modeled but wrongly in my eyes.

AndWac:
Note 4: Your PR would have 2 commits: one to add the new folder (pyparsing-legacy or pyparsing-35) with its contents and a separate one with the updated pyparsing in the original folder.
You can look at the folders below the python folder. We already have some examples.

shuryanc:
Thanks, will do the correction.

components/python/pyparsing-35/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,39 @@
...
+file path=usr/lib/python3.7/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.7.egg-info/PKG-INFO
+file path=usr/lib/python3.7/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.7.egg-info/SOURCES.txt
+file path=usr/lib/python3.7/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.7.egg-info/dependency_links.txt
+file path=usr/lib/python3.7/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.7.egg-info/top_level.txt
+file path=usr/lib/python3.7/vendor-packages/pyparsing.py
+file path=usr/lib/python3.9/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.9.egg-info/PKG-INFO
+file path=usr/lib/python3.9/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.9.egg-info/SOURCES.txt
+file path=usr/lib/python3.9/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.9.egg-info/dependency_links.txt
+file path=usr/lib/python3.9/vendor-packages/pyparsing-$(COMPONENT_VERSION)-py3.9.egg-info/top_level.txt
+file path=usr/lib/python3.9/vendor-packages/pyparsing.py
```
AndWac, line 39: You'll need to regenerate the sample-manifest because here aren't versions for python-3.7 and 3.9.

components/python/pyparsing/Makefile
```diff
@@ -59,7 +85,14 @@ COMPONENT_TEST_MASTER = $(COMPONENT_TEST_RESULTS_DIR)/results-all.master
 
 ASLR_MODE = $(ASLR_NOT_APPLICABLE)
 
+# Manually added dependencies
+REQUIRED_PACKAGES += library/python/pep517-37
```
AndWac, line 91: You have missed the dependencies on
library/python/build-37
library/python/build-39

PR #8206, opened 2022-05-19
Added python module installer-0.5.1

shuryanc:
Added python module installer-0.5.1

### DIFF
--- a/components/python/installer/Makefile
+++ b/components/python/installer/Makefile
@@ -1,0 +1,58 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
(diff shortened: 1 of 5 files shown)


components/python/installer/Makefile
```diff
@@ -0,0 +1,54 @@
...
+COMPONENT_LICENSE=      MIT
+COMPONENT_LICENSE_FILE= LICENSE
+
+PYTHON_VERSIONS=	3.7 3.9
+
+include $(WS_MAKE_RULES)/common.mk
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += library/python/packaging
+REQUIRED_PACKAGES += library/python/ply
```
AndWac: The dependencies aren't correct.
Depending on the base package is not correct if there are specialised packages, eg. packaging-37 and packaging-39.
So, for packaging it should be
REQUIRED_PACKAGES += library/python/packaging-37
REQUIRED_PACKAGES += library/python/packaging-39

Also setuptools, toml and ply have python version dependent packages that need to be required instead of their base packages.

PR #8207, opened 2022-05-19
Added pyqt-builder-1.12.2

shuryanc:
Added pyqt-builder-1.12.2

### DIFF
--- a/components/python/PyQt-builder/Makefile
+++ b/components/python/PyQt-builder/Makefile
@@ -1,0 +1,54 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright 2022 Daniel Chan
+#
+
+BUILD_BITS=			64
+BUILD_STYLE=		setup.py
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		PyQt-builder
+COMPONENT_VERSION=	1.12.2
+COMPONENT_SUMMARY=	The PEP 517 compliant PyQt build system
+COMPONENT_PROJECT_URL=	https://pypi.org/project/PyQt-builder/
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= sha256:f62bb688d70e0afd88c413a8d994bda824e6cebd12b612902d1945c5a67edcd7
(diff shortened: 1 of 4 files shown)


components/python/PyQt-builder/Makefile
```diff
@@ -0,0 +1,49 @@
...
+COMPONENT_CLASSIFICATION=	Development/Python
+COMPONENT_LICENSE=      GPLv3
+COMPONENT_LICENSE_FILE= LICENSE-GPL3
+
+PYTHON_VERSIONS=	3.9
+
+include $(WS_MAKE_RULES)/common.mk
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += library/python/setuptools-39
```
AndWac, line 53: You seem to have missed the dependency on
library/python/sip-39
(check the package's metadata, eg. setup.cfg and setup.py).

components/python/PyQt-builder/Makefile
```diff
@@ -0,0 +1,49 @@
...
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= sha256:f62bb688d70e0afd88c413a8d994bda824e6cebd12b612902d1945c5a67edcd7
+COMPONENT_ARCHIVE_URL=	$(call pypi_url)
+COMPONENT_FMRI=	library/python/pyqtbuilder
+COMPONENT_CLASSIFICATION=	Development/Python
+COMPONENT_LICENSE=      GPLv3
+COMPONENT_LICENSE_FILE= LICENSE-GPL3
+
+PYTHON_VERSIONS=	3.9
+
```
AndWac, line 45: This looks like simple tool and not a framework. So we might only want to have a single package pyqt-builder and not multiple packages (pyqt-builder, pyqt-build-39).
If so, just use
PYTHON_VERSION= 3.9
here and omit the -pyver part on the manifest. You can look at out ansible package for an example.

components/python/PyQt-builder/PyQt-builder.p5m
```diff
@@ -0,0 +1,84 @@
...
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
```
AndWac, line 24: setup.py states that installation of this tool requires packaging and sip (>= 6.3 and < 7). Thus, you'll need to add runtime dependencies here:

depend type=require fmri=library/python/packaging-39
depend type=require fmri=library/python/sip-39

components/python/PyQt-builder/PyQt-builder.p5m
```diff
@@ -0,0 +1,87 @@
...
+depend type=require fmri=library/python/packaging-39
+depend type=require fmri=library/python/sip
+
```
AndWac, line 27: This should be sip-39 because this package is being build with python-3.9.

shuryanc: I think sip is a tool (sip has some executable files under /usr/bin), so I changed it to sip, how do you think?

AndWac: That would be ok then. Did you check that sip works with other python versions, too?
We will probably drop python 3.5 as soon as IPS works with 3.9. Probably 3.7 gets dropped next and 3.10 or 3.11 added.

shuryanc: Yes, tested building pyqt-builder by python3.7 and the 3.9 sip still works. I will try to make them to adapt new version of python when higher version is shipped.

PR #8225, opened 2022-05-21
Added pyparsing-35

shuryanc:
Added pyparsing-35 to prevent from broken dependency after pyparsing update for python 3.7 and python 3.9


AndWac:
The commit from this PR should be in #8205 as a 2nd commit. This would allow for easier testing and integration. Plus, they belong together and should not be merged separately.

PR #8227, opened 2022-05-21
Added python module sip4

shuryanc:
Added python module sip4


AndWac:
Where is this package needed?

shuryanc:
Will be needed by krita

PR #8229, opened 2022-05-21
Added PyQt5-sip

shuryanc:
Added PyQt5-sip, will be needed by PyQt5 in order for sip to work with it properly.

### DIFF
--- a/components/python/PyQt5-sip/Makefile
+++ b/components/python/PyQt5-sip/Makefile
@@ -1,0 +1,62 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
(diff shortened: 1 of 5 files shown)


AndWac:
If this is a dependency for PyQT5 why isn't it in its (runtime) dependencies?

components/python/PyQt5-sip/PyQt5-sip.p5m
```diff
@@ -0,0 +1,31 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend type=require fmri=library/python/pyqt5
+
```
AndWac: From your original PR comment I would have expected to have a dependency on pyqt5-sip in the manifest of pyqt5.

AndWac:
We now have a dependency cycle:
Creating Plan (Solver setup): /
pkg install: No matching version of library/python/pyqt5 can be installed:
  Reject:  pkg://openindiana.org/library/python/pyqt5@5.15.6-2022.0.0.0
           pkg://openindiana.org/library/python/pyqt5@5.15.6-2022.0.0.0
  Reason:  No version for 'require' dependency on library/python/pyqt5-sip can be found

pyqt5-sip depends on pyqt5 and pyqt5 runtime depends on pyqt5-sip...

PR #8230, opened 2022-05-21
bump to gcc-11.3.0, make dirty patches clean.

klausz65:
Patch: 0020-Use-compare-debug-to-pass-stage2-stage3-comparison-a.patch has been up-streamed.


components/developer/gcc-11/Makefile
```diff
@@ -18,12 +18,13 @@
...
-COMPONENT_VERSION= 11.2.0
-COMPONENT_REVISION= 1
+COMPONENT_VERSION= 11.3.0
 COMPONENT_SRC= gcc-releases-gcc-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE= $(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.gz
 COMPONENT_ARCHIVE_URL=	https://github.com/gcc-mirror/gcc/archive/releases/$(COMPONENT_ARCHIVE)
-COMPONENT_ARCHIVE_HASH=	sha256:5efe343a97e95fe9031aa23e103755ec3bc56855c865c4b6ffcbad6c25675514
+COMPONENT_ARCHIVE_HASH=	sha256:5b117751dd46ebfe9774da9efc977c094f1e0a7061c887d5f8b3fe00725186d2
+
+COMPONENT_BUILD_ARGS=	-j20
```
AndWac: Please remove this. We set the build parallelism in the jenkins job.

alarcher:
This commit also touches .mk files and other files silently without explanation.
I do not see why you call the patches "dirty", can you explain?

All patches from this changeset hat wrong offsets, therefore I corrected those offsets, and obsoleted: 0020-Use-compare-debug-to-pass-stage2-stage3-comparison-a.patch because it is already upstreamed. I will remove the .mk files and provide them with a extra #PR.

components/developer/gcc-11/Makefile
klausz65: Okay, will remove COMPONENT_BUILD_ARGS = j20

AndWac (approved):

components/developer/gcc-11/Makefile
```diff
@@ -56,16 +55,23 @@ CONFIGURE_LIBDIR.64 = $(CONFIGURE_PREFIX)/lib
 CONFIGURE_OPTIONS+= --with-diagnostics-urls=auto-if-env
 CONFIGURE_OPTIONS+= --enable-plugins
 CONFIGURE_OPTIONS+= --enable-objc-gc
-CONFIGURE_OPTIONS.i386+= --enable-initfini-array
+CONFIGURE_OPTIONS.sparc+= --with-mulhigh-size=2048
+CONFIGURE_OPTIONS.sparc+= --with-cpu=ultrasparc3
+CONFIGURE_OPTIONS.sparc+= --enable-float128
```
alarcher, line 60: Out of curiosity, is float128 enabled also on amd64?

klausz65: From the top of my head I think yes, per default.

alarcher: There is simply no support, __Float128 is defined nowhere.

PR #8231, opened 2022-05-21
sxhkd: add version 0.6.2

torculus:
Adding Simple X Hotkey Daemon (sxhkd) so that users can keep their custom keybindings across different window managers. It's also a dependency for bspwm, which I hope to add as well

### DIFF
--- a/components/desktop/sxhkd/Makefile
+++ b/components/desktop/sxhkd/Makefile
@@ -1,0 +1,40 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Benjamin S. Osenbach
+#
+
+BUILD_BITS=64 # for binaries or 32_and_64 for libraries
+BUILD_STYLE=justmake
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=sxhkd
+COMPONENT_VERSION=0.6.2
+COMPONENT_SUMMARY= Simple X hotkey daemon
+COMPONENT_PROJECT_URL=https://github.com/baskerville/sxhkd
+COMPONENT_FMRI=application/sxhkd
+COMPONENT_CLASSIFICATION=Applications/System Utilities
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=https://github.com/baskerville/sxhkd/archive/refs/tags/$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=sha256:1edc8b1a8b3631c10ba9cb9df1181830dacbbdf77adb558e31d5dd2029637386
+COMPONENT_LICENSE=BSD
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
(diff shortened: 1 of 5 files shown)


components/desktop/sxhkd/Makefile
```diff
@@ -0,0 +1,37 @@
...
+COMPONENT_PROJECT_URL=https://github.com/baskerville/sxhkd
+COMPONENT_FMRI=application/sxhkd
+COMPONENT_CLASSIFICATION=Applications/System Utilities
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=https://github.com/baskerville/sxhkd/archive/refs/tags/$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=sha256:1edc8b1a8b3631c10ba9cb9df1181830dacbbdf77adb558e31d5dd2029637386
+COMPONENT_LICENSE=MIT
+COMPONENT_LICENSE_FILE=sxhkd.license
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
```
AndWac, line 30: $(COMPONENT_NAME).license is the default for COMPONENT_LICENSE_FILE and thus you don't explicitly set it here.

components/desktop/sxhkd/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,34 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/local/bin/sxhkd
+file path=usr/local/share/doc/sxhkd/examples/background_shell/profile
```
AndWac: Please change the base path. If part of oi-userland the package should install itself into /usr and not /usr/local.

components/desktop/sxhkd/Makefile
torculus: Thanks. I unset COMPONENT_LICENSE_FILE and noticed that the license was incorrect (was supposed to be BSD)

components/desktop/sxhkd/manifests/sample-manifest.p5m
torculus: Okay, I added a patch to use `PREFIX=/usr` instead of `/usr/local`. Seemed like the most elegant solution (1 line of code changed instead of bundling multiple statements into the build action)

components/desktop/sxhkd/patches/01-fix-usr-local.patch
```diff
@@ -0,0 +1,11 @@
+--- sxhkd-0.6.2/Makefile.orig	2020-08-02 00:53:02.000000000 +0000
++++ sxhkd-0.6.2/Makefile	2022-05-23 07:45:39.374284976 +0000
+@@ -7,7 +7,7 @@
+ LDFLAGS  ?=
+ LDLIBS    = $(LDFLAGS) -lxcb -lxcb-keysyms
+ 
+-PREFIX    ?= /usr/local
++PREFIX    ?= /usr
+ BINPREFIX ?= $(PREFIX)/bin
```
AndWac: I'd prefer a solution inside our Makefile (setting the PREFIX there as the use of ?= indicates that this is intended).
The reason for this is that your patch needs adoption every time the Makefile changes near to your patch while setting PREFIX in out Makefile would continue to work.

torculus: Done! Thanks for your help, this is more elegant than using a patch

PR #8235, opened 2022-05-23
shared-macros.mk: add 5.34 to PERL_VERSIONS and PERL_64_ONLY_VERSIONS

mtelka:


AndWac:
Are all perl modules available now as the comment stated as necessary?

mtelka:
> Are all perl modules available now as the comment stated as necessary?

Yes.

PR #8238, opened 2022-05-23
shared-macros.mk: tested-and-compared targets should fail if test res…

mtelka:
…ults are missing

Without this if you do `rm -rf test; mkdir test; touch test/results-FOO.master` (in a component directory) then `gmake test` will silently pass.  This will help to catch cases when some master test results are missing, but others are there.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -488,47 +488,39 @@ COMPONENT_TEST_TRANSFORMS = \
 # set the default commands used to generate the file containing the set
 # of transforms to be applied to the test results to try to normalize them.
 COMPONENT_TEST_CREATE_TRANSFORMS = \
-	if [ -e $(COMPONENT_TEST_MASTER) ]; \
-	then \
-		print "\#!/bin/sh" > $(COMPONENT_TEST_TRANSFORM_CMD); \
-        	print '$(COMPONENT_TEST_TRANSFORMER) ' \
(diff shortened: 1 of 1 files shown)


AndWac:
This doesn't work as expected for developer/ninja:
With your changes I'll get an error 1:
...
esults"; if [ -s /export2/home/andreas/oi-userland/components/developer/ninja/build/test/amd64/test-64-diffs ]; then print "Differences found."; /bin/cat /export2/home/andreas/oi-userland/components/developer/ninja/build/test/amd64/test-64-diffs; exit 2; else print "No differences found."; fi
gmake: *** [/export2/home/andreas/oi-userland/make-rules/justmake.mk:89: /export2/home/andreas/oi-userland/components/developer/ninja/build/amd64/.tested-and-compared] Error 1

This doesn't happen without your changes.

mtelka:
> This doesn't work as expected for developer/ninja: [...snip...]

No, this works exactly as expected :-).  The problem is in the `developer/ninja` package.  Its `Makefile` contains this:

`COMPONENT_TEST_MASTER = $(COMPONENT_TEST_RESULTS_DIR)/results-all.master`

but such file is not there, instead there is [results-64.master](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/developer/ninja/test/results-64.master) only.

AndWac:
Haha, ok. I didn't notice that discrepancy.

PR #8262, opened 2022-05-26
Security update to ntfsprogs

jpandre:
Security version 2022.5.17 (May 26, 2022)
- improved defence against maliciously tampered NTFS partitions
- improved defence against improper use of options
- updated the documentation


AndWac:
Just a note: You only need to explicitly set IPS_COMPONENT_VERSION if COMPONENT_VERSION is not suitable for IPS, eg. because it contains leading zeroes or non-digits. So, it would be sufficient here to just use COMPONENT_VERSION=2022.5.17.

PR #8267, opened 2022-05-26
jxrlib: Fixed fmri and updated build path for a jar file

shuryanc:
Fixed name and moved build path for a jar file

### DIFF
--- a/components/library/jxrlib/Makefile
+++ b/components/library/jxrlib/Makefile
@@ -21,7 +21,7 @@ COMPONENT_NAME=		jxrlib
 COMPONENT_VERSION=	0.2.4
 COMPONENT_SUMMARY=	Decode and encode JPEG-XR data from Java
 COMPONENT_PROJECT_URL=	https://github.com/glencoesoftware/jxrlib
-COMPONENT_FMRI=		library/jxlib
+COMPONENT_FMRI=		library/jxrlib
 COMPONENT_CLASSIFICATION=System/Libraries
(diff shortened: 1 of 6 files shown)


components/library/jxrlib/Makefile
```diff
@@ -19,9 +19,10 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		jxrlib
 COMPONENT_VERSION=	0.2.4
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	Decode and encode JPEG-XR data from Java
 COMPONENT_PROJECT_URL=	https://github.com/glencoesoftware/jxrlib
-COMPONENT_FMRI=		library/jxlib
+COMPONENT_FMRI=		library/jxrlib
 COMPONENT_CLASSIFICATION=System/Libraries
```
AndWac, line 25: When you change COMPONENT_FMRI you'll need to obsolete the old package. You can do that with a local history file. In this case you should use the special syntax to forward from the old to the new fmri. In this case you also don't need to set the COMPONENT_REVISION because you'll create a new package.

components/library/jxrlib/history
```diff
@@ -0,0 +1 @@
+library/jxlib@0.2.4,5.11-2022.0.0.0 library/jxrlib
```
AndWac: When you obsolete a package make sure that its COMPONENT_REVISION here is higher than the last released one:
library/jxlib@0.2.4,5.11-2022.0.0.1 library/jxrlib

PR #8268, opened 2022-05-26
freeimage: Changed to use distro packages rather than the packed files

shuryanc:
Changed freeimage to use distro pacakges rather than the packed files.


components/library/FreeImage/Makefile
```diff
@@ -35,17 +36,58 @@ COMPONENT_LICENSE_FILE=	license-gplv3.txt
...
+
+FIFLAGS=-fPIC
+FIFLAGS+=-fexceptions
+FIFLAGS+=-fvisibility=default
+CFLAGS+= $(FIFLAGS)
+CXXFLAGS+=$(FIFLAGS)
 CXXFLAGS+=-Wno-ctor-dtor-privacy
 
+# Build dependencies
+REQUIRED_PACKAGES += image/library/libjpeg8-turbo
```
AndWac: We cannot distinguish between build, runtime and automatically detected packages. We would need a new name, eg. REQUIRED_BUILD_PACKAGES and some handling for it in the make-rules. We only manually add build or test dependencies if gmake REQUIRED_PACKAGES cannot detect them automatically. The detection uses a simple mechanism and fails to detect some kinds of dependencies.

PR #8280, opened 2022-05-31
fix proftpd firewall settings

oicontrib:
Hi.
I think the file 'components/network/proftpd/svc-ftp' is outdated.
The function 'generate_rules' currently has no parameter for "any" (see https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/svc/shell/ipf_include.sh#L707)

The line "shell/ksh93" was added to components/network/proftpd/pkg5 by the build process. I am not sure if the ksh93 should be in dependencies.


AndWac:
If a package should be re-deployed either COMPONENT_VERSION must be incremented (and COMPONENT_REVISION reset) or COMPONENT_REVISION must be incremented. Here you should add
COMPONENT_REVISION= 1
to the Makefile.

oicontrib:
> Here you should add COMPONENT_REVISION= 1 to the Makefile.

Thanks!

PR #8368, opened 2022-06-02
Adding Golang 118

Toasterson:

PR #8370, opened 2022-06-02
Switch cloney to hardlinks instead of softlinks

Toasterson:
Softlinks are causing more and more problems when using cloney. As the whole repo must be within the same Filesystem anyway it is safe to switch to hardlinks


Toasterson:
Required for #8368

alarcher:
Does it break gmake clean?

AndWac:
Looks like gmake clean still wroks, at least with the packages I have checked.

AndWac (approved):

PR #8373, opened 2022-06-03
babl archive download URL should be changed

gounthar:
The `make download` command was giving me this error:

```bash
 ssl.CertificateError: hostname 'ftp.gimp.org' doesn't match either of 'download.gimp.org', 'gimp.org'
```

The new `URL` works.


gounthar:
Thanks 👍

PR #8384, opened 2022-06-05
nim: Add package

ghost:
Add package for the Nim programming language.
Nim itself seems to be fully functional, as I have tested by building the Nim program Nitter.
The tests fail with really weird errors, however, so I disabled them for now.


ghost:
I added a patch for an issue with SSL connections that also appears on NetBSD and isn't yet fixed upstream: https://github.com/nim-lang/Nim/issues/17658

PR #8391, opened 2022-06-06
redis: Update to 7.0.0

ghost:
This builds, installs and runs fine here (tried with another application that uses redis) but the tests fail, always at a different point with some weird I/O Exception. Please try before merging.


ghost:
Please use version 6.2.7 (#8395) instead.

PR #8392, opened 2022-06-06
Introduce cloney mode

Toasterson:
It will need some testing

### DIFF
--- a/tools/cloney
+++ b/tools/cloney
@@ -32,9 +32,13 @@ srcdir=$1
 destdir=$2
 
 PATH=/usr/bin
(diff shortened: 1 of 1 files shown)


tools/cloney
```diff
@@ -32,9 +32,11 @@ srcdir=$1
 destdir=$2
 
 PATH=/usr/bin
+CLONEY_MODE="symlink"
```
AndWac, line 38: Shouldn't this be
CLONEY_MODE?="symlink"
?

Toasterson: Yes, I apparently killed that line out of my commit. Added

PR #8466, opened 2022-06-13
Added imath as a replacement of ilmbase

shuryanc:
IlmBase has been superceded by Imath since OpenEXR 3.0. 
The affected packages will be pushed.

pkg search -r -H -o pkg.name "depend:require:library/ilmbase" | sort -u
image/editor/gimp
library/audio/gstreamer1/plugin/bad
library/libfreeimage
SUNWilmbase

### DIFF
--- a/components/library/imath/Makefile
+++ b/components/library/imath/Makefile
@@ -1,0 +1,45 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Daniel Chan
+#
+
+BUILD_BITS=		32_and_64
+BUILD_STYLE=	cmake
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		Imath
+COMPONENT_VERSION=	3.1.5
+COMPONENT_FMRI=		library/imath
+COMPONENT_CLASSIFICATION=System/Libraries
+COMPONENT_PROJECT_URL=	https://github.com/AcademySoftwareFoundation/$(COMPONENT_NAME)
+COMPONENT_SUMMARY=	C++ representation of 2D and 3D vectors
+COMPONENT_SRC=			$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:1e9c7c94797cf7b7e61908aed1f80a331088cc7d8873318f70376e4aed5f25fb
+COMPONENT_ARCHIVE_URL=  $(COMPONENT_PROJECT_URL)/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_LICENSE=	BSD 3-Clause
+COMPONENT_LICENSE_FILE=	LICENSE.md
+
+TEST_TARGET=		$(NO_TEST)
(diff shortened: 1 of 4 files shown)


alarcher:
@shuryanc Can you check whether the patch is still needed?

components/meta-packages/history/history
```diff
@@ -896,6 +896,7 @@ library/gnome/print/gnome-print/gnome-print-papi@2.30.0,5.11-2014.0.0.0
 library/gnome/print/gnome-print@2.30.0,5.11-2014.0.0.0
 library/gnu-libiconv@1.14,5.11-2017.0.0.1 noincorporate
 library/gnutls@2.12.24,5.11-2018.0.0.4
+library/ilmbase@2.2.0,5.11-2020.0.1.0
 library/java/cairo-java@1.0.8,5.11-2017.0.0.0
```
AndWac: You'll need to increment the COMPONENT_REVISION here, so make it
library/ilmbase@2.2.0,5.11-2020.0.1.1

AndWac:
You'll need to split this PR into two, because we cannot obsolete ilmbase before Imath has been added and the dependent packages have been rebuilt with Imath. The Jenkins run would fail if any of the dependencies is installed.
Thus, we need to
1. add Imath
2. change the packages to use Imath instead of ilmbase
3. obsolete ilmbase

shuryanc:
> You'll need to split this PR into two, because we cannot obsolete ilmbase before Imath has been added and the dependent packages have been rebuilt with Imath. The Jenkins run would fail if any of the dependencies is installed. Thus, we need to
> 
>     1. add Imath
> 
>     2. change the packages to use Imath instead of ilmbase
> 
>     3. obsolete ilmbase

Thanks



> @shuryanc Can you check whether the patch is still needed?

No patch is needed. As least it works perfectly for 64 bits.

components/library/imath/Makefile
```diff
@@ -0,0 +1,44 @@
...
+COMPONENT_NAME=		Imath
+COMPONENT_VERSION=	3.1.5
+COMPONENT_FMRI=		library/imath
+COMPONENT_PROJECT_URL=	https://github.com/AcademySoftwareFoundation/$(COMPONENT_NAME)
+COMPONENT_SUMMARY=	C++ representation of 2D and 3D vectors
+COMPONENT_SRC=			$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:1e9c7c94797cf7b7e61908aed1f80a331088cc7d8873318f70376e4aed5f25fb
+COMPONENT_ARCHIVE_URL=  $(COMPONENT_PROJECT_URL)/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_LICENSE=	BSD 3-Clause
```
AndWac, line 30: COMPONENT_CLASSIFICATION is referenced from the manifest but not defined.

alarcher:
> @shuryanc Can you check whether the patch is still needed?

@shuryanc @AndWac I do not think you have seen my question whether the only patch in ilmbase is still required in imath to ensure that the code works as expected.

AndWac:
How to proceed from here? We seem to be stuck and cannot move forward or
backward. Other updates and rebuilds are waiting...


Am 6/15/22 um 14:49 schrieb Aurélien Larcher:

shuryanc:
I don't understand. imath doesn't have similar code like the patch and the patch cannot be applied to imath. And my gmake test was good. Correct me and provide some hints to me so that I can make thing moves forwarding.

alarcher:
> I don't understand. imath doesn't have similar code like the patch and the patch cannot be applied to imath. And my gmake test was good. Correct me and provide some hints to me so that I can make thing moves forwarding.

The question was whether the patch sill applies. If the answer is no then there is nothing more to add :)

shuryanc:
The patch doesn't apply. Thanks.

PR #8467, opened 2022-06-13
Updated openexr to 3.1.5

shuryanc:
Update openexr to 3.1.5 which based on imath.

Affected packages have been tested and seems good (some requires imath fix):
pkg search -r -H -o pkg.name "depend:require:library/openexr" | sort -u
image/editor/gimp
image/imagemagick
library/audio/gstreamer1/plugin/bad
library/libfreeimage


AndWac:
We have a problem with ilmbase and imath I didn't catch. The cannot be installed at the same time because they deliver clashing links:

The requested change to the system attempts to install multiple actions
for link 'usr/lib/amd64/libImath.so' with conflicting attributes:

    1 package delivers 'link path=usr/lib/amd64/libImath.so target=libImath-2_2.so.12.0.0':
        pkg://openindiana.org/library/ilmbase@2.2.0,5.11-2020.0.1.0:20200330T121826Z
    1 package delivers 'link path=usr/lib/amd64/libImath.so target=libImath-3_1.so':
        pkg://openindiana.org/library/imath@3.1.5,5.11-2022.0.0.0:20220614T183609Z

shuryanc:
Yes, imath and ilmbase are both developed by AcademySoftwareFoundation and imath is to replace ilmbase. So they cannot coexist and require removal of ilmbase before installing imath.

AndWac:
Then this needs manual intervention on the build server. I need to remove ilmbase and all its dependencies before we can continue.

PR #8468, opened 2022-06-13
FreeImage: now depends on imath

shuryanc:
Due to replacement of ilmbase by imath, freeimage needs to be fixed.


components/library/FreeImage/pkg5
```diff
@@ -6,7 +6,7 @@
         "image/library/libraw",
         "image/library/libtiff",
         "image/library/openjpeg",
-        "library/ilmbase",
+        "library/imath",
         "library/jxrlib",
```
AndWac, line 10: How is the dependency on imath detected on the dependent packages? Do they need some "help", eg. by new flags?

shuryanc: I use _pkg search -r -H -o pkg.name "depend:require:library/ilmbase" | sort -u_ to search for the packages that required imath. After building and installing imath, I use gmake REQUIRED_PACKAGES to generate the required packages for those packages that originally required ilmbase.

PR #8469, opened 2022-06-13
gst-plugins-bad1 now depends on imath

shuryanc:
Due to replacement of ilmbase by imath, library/audio/gstreamer1/plugin/bad needs to be fixed.


AndWac:
This is now obsolete as gst-plugin-bad1 has been updated.

shuryanc:
This is fixed.

PR #8470, opened 2022-06-13
Gimp fix after replacing ilmbase by imath

shuryanc:
Due to replacement of ilmbase by imath, gimp needs to be fixed.

### DIFF
--- a/components/image/gimp/pkg5
+++ b/components/image/gimp/pkg5
@@ -21,7 +21,6 @@
         "library/desktop/pango",
         "library/glib2",
         "library/graphics/libmng",
-        "library/ilmbase",
         "library/lcms2",
         "library/libmypaint",
         "library/libpoppler",
--- a/components/image/gimp/Makefile
+++ b/components/image/gimp/Makefile
@@ -18,6 +18,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		gimp
 COMPONENT_VERSION=	2.10.32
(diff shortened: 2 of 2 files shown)


components/image/gimp/pkg5
```diff
@@ -20,7 +21,7 @@
         "library/desktop/pango",
         "library/glib2",
         "library/graphics/libmng",
-        "library/ilmbase",
+        "library/imath",
         "library/lcms2",
```
AndWac, line 24: Did you run "gmake REQUIRED_PACKAGES" to update the list?
I run it and it didn't detect the dependency on imath (although installed).

shuryanc:
Basically, imath should be needed by the packages that required ilmbase. But seems GIMP doesn't need imath since openexr 3.1.5.

AndWac:
I'd rather say gimp doesn't detect it anymore. I expect this to be an optional feature.

shuryanc:
I think I understand the reason. gimp requires libIex.so, which was in ilmbase. In v3.1.5, the libIex.so is in openexr instead of imath.

AndWac:
This is now obsolete as gimp has been updated.

AndWac:
We can only proceed if you fix the conflicts here and for gst-plugins-bad1. These packages have been changed after your PR's have been created.

shuryanc:
Sure. Will update this commit. Just busy with life recently and get things delayed.

shuryanc:
This is fixed.

PR #8484, opened 2022-06-16
Add Bit::Vector perl module

fritzkink:


components/perl/Bit-Vector/Makefile
```diff
@@ -0,0 +1,50 @@
...
+        '-e "/^PERL/d"' \
+        '-e "/^make/d"'
+
+REQUIRED_PACKAGES += runtime/perl-522
+REQUIRED_PACKAGES += runtime/perl-524
+REQUIRED_PACKAGES += runtime/perl-534
+REQUIRED_PACKAGES += library/perl-5/carp-clan-522
+REQUIRED_PACKAGES += library/perl-5/carp-clan-524
+REQUIRED_PACKAGES += library/perl-5/carp-clan-534
+
```
AndWac, line 43: The latest makemaker.mk changes make this superfluos. It is sufficient to only add the base package as a dependency. That is, library/perl-5/carp-clan
For an example look at Test-Most/Makefile

mtelka: > The latest makemaker.mk changes make this superfluos. It is sufficient to only add the base package as a dependency. That is, library/perl-5/carp-clan For an example look at Test-Most/Makefile

Yes.  Please note that to make this working you need to use `PERL_REQUIRED_PACKAGES` instead of `REQUIRED_PACKAGES`.  So in this case I recommend to use this line:

`PERL_REQUIRED_PACKAGES += library/perl-5/carp-clan`

fritzkink: thanks a lot for the hints.

PR #8509, opened 2022-06-19
Allow changing cloney args from Makefiles

Toasterson:


Toasterson:
Tested with golang 1.18 and jenkins-core-weekly. Switch works well.

PR #8512, opened 2022-06-21
developer/git: add subpackage with git-subtree

mtelka:


components/developer/git/Makefile
```diff
@@ -69,17 +70,13 @@ COMPONENT_BUILD_ENV+=	NO_PERL_MAKEMAKER=1
 COMPONENT_BUILD_ENV+=	V=1
 COMPONENT_INSTALL_ENV+=	NO_PERL_MAKEMAKER=1
 
-#
-# perl's install path for sparc is based off the value of the
-# "arch" command.  So we must package that way also.
-#
-PKG_MACROS +=		P_ARCH=$(shell arch)
```
AndWac, line 76: Do we loose SPARC support here or will it still be supported?

mtelka: It still should be supported, but I have no way to test it.  This change should make things more clear and simplify future updates of our (primary) perl.

Anyway, if there is some issue with SPARC support after this change I'll try to solve it.

PR #8519, opened 2022-06-25
Allow changing cloney args from Makefile in more cases

AndWac:
This completes call to cloney for existing invocations.
Note, that we should add more, eg. for BUILD_STYLE=cmake (pgagent could make use of it).


Toasterson (approved):
I don't know why my grep did not find these but LGTM

PR #8525, opened 2022-06-25
putty: update to 0.77 and provide it in its entirety

AndWac:
I didn't find a way to disable gtk anymore after the project has switched to cmake.
Thus, I have obsoleted putty-tools and introduced putty in its entirety.


Toasterson:
Isn't the new Manifest missing now?

AndWac:
Yes, typical error: new file not added.

PR #8532, opened 2022-06-26
Cleanup python mediator

AndWac:
This is the first step to clean the python mediator. The overall aim is to get rid of mediator=python3 and replace it with mediator=python.
/usr/bin/python and /usr/bin/amd64/python will be symlinks to python-2.7. Only python3 is mediated.


alarcher:
@AndWac This is a problem because some Xorg packages expect to see /usr/bin/python3; however since the mediator for python3 was removed and the python mediator is still set to 2.7 the change breaks all the packages relying on the existence of /usr/bin/python3. Some python3 mediators were incorrect but some of them actually had a function. How do you suggest we should proceed? I do not really feel like patching all the component to replace /usr/bin/python3 with something else, but I am not sure we can change the python mediator, or can we?

AndWac:
Am 14.07.22 um 11:18 schrieb Aurélien Larcher:
>
> @AndWac <https://github.com/AndWac> This is a problem because some
> Xorg packages expect to see /usr/bin/python3; however since the
> mediator for python3 was removed and the python mediator is still set
> to 2.7 the change breaks all the packages relying on the existence of
> /usr/bin/python3. Some python3 mediators were incorrect but some of
> them actually had a function. How do you suggest we should proceed? I
> do not really feel like patching all the component to replace
> /usr/bin/python3 with something else, but I am not sure we can change
> the python mediator, or can we?
>
We can change it. In fact, I already did on the build server. We
discussed the situation in our last maintainer meeting. Gimp uses its
python2 version explicitly so there is no problem to be expected.

alarcher:
> We can change it. In fact, I already did on the build server. We discussed the situation in our last maintainer meeting. Gimp uses its python2 version explicitly so there is no problem to be expected.

Great news :) this will help a lot. It would be nice to have the default mediator settings reflect what is on the build server.
Thanks :)

alarcher:
A lot of components are broken now, it will take some time to fix them before generating another gcc-1x iso.

AndWac:
Am 14.07.22 um 11:29 schrieb Aurélien Larcher:
>
>     We can change it. In fact, I already did on the build server. We
>     discussed the situation in our last maintainer meeting. Gimp uses
>     its python2 version explicitly so there is no problem to be expected.
>
> Great news :) this will help a lot. It would be nice to have the
> default mediator settings reflect what is on the build server.
>
It would be problematic to force it, because there is no way back to
2.7. Even if you change the mediator back to 2.7 it wouldn't generate
all symlinks.
I tried to explain this in my Flag-Day posting on openindiana.

PR #8546, opened 2022-07-01
libass: update to 0.16.0 and enable assembly optimisations

TheOneric:
0.16.0 did not change or remove any interfaces *(except on MS Windows)*.
Binaries linking against assembly-enabled libass builds were tested locally.

### DIFF
--- a/components/library/libass/Makefile
+++ b/components/library/libass/Makefile
@@ -18,13 +18,13 @@ BUILD_BITS=			32_and_64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libass
-COMPONENT_VERSION=	0.15.2
+COMPONENT_VERSION=	0.16.0
 COMPONENT_FMRI=		library/video/libass
 COMPONENT_SUMMARY=	Portable renderer for the ASS/SSA (Substation Alpha) subtitle format
 COMPONENT_CLASSIFICATION=System/Multimedia Libraries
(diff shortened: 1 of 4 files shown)


components/library/libass/Makefile
```diff
@@ -18,23 +18,28 @@ BUILD_BITS=			32_and_64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libass
-COMPONENT_VERSION=	0.15.2
+COMPONENT_VERSION=	0.16.0
+COMPONENT_REVISION=	1
 COMPONENT_FMRI=		library/video/libass
```
AndWac, line 22: COMPONENT_REVISION is superfluous here. It only need to be set/incremented if COMPONENT_VERSION doesn't change. If COMPONENT_VERSION is being incremented COMPONENT_REVISION shall be reset.

TheOneric: The `COMPONENT_VERSION` change and `COMPONENT_REVISION` addition are in two different commits. Should I drop it anyway?

mtelka: Since both commits are in one PR, they will be squashed during merge.  If you want to see two commits in git you need to create two PRs.

TheOneric: Pre-squashed the commits and removed `COMPONENT_VERSION`.

PR #8552, opened 2022-07-02
added library/cereal

erwinlem:
This is a dependency for Domoticz, will do a pull request for that once merged.

### DIFF
--- a/components/library/cereal/Makefile
+++ b/components/library/cereal/Makefile
@@ -1,0 +1,36 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 erwinlem
+#
+
+BUILD_STYLE=cmake
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		cereal
+COMPONENT_VERSION=	1.3.2
+COMPONENT_SUMMARY=	A C++11 library for serialization
+COMPONENT_PROJECT_URL=	https://uscilab.github.io/cereal/
+COMPONENT_FMRI=		library/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=Development/C++
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=	https://github.com/USCiLab/cereal/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:16a7ad9b31ba5880dac55d62b5d6f243c3ebc8d46a3514149e56b5e7ea81f85f
+COMPONENT_LICENSE=	BSD3CLAUSE
+COMPONENT_LICENSE_FILE=	LICENSE
+
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
+include $(WS_MAKE_RULES)/common.mk
+
+CMAKE_OPTIONS+= -DBUILD_TESTS=OFF
+
+# Auto-generated dependencies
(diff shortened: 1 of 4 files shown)


components/library/cereal/Makefile
```diff
@@ -0,0 +1,34 @@
...
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=	https://github.com/USCiLab/cereal/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:16a7ad9b31ba5880dac55d62b5d6f243c3ebc8d46a3514149e56b5e7ea81f85f
+COMPONENT_LICENSE=	BSD3CLAUSE
+COMPONENT_LICENSE_FILE=	LICENSE
+
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
+include $(WS_MAKE_RULES)/common.mk
+
+CMAKE_OPTIONS+= -DBUILD_TESTS=OFF
```
AndWac, line 34: Did you run "gmake REQUIRED_PACKAGES"? If so, there should be at least a generated comment here.

components/library/cereal/Makefile
```diff
@@ -0,0 +1,34 @@
...
+BUILD_STYLE=cmake
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		cereal
+COMPONENT_VERSION=	1.3.2
+COMPONENT_SUMMARY=	A C++11 library for serialization
+COMPONENT_PROJECT_URL=	https://uscilab.github.io/cereal/
+COMPONENT_FMRI=		library/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
```
AndWac, line 24: Please add a COMPONENT_CLASSIFICATION

components/library/cereal/Makefile
erwinlem: I did, it was empty. Added the separator so it's clear I did.

AndWac:
This runs into a build error quite early:
CMake Error at /usr/lib/cmake/Boost-1.78.0/BoostConfig.cmake:141 (find_package):
  Found package configuration file:

    /usr/lib/cmake/boost_serialization-1.78.0/boost_serialization-config.cmake

  but it set boost_serialization_FOUND to FALSE so package
  "boost_serialization" is considered to be NOT FOUND.  Reason given by
  package:

  No suitable build variant has been found.

  The following variants have been tried and rejected:

  * libboost_serialization.so.1.78.0 (32 bit, need 64)

  * libboost_serialization.a (32 bit, need 64)

Call Stack (most recent call first):
  /usr/lib/cmake/Boost-1.78.0/BoostConfig.cmake:262 (boost_find_component)
  /usr/share/cmake-3.23/Modules/FindBoost.cmake:594 (find_package)
  CMakeLists.txt:111 (find_package)

erwinlem:
I removed 32 bit support on my version of Boost, because i ran into thesame problem with porting Ceph. For some reason cmake finds the 32bit libraries instead of 64bit.  BoostConfig.cmake has a 32 and 64 bit version, cmake picks the wrong one for some reason. What is the proper way to solve this?

For now i disabled the performance tests which "solves" the problem.

AndWac:
We have a similar problem with other packages, too. We need to find a way to teach cmake to use the 64-bit library.

erwinlem:
CMAKE_LIBRARY_ARCHITECTURE=amd64 is the proper way to fix this. That changes the search path for cmake. I think this should be set by default for all cmake amd64 builds.

I also noticed in cmake.mk that LDFLAGS & PKG_CONFIG_PATH are added to options and not to the environment. Was that a mistake? It does generate warnings and makes no sense.

AndWac:
You are most probably right. Can you propose fixes or enhancements and create a PR with them, please?

components/library/cereal/Makefile
```diff
@@ -0,0 +1,37 @@
...
+COMPONENT_ARCHIVE_URL=	https://github.com/USCiLab/cereal/archive/refs/tags/v$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH=	sha256:16a7ad9b31ba5880dac55d62b5d6f243c3ebc8d46a3514149e56b5e7ea81f85f
+COMPONENT_LICENSE=	BSD3CLAUSE
+COMPONENT_LICENSE_FILE=	LICENSE
+
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
+include $(WS_MAKE_RULES)/common.mk
+
+CMAKE_OPTIONS+= -DCMAKE_LIBRARY_ARCHITECTURE=amd64
+CMAKE_OPTIONS+= -DBUILD_TESTS=OFF
```
AndWac, line 34: Do you want to remove now that after your cmake.mk PR has been merged?

PR #8586, opened 2022-07-08
shared-macros.mk: set default USE_COMMON_TEST_MASTER to true

mtelka:
Most components are (or should be) built 64-bit only so they (should) use single test results anyway.  Those that are not, and needs separate test results based on bitness (or in case of perl modules: based on perl version), should state this explicitly in their Makefile.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -455,7 +455,7 @@ COMPONENT_TEST_BUILD_DIR =	$(BUILD_DIR)/test/$(MACH$(BITS))
 COMPONENT_TEST_RESULTS_DIR =	$(COMPONENT_DIR)/test
 
 # set the default master test results file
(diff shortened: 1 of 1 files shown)


alarcher:
Almost all the libaries are built with BUILD_32_and_64 and those running a testsuite are likely to use (results-64.master, results-32.master). While I agree that it would be easier, this change breaks silently tests for a number of components. You could make an announcement on the developer list so that people do not lose time trying to understand what went wrong with the testsuite as I just did...

alarcher:
The idea of common master test result is that all architectures with card(architectures) > 1, yield the same results, not that there is only architecture tested. In this later case the specific name was used.

alarcher:
I am not saying that this is wrong, just that you have changed the semantics and therefore a number of components are now broken, so it is more a matter of deciding what should be the correct definition and ensuring consistency.

PR #8609, opened 2022-07-10
ImageMagick - fixed build due to new default perl version 5.34

fritzkink:


mtelka:
Please hold on with this.

mtelka:
Some explanation:  Currently it is possible that some perl related components are not buildable.  If not, many perl related components needs at least rebuild to get updated to perl 5.34.  I know about that and I'm working to make the transaction to new default perl (and to obsolete 5.22 and 5.24, and also start to build perl modules for 5.36) as easy as possible (with future similar upgrades in mind).

This situation is transient and it will be resolved (hopefully) soon.

Thanks for understanding.

components/image/imagemagick/Makefile
```diff
@@ -43,7 +43,11 @@ COMPONENT_LICENSE= Apache 2.0 license
 COMPONENT_LICENSE_FILE= LICENSE
 
 CONFIGURE_FIX_LIBTOOL_RPATH= no
-PERL_VERSION= 5.22
+
+PERL.64=                /usr/perl5/5.34/bin/perl
+PERL.32=                /usr/perl5/5.22/bin/perl
+PERL=                   $(PERL.$(BITS))
```
mtelka, line 49: Please remove these PERL definitions.  It should work without them.

mtelka:
I also miss pkg5 and manifests update.

mtelka:
Please fix the PR title too (5.32 -> 5.34).

components/image/imagemagick/Makefile
mtelka: By 'should' I mean "you should make this working without hardcoded perl paths and/or versions", not "if you remove them it will still work" :-).  Thanks.

fritzkink: no I tried differently but this was the only solution working for me.

fritzkink:
there is no change in manifest and pkg5.

AndWac:
How shall we proceed with this? Without these fixes the package doesn't build.
mtelka: Will you propose a better solution or shall I merge?

mtelka:
> Will you propose a better solution or shall I merge?

Yes, I'll look at it and if possible propose better solution.

mtelka:
see #8636

AndWac:
Superceded by  #8636

PR #8619, opened 2022-07-11
perl/XML-NamespaceSupport: move from xml-namespacesupport; rebuild fo…

mtelka:
…r perl 5.36 and to obsolete packages for perl 5.22 and 5.24


AndWac:
How far do you plan to obsolete perl-5.22? It's our only 32-bit perl and removing it completely may have some impacts on other packages.

mtelka:
> How far do you plan to obsolete perl-5.22? It's our only 32-bit perl and removing it completely may have some impacts on other packages.

If there is any issue with 64-bit only perl, we (I) will solve it.  The solution will depend on the actual issue we will encounter.  And yes, I plan to completely obsolete perl 5.22.

The worst case scenario is to build some never perl for 32-bits, but I believe it won't be needed.

PR #8691, opened 2022-07-13
git: update to 2.37.1

AndWac:
gmake test run fine


mtelka:
Yes!  This also switches from perl 5.22 to perl 5.34.  For well written Makefiles such switch should be done completely automatically during (re)build.

PR #8724, opened 2022-07-15
openjdk-18: Add package

ghost:
Add package for OpenJDK version 18, based on the OpenJDK 11 package and patches from Tribblix.
This builds and runs fine locally on my computer, but may fail on the build server for the same reason as my OpenJDK 11 fixes.
I recommend solving that problem before merging.


AndWac:
OpenJDK-18 is not an LTS version, while OpenJDK-17 is. Wouldn't it be better to add OpenJDK-17 instead?

ghost:
I always try to keep packages as up to date as possible.
OpenJDK 18 is currently the newest version we can built because Tribblix offers OpenJDK 17 and OpenJDK needs version-1 to build.
My plan is to build packages for OpenJDK 19 and then 20 based on our own OpenJDK 18 and then 19 package instead of Tribblix binaries.

PR #8728, opened 2022-07-15
Prefer 32-bit build for legacy components

mtelka:
This is an attempt to solve issue we discussed in #8540.  Untested.  Please test before merge.


Toasterson:
Can you solve the merge confilct?

mtelka:
> Can you solve the merge confilct?

Done, I think.  Thanks.

Toasterson:
Tested with components under openindian directory. So far all failures also happen with or without this change. @mtelka do you know any component I could test with? Otherwise I would say merge and fix the failures.

mtelka:
> Tested with components under openindian directory. So far all failures also happen with or without this change.

What fails there? For example `openindiana/illumos-gate` does not fail without this change.

> do you know any component I could test with?

Sorry, no.

> Otherwise I would say merge and fix the failures.

I prefer just fix failures (without merging this) :-).

Toasterson:
Ok, I am a bit lost what to do here. Shall we merge and see what happens or close and fix legacy components to not be legacy anymore?

mtelka:
> Ok, I am a bit lost what to do here. Shall we merge and see what happens or close and fix legacy components to not be legacy anymore?

I prefer not to merge this, but fix all problems instead.  To do so, we need a list of components with problems.  I'm not aware of any component failing because of this.  Everything I encountered I just fixed.

AndWac:
Alas it's tedious to find failing packages as some package builds will only fail during mass builds (eg. by running gmake publish in the top level folder) and build ok when build separately.
I hope that #9560 will fix at least some of those problems.
I am also in favor to close this PR and find and fix failing builds systematically.

Toasterson:
Then we have a decision :) Closing.

PR #8734, opened 2022-07-15
Fix developer components

alarcher:
@AndWac @mtelka Some fixes with updates of component Makefiles.


components/developer/swig/Makefile
```diff
@@ -79,23 +79,17 @@ COMPONENT_TEST_TARGETS = \
...
-build:		$(BUILD_32_and_64)
-
-install:	$(INSTALL_32_and_64)
-
-test:		$(TEST_32_and_64)
-
-REQUIRED_PACKAGES += developer/java/openjdk7
+# Build dependencies
+PERL_REQUIRED_PACKAGES += runtime/perl
+REQUIRED_PACKAGES += developer/java/openjdk11
```
AndWac: We only have runtime/java/openjdk11

components/developer/swig/Makefile
```diff
@@ -79,23 +79,17 @@ COMPONENT_TEST_TARGETS = \
...
-build:		$(BUILD_32_and_64)
-
-install:	$(INSTALL_32_and_64)
-
-test:		$(TEST_32_and_64)
-
+# Build dependencies
+PERL_REQUIRED_PACKAGES += runtime/perl
 REQUIRED_PACKAGES += developer/java/openjdk7
 REQUIRED_PACKAGES += developer/parser/bison
```
AndWac, line 85: Doesn't openjdk11 work?  runtime/java(openjdk11 should provide everything needed. We might be able to obsolete openjdk7.

alarcher: I'll check now

components/developer/swig/Makefile
alarcher: @AndWac Basically I think most components can be fixed using 'gmake update' and some of them can be made 64-bit only.

alarcher:
@AndWac Build of gcc49 failed with a sed error on the bot but succeeds in my zone... weird.

AndWac:
The bot is not fully accurate and up-to-date. This makes it not so reliable (in both directions).

components/developer/swig/Makefile
```diff
@@ -79,23 +79,17 @@ COMPONENT_TEST_TARGETS = \
...
-
-install:	$(INSTALL_32_and_64)
-
-test:		$(TEST_32_and_64)
-
-REQUIRED_PACKAGES += developer/java/openjdk7
+# Build dependencies
+PERL_REQUIRED_PACKAGES += runtime/perl
+REQUIRED_PACKAGES += runtime/java/openjdk11
 REQUIRED_PACKAGES += developer/parser/bison
```
AndWac, line 85: developer/parser/bison has been removed and replaced by
developer/parser/bison-37
or
developer/parser/bison-38

alarcher: Hmm, this is unfortunate, however developer/parser/bison will install developer/parser/bison-37. I do not feel confortable using a versionedfmri for such an essential package. Otherwise anytime we bump we'll have to change the dependency in each component. Probably developer/parser/bison should always point to the bison-of-the-day.

components/developer/pkg-config/Makefile
```diff
@@ -24,27 +24,27 @@
 # Copyright (c) 2017, Aurelien Larcher
 #
 
+BUILD_STYLE=configure
+BUILD_BITS=64
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 29: This breaks firefox:
checking for ar... /usr/gnu/bin/ar
checking for pkg_config... not found
DEBUG: pkg_config: Looking for /usr/bin/amd64/pkg-config
ERROR: Cannot find pkg_config
gmake: *** [/export/home/andreas/oi-userland/make-rules/configure.mk:175: /export/home/andreas/oi-userland/components/web/firefox/build/amd64/.configured] Error 1

This seems to be coming from the latest pkg-config changes which aren't present yet.

alarcher: Well the firefox Makefile is incorrect:
`echo "ac_add_options PKG_CONFIG=/usr/bin/amd64/pkg-config" >> $(MOZCONFIG) ;`

The PATH is set so that the right binaries are used in 64-bit mode, hardcoding the path is wrong.

alarcher: This does not make sense as both pkg-config binaries are identical.

components/developer/swig/Makefile
AndWac: Yes, I know. The problem was that I updated bison 3.8 and later found out that it broke some packages. Thus, I changed it fmri to bison-38 and introduced the old version as bison-37. We had similar situations for automake and some other essential packages, too.

components/developer/pkg-config/Makefile
AndWac: I see, we need to fix it later. I have already merged the change because it works without your developer-fixes.

components/developer/swig/Makefile
alarcher: OK perfect :) the component as it is does not cause problem and uses bison-37 so let us leave it as it is...

components/developer/pkg-config/Makefile
alarcher: No problem, this is not a big deal :)

PR #8769, opened 2022-07-21
Fix components

alarcher:
Continuing here to distinguish with gcc-10 fixes since I have now a dedicated build zone.
@AndWac I need to review the few manifest issues you reported earlier.


alarcher:
@AndWac This batch should be ready.

components/x11/urxvt/pkg5
```diff
@@ -19,4 +19,4 @@
         "terminal/urxvt"
     ],
     "name": "rxvt-unicode"
```
AndWac, line 21: pkg5 misses the perl update

alarcher: The file was updated but for some reason it is this perl that is detected: I have run it twice again to check that I did not forget. I suspect something is not updated yet and that makes the dependency produced to the old one.

AndWac: On my system it is properly detected.

AndWac (approved):

components/x11/urxvt/pkg5
alarcher: Maybe I should update the system to a more recent build?

AndWac:
Am 21.07.22 um 21:25 schrieb Aurélien Larcher:
>
>
>
> In components/x11/urxvt/pkg5
> <https://github.com/OpenIndiana/oi-userland/pull/8769#discussion_r927031196>:
>
> > @@ -19,4 +19,4 @@
>           "terminal/urxvt"
>       ],
>       "name": "rxvt-unicode"
>
> Maybe I should update the system to a more recent build?
>
>
Definitely.

We seem to have a new problem on the build server. The last three
jenkins runs were successful but the new packages didn't make it into
the repository.
The run times were also quite shorter than the ones before. :(

> DefiniMessage ID:
> ***@***.***>
>

PR #8770, opened 2022-07-21
Added domoticz

erwinlem:


erwinlem:
Seems like the buildserver runs out of memory, my lowly atom with 64GB is able to compile this.
Can anybody else test this?

alarcher:
> Seems like the buildserver runs out of memory, my lowly atom with 64GB is able to compile this. Can anybody else test this?

OK I will do it tonight.

AndWac (approved):

PR #8772, opened 2022-07-21
components.ignore: fix typo

AndWac:


alarcher:
@AndWac Missing escaped slash

components/components.ignore
```diff
@@ -28,4 +28,4 @@
 # Our build server fails to build the following packages (the builds succeed on other build systems):
 /^runtime\/openjdk-11$/d
 /^runtime\/sbcl$/d
-/^web/icedtea-web/d
+/^web/icedtea-web$/d
```
alarcher: `/^web\/icedtea-web$/d`

AndWac: That's something where I also easily fail.

PR #8780, opened 2022-07-23
minimal_install: remove system/library/processor, obseleted by illumos#14802

richlowe:
This probably fixes building the OI installers, but also anything else using minimal_install.

Please do check this out though, I'm not expert in these things any longer.


AndWac:
It is not sufficient to only change minimal. You also need to change the Makefile of the package(s) that include it, In this case you'll need to increment COMPONENT_REVISION of install-types/Makefile.
Without changing a Makefile our jenkings wouldn't be triggered. And changes to a package should at least increment its COMPONENT_REVISION.

richlowe:
Sigh, this time I didn't completely forget! I just also didn't stage the makefile.  Sorry

PR #8784, opened 2022-07-25
Fix components

alarcher:


components/desktop/gnome3/geocode-glib/Makefile
```diff
@@ -13,35 +13,36 @@
...
-include $(WS_MAKE_RULES)/ips.mk
+include $(WS_MAKE_RULES)/common.mk
 
 PATH=$(PATH.gnu)
 
 CONFIGURE_OPTIONS += --sysconfdir=/etc
 CONFIGURE_OPTIONS += --disable-static
 # Required for other components.
+CONFIGURE_OPTIONS.32 += --disable-introspection
 CONFIGURE_OPTIONS.64 += --enable-introspection
```
AndWac: Is this still being necessary after you made it 64-bit only?

AndWac (approved):

components/desktop/gnome3/geocode-glib/Makefile
alarcher: No, you are right :)

PR #8790, opened 2022-07-25
libgtop: fix build and drop 32-bit

AndWac:
32-bit can be dropped because its dependencies are all 64-bit now.


alarcher:
Great :)

PR #8793, opened 2022-07-26
arpsend: update to 1.2.4

dbh201:


components/cluster/arpsend/patches/arpsend-1.2.4-ccaputo-my_eaddr.patch
```diff
@@ -0,0 +1,11 @@
+--- src/arpsend.c.original      2021-04-02 18:21:52.000000000 +0000
```
alarcher: Coud you just preprend a bogus directory name to avoid adding PATCH_LEVEL=0. Better have the patches in the right format than fiddle with macros in the Makefile.

PR #8795, opened 2022-07-27
clamav - update to version 0.105.1

fritzkink:


components/sysutils/clamav/Makefile
```diff
@@ -23,7 +23,7 @@ COMPONENT_NAME=		clamav
 ### But our IPS version should better have same number of components.
```
AndWac, line 23: You should add
USE_COMMON_TEST_MASTER= no
to the Makeflle in order make the tests pass.

PR #8954, opened 2022-08-05
Regular packages should depend on runtime/perl instead of runtime/per…

mtelka:
…l-$(PLV)


mtelka:
This will save rebuild of every regular package that depend on perl when we add new perl version or obsolete an old one.

PR #8957, opened 2022-08-05
texinfo: rebuild to update perl dependencies

mtelka:


mtelka:
Please merge after #8954.

PR #9002, opened 2022-08-09
xscreensaver: update to 6.04

dbh201:
Tests done with Virtualbox with VBoxVGA and 3D acceleration enabled :
 - lockscreen works
 - lockscreen when xscreensaver run as root works (with a warning message)
 - hacks are disabled when xscreensaver is run as root
 - some hacks tested and working
 - some gl hacks tested and working
 - xscreensaver starts automatically at login

Reverted to the older lockscreen (using a dithered OI-logo), until the 06-gtk-lock.patch for 5.44 is modified to work with 6.x.


dbh201:
The 11-pam_audit.patch is also being finicky, and needs to be modified to work with 6.x for Solaris' custom auditing functionality to work. However, PAM reports directly to syslog as well and I have confirmed authpriv.notice messages are generated by xscreensaver-auth.

PR #9003, opened 2022-08-09
postgresql-14: rebuild for perl 5.34

mtelka:


components/database/postgresql-14/Makefile
```diff
@@ -122,17 +122,18 @@ COMPONENT_TEST_TRANSFORMS= \
 REQUIRED_PACKAGES += library/libxslt
```
AndWac: library/libxslt is now automatically detected and can removed here.

mtelka: > library/libxslt is now automatically detected and can removed here.

Yep. Thanks for spotting it.  Fixed.

PR #9019, opened 2022-08-10
Added plib 1.8.5

jakegoerzen:
This is an initial commit of PLIB version 1.8.5 a set of portable game libraries, which is one of the requirements to building flightgear and some other simulation software.

### DIFF
--- a/components/library/plib/Makefile
+++ b/components/library/plib/Makefile
@@ -1,0 +1,43 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL)". You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2021 Jake Goerzen
+
+BUILD_BITS=		64
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		plib
+COMPONENT_VERSION=	1.8.5
+COMPONENT_PROJECT_URL=	http://plib.sourceforge.net
+COMPONENT_SUMMARY=	PLIB - A Suite of Portable Game Libraries.
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:485b22bf6fdc0da067e34ead5e26f002b76326f6371e2ae006415dea6a380a32
+COMPONENT_ARCHIVE_URL=	http://plib.sourceforge.net/dist/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI =	library/plib
+COMPONENT_LICENSE=	LGPL
+COMPONENT_LICENSE_FILE=	COPYING
+COMPONENT_CLASSIFICATION =	System/Libraries
+
+include $(WS_MAKE_RULES)/common.mk
+
+COMPONENT_PRE_CONFIGURE_ACTION = cd $(COMPONENT_SRC) && autoreconf -if
+
+CONFIGURE_OPTIONS += --disable-static
+
+# Auto-generated dependencies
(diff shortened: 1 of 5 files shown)


components/library/plib/Makefile
```diff
@@ -0,0 +1,34 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2021 Jake Goerzen
+
+BUILD_BITS=			32_and_64
+
```
AndWac, line 15: Do you really need 32 Bit? We want to get rid of 32 bit applications and I am reluctant to merge new packages for 32 bit.

components/library/plib/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,90 @@
...
+file path=usr/include/plib/ssgaScreenDump.h
+file path=usr/include/plib/ssgaShapes.h
+file path=usr/include/plib/ssgaSky.h
+file path=usr/include/plib/ssgaSphere.h
+file path=usr/include/plib/ssgaWaveSystem.h
+file path=usr/include/plib/ssgconf.h
+file path=usr/include/plib/ul.h
+file path=usr/include/plib/ulRTTI.h
+file path=usr/lib/$(MACH64)/libplibfnt.a
+file path=usr/lib/$(MACH64)/libplibjs.a
```
AndWac: We usually don't deliver static libs. Isn't it possible to create shared libs for this package?

mtelka: Isn't this dropped automatically by transforms/archive-libraries-drop?

components/library/plib/Makefile
jakegoerzen: As far as I know 32 Bit is not need.  I'll switch to build with BUILD_BITS= 64

mtelka: Then you could simply remove the `BUILD_BITS=` line because 64 bit is default now.

components/library/plib/manifests/sample-manifest.p5m
jakegoerzen: PLIB doesn't have an --enable-shared option to configure the build with.  I'd have to modify it's build to get shared libraries to be produced.  What to you suggest?

mtelka: It looks like plib intentionally do not support shared libs.  Strange.

components/library/plib/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,90 @@
...
+file path=usr/lib/libplibpsl.a
+file path=usr/lib/libplibpu.a
+file path=usr/lib/libplibpuaux.a
+file path=usr/lib/libplibpw.a
+file path=usr/lib/libplibsg.a
+file path=usr/lib/libplibsl.a
+file path=usr/lib/libplibsm.a
+file path=usr/lib/libplibssg.a
+file path=usr/lib/libplibssgaux.a
+file path=usr/lib/libplibul.a
```
AndWac: Please re-run gmake sample-manifest.
If I do it, I'll get some differences:
 file path=usr/lib/$(MACH64)/libplibul.a
-file path=usr/lib/libplibfnt.a
-file path=usr/lib/libplibjs.a
-file path=usr/lib/libplibnet.a
-file path=usr/lib/libplibpsl.a
-file path=usr/lib/libplibpu.a
-file path=usr/lib/libplibpuaux.a
-file path=usr/lib/libplibpw.a
-file path=usr/lib/libplibsg.a
-file path=usr/lib/libplibsl.a
-file path=usr/lib/libplibsm.a
-file path=usr/lib/libplibssg.a
-file path=usr/lib/libplibssgaux.a
-file path=usr/lib/libplibul.a

Make sure to propagate the changes to the real manifest and check whether this still is what you'll need.

components/library/plib/manifests/sample-manifest.p5m
jakegoerzen: I found a bug report on plib with a patch to enable shared libraries:

https://sourceforge.net/p/plib/bugs/42/

AndWac:
pkg5 needs to be updated. After this update please squash the commits and only keep a single, relevant comment.

PR #9020, opened 2022-08-11
Added openscenegraph 3.6.5

jakegoerzen:
Initial commit of OpenSceneGraph 3.6.5

This is a pre-requisite library to building flightgear, simgear and other simulation software.


jakegoerzen:
I can't generate a pkg5 file for this one because when I run gmake pre-publish I get and error:

Lint engine setup...
Starting lint run...
ERROR userland.action001.2        64-bit object 'usr/lib/libOpenThreads.so.3.3.1' in 32-bit path
ERROR userland.action001.2        64-bit object 'usr/lib/libosg.so.3.6.5' in 32-bit path
ERROR userland.action001.2        64-bit object 'usr/lib/libosgAnimation.so.3.6.5' in 32-bit path
ERROR userland.action001.2        64-bit object 'usr/lib/libosgDB.so.3.6.5' in 32-bit path
ERROR userland.action001.2        64-bit object 'usr/lib/libosgFX.so.3.6.5' in 32-bit path
ERROR userland.action001.2        64-bit object 'usr/lib/libosgGA.so.3.6.5' in 32-bit path
...

Do I need to pass something to CMAKE_OPTIONS to install into /usr/lib/$(MACH64) ?

AndWac:
Yes, you'll need to teach it that our 64 bit libs reside in /usr/lib/$(MACH64). That would also make your transforms (despite the one for libGL.so.1) obsolete. You can search for solutions in existing packages that use BUILD_STYLE= cmake.
Some other packages use CMAKE_OPTIONS.64 += -DLIB_SUFFIX=/$(MACH64)  but that doesn't seem to work here.

AndWac:
Another alternative would be to patch the CMakeLists.txt file such that it deploys the plugin files into the correct folder.

Furthermore, on my build system three additional (optional) dependencies have been detected and used: FFmpeg, LibVNCServer, and SDL. They might be valuable additions, but dependent on FFmpeg would mean that the whole package has to be moved into the encumbered repository (and thus all packages that depend on it later). Nevertheless a thing to think about.
We also have openexr which is not properly detected yet.

AndWac:
You still need to deal with ffmpeg, sdl., and vnc.
On my build system which is almost identical to the official one, running gmake REQUIRED_PACKAGES gives
 REQUIRED_PACKAGES += library/giflib
 REQUIRED_PACKAGES += library/libpoppler
+REQUIRED_PACKAGES += library/libvncserver
+REQUIRED_PACKAGES += library/sdl
 REQUIRED_PACKAGES += system/library/fontconfig
 REQUIRED_PACKAGES += system/library/freetype-2
+REQUIRED_PACKAGES += video/ffmpeg
 REQUIRED_PACKAGES += web/curl
 REQUIRED_PACKAGES += x11/library/libx11
 REQUIRED_PACKAGES += x11/library/libxinerama
So, the modules for ffmpeg, sdl, and vnc will be build at the moment but won't be packaged.
So you still need to decide whether you want to support them or not. In case you don't want them you should explicitly disable them.

jakegoerzen:
I'll install those packages on my build system and rebuild.  I'm pretty sure ffmpeg is used for saving/converting a video screenshot stream so if we don't want that option I could leave it out so we don't need to use encumbered packages.  I still need to investigate openexr inclusion as well.

AndWac:
If this package depends on anything that is in encumbered (like ffmpeg) this also needs to be moved to the encumbered folder and thus will also end in the encumbered repository.
You should try to run "gmake publish". It won't succeed because it violates the requirements.
You'll just need to move everything into its new place, eg. something like 
cd components/library
git mv openscenegraph ../encumbered/

should do it.

AndWac (approved):

PR #9029, opened 2022-08-12
Add Mojo::IOLoop::ForkCall perl module

mtelka:


AndWac:
The summary contains "(DEPRECATED)". Can you elaborate why do you want this package?

mtelka:
It is needed for znapzend.

PR #9055, opened 2022-08-15
Added python/bleach

erwinlem:
Started porting matrix-synapse, this is one of the dependencies. 
I did `make REQUIRED_PACKAGES` but it did not pickup python as a dependency, is this correct?

### DIFF
--- a/components/python/bleach/Makefile
+++ b/components/python/bleach/Makefile
@@ -1,0 +1,47 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# This file was automatically generated using the following command:
+#   $WS_TOOLS/python-integrate-project bleach
+#
+
+BUILD_STYLE = setup.py
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		bleach
+COMPONENT_VERSION =		5.0.1
+COMPONENT_SUMMARY =		bleach - An easy safelist-based HTML-sanitizing tool.
+COMPONENT_PROJECT_URL =		https://github.com/mozilla/bleach
+COMPONENT_ARCHIVE_URL =		\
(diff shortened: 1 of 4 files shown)


AndWac:
Yes, our dependency checker is weak to detect scripting language dependencies.

components/python/bleach/Makefile
```diff
@@ -0,0 +1,45 @@
...
+PYTHON_VERSIONS=	3.7 3.9
+
+include $(WS_MAKE_RULES)/common.mk
+
+## pytest fails, most likely outdated.
+#COMPONENT_TEST_CMD =	$(PYTHON) -m pytest
+#COMPONENT_TEST_DIR=	$(SOURCE_DIR)/tests
+#COMPONENT_TEST_ARGS =
+#COMPONENT_TEST_MASTER = $(COMPONENT_TEST_RESULTS_DIR)/results-all.master
+
```
AndWac: When dealing with python packages you'll need to analyse its meta data. Typically you'll find them in setup.py and/or setup.cfg.
For this package you'll find
INSTALL_REQUIRES = [
    # html5lib requirements
    "six>=1.9.0",
    "webencodings",
]
in setup.py. Thus, you'll need to create packages for webencodings first as we don't have them yet. Furthermore, you'll need to add runtime dependencies in the manifest file for them and the appropriate python runtimes.
You can find some examples for this in other python packages, eg. in pytest.
This package also supports the tox test framework. You can find examples on how to use it also in other python packages, eg. wheel.

mtelka:
Please use `python-integrate-project` to integrate this project.  Thanks.

erwinlem:
``python-integrate-projec`` works really well, but not all packages for testing are there so i will skip that for now.

mtelka:
I do not recommend to merge this yet since `python-integrate-project` failed with ERRORs.  The merge should be postponed until all packages needed for testing are merged first and tests are run.  The list of missing packages is: black, hashin, mypy, pip-tools, sphinx, twine.  I already had three of them in my queue and I just added rest to the queue too.

mtelka:
Hmmm, it is not possible to get bleach tested without bleach itself, so I've no objection to merge this as is (although I'd prefer a bit better integration, but this could be improved later).

erwinlem:
I tried adding the packages for testing as wel but this spirals out of control into a big yak shaving fest. I have Synapse running but ``python-integrate-project`` fails at sample manifest (works fine for matrix_common). I will look later into what is happening under the hood.

Maybe we should pickup the tests with the next bleach update when we have all the dependencies available.

mtelka (approved):

PR #9057, opened 2022-08-15
perl - added package Time::HiRes

fritzkink:


AndWac:
This package seems to depend on text/jq

fritzkink:
added missing dependencies

PR #9061, opened 2022-08-15
perl - added package NetAddr::IP

fritzkink:


mtelka:
The `NetAddr::IP` module required `perl-integrate-module.conf` and minor update of `perl-integrate-module` script itself. See #9069 and  #9070.

Please close this PR.  Thanks.

PR #9063, opened 2022-08-15
perl - added package Encode-Detect

fritzkink:


mtelka:
The `Encode::Detect` perl module requires custom `perl-integrate-module.conf` file to integrate properly using the `perl-integrate-module` script.  I created new PR #9068 for it.  Please close this one.  Thanks.

PR #9062, opened 2022-08-15
perl - added package Razor2

fritzkink:

PR #9064, opened 2022-08-15
perl - added package Clone

fritzkink:


mtelka:
@fritzkink  This module needs custom simple `perl-integrate-module.conf` to set proper `COMPONENT_TEST_TRANSFORMS` to integrate successfully using the `perl-integrate-module` script.  The `perl-integrate-module.conf` file should look similarly as `components/perl/PkgConfig/perl-integrate-module.conf`. I also suggest to split the transforms to separate lines, like:
```
COMPONENT_TEST_TRANSFORMS += "-e '/^SV/d'"
COMPONENT_TEST_TRANSFORMS += "-e '/PV/d'"
...
```
for better readability (please note usage of `'` and `"`; this seems to be the best way how to use them for sed rules in Makefiles).

To add `perl-integrate-module.conf` to the perl module integration is a bit tricky.  I use steps I suggested in #9062 with modified step 3:

3a. run `$WS_TOOLS/perl-integrate-module MODULE` (where `MODULE` is `Clone` in this case)
3b. if the run failed (`ERROR` or `FATAL` reported by the `perl-integrate-module` script, or you are not satisfied with the result), then you need custom `perl-integrate-module.conf`; if the integration is okay (you are satisfied with the `git show` output), go to 3h.
3c. make sure there is nothing staged for git (use `git status` to confirm that).  In a case the `perl-integrate-module` script failed with `FATAL` you'll probably find that there is something staged, so you need to cleanup using `git reset --hard`.
3d. create (or modify, if this is your 2nd or 3rd or ... iteration) `perl-integrate-module.conf` in the newly created directory for the perl module (in this case you need to create/modify `components/perl/Clone/perl-integrate-module.conf`).
3e. add `perl-integrate-module.conf` to git (`git add components/perl/Clone/perl-integrate-module.conf` then `git commit -m "perl-integrate-module.conf for Clone"`) - make sure the changeset contains `components/perl/Clone/perl-integrate-module.conf` only and nothing else
3f. remove the changeset with `Add Clone perl module` message from your git history (`git rebase -i HEAD~2` should usually do the right thing); if this is your 2nd or 3rd or ... iteration you'll probably need `git rebase -i HEAD~3` - in this case I suggest to squash all `perl-integrate-module.conf` commits to one.
3g. go to 3a
3h. squash your `perl-integrate-module.conf` commit(s) to the `Add Clone perl module` commit: `git rebase -i HEAD~2` then move the `pick ... perl-integrate-module.conf for Clone` line below the `pick ... Add Clone perl module` line and change `pick ... perl-integrate-module.conf for Clone` to `f ... perl-integrate-module.conf for Clone`.
3i. continue to 4 (see #9062)

Thanks.

PS: Yes, the only file you need to manually touch for proper perl module integration is `perl-integrate-module.conf`.  While still most modules integrates properly even without `perl-integrate-module.conf`.  Please never manually edit Makefile, *.p5m, etc. They should all be created automatically using the `perl-integrate-module` script.

fritzkink:
I don't understand this line in your integrate script:

`cat perl-integrate-module.conf | gsed -e '0,/^%include-2%/d' -e '/^%/,$d' | gsed -e '1s/^./\n&/'`

when I run this on the command line nothing remains, rendering the intended input useless. Content of perl-integrate-module.conf:

`$ cat perl-integrate-module.conf
COMPONENT_TEST_TRANSFORMS += "-e '/^SV/d'"
COMPONENT_TEST_TRANSFORMS += "-e '/PV/d'"
COMPONENT_TEST_TRANSFORMS += "-e '/MG_PTR/d'"
COMPONENT_TEST_TRANSFORMS += "-e '/MAGIC/d'"
`

mtelka:
> I don't understand this line in your integrate script:
> 
> `cat perl-integrate-module.conf | gsed -e '0,/^%include-2%/d' -e '/^%/,$d' | gsed -e '1s/^./\n&/'`
> 
> when I run this on the command line nothing remains, rendering the intended input useless. Content of perl-integrate-module.conf:
> 
> `$ cat perl-integrate-module.conf COMPONENT_TEST_TRANSFORMS += "-e '/^SV/d'" COMPONENT_TEST_TRANSFORMS += "-e '/PV/d'" COMPONENT_TEST_TRANSFORMS += "-e '/MG_PTR/d'" COMPONENT_TEST_TRANSFORMS += "-e '/MAGIC/d'" `

You need to add `%include-3%` as first line in your `perl-integrate-module.conf`. The `%include-3%` line is a tag starting the include snippet.  The snippet ends either at end of file, or at next tag (tags are all lines starting with `%`).

There are three supported `include` tags: `%include-1%` (this will go right before `shared-macros.mk` include in the final `Makefile`), `%include-2%` (right before `common.mk` include in the final `Makefile`), and `%include-3%` (right after `common.mk` include in the final `Makefile`).

Please look at [CGI/perl-integrate-module.conf](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/perl/CGI/perl-integrate-module.conf) and [CGI/Makefile](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/perl/CGI/Makefile) to see an example of `%include-1%` and `%include-3%`.

fritzkink:
thanks for the final comment, should be ok now.

fritzkink:
now I have the next challenges:

`$WS_TOOLS/perl-integrate-module Mail::DKIM
`

here complains about version. how to deal with it:

`ERROR: Unsupported version format: 1.20220520
`

and

`$WS_TOOLS/perl-integrate-module Mail::SPF
`

here download path ist different from automatically built by script:

`https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/Mail-SPF-2.9.0.tar.gz
`

and in reality:

`https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/mail-spf/Mail-SPF-v2.9.0.tar.gz
`

components/perl/Clone/perl-integrate-module.conf
```diff
@@ -0,0 +1,5 @@
+%include-3%
```
mtelka, line 16: This file lacks the license header.  I believe all files, especially manually created, should have the license header to avoid confusion.

mtelka (changes requested):

mtelka:
> now I have the next challenges:
> 
> `$WS_TOOLS/perl-integrate-module Mail::DKIM `
> 
> here complains about version. how to deal with it:
> 
> `ERROR: Unsupported version format: 1.20220520 `

Please read https://github.com/mtelka/oi-userland/blob/pim-doc/doc/perl-modules.md for information how to solve similar problems.

> 
> and
> 
> `$WS_TOOLS/perl-integrate-module Mail::SPF `
> 
> here download path ist different from automatically built by script:
> 
> `https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/Mail-SPF-2.9.0.tar.gz `
> 
> and in reality:
> 
> `https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/mail-spf/Mail-SPF-v2.9.0.tar.gz `

I think something like this in the `perl-integrate-module.conf` file should solve the problem:

```
%include-2%
COMPONENT_ARCHIVE_URL = https://cpan.metacpan.org/authors/id/$(COMPONENT_CPAN_AUTHOR_URL)/mail-spf/$(COMPONENT_ARCHIVE)

**Comment by mtelka** [comment]
> > `$WS_TOOLS/perl-integrate-module Mail::SPF `
> > here download path ist different from automatically built by script:
> > `https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/Mail-SPF-2.9.0.tar.gz `
> > and in reality:
> > `https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/mail-spf/Mail-SPF-v2.9.0.tar.gz `
> 
> I think something like this in the `perl-integrate-module.conf` file should solve the problem:
> 
> ```
> %include-2%
> COMPONENT_ARCHIVE_URL = https://cpan.metacpan.org/authors/id/$(COMPONENT_CPAN_AUTHOR_URL)/mail-spf/$(COMPONENT_ARCHIVE)
> ```

Or just wait until #9210 is merged.

BTW, you'll need to set `COMPONENT_SRC` too.  Similarly as it is done in `Mouse/perl-integrate-module.conf` and `Net-SNMP/perl-integrate-module.conf`.

**Comment by mtelka** [comment]
@fritzkink, please update your machine to latest perl packages.  You should have installed exactly these versions:

```
# pkg list runtime/perl-53{4,6}
runtime/perl-534                                  5.34.1-2022.0.0.3          i--
runtime/perl-536                                  5.36.0-2022.0.0.2          i--
#
```

Then please re-run the `perl-integrate-module` script again.  This will update `.p5m` files.  If you won't do that the Clone package build would fail once merged.

Thank you.

File: components/perl/Clone/perl-integrate-module.conf [hunk 2]
```diff
@@ -0,0 +1,16 @@
+%hook-no-license%
```
mtelka: Why do you need `%hook-no-license%`?  Even without it the results seem acceptable.

And, this file still lacks the license header.  By license header I mean this:

```
$ head -n15 components/perl/Authen-PAM/perl-integrate-module.conf
#
# This file and its contents are supplied under the terms of the
# Common Development and Distribution License ("CDDL"), version 1.0.
# You may only use this file in accordance with the terms of version
# 1.0 of the CDDL.
#
# A full copy of the text of the CDDL should have accompanied this
# source.  A copy of the CDDL is also available via the Internet at
# http://www.illumos.org/license/CDDL.
#

#
# Copyright 2022 Marcel Telka
#

$
````

mtelka (changes requested):

mtelka:
Also, my 6 days old comment above still applies: https://github.com/OpenIndiana/oi-userland/pull/9064#issuecomment-1243516330

fritzkink:
i don't see the difference (below output on my machine):
` pkg list runtime/perl-53{4,6}`
`NAME (PUBLISHER)                                  VERSION                    IFO`
`runtime/perl-534                                  5.34.1-2022.0.0.3          i--`
`runtime/perl-536                                  5.36.0-2022.0.0.2          i--`

components/perl/Clone/perl-integrate-module.conf
fritzkink: because of this:

`head Clone.license
---------

Copyright 2001-2019 Ray Finch. All Rights Reserved.

This module is free software; you can redistribute it and/or
modify it under the same terms as Perl itself.

===========================================================================

This program is free software; you can redistribute it and/or modify
`
If it makes you happy I added the header.

mtelka:
Please do these:

1) Replace the 'Copyright' line in the `perl-integrate-module.conf` file by your name, because I do not own copyright to that file.  It is your file.  If you do not want to see your name there, just remove the 'Copyright' line.
2) Remove `%hook-no-license%` from `perl-integrate-module.conf`.
3) Re-run the `perl-integrate-module` script to get everything updated (especially p5m files, otherwise the build will fail once merged).

Thanks.

mtelka (approved):

PR #9107, opened 2022-08-22
nwam-manager: remove obsoleted package from REQUIRED_PACKAGES

mtelka:
Please note the build of `nwam-manager` failed for me without https://github.com/OpenIndiana/nwam-manager/pull/5

### DIFF
--- a/components/openindiana/nwam-manager/nwam-manager.p5m
+++ b/components/openindiana/nwam-manager/nwam-manager.p5m
@@ -14,6 +14,7 @@
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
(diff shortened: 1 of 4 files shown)


components/openindiana/nwam-manager/nwam-manager.p5m
```diff
@@ -27,7 +28,7 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 file path=etc/gconf/schemas/nwam-manager.schemas
 file path=etc/xdg/autostart/nwam-manager.desktop
-file path=usr/bin/nwam-manager-properties mode=0555
+file path=usr/bin/nwam-manager-properties
 file path=usr/lib/nwam-manager mode=0555
```
AndWac, line 32: Why doesn't nwam-manager-properties needed to be executable?

mtelka: > Why doesn't nwam-manager-properties needed to be executable?

Because nobody should execute such strange command.  No, just kidding :-).

The reason is that we do not need explicit mode for _normal_ binaries because https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/transforms/defaults#L113

PR #9113, opened 2022-08-22
mariadb-106: initial commit, version 10.6.9

dbh201:


components/database/mariadb-106/Makefile
```diff
@@ -0,0 +1,142 @@
...
+#
+
+BUILD_BITS= 64
+BITS=		64
+BUILD_STYLE=cmake
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		mariadb
+COMPONENT_MAJOR_VERSION=10.6
```
pyhalov, line 26: 10.9.2 is here. Should we ship it instead?

dbh201: I picked 10.6 to package primarily because it's the last long-term release, supported until 2026 (see https://mariadb.org/about/#maintenance-policy). 10.7+ have much shorter lifecycles. Also, I just checked around a bit and CentOS, Fedora, Debian, Ubuntu, Alpine, and FreeBSD are all on 10.5 or 10.6 in their latest distro releases as well.

mtelka: +1 for shipping 10.6.x and to skip 10.7, 10.8, 10.9.

pyhalov: OK

PR #9130, opened 2022-08-25
glib: remove perl from REQUIRED_PACKAGES; add few fixes to improve te…

mtelka:
…st results


mtelka:
The failing tests that are left for future work are: appmonitor, converter-stream, testfilemonitor, trash, gsettings.

PR #9137, opened 2022-08-26
Package history cleanup

mtelka:
illumos did similar cleanup more than [5 years ago](https://www.illumos.org/issues/8002) making smooth direct upgrade from OpenSolaris impossible.
These packages were usually obsoleted/renamed more than 10 years ago.


AndWac (approved):

Toasterson (approved):
LGTM, I am not aware that anybody has used this upgrade path in the last 5 years

PR #9154, opened 2022-08-28
lightdm: workaround for #14940

dbh201:
This commit disables user switching in lightdm for the default seat.


AndWac:
Are you sure that this is a common problem and you are not the only one affected by it?

dbh201:
I can confirm this issue is present in a fresh, updated install of OI on VirtualBox. I can't say how common it is, but IMHO it's a bit of a stability risk. I put most of what I found on the bug tracker just now.

PR #9167, opened 2022-08-29
install-types: any perl version should be sufficient

mtelka:
In addition I believe we do not need to list `runtime/perl-XXX/module/sun-solaris` here explicitly since it should be pulled in via SUNWcs implicitly anyway.

### DIFF
--- a/components/meta-packages/install-types/includes/minimal
+++ b/components/meta-packages/install-types/includes/minimal
@@ -133,8 +133,7 @@ depend type=require fmri=package/pkg
 depend type=require fmri=package/svr4
 depend type=require fmri=release/name
 depend type=require fmri=release/notices
-depend type=require fmri=runtime/perl-534
-depend type=require fmri=runtime/perl-534/module/sun-solaris
+depend type=require fmri=runtime/perl
 depend type=require fmri=service/fault-management
 depend type=require fmri=service/file-system/smb
 depend type=require fmri=service/management/sysding
--- a/components/meta-packages/install-types/Makefile
+++ b/components/meta-packages/install-types/Makefile
@@ -16,7 +16,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		install-types
 COMPONENT_VERSION=	0.1
-COMPONENT_REVISION=	51
+COMPONENT_REVISION=	52
 COMPONENT_SUMMARY=	A meta-packages that install common applications for ISOs
 
 include $(WS_MAKE_RULES)/ips.mk


components/meta-packages/install-types/includes/minimal
```diff
@@ -133,8 +133,7 @@ depend type=require fmri=package/pkg
 depend type=require fmri=package/svr4
 depend type=require fmri=release/name
 depend type=require fmri=release/notices
-depend type=require fmri=runtime/perl-534
-depend type=require fmri=runtime/perl-534/module/sun-solaris
+depend type=require fmri=runtime/perl
 depend type=require fmri=service/fault-management
```
AndWac, line 137: How will sun-solaris be installed then?

mtelka: Both `SUNWcs` and `system/extended-system-utilities` requires `sun-solaris`, so I expect the package will be installed automatically to satisfy dependencies.  But, honestly, I do not know how exactly this installation related stuff works so I could be easily wrong and something will fail during the installation.

Similarly I assume the `runtime/perl` meta-package will automatically bring in the right version of `runtime/perl-XXX` based on dependencies of `runtime/perl`.

If this my assumption is not correct, then we cannot do this, obviously.

AndWac:
Alas this doesn't seem to work. I have installed a VM from a newly generated ISO text image. After installation system/intrd (interrupt balancer) fails to start. Nevertheless runtime/perl-536/modules/sun-solaris is installed. From the log:
[ Aug. 30 18:48:00 Executing start method ("/lib/svc/method/svc-intrd"). ]
Can't locate Sun/Solaris/Kstat.pm in @INC (you may need to install the Sun::Solaris::Kstat module) (@INC contains: /usr/perl5/site_perl/5.34/i86pc-solaris-thread-multi-64 /usr/perl5/site_perl/5.34 /usr/perl5/vendor_perl/5.34/i86pc-solaris-thread-multi-64 /usr/perl5/vendor_perl/5.34 /usr/perl5/5.34/lib/i86pc-solaris-thread-multi-64 /usr/perl5/5.34/lib) at /usr/lib/intrd line 66.
So, it looks like the wrong version has been installed.

mtelka:
> Alas this doesn't seem to work. I have installed a VM from a newly generated ISO text image. After installation system/intrd (interrupt balancer) fails to start. Nevertheless runtime/perl-536/modules/sun-solaris is installed. From the log: [ Aug. 30 18:48:00 Executing start method ("/lib/svc/method/svc-intrd"). ] Can't locate Sun/Solaris/Kstat.pm in @inc (you may need to install the Sun::Solaris::Kstat module) (@inc contains: /usr/perl5/site_perl/5.34/i86pc-solaris-thread-multi-64 /usr/perl5/site_perl/5.34 /usr/perl5/vendor_perl/5.34/i86pc-solaris-thread-multi-64 /usr/perl5/vendor_perl/5.34 /usr/perl5/5.34/lib/i86pc-solaris-thread-multi-64 /usr/perl5/5.34/lib) at /usr/lib/intrd line 66. So, it looks like the wrong version has been installed.

This is because system-perl mediator was not switched to 5.36 yet (see #9169).  Since `runtime/perl-536/modules/sun-solaris` was automatically installed, everything works as expected and this change is okay.  Thank you.

PR #9169, opened 2022-08-30
Switch default perl from 5.34 to 5.36

mtelka:


mtelka:
Build of perl-534 and perl-536 fails for unknown reason.  Please look at it.  If it is something that should be fixed please share build logs.  Thanks.

PR #9174, opened 2022-08-31
transforms/defaults: set perl mediator priority for runtime/perl pack…

mtelka:
…ages only

### DIFF
--- a/transforms/defaults
+++ b/transforms/defaults
@@ -153,7 +153,9 @@ set name=variant.arch value=$(MACH)
 #
 # Set the default Perl for mediated links
 #
-<transform link mediator=perl mediator-version=5.36 -> default mediator-priority vendor>
+<transform link mediator=perl mediator-version=5.36 -> default tmp.fmri %{pkg.fmri} >
+<transform link tmp.fmri=.*runtime/perl -> default mediator-priority vendor>
+<transform link tmp.fmri=.* -> delete tmp.fmri .* >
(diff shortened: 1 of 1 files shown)


Toasterson:
Do we have other `link mediator=perl mediator-version=5.36` actions that got affected? I only see normal updates here. If we need to update more packages we will need Revision bumps as this only takes affect for rebuilt packages.

mtelka:
> Do we have other `link mediator=perl mediator-version=5.36` actions that got affected? I only see normal updates here. If we need to update more packages we will need Revision bumps as this only takes affect for rebuilt packages.

All perl related packages (those that deliver versioned man pages or binaries for perl; so usually perl modules) that were (re)built after 60713aad6 (merged on  2022-08-13) are affected (they have `mediator-priority=vendor` but they shouldn't).  Fortunately, this is currently harmless and would cause a problem only in a case when we would try to downgrade our default perl (which we do not plan to).  For example, if we would like to switch back to default perl 5.34 now, we would need to rebuild all affected packages to switch to perl 5.34 successfully.

Since this is harmless, there is no need to rebuild affected packages right now.  They will be rebuilt later either during their upgrade to their new version,  or after we integrate new perl major version.

alarcher:
@mtelka Let me know when you would like this PR to be merged.

Toasterson:
Thanks for the explanation. Merging now.

PR #9184, opened 2022-09-05
Introducing zrepl with version 0.5.0

jgmills:
Zrepl is a one-stop ZFS backup and replication solution, written in the GO language.  The OI version has man pages and sample configuration files.  It also has SMF support.  All errors from the daemon appear in /var/svc/log/application-backup-zrepl:default.log .  The zrepl client also has RBAC support.

The build, install, and publish steps were all successful, as was the package install.  There are no built-in tests.  External tests, a simulated backup, were also successful.

The configuration file is /etc/zrepl/zrepl.yml .  The default file must be modified to suit the circumstances of your systems.  You can test it with "zrepl configcheck" run as root.

To run the zrepl service, enable svc:/application/backup/zrepl:default and watch the service log file.  If it reports "... has been modified", add '"atime": "off"' to the configuration, as shown in one of the sample config files.  Likewise, if it reports "cannot create placeholder filesystem", add "encryption: inherit" to the configuration.

You may notice other errors that do not require a configuration change.  If the SMF log file reports "pool must be upgraded", you must upgrade the pool.  The usual command is "zpool upgrade ...".  The client commands like "zrepl status" and "zrepl version" require the RBAC privilege file_dac_search .  You can provide that privilege by prefixing the command with "pfexec".  Finally, if the backup filesystems exist but are not mounted they can be mounted manually by ZFS.  You will need to override or remove the property "mountpoint=none" before the ZFS mount will succeed.

### DIFF
--- a/components/sysutils/zrepl/Makefile
+++ b/components/sysutils/zrepl/Makefile
@@ -1,0 +1,81 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Gary Mills
+#
+BUILD_BITS=	64
+BUILD_STYLE=	justmake
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		zrepl
+COMPONENT_VERSION=	0.5.0
+COMPONENT_SUMMARY=	A one-stop, integrated solution for ZFS replication
+COMPONENT_PROJECT_URL=	https://zrepl.github.io/
+COMPONENT_FMRI=		backup/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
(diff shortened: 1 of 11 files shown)


alarcher:
@jgmills Thank you, I'll look into it then merge.

AndWac:
The sample-manifest seem to be created with an older version of our tools as it is missing the line
It would be nice if this line could be added to the sample-manifest and the package's manifest.

jgmills:
Yes, they were created with quite an old version.  I'll update the PR to include the added line in a few days.

jgmills:
I don't think this is going to work.  I can add the line to sample-manifest with no problem, but when I add it to zrepl.p5m, I get this error from pkgmogrify:

```
missing value:
    set name=pkg.human-version value=
```

That's with no change to Makefile.  I can fix it by enclosing the null value in double-quotes, but I still don't know what will happen with the new support files.  I think I need to rebase.  How do I do that and still preserve any changes I've made?

AndWac:
You'll need to update your oi-userland repo. It doesn't work on your side becase HUMAN_VERSION gets a default
shared-macros.mk:HUMAN_VERSION ?=       $(COMPONENT_VERSION)

If you don't update (and merge your own branches) you'll get more and more problems.

mtelka:
> How do I do that and still preserve any changes I've made?

Something like this should work:

```
git checkout oi/hipster
git pull
git checkout zrepl
git rebase oi/hipster
```

jgmills:
Thanks.  That seemed to work.  I had to append "upstream oi/hipster" to the "git pull" command, but everything else was verbatim.

jgmills:
Okay, I've rebased my OI repository and updated my OS.  I just did a forced push with a few changed files.  As far as I can tell, it's correct now.

PR #9188, opened 2022-09-07
geoip-database: automatic version on rebuild

dbh201:


erwinlem:
Could you also make sure the file GeoIP.dat.gz is mirrored on dlc.openindiana.org? That file has no version number so it will be overwritten tomorrow. And that prevents me from building it because the hash is different.

dbh201:
The hashes aren't all that useful since mailfud.org doesn't put their checksums with the updates. Would it be better to just do the Makefile like this? This Makefile can just be rebuilt on a schedule to keep the package updated. Not sure how it's going to affect version-locking though.

AndWac:
We should try to get secure checksums. If they aren't supported by the original project we should have at least our own in order to prevent later corruptions or manipulations.

dbh201:
So, to update this from here on, just change the version number and delete the sha256sums.txt file. Running `gmake publish` should do the rest.

PR #9195, opened 2022-09-10
virtualbox: update to 6.1.38

dbh201:


components/sysutils/virtualbox/manifests/sample-manifest.p5m
```diff
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2021 <contributor>
+# Copyright 2022 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
```
mtelka, line 17: Please update both `virtualbox-additions.p5m` and `virtualbox.p5m` similarly.

components/sysutils/virtualbox/Makefile
```diff
@@ -14,30 +14,39 @@
...
 COMPONENT_SUMMARY=      VirtualBox - general-purpose full virtualizer
 COMPONENT_PROJECT_URL=  https://www.virtualbox.org/
 COMPONENT_FMRI=         system/virtualbox
 COMPONENT_CLASSIFICATION= System/Virtualization
+
+COMPONENT_DOWNLOAD_URL= https://download.virtualbox.org/virtualbox/$(COMPONENT_VERSION)
 COMPONENT_SRC=			$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+
+# Grab checksum file
+$(shell cd $(USERLAND_ARCHIVES) && [ -f $(COMPONENT_SRC).sums ] || curl -o $(COMPONENT_SRC).sums https://www.virtualbox.org/download/hashes/$(COMPONENT_VERSION)/SHA256SUMS )
```
mtelka: This is nice way how to detect new values for `SDK_BUILD_NUMBER` and `COMPONENT_ARCHIVE_HASH`, but I believe we still need the actual values recorded here in the `Makefile`.  This would help to detect cases when upstream replace the source tarball with some newer/better/fixed version.  If we won't detect that we could get surprising results.

Thanks.

mtelka (changes requested):

PR #9196, opened 2022-09-10
Add checksum file support to prep-download.mk + some sanity checks

dbh201:


AndWac:
How should a new source file be downloaded with this?

dbh201:
The hash/sig variables take priority, so no changes are required unless both variables are missing from the Makefile.
If both are missing then you need to add 'ALLOW_UNVERIFIED_DOWNLOADS=yes' to the Makefile.

I have a couple ideas for improving it but I haven't fleshed them out yet. At the moment it's just an option to download the checksum file yourself and add it to the component directory, but I want to do automatic downloading of checksum files (based on a list of typical filenames, or COMPONENT_CHECKSUM_FILE), putting them in $(COMPONENT_DIR)/checksums/ and then adding them to the clobber list.

AndWac:
I have seen this variable. I think we should at least document its usage in our doc/makefile-target.txt

dbh201:
I moved a lot of this functionality into userland-fetch. I'll have to look this over and make some changes. I'll be sure to explain any new variables in the docs.

PR #9199, opened 2022-09-10
Add Razor2::Client::Agent perl module

fritzkink:


components/perl/Razor2-Client-Agent/Makefile
```diff
@@ -0,0 +1,37 @@
...
+#   $WS_TOOLS/perl-integrate-module Razor2::Client::Agent
+#
+
+BUILD_STYLE = makemaker
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_PERL_MODULE =		Razor2::Client::Agent
+HUMAN_VERSION =			2.86
+COMPONENT_REVISION =		1
```
mtelka, line 23: This line should not be there.  I'm curious why it appeared here.  Did you run  the `perl-integrate-module` script twice and then squashed both changesets?  Or maybe there is a bug in the `perl-integrate-module` script...

Anyway, this is harmless, so no problem to leave it here.

mtelka (approved):
This apparently replaces #9062, so #9062 could be closed.

components/perl/Razor2-Client-Agent/Makefile
AndWac: Shall I merge this as is or wait for a correction?

mtelka: > Shall I merge this as is or wait for a correction?

Correction is not needed, but I'd like to know the answer to my question.
So feel free to merge.

Thanks.

PR #9205, opened 2022-09-11
Add DB_File perl module

fritzkink:


mtelka (changes requested):
DB_File is already part of `runtime/perl-XXX` packages.  Please run `pkg search DB_File.pm` to confirm.  Or, do I miss something?

PR #9207, opened 2022-09-11
Add Net::CIDR::Lite perl module

fritzkink:


components/perl/Net-CIDR-Lite/test/results-all.master
```diff
@@ -0,0 +1,7 @@
+t/base.t .... ok
+t/more.t .... ok
+t/pod.t ..... skipped: Test::Pod 1.00 required for testing POD
+t/podcov.t .. skipped: Test::Pod::Coverage required for testing pod coverage
```
mtelka, line 4: This would look nicer if you have installed `test-pod` and `test-pod-coverage` packages, but since both tests passes it is not a big deal :-).

mtelka (approved):

PR #9213, opened 2022-09-12
fuse : adapt to kernel changes

jpandre:
Being a kernel module, fuse depends on kernel structures defined by kernel headers.
This is to adapt to a recent change in <sys/thread.h> with no change to fuse proper. Actually this is a workaround similar to pull request #6069 to cope with a missing dependency.


mtelka:
The build failed.  I'm working on a fix.

AndWac:
openindiana/fuse failed. I have manually restarted the build.

Am 12.09.22 um 14:55 schrieb Marcel Telka:

mtelka:
There are several problems with current fuse.  One of them is that it tries to use gcc-4.4.4 which could be not installed (since it is not in REQUIRED_PACKAGES).  I created #9215 to switch fuse to gcc-7 (the same compiler used to build illumos-gate).

PR #9215, opened 2022-09-12
fuse: switch to gcc-7

mtelka:


3eka:
I wonder how much work would it be to create *golang* OI fuse module from this one?

PR #9220, opened 2022-09-12
botan: bump version 2

cgrzemba:
Update library botan to version 2. It is a requisite for keepassxc.

I kept the library name botan and do not changed to botan2 because there is no package depend on library/botan until now.


AndWac:
ok we have a problem with our python mediator settings. You have the old default 2.7 and on our build server I have manually changed it to python-3.9.
This creates a problem here:
-file path=usr/lib/$(MACH64)/python2.7/site-packages/botan2.py
+file path=usr/lib/$(MACH64)/python3.9/site-packages/botan2.py

AndWac:
I have created #9221 for this.

cgrzemba:
I pushed a improved recipe related the Python version problem.

AndWac:
gmake pre-publish fails:
/usr/bin/pkgdepend generate -m -d /export/home/andreas/oi-userland/components/library/botan/build/prototype/i386/mangled -d /export/home/andreas/oi-userland/components/library/botan/build/prototype/i386 -d /export/home/andreas/oi-userland/components/library/botan/build -d /export/home/andreas/oi-userland/components/library/botan -d Botan-2.19.2 -d /export/home/andreas/oi-userland/components/library/botan/files /export/home/andreas/oi-userland/components/library/botan/build/manifest-i386-botan.mangled >/export/home/andreas/oi-userland/components/library/botan/build/manifest-i386-botan.depend
Couldn't find 'usr/lib/python3.9/vendor-packages/__pycache__/botan2.cpython-39.pyc' in any of the specified search directories:
	/export/home/andreas/oi-userland/components/library/botan
	/export/home/andreas/oi-userland/components/library/botan/Botan-2.19.2
	/export/home/andreas/oi-userland/components/library/botan/build
	/export/home/andreas/oi-userland/components/library/botan/build/prototype/i386
	/export/home/andreas/oi-userland/components/library/botan/build/prototype/i386/mangled
	/export/home/andreas/oi-userland/components/library/botan/files
gmake: *** [/export/home/andreas/oi-userland/make-rules/ips.mk:499: /export/home/andreas/oi-userland/components/library/botan/build/manifest-i386-botan.depend] Error 1

cgrzemba:
hmm, 
this file is not in manifest and pyc files should never be contents of package ?

AndWac:
Yes, probably created by a tranform from another file.

cgrzemba:
Oh, yes: autopyc
this transform assembles a dependency: botan2.cpython-39.pyc 
for what reason?

cgrzemba:
I'm wrong, Solaris ships for Python packages compiled files too. So transform is correct but the compile step is missing here.

AndWac:
Running gmake sample-manifest here gives a small difference:
-file path=usr/lib/$(MACH64)/python3.9/site-packages/botan2.py
+file path=usr/lib/python3.9/vendor-packages/botan2.py

gmake pre-publish runs fine, though. Shall I merge this as is?

components/library/botan/Makefile
```diff
@@ -47,24 +61,14 @@ CONFIGURE_OPTIONS+= --includedir=/usr/include
...
 # Auto-generated dependencies
 REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
 REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
 REQUIRED_PACKAGES += compress/bzip2
-REQUIRED_PACKAGES += library/security/openssl
 REQUIRED_PACKAGES += library/zlib
-REQUIRED_PACKAGES += shell/ksh93
 REQUIRED_PACKAGES += system/library
 REQUIRED_PACKAGES += system/library/math
+REQUIRED_PACKAGES += runtime/python-$(subst .,,$(PYTHON_VERSION))
```
mtelka: Was this line really auto-generated?

components/library/botan/Makefile
```diff
@@ -13,26 +13,40 @@
 # Copyright 2021 Andreas Wacknitz
 #
 
-BUILD_BITS= 64_and_32
+BUILD_BITS= 64
```
mtelka: This is default now, so it could be removed.

components/library/botan/Makefile
```diff
@@ -13,26 +13,40 @@
...
 
 COMPONENT_PRE_CONFIGURE_ACTION = \
         ($(CLONEY) $(SOURCE_DIR) $(@D))
 
-CONFIG_SHELL =  /usr/bin/python
+COMPONENT_POST_INSTALL_ACTION = \
+	( cd $(PROTO_DIR) && \
+	mkdir -p .$(PYTHON_LIB) && \
+	mv usr/lib/$(MACH64)/python$(PYTHON_VERSION)/site-packages/botan2.py .$(PYTHON_LIB) && \
+	$(CONFIG_SHELL) -m py_compile .$(PYTHON_LIB)/botan2.py )
```
mtelka, line 43: Isn't possible to achieve the same results using some simpler approach?

components/library/botan/Makefile
```diff
@@ -13,26 +13,40 @@
...
+
 CONFIGURE_SCRIPT= $(SOURCE_DIR)/configure.py
 
 CONFIGURE_OPTIONS.64= --cpu=x86_64
```
mtelka: Since you dropped 32-bit this could be simplified.

mtelka (changes requested):

components/library/botan/Makefile
cgrzemba: no, it was my attempt to resolve the pyc dependency, but basically correct

mtelka: > no, it was my attempt to resolve the pyc dependency, but basically correct

If you run `gmake REQUIRED_PACKAGES` again, will it be replaced by runtime/python-39, or simply removed?  If removed, you need to move this up (above the `Auto-generated dependencies` comment).  If replaced, then you could leave it as is, but better would be to replace it,  because it will get replaced anyway later.

If it was replaced by runtime/python-XX, where XX is something else than 39, then you need to do something to solve that.

components/library/botan/Makefile
cgrzemba: The usr/lib/$(MACH64) path is correct and neccessary required for the other parts of the component.

components/library/botan/Makefile
cgrzemba: The component has no option for setting the PYTHONPATH for installing. Its simply guessing the site-packages path. The alternative would be hacking the install script. The python module compiling is necessary any way.
So I prefer this three lines.

components/library/botan/Makefile
mtelka: > The usr/lib/$(MACH64) path is correct and neccessary required for the other parts of the component.

~~I'm not trying to say what you did is incorrect (OTOH, this does not imply it is correct :-); I simply do not know).  It just looks too complex.  Isn't there some other way to place the `botan2.py` to proper place?  Some option to configure script, some simple patch of Makefiles, etc.~~

How is `usr/lib/$(MACH64)` related to `CONFIGURE_OPTIONS.64= --cpu=x86_64`?

components/library/botan/Makefile
cgrzemba: gmake REQUIRED_PACKAGES will add

so I will choose this too.

cgrzemba:
I stumble over the license text in the moment:

Copyright (C) 1999-2022 The Botan Authors
All rights reserved.

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:

1. Redistributions of source code must retain the above copyright notice,
   this list of conditions, and the following disclaimer.

2. Redistributions in binary form must reproduce the above copyright
   notice, this list of conditions, and the following disclaimer in the
   documentation and/or other materials provided with the distribution.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
POSSIBILITY OF SUCH DAMAGE.

This is not BSD? It is proprietary?

mtelka:
> I stumble over the license text in the moment:
> 
> Copyright (C) 1999-2022 The Botan Authors All rights reserved.
> 
> Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
> 
>     1. Redistributions of source code must retain the above copyright notice,
>        this list of conditions, and the following disclaimer.
> 
>     2. Redistributions in binary form must reproduce the above copyright
>        notice, this list of conditions, and the following disclaimer in the
>        documentation and/or other materials provided with the distribution.
> 
> 
> THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
> 
> This is not BSD? It is proprietary?

This is `BSD 2-Clause "Simplified" License`, with SPDX identifier `BSD-2-Clause`.  See https://spdx.org/licenses/BSD-2-Clause.html

components/library/botan/Makefile
```diff
@@ -13,58 +13,61 @@
...
 COMPONENT_SRC=		Botan-$(COMPONENT_VERSION)
-COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tgz
-COMPONENT_ARCHIVE_HASH=	sha256:6847ffb64b8d2f939dccfecc17bd2c80385d08f7621e2c56d3a335118e823613
-COMPONENT_ARCHIVE_URL=	https://botan.randombit.net/releases/old/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.xz
+COMPONENT_ARCHIVE_HASH=	\
+	sha256:3af5f17615c6b5cd8b832d269fb6cb4d54ec64f9eb09ddbf1add5093941b4d75
+COMPONENT_ARCHIVE_URL=	https://botan.randombit.net/releases/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI=library/security/botan
+COMPONENT_CLASSIFICATION=org.opensolaris.category.2008:System/Security
```
mtelka: The `org.opensolaris.category.2008:` part should not be there.  See `grep COMPONENT_CLASSIFICATION make-rules/ips.mk`.

components/library/botan/Makefile
```diff
@@ -13,58 +13,61 @@
...
 
-BUILD_BITS= 64_and_32
-#USE_OPENSSL11= yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		botan
-COMPONENT_VERSION=	1.10.17
-COMPONENT_LICENSE=	BSD
+COMPONENT_VERSION=	2.19.2
+COMPONENT_LICENSE=	BSD 2-Clause "Simplified" License
```
mtelka: I suggest to use the short SPDX identifier here.  It is more accurate and less confusing.

components/library/botan/botan.p5m
```diff
@@ -14,293 +14,401 @@
 # Copyright 2021 Andreas Wacknitz
 #
 
-set name=pkg.fmri value=pkg:/library/security/botan@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+# Drop static libraries:
+<transform path=usr/lib/.*\.a -> drop>
```
mtelka: Isn't this already handled by `transforms/archive-libraries-drop`?

mtelka (changes requested):

components/library/botan/botan.p5m
cgrzemba: it seems that the transforms/archive-libraries-drop is not referenced by make-rules/ips.mk

mtelka: > it seems that the transforms/archive-libraries-drop is not referenced by make-rules/ips.mk

Interesting.  It looks like it is completely unused.  Anyway, you could use it by adding `PKGMOGRIFY_TRANSFORMS += $(WS_TOP)/transforms/archive-libraries-drop` to your `Makefile`.  Up to you.

components/library/botan/Makefile
```diff
@@ -13,58 +13,61 @@
...
-CONFIGURE_OPTIONS+= --disable-modules=ecdsa
+CONFIGURE_OPTIONS+= --os solaris
 CONFIGURE_OPTIONS+= --with-bzip2
 CONFIGURE_OPTIONS+= --with-zlib
+CONFIGURE_OPTIONS+= --with-python-versions=$(PYTHON_VERSION)
 
-COMPONENT_POST_INSTALL_ACTION= \
-            (cd $(PROTOUSRDIR) ;  \
-             $(MV)  include/botan-1.10/botan/build.h include/botan-1.10/botan/build-$(BITS).h )
+test: $(TEST_64)
```
mtelka: Is this line really needed here?

components/library/botan/botan.p5m
AndWac: We cannot drop static libraries automatically because we have rare cases where we only have static libs. So, the rule of thumb is to drop them if shared libs are available.

mtelka: > We cannot drop static libraries automatically because we have rare cases where we only have static libs. So, the rule of thumb is to drop them if shared libs are available.

Or we can do it this way: #9236

mtelka:
@cgrzemba, I think you should run `gmake sample-manifest` to update your p5m files.  It is always good idea to run this to make sure nothing changed:

```
gmake sample-manifest
gmake REQUIRED_PACKAGES
gmake test
gmake publish
git status .
```

components/library/botan/manifests/sample-manifest.p5m
```diff
@@ -22,276 +23,387 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
...
+file path=usr/include/botan-2/botan/xmss_wots.h
+file path=usr/include/botan-2/botan/xmss_wots_parameters.h
+file path=usr/include/botan-2/botan/xmss_wots_privatekey.h
+file path=usr/include/botan-2/botan/xmss_wots_publickey.h
+file path=usr/include/botan-2/botan/xtea.h
+file path=usr/include/botan-2/botan/xts.h
+file path=usr/include/botan-2/botan/zfec.h
+file path=usr/include/botan-2/botan/zlib.h
+file path=usr/lib/$(MACH64)/libbotan-2.a
+link path=usr/lib/$(MACH64)/libbotan-2.so target=libbotan-2.so.19.19.2
```
AndWac, line 338: If you run gmake sample-manifest with #9236 there shouldn't be any static lib here.

components/library/botan/botan.p5m
```diff
@@ -14,293 +14,397 @@
...
+file path=usr/include/botan-2/botan/xmss_wots.h
+file path=usr/include/botan-2/botan/xmss_wots_parameters.h
+file path=usr/include/botan-2/botan/xmss_wots_privatekey.h
+file path=usr/include/botan-2/botan/xmss_wots_publickey.h
+file path=usr/include/botan-2/botan/xtea.h
+file path=usr/include/botan-2/botan/xts.h
+file path=usr/include/botan-2/botan/zfec.h
+file path=usr/include/botan-2/botan/zlib.h
+file path=usr/lib/$(MACH64)/libbotan-2.a
+link path=usr/lib/$(MACH64)/libbotan-2.so target=libbotan-2.so.19.19.2
```
AndWac, line 339: This will fail to publish on the build server.

components/library/botan/manifests/sample-manifest.p5m
mtelka: > If you run gmake sample-manifest with #9236 there shouldn't be any static lib here.

Exactly. @cgrzemba, you should make sure your git botan branch is up to date with oi/hipster since you removed transforms from your p5m file.

components/library/botan/botan.p5m
cgrzemba: Why will it fail?

AndWac: You are right, it won't fail. It would devliver an unwanted static lib. BTW: I have created #9242 which will make dropping static libs optional after some discussions.

mtelka:
@cgrzemba, you should always make sure the `diff manifests/sample-manifest.p5m botan.p5m` output is as shortest as possible.

components/library/botan/botan.p5m
AndWac, line 339: In the old manifest we had explicit transforms to drop static libs. Without a transform rule you should manually remove the static library.

PR #9226, opened 2022-09-13
Default DB version should be the newest one

mtelka:


mtelka:
mariadb-10.1 is no longer supported.  It was EOLed almost 2 years ago.  We should start to obsolete it.

AndWac:
This would be a paradigm shift. In the past we had the oldest DB as our default.

mtelka:
> This would be a paradigm shift. In the past we had the oldest DB as our default.

Yes, I know.  I edited comments saying so. :-)

The previous (defensive) default versions are perfect when we would aim for very high stability and as less changes as possible, but it is hard to maintain and makes switch to newer DB versions harder than they needs to be (exactly as I wrote in the comment: we would detect incompatibilities late giving us less time to solve them).

AndWac:
I am all for it. In the past such a change would have been accompanied with another PR rebuilding all depending packages.

mtelka:
I prefer to defer the rebuild once we need to rebuild such components for any reason (IOW, lazy rebuild).  Maybe I'll do so later for mariadb-10.1 dependencies so we can obsolete mariadb-10.1 ASAP.

We do mass rebuilds in a case when some upgraded library would make binaries (from other packages) non-working due to incompatible changes in the library.  In case of this PR there is no such incompatibility.  The only issue we could have after this PR is merged are few components that could be no longer buildable, but this should be easily fixable when we encounter them by simply adding `PG_VERSION=XXX` or `MYSQL_VERSION=XXX` to their Makefile.

AndWac:
Aurélien tries to rebuild every component with gcc-10. Such postponed changes might add more work for him and thus will hinder a switch to a newer gcc. We need to keep that in mind.

mtelka:
> Aurélien tries to rebuild every component with gcc-10. Such postponed changes might add more work for him and thus will hinder a switch to a newer gcc. We need to keep that in mind.

Sorry, that's not true for at least hunspell.  It was not buildable since [November 2020](https://github.com/OpenIndiana/oi-userland/commit/e847e8bef060246f5231f4cb1aaaeee9fc9cca56), until I fixed it (as a part of perl update work) about a [month ago](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/text/hunspell/patches/04-gettext.patch).

OTOH, when there are any non-buildable components found (and reported) after this PR, I'll be happy to fix them.

alarcher:
@mtelka There is a reason why gcc has not been updated recently and it is because people break the userland too often. I do not mind being pragmatic but at least let us try to keep the damage at a minimum. Identifying and listing the affected components in a ticket could help, not just leaving them blindly in a broken state.

PR #9236, opened 2022-09-14
make-rules/ips.mk: drop static libraries by default

mtelka:


AndWac:
Ok, with this integrated we need to find the packages that want static libs.

mtelka:
> Ok, with this integrated we need to find the packages that want static libs.

Yes, but we do not need to hurry.  This affects sample-manifest only.  So this will propagate slowly.

alarcher:
@mtelka I probably does not because scientific libraries need to deliver static libraries. At least these components may be broken. Same with boost.

mtelka:
> @mtelka I probably does not because scientific libraries need to deliver static libraries. At least these components may be broken. Same with boost.

Sorry, but nothing will be broken after this. This changes `manifests/sample-manifest.p5m` only, not component's `.p5m` file, nor the generated package.  It is up to component integrator if, when and how this will propagate to real packages.

alarcher:
> > @mtelka I probably does not because scientific libraries need to deliver static libraries. At least these components may be broken. Same with boost.
> 
> Sorry, but nothing will be broken after this. This changes `manifests/sample-manifest.p5m` only, not component's `.p5m` file, nor the generated package. It is up to component integrator if, when and how this will propagate to real packages.

@mtelka @AndWac You are breaking the existing behaviour as static libraries will disappear from the sample-manifest if the switch is not added to the component. What is the point then besides removing libraries that are present in the workspace and may be added to the manifest? To me it looks like solving a non-problem and creating a confusing situation. Next time someone updates a component delivering static libraries they therefore need to be aware of this hidden switch. Recent changes go towards convenience at the expense of consistency: at least they should be document instead of driving by commit. If you want to change the "paradigm" document the new expected otherwise we are going to chase several rabbits trying to fix components that have been broken or affected by silent modifications, sometimes for questionable benefits.
Besides that let me state that you can deal with the compiler upgrades on your own. I am not going to spend hours fixing stuff again for nothing.

mtelka:
@alarcher, please share list of components that are failing or have some other problems and I'll fix them.  Thanks.

AndWac:
Ok, should we revert this change?

Am 14.09.22 um 20:10 schrieb Aurélien Larcher:

mtelka:
> Ok, should we revert this change?

Why?  There is absolutely no reason for that.

AndWac:
Offending Aurélien is a good reason for that.

Am 15.09.22 um 10:54 schrieb Marcel Telka:

mtelka:
> Offending Aurélien is a good reason for that.

:-).  Ok.  Do whatever you want.

AndWac:
Ok, I inverted the logic with #9242. So one can make use of
DROP_STATIC_LIBRARIES intentionally.

Am 15.09.22 um 10:59 schrieb Marcel Telka:

mtelka:
> Ok, I inverted the logic with #9242. So one can make use of DROP_STATIC_LIBRARIES intentionally.

I believe defaults should be the most common cases to make our life easier and Makefiles as short as possible.  But as I said: do whatever you want. :-)

AndWac:
You are right but this is the least deviation from the old behavior.
Next time I will try to discuss such changes or at least give people
some time to react on proposed changes.

Am 15.09.22 um 11:17 schrieb Marcel Telka:

mtelka:
BTW, when you merge #9242, you'll break botan (similarly as there are some unknown components broken now after #9236). :-)

mtelka:
The following command shows there are currently 90 packages delivering `*.a` files.  I believe most of them should not.

```
pkg search -H '*.a' | awk '{print $4}' | sed -e 's/@.*$//g' -e 's/^pkg:\///g' | sort -u | wc -l
```

pyhalov:
It will clearly break components. Some of them deliver only static libraries. For example, You definately need postgresql static libraries.

mtelka:
> It will clearly break components. Some of them deliver only static libraries. For example, You definately need postgresql static libraries.

@pyhalov, please define 'break', because `gmake publish` will produce same results as before.

pyhalov:
Look - https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/database/postgresql-12/postgresql12-developer.p5m#L869 . There are PostgreSQL libraries, which are shipped only as static ones. They are needed to compile PostgreSQL extensions.

pyhalov:
The awful thing here - you'll not notice breakage - component will publish, but will fail to deliver some files, which it is expected to ship. So this should be turned on carefully on component per component basis.

pyhalov:
Perhaps, better invest time in turning on SYSTEM_TESTs, they could be useful to find such issues. Sorry, talking is cheap :)

mtelka:
Sorry, @pyhalov, you apparently missed what is (was) this change about.  This PR didn't change the resulting package, so in your postgresql example, when you run `gmake publish` there you'll get static libraries packaged.  Exactly as before.

mtelka:
> Perhaps, better invest time in turning on SYSTEM_TESTs, they could be useful to find such issues. Sorry, talking is cheap :)

Please elaborate.  What you mean by 'turning on SYSTEM_TESTs'?  Thanks.

pyhalov:
solaris-userland has [system-test ](https://github.com/oracle/solaris-userland/commit/413f92294e7aa7e3eaae210239b6c6cf9b38fb50) target, which is intended to test installed packages.

mtelka:
> solaris-userland has [system-test ](https://github.com/oracle/solaris-userland/commit/413f92294e7aa7e3eaae210239b6c6cf9b38fb50) target, which is intended to test installed packages.

Thanks for the pointer.  That would be nice to have, but currently not so important for me to spend time on it.  Sorry.

PR #9240, opened 2022-09-14
make-rules: run pkglint via unversioned python

mtelka:


pyhalov:
Do we deliver pkg modules for all pythons? If not, this can lead to pkglint failure if python version, selected by mediator, lacks it.

mtelka:
> Do we deliver pkg modules for all pythons? If not, this can lead to pkglint failure if python version, selected by mediator, lacks it.

No, we don't.  On (official) build machines it is expected that all mediators are default (unset).

If you feel it is a problem, then we should start to deliver perl modules from illumos-gate for all perl versions to make sure that intrd works even when a random user sets the system-perl mediator to non-default value.  This is far more dangerous (all users are possibly affected) than pkglint executed during the oi-userland build (only few experienced developers are possibly affected).

PR #9249, opened 2022-09-16
userland tools: use python-3.9 instead of 3.5

AndWac:
After pkg5 #116 has been merged we need to use python-3.9.

### DIFF
--- a/tools/bass-o-matic
+++ b/tools/bass-o-matic
@@ -1,4 +1,4 @@
-#!/usr/bin/python3.5
+#!/usr/bin/python3.9
 #
 # CDDL HEADER START
 #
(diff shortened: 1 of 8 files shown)


mtelka:
Maybe it would be better to simply use `/usr/bin/python` instead to avoid similar issues in future.

AndWac:
I prefer to be explicit here instead of relying on mediator settings.

mtelka:
> I prefer to be explicit here instead of relying on mediator settings.

Okay. Up to you :-).  But you are asking for future problems with this.

AndWac:
Yes, but I think they are easily solvable.

PR #9270, opened 2022-09-17
Add Error perl module

fritzkink:


components/perl/Error/perl-integrate-module.conf
```diff
@@ -0,0 +1,4 @@
+%include-2%
+COMPONENT_VERSION=              0.17029
+HUMAN_VERSION=			1.17029
```
mtelka: Sorry, this is wrong.  HUMAN_VERSION must always match the version as seen at CPAN.

This module needs some tweak in `perl-version-convert` to integrate properly.  Please try to run `perl-integrate-module` again, once #9273 is merged (and without any `perl-integrate-module.conf`).

mtelka (changes requested):

mtelka (approved):

PR #9287, opened 2022-09-20
obsolete mongodb-40

AndWac:
Mongodb-40 doesn't build anymore for some time and nobody cared. So let's obsolete it.


mtelka:
+1

Also, [MongoDB 4.0 reached EOL in April 2022](https://www.mongodb.com/support-policy/lifecycles), so obsolete is the right thing to do.

PR #9290, opened 2022-09-21
gnupg;gnutls-3: fix keyserver errors

dbh201:
BUG 1: gnupg was compiled to use libdns, which caused gnupg programs to be unable to resolve dns names.
BUG 2: gnutls-3 was not compiled with the system certificate store locations, which caused gnupg (and probably others) to be unable to validate certificates.

For the commit fixing bug 2, I have also disabled gnutls 32-bit. If 32-bit is required somewhere else (I haven't checked), the autogen package must be rebuilt to include 32-bit as well.

### DIFF
--- a/components/library/gnutls-3/Makefile
+++ b/components/library/gnutls-3/Makefile
@@ -31,7 +31,7 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		gnutls
 COMPONENT_MJR_VERSION=	3.6
 COMPONENT_VERSION=	$(COMPONENT_MJR_VERSION).16
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_PROJECT_URL=  https://gnutls.org/
 COMPONENT_SUMMARY=	GNU transport layer security library
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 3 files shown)


components/library/gnutls-3/Makefile
```diff
@@ -25,13 +25,13 @@
 # Copyright (c) 2022, Andreas Wacknitz
 #
 
-BUILD_BITS=		64_and_32
+BUILD_BITS=		64
 include ../../../make-rules/shared-macros.mk
```
AndWac: When removing 32 bit we'll need to check and change all dependent packages (recursively):
pkg search -r -H -o pkg.name "depend:require:library/gnutls-3" | sort -u
application/cluster/pacemaker
crypto/gnupg
database/yaz
desktop/nntp/pan
desktop/remote-desktop/rdesktop
desktop/remote-desktop/tigervnc
diagnostic/wireshark/wireshark-common
editor/gnu-emacs/gnu-emacs-gtk
editor/gnu-emacs/gnu-emacs-x11
gnome/vino
library/desktop/gtk-vnc
library/desktop/vte-291
library/glib-networking
library/libmicrohttpd
library/libzip
library/print/cups-libs
library/samba/libsmbclient
library/xmlsec
mail/claws-mail
mail/mailutils
network/chat/weechat
print/qpdf
service/network/samba
system/rsyslog
web/aria2
x11/server/xvnc

As you can see there are some 64 bit-only packages already but some deliver also 32 bit.
Alas it looks like we cannot drop 32 bit gnutls-3 at the moment...

dbh201: I had to change autogen to include 32-bit libopts so this can compile. #9300

components/sysutils/gnupg/Makefile
```diff
@@ -68,6 +68,7 @@ CONFIGURE_OPTIONS  +=		--infodir=$(CONFIGURE_INFODIR)
 CONFIGURE_OPTIONS  +=		--enable-nls
 CONFIGURE_OPTIONS  +=		--enable-largefile
```
AndWac, line 70: You also need to set COMPONENT_REVISION=1 here.

PR #9301, opened 2022-09-21
Obsolete glib1 and gtk1

mtelka:


mtelka:
Both `glib-1.2.10` and `gtk-1.2.10`  were released more than 20 years ago, they are unmaintained for many many years now, we do not have build recipes for them in oi-userland, and nothing depends on them in our pkg repo.  The only reasonable way to go is to obsolete both packages.

PR #9310, opened 2022-09-22
mariadb - update to version 10.6.10

fritzkink:
now with openssl 1.1


AndWac:
gmake publish:
ERROR userland.action001.2        64-bit object 'usr/mariadb/10.6/mysql-test/suite/plugins/pam/pam_mariadb_mtr.so' in 32-bit path
gmake: *** [/export/home/andreas/oi-userland/make-rules/ips.mk:560: /export/home/andreas/oi-userland/components/database/mariadb-106/build/.linted-i386] Error 1

components/database/mariadb-106/Makefile
```diff
@@ -16,21 +16,23 @@
 # Copyright 2022, Daniel Bell
 #
 
-BUILD_BITS= 64
-BITS=		64
-BUILD_STYLE=cmake
+BUILD_BITS= 		64
+BITS=			64
```
mtelka: You do not need to set `BUILD_BITS` and `BITS` to 64.  Both are default.

PR #9315, opened 2022-09-23
Obsolete WBEM

mtelka:


mtelka:
Reasons why we want this: https://openindiana.org/pipermail/oi-dev/2022-September/033251.html

Toasterson:
LGTM, I think anybody using WBEM would have the resources to compile it themselves and add a package if they realy desire to. Which they always can.

Toasterson (approved):

PR #9357, opened 2022-09-26
bash: Update to 5.2

ghost:
Update bash to the latest version 5.2.


components/shell/bash/bash.p5m
```diff
@@ -21,6 +21,7 @@
 # Copyright (c) 2011, 2016, Oracle and/or its affiliates. All rights reserved.
 # Copyright 2017, Aurelien Larcher
 # Copyright 2019, Michal Nowak
+# Copyright 2022, Niklas Poslovski
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
```
AndWac, line 27: Please also add
here.

AndWac:
Did you run "gmake test" and weren't there any changes?
At leasr results-64.master should be renamed to results-all.master.

ghost:
Human-version is already in line 33 of the file.
I already ran gmake test and copied the new results to results-64.master but gmake test didn't correctly reply with Differences found/No differences found so I'm not sure if it works as expected.
I renamed the file to results-all.master,but when running gmake test again that doesn't make any difference here.

AndWac:
We have a new variable USE_COMMON_TEST_MASTER with yes as its default. Thus, without changing that in the Makefile gmake test expects test/results-all.master. If that's not the case it still compares the results but gives an error even if there are no differences. I don't have access to my OI systems at the moment so I cannot check and analyse further.

PR #9358, opened 2022-09-26
Add Mail::SPF perl module

fritzkink:
@mtelka please check why tests are not working. Also please check why *.p5m was not fully build correctly and whether my adaptions in perl-integrate-module.conf are reasonable.

### DIFF
--- a/components/perl/Mail-SPF/Makefile
+++ b/components/perl/Mail-SPF/Makefile
@@ -1,0 +1,47 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# This file was automatically generated using the following command:
+#   $WS_TOOLS/perl-integrate-module Mail::SPF
+#
+
+BUILD_STYLE = modulebuild
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_PERL_MODULE =		Mail::SPF
+HUMAN_VERSION =			2.9.0
+COMPONENT_SUMMARY =		Mail::SPF - An object-oriented implementation of Sender Policy Framework
+COMPONENT_ARCHIVE_URL=		https://cpan.metacpan.org/authors/id/J/JM/JMEHNLE/mail-spf/Mail-SPF-v2.9.0.tar.gz
+COMPONENT_ARCHIVE_HASH =	\
+	sha256:61cb5915f1c7acc7a931ffc1bfc1291bdfac555e2a46eb2391b995ea9ecb6162
+COMPONENT_LICENSE =		BSD-3-Clause
+COMPONENT_LICENSE_FILE =	LICENSE
+
+COMPONENT_SRC =                 $(COMPONENT_NAME)-v$(HUMAN_VERSION)
+
+
+include $(WS_MAKE_RULES)/common.mk
+
+# see https://rt.cpan.org/Ticket/Display.html?id=93241
+COMPONENT_PREP_ACTION =+ $(RM) -f $(SOURCE_DIR)/t/90-author-pod-validation.t
+GENERATE_EXTRA_CMD += | grep -v spfd
+
+
+# Auto-generated dependencies
+PERL_REQUIRED_PACKAGES += library/perl-5/error
+PERL_REQUIRED_PACKAGES += library/perl-5/module-build
+PERL_REQUIRED_PACKAGES += library/perl-5/net-dns
+PERL_REQUIRED_PACKAGES += library/perl-5/net-dns-resolver-programmable
+PERL_REQUIRED_PACKAGES += library/perl-5/netaddr-ip
+PERL_REQUIRED_PACKAGES += library/perl-5/uri
+PERL_REQUIRED_PACKAGES += runtime/perl
--- a/components/perl/Mail-SPF/perl-integrate-module.conf
+++ b/components/perl/Mail-SPF/perl-integrate-module.conf
@@ -1,0 +1,23 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Friedrich Kink
+#
+
+%include-2%
+COMPONENT_SRC =                 $(COMPONENT_NAME)-v$(HUMAN_VERSION)
+
+%include-3%
+# see https://rt.cpan.org/Ticket/Display.html?id=93241
+COMPONENT_PREP_ACTION =+ $(RM) -f $(SOURCE_DIR)/t/90-author-pod-validation.t
+GENERATE_EXTRA_CMD += | $(GNU_GREP) -v spfd
+
--- a/components/perl/Mail-SPF/Mail-SPF-PERLVER.p5m
+++ b/components/perl/Mail-SPF/Mail-SPF-PERLVER.p5m
@@ -1,0 +1,87 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# This file was automatically generated using perl-integrate-module
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)-$(PLV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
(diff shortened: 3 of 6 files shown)


mtelka:
> @mtelka please check why tests are not working. Also please check why *.p5m was not fully build correctly and whether my adaptions in perl-integrate-module.conf are reasonable.

The first problem you have is license.  You need `%hook-no-license%` to solve that.

Since this module primarily supports `modulebuild` you should switch to `BUILD_STYLE = modulebuild`.  This should solve the problem with installation and sample-manifest.

For an example how to do that (all of the above) please look at `Encode-Detect/perl-integrate-module.conf`.

For failed tests you need to find the root cause of the failure and solve that.

There is also one additional problem: `usr/sbin/spfd`.  This file cannot be installed like that because you'll end up with two packages (-534 and -536) containing the same file.  Such packages would be impossible to install together.

fritzkink:
Thanks should be ok now.

components/perl/Mail-SPF/perl-integrate-module.conf
```diff
@@ -0,0 +1,34 @@
...
+%include-2%
+COMPONENT_SRC =                 $(COMPONENT_NAME)-v$(HUMAN_VERSION)
+
+%include-3%
+# see https://rt.cpan.org/Ticket/Display.html?id=93241
+COMPONENT_PREP_ACTION =+ rm -f $(SOURCE_DIR)/t/90-author-pod-validation.t
+
+%hook-manifest%
+# this daemon is not needed for the package
+/usr/gnu/bin/sed -i -e '/spfd$/d' "$DISTRIBUTION-PERLVER.p5m"
```
mtelka: I suggest to add something like:

```
GENERATE_EXTRA_CMD += | grep -v spfd
```

into `%include-3%`.  This is far simpler and it will also remove the file from sample manifest.

components/perl/Mail-SPF/perl-integrate-module.conf
```diff
@@ -0,0 +1,34 @@
...
+%include-1%
+# Force modulebuild build style
+BUILD_STYLE = modulebuild
+
+%include-2%
+COMPONENT_SRC =                 $(COMPONENT_NAME)-v$(HUMAN_VERSION)
+
+%include-3%
+# see https://rt.cpan.org/Ticket/Display.html?id=93241
+COMPONENT_PREP_ACTION =+ rm -f $(SOURCE_DIR)/t/90-author-pod-validation.t
```
mtelka: Here it would be far better to add a patch that either fixes the test (the patch is attached to the ticket), or skips the test (a suggestion how to do that is in the ticket too).

components/perl/Mail-SPF/perl-integrate-module.conf
```diff
@@ -0,0 +1,34 @@
...
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Friedrich Kink
+#
+
+%hook-no-license%
+printf "%s\n" "$COPYRIGHT" > "$DISTRIBUTION.license"
+LICENSE="MPL-1.1"
```
mtelka: The license is definitely not MPL-1.1.  It looks like it is some kind of BSD, but you need to find which one: https://spdx.org/licenses/

mtelka (changes requested):

components/perl/Mail-SPF/perl-integrate-module.conf
fritzkink: No I don't, the author expressed his thoughts in a file called LICENSE which we should respect. Assuming something else without feedback from the author is simply wrong. I'll remove of course MPL-1.1 but the Mail-SPF.license file contains only the original license file.

components/perl/Mail-SPF/perl-integrate-module.conf
fritzkink: Patching as suggested did not resolve completely the issue (different results between the perl versions). I'm not aware that openindiana provides distroprefs for perl where the mentioned file should land. Deleting the file simply also skips the test.

components/perl/Mail-SPF/perl-integrate-module.conf
fritzkink: ok

components/perl/Mail-SPF/perl-integrate-module.conf
mtelka: > No I don't, the author expressed his thoughts in a file called LICENSE which we should respect. Assuming something else without feedback from the author is simply wrong. I'll remove of course MPL-1.1 but the Mail-SPF.license file contains only the original license file.

Sorry, I do not understand.  Please elaborate.  Do you agree that the `LICENSE` file (and `Mail-SPF.license` file) contains some kind of BSD-style license?

fritzkink: This section is part of almost any license terms. And "some kind of" is exactly what I mean with assumption.

components/perl/Mail-SPF/perl-integrate-module.conf
```diff
@@ -0,0 +1,34 @@
...
+USE_DEFAULT_PERL_LICENSE=0
+
+%include-1%
+# Force modulebuild build style
+BUILD_STYLE = modulebuild
```
mtelka: This could be safely dropped after #9377 is merged.

components/perl/Mail-SPF/perl-integrate-module.conf
mtelka: > This section is part of almost any license terms. And "some kind of" is exactly what I mean with assumption.

What you refer to with "this section"?

Sorry, please elaborate about "assumption" more.  I'm afraid I still do not understand what you try to say by all of that.

And, I still miss answer to my question: Do you agree that the `LICENSE` file (and `Mail-SPF.license` file) contains some kind of BSD-style license?

If I know the answer to that question maybe I'd better understand what problem you see here and there.

Thanks.

fritzkink: With "this section" I mean the stanza in capital letters in the license file. 
License is a strong legal term. If you change the license because you think it contains "some kind of" what ever other license without asking for feedback from the original author(s) you are liable for the consequences. Just a simple example:
If the license just contains "You can use the source for what ever you want but you are by any means not allowed to change the code" and you replace it because it contains "some kind of" whatever known license and this 15 page license contains a small stanza stating "in case of a bug you are allowed to fix it", you turn it into the opposite of what the author(s) intended.  And now, who is liable in case of a legal proceedings?

Basically it is not about whether I agree to your statement "contains some kind of BSD-style license". It is about we are not allowed to change whatever license.

components/perl/Mail-SPF/perl-integrate-module.conf
```diff
@@ -0,0 +1,28 @@
...
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Friedrich Kink
+#
+
+%hook-no-license%
+printf "%s\n" "$COPYRIGHT" > "$DISTRIBUTION.license"
+LICENSE="$DISTRIBUTION license"
```
mtelka: Please wait until #9451 is merged, then rebase on top of it, remove this line and re-run the `perl-integrate-module` script.  Thanks.

mtelka (changes requested):

AndWac:
@mtelka: is this ready for merge now?

components/perl/Mail-SPF/perl-integrate-module.conf
mtelka: Please elaborate.  I just tried to run `perl-integrate-module` with this patch: https://rt.cpan.org/Ticket/Attachment/1603414/857647/Util.pm-patch.gz and `COMPONENT_PREP_ACTION` line removed from `perl-integrate-module.conf` and the run passed without any warning (or error).

mtelka: If this line will stay here (I believe it shouldn't), please replace `rm -f` by `$(RM)`.

components/perl/Mail-SPF/perl-integrate-module.conf
```diff
@@ -0,0 +1,23 @@
...
+# Copyright 2022 Friedrich Kink
+#
+
+%include-2%
+COMPONENT_SRC =                 $(COMPONENT_NAME)-v$(HUMAN_VERSION)
+
+%include-3%
+# see https://rt.cpan.org/Ticket/Display.html?id=93241
+COMPONENT_PREP_ACTION =+ rm -f $(SOURCE_DIR)/t/90-author-pod-validation.t
+GENERATE_EXTRA_CMD += | grep -v spfd
```
mtelka: Please replace `grep` by `$(GNU_GREP)`.

mtelka (changes requested):

fritzkink:
Are there any further comments/suggerstions?

components/perl/Mail-SPF/Makefile
```diff
@@ -0,0 +1,48 @@
...
+#
+
+BUILD_STYLE = modulebuild
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_PERL_MODULE =		Mail::SPF
+HUMAN_VERSION =			2.9.0
+COMPONENT_REVISION =		6
+COMPONENT_SUMMARY =		Mail::SPF - An object-oriented implementation of Sender Policy Framework
```
AndWac, line 23: As we don't have any release of this package yet the COMPONENT_REVISION should be 0 or unset.

fritzkink: Sorry overseen. Fixed now.

AndWac (approved):

mtelka:
> Are there any further comments/suggerstions?

You didn't respond yet to my comment https://github.com/OpenIndiana/oi-userland/pull/9358#discussion_r986140941.

mtelka:
@fritzkink, please fix the following problem.  Thanks.

```
# pkg install -v library/perl-5/mail-spf
Creating Plan (Checking for conflicting actions): /           Packages to install:  3
       Create boot environment: No
          Rebuild boot archive: No

Changed packages:
openindiana.org
  library/perl-5/mail-spf
    None -> 2.9.0-2022.0.0.0
  library/perl-5/mail-spf-534
    None -> 2.9.0-2022.0.0.0
  library/perl-5/mail-spf-536
    None -> 2.9.0-2022.0.0.0

pkg install: The following packages deliver conflicting action types to usr/bin/spfquery:

  file:
    pkg://openindiana.org/library/libspf2@1.2.11,5.11-2022.0.0.0:20221001T152656Z
  link:
    pkg://openindiana.org/library/perl-5/mail-spf-534@2.9.0,5.11-2022.0.0.0:20221030T185222Z
    pkg://openindiana.org/library/perl-5/mail-spf-536@2.9.0,5.11-2022.0.0.0:20221030T185225Z

#
```

fritzkink:
thank you for catching this conflict. New pull request uploaded.

PR #9363, opened 2022-09-27
smartmontools: update-smart-drivedb is still full of bashisms; fix gpg

mtelka:
Sorry for one more fix in the same area.  I had to test harder before.


AndWac:
This fails now:
ency '
    depend type=require fmri=__TBD pkg.debug.depend.file=bash \
        pkg.debug.depend.path=usr/bin \
        pkg.debug.depend.reason=usr/sbin/update-smart-drivedb \
        pkg.debug.depend.type=script'.
gmake: *** [/jenkins/jobs/oi-userland/workspace/make-rules/ips.mk:523: /jenkins/jobs/oi-userland/workspace/components/sysutils/smart

PR #9368, opened 2022-09-27
smartmontools: update REQUIRED_PACKAGES

mtelka:


components/sysutils/smartmontools/Makefile
```diff
@@ -27,7 +27,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= smartmontools
 COMPONENT_VERSION= 7.3
-COMPONENT_REVISION =	2
+COMPONENT_REVISION =	3
 COMPONENT_SUMMARY= smartmontools contains utilities that control and monitor storage devices
```
AndWac, line 31: You don't need to change COMPONENT_REVISION again because the last change hasn't been published.

mtelka: > You don't need to change COMPONENT_REVISION again because the last change hasn't been published.

Yes, I know, but it is harmless (and safe) to do so.

PR #9370, opened 2022-09-27
ipmitool: Update to 1.8.19

ghost:
Update the ipmitool package to the latest version 1.8.19.
The version that we currently offer in our repository is listed as vulnerable at Repology.


components/sysutils/ipmitool/Makefile
```diff
@@ -13,31 +13,30 @@
 # Copyright 2017 Gary Mills
 # Copyright 2016 Alexander Pyhalov
 # Copyright 2019 Michal Nowak
+# Copyright 2022 Niklas Poslovski
 #
 
 PREFERRED_BITS=		64
```
AndWac: PREFERRED_BITS is not necessary if the package is only being build in one variant.

ghost: Should I add BUILD_BITS instead or is 64bit default now so that I can completely remove the line?

AndWac: BUILD_BITS=64 is the new default.

PR #9372, opened 2022-09-27
gnutls: Update to 3.7.7

ghost:
Update the gnutls package to the latest version 3.7.7.


AndWac:
What about backwards compatibility? Have you tested the dependend packages? Do they need to be rebuild?

ghost:
I tested three of the packages which depend on it and they all can still successfully connect to SSL-encrypted networks.
I also looked into the commit history and it seems there were never rebuilds after gnutls updates.

PR #9380, opened 2022-09-28
Rebuild for json-c

ghost:
Rebuild the packages libmypaint and clamav to be compatible with the updated json-c version.
Both packages are the newest version already.


AndWac:
Sorry, I merged bind too early. You'll need to add it here...

PR #9387, opened 2022-09-28
Obsolete library/python/mkdocs{,-27}

mtelka:
None package depends on `mkdocs` and `mkdocs-27` so we can safely obsolete them. Please note python 2.7 is no longer supported for almost three years now.


alarcher:
@mtelka Same here, this components was added to generate our own documentation. Is it so that we do not want to allow building the documentation on our own system?

mtelka:
> @mtelka Same here, this components was added to generate our own documentation. Is it so that we do not want to allow building the documentation on our own system?

Do you have a pointer where this is used to generate our own documentation? I do not see any occurrence of mkdocs in the oi-userland that would suggest it is needed/used. Thanks.

PR #9392, opened 2022-09-28
Obsolete library/python/cython{,-27,-35}

mtelka:
No package depends on `cython` and/or `cython-27` and/or `cython-35` so we can safely obsolete them. Please note both python 2.7 and 3.5 are no longer supported for at least two years now.


alarcher:
@mtelka Why not update it instead of obsoleting. The project is active and is used by Scipy. Could removing it all from the tree add and extra cost and possibly cause reintroduction under another name?

mtelka:
> @mtelka Why not update it instead of obsoleting. The project is active and is used by Scipy. Could removing it all from the tree add and extra cost and possibly cause reintroduction under another name?

It is just simpler to obsolete than to update.  And since there was no known consumer, then the obsoletion was the obvious option chosen.  BTW, what is Scipy?  I see no reference to it in oi-userland.  Is it [scipy.org](https://scipy.org/)?  How is it related to oi-userland?

alarcher:
> > @mtelka Why not update it instead of obsoleting. The project is active and is used by Scipy. Could removing it all from the tree add and extra cost and possibly cause reintroduction under another name?
> 
> It is just simpler to obsolete than to update. And since there was no known consumer, then the obsoletion was the obvious option chosen. BTW, what is Scipy? I see no reference to it in oi-userland. Is it [scipy.org](https://scipy.org/)? How is it related to oi-userland?

But it is harder to readd tha update :) based on your list of already obsoleted components we should prioritise readding components that have no internal consumer but are consumed. I know it is questionable to keep the status quo but giving a chance not to break promises before removal (say with a deadline) could be a compromise.

mtelka:
https://github.com/OpenIndiana/oi-userland/pull/10459

PR #9402, opened 2022-09-28
Obsolete library/python/mkdocs-bootstrap{,-27}

mtelka:
No package depends on `mkdocs-bootstrap` and/or `mkdocs-bootstrap-27` so we can safely obsolete them. Please note python 2.7 is no longer supported for almost three years now.


alarcher:
@mtelka @AndWac Is it still not the package used for generating our own documentation?

mtelka:
> @mtelka @AndWac Is it still not the package used for generating our own documentation?

Do you have a pointer where this is used to generate our own documentation?  I do not see any occurrence of `mkdocs` in the oi-userland that would suggest it is needed/used.  Thanks.

mtelka:
`mkdocs-bootstrap` will be added back: #9488

AndWac:
Sorry for being late in this discussion. Alas we are too few to establish a better process around merging PRs (and steering OI in certain directions) than having me as more or less the only one merging PRs. While we have several people who can merge PRs most of them seem to be busy with other things to care on a regular basis. We even have very few people creating PRs at all. And so I am happy for every PR that is being created for OI, especially in areas where I am still unexperienced. Sometimes not all requirements can be met (or are known to me). Nevertheless I want to keep OI development active and take the risk of introducing problems. The alternative seem to be worse to me,
The python eco system is a mess and having 4 versions of python under active support is a nightmare. So I was happy to merge the PRs obsoleting python 2.7 and 3.5. I would even prefer to also get rid of python 3.7 just to be able to concentrate on a single version. I count this as the many drawbacks of the rolling release model we are using. I wish we would have two versions of OI, eg. stable and development in order to suit more people’s interests.

Regards,
Andreas

PR #9408, opened 2022-09-28
brltty: Update to 6.5

ghost:
Update the brltty package to the latest version 6.5.
A rebuild is required for the liblouis update anyway so I took the chance :)


AndWac:
You should always create a separate PR for rebuilding dependend packages and refer to the corresponding PR when you create a PR for a package update. Furthermore it is recommended to write some information about what and how you have tested an update.

ghost:
brltty depends on liblouis so it needs a rebuild after merging #9407 but instead of simply doing a rebuild, I updated this package as well, so you only need to merge this one after #9407 has been built on the build server.
I checked that it compiles successfully and that I can publish a package for it.
There are no tests available for this package, so I couldn't test it further.

PR #9411, opened 2022-09-28
libpcap: Update to 1.10.1

ghost:
Update the libpcap package to the latest version 1.10.1.

### DIFF
--- a/components/library/libpcap/libpcap.p5m
+++ b/components/library/libpcap/libpcap.p5m
@@ -11,18 +11,20 @@
 
 #
 # Copyright 2016 Adam Stevko
+# Copyright 2022 Niklas Poslovski
 #
 
-set name=pkg.fmri value=pkg:/system/library/libpcap@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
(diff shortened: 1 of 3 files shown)


AndWac:
Have you checked that all dependent packages are 64 bit only?

ghost:
I messed up the FMRI first and therefore couldn't see any package that depends on it.
I'll have to re-check everything now,maybe some of them even need to be rebuilt or have newer versions available.
I'll also check if they're all 64bit-only already...

ghost:
There are quite a lot of dependents and some of them are still available in 32bit.
Maybe some of them can be upgraded to 64bit-only later, but for now it's simpler to keep the 32bit compatibility of libpcap.
At least everything still works without rebuilding.

components/library/libpcap/libpcap.p5m
```diff
@@ -11,40 +11,49 @@
...
+file path=usr/include/pcap/socket.h
 file path=usr/include/pcap/usb.h
 file path=usr/include/pcap/vlan.h
+file path=usr/lib/$(MACH64)/libpcap.a
 link path=usr/lib/$(MACH64)/libpcap.so target=libpcap.so.1
 file path=usr/lib/$(MACH64)/libpcap.so.$(COMPONENT_VERSION)
 link path=usr/lib/$(MACH64)/libpcap.so.1 target=libpcap.so.$(COMPONENT_VERSION)
+file path=usr/lib/$(MACH64)/pkgconfig/libpcap.pc
+file path=usr/lib/libpcap.a
 link path=usr/lib/libpcap.so target=libpcap.so.1
```
AndWac, line 51: Here we should drop the static libs.

PR #9434, opened 2022-09-29
libspf2: Update to 1.2.11

ghost:
Update the libspf2 package to the latest version 1.2.11.
I tested that the package libmilter which depends on libspf still runs with it, and it does so without needing a rebuild.

### DIFF
--- a/components/library/libspf2/libspf2.p5m
+++ b/components/library/libspf2/libspf2.p5m
@@ -11,11 +11,12 @@
 
 #
 # Copyright 2022 Friedrich Kink
+# Copyright 2022 Niklas Poslovski
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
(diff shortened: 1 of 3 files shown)


components/library/libspf2/libspf2.p5m
```diff
@@ -45,6 +46,7 @@ file path=usr/include/spf2/spf_record.h
 file path=usr/include/spf2/spf_request.h
 file path=usr/include/spf2/spf_response.h
 file path=usr/include/spf2/spf_server.h
+file path=usr/lib/$(MACH64)/libspf2.a
 link path=usr/lib/$(MACH64)/libspf2.so target=libspf2.so.2.1.0
```
AndWac: We don't want to deliver static libraries. We only deliver them when no alternative is available.

mtelka: @nipos please set `DROP_STATIC_LIBRARIES` to `yes` in the `Makefile` (before the `common.mk` include) so the static library is removed from `sample-manifest.p5m` too.  Thanks.

PR #9459, opened 2022-10-02
openblas: Update to 0.3.21

ghost:
Update the openblas package to the latest version 0.3.21.
OpenBLAS doesn't offer tests but I built the latest arpack version against the openblas update and its tests are successful.

### DIFF
--- a/components/scientific/openblas/manifests/sample-manifest.p5m
+++ b/components/scientific/openblas/manifests/sample-manifest.p5m
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2017 <contributor>
+# Copyright 2022 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
 set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
@@ -24,6 +25,7 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 file path=usr/include/openblas/cblas.h
 file path=usr/include/openblas/f77blas.h
+file path=usr/include/openblas/lapack.h
 file path=usr/include/openblas/lapacke.h
 file path=usr/include/openblas/lapacke_config.h
 file path=usr/include/openblas/lapacke_mangling.h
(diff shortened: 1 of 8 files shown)


components/scientific/openblas/manifests/sample-manifest.p5m
```diff
@@ -22,8 +23,28 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+file path=opt/OpenBLAS/include/openblas/cblas.h
+file path=opt/OpenBLAS/include/openblas/f77blas.h
```
AndWac: There is something wrong here: We don't deliver files to /opt!

ghost: It looks like the new version tries to install the same files twice - In /opt and in their correct location.
I already removed the /opt entries from openblas.p5m and thought that should be sufficient.
I have no idea how I should stop it from trying to install to /opt.
When I remove the PREFIX from the install args, it installs *only* to /opt and in the current form it wants to install to both.

AndWac: It is sufficient. It's just irritating that sample-manifest contains theses files. Something seems to be wrong. Maybe you can try to find the reason for it and fix it.

ghost: I moved DROP_STATIC_LIBRARIES a few lines up and build everything again, now the /opt entries are gone.
I can't really imagine that this is the reason for the difference (maybe it somehow didn't correctly clean the prototype directory before or something like that) but at least it's correct now.

alarcher:
Numerical libraries are usually also shipped as static libs as it is still common in scientific computing projects (like the ones I work on).
However it seems that there is an incentive here to drop them.
At least whenever such decision is made it is good to check what other distributions do and be consistent with what people expect.
There is no « one rule fits all » especially in academia and static libs were provided for a good reason.
I let you make the decision based on your findings.

alarcher:
@nipos Thanks for doing this work, very much appreciated :)

AndWac:
Dropping static libraries is a rule with possible exceptions. I didn't know that it is common in scientific projects to deliver static libs. So, we have an exception here and thus, we should deliver the static lib.

mtelka:
> Dropping static libraries is a rule with possible exceptions. I didn't know that it is common in scientific projects to deliver static libs. So, we have an exception here and thus, we should deliver the static lib.

Then something like `DROP_STATIC_LIBRARIES = no` should be added to the `Makefile` with short comment explaining why we do so.

ghost:
I didn't know that static libs should be kept for mathematical libraries.
I have added them to the manifest again.

PR #9460, opened 2022-10-02
arpack: Update to 3.8.0

ghost:
Update the arpack package to the latest version 3.8.0.
Arpack depends on OpenBLAS and needs to be rebuilt after merging #9459 anyway so I took the chance to update this one too.
I successfully built the package and ran the tests.


AndWac:
Please re-add the static lib here, too.

ghost:
Since I switched to building with CMake, it doesn't build the static lib by default.
I tried to enable it with some CMake Options, but it looks like I can only choose between either the shared lib or the static lib, but not both.
I could try going back to the old configure and make build style, but that didn't play nice with the updated version and CMake was the easier solution.
Please tell me if it's okay not to have the static lib in that case (I still have a little fix then, don't merge yet) or if I should try finding a solution for the configure build style.

AndWac:
If it's not possible to provide a static lib then leave it like it is. But I want alarcher's opinion as a possible consumer.

alarcher:
@nipos Thanks for the information and for your efforts, I can fix it later. Go ahead and merge :)

ghost:
I have uploaded my fix mentioned above now.
It replaces the old configure options with cmake options so that at least ICB is enabled again.
The static lib is still missing however.

PR #9466, opened 2022-10-02
rrdtool - rebuild for standard perl (5.36)

fritzkink:


components/image/rrdtool/test/results-all.master
```diff
@@ -6,8 +6,8 @@ FAIL: modify5
 FAIL: tune1
 FAIL: tune2
 PASS: graph1
-PASS: graph2
-PASS: rpn1
+FAIL: graph2
+FAIL: rpn1
```
mtelka: Why these new failures?  Please root cause.  Maybe there is some package dependency (REQUIRED_PACKAGE) missing?

mtelka (changes requested):

components/image/rrdtool/test/results-all.master
fritzkink: no. it is a problem with the current pango version (meaning will happen with the current rrdtool version, too):

(gdb) bt
#0  0x00007fff7d39259b in pango_font_get_scale_factors () at /usr/lib/64/libpango-1.0.so.0
#1  0x00007fff7d3a486c in pango_layout_line_get_extents_and_height () at /usr/lib/64/libpango-1.0.so.0
#2  0x00007fff7d3aac46 in pango_layout_get_extents () at /usr/lib/64/libpango-1.0.so.0
#3  0x00007fff7d3ab29c in pango_layout_get_size () at /usr/lib/64/libpango-1.0.so.0
#4  0x00007fff7d3a834d in pango_layout_check_lines.part () at /usr/lib/64/libpango-1.0.so.0
#5  0x00007fff7d3aaed4 in pango_layout_get_extents () at /usr/lib/64/libpango-1.0.so.0
#6  0x00007fff7d3ab22a in pango_layout_get_pixel_extents () at /usr/lib/64/libpango-1.0.so.0
#7  0x00007fff7d201910 in gfx_get_text_width () at /usr/src/oi-userland/components/image/rrdtool/build/amd64/src/.libs/librrd.so.8
#8  0x00007fff7d1f6643 in graph_size_location () at /usr/src/oi-userland/components/image/rrdtool/build/amd64/src/.libs/librrd.so.8
#9  0x00007fff7d1f6d21 in graph_paint_timestring () at /usr/src/oi-userland/components/image/rrdtool/build/amd64/src/.libs/librrd.so.8
#10 0x00007fff7d1f87de in rrd_graph_v () at /usr/src/oi-userland/components/image/rrdtool/build/amd64/src/.libs/librrd.so.8
#11 0x0000000000403804 in HandleInputLine ()
#12 0x0000000000403efb in main ()
(gdb)

mtelka: > no. it is a problem with the current pango version (meaning will happen with the current rrdtool version, too):

Does it mean that with pango 1.50.7 (the pango version available at time of rrdtool update to 1.8.0) the test passed and now with pango 1.50.10 the test fails?  If so, and if the problem is really in pango, then I believe pango should be fixed.

BTW, how the stack dump proves the issue is in pango?

Thanks.

AndWac: It might have to do with the fact that pango-1.44 dropped the support for old bitmap fonts (PCF, BDF).

AndWac: I have checked the old rrdtool version on tn the build server. Both tests pass with the new pango.

fritzkink: I've just repeated the build with the original settings and got the same additional failing tests like with perl 5.36. This makes sense as perl is not involved at all for these tests and nothing else changed. So what happens on the build server? Are these test failing, too, or is it only on my server?

AndWac: On the build server the tests results are identical to the expected results in test/results-64.master. In other words: they pass.

fritzkink: Hmm, so it means they are only failing on my system?

AndWac: We only have two data points yet ;-)
Do you have another machine/installation you can run the tests on?

fritzkink: No, unfortunately not. Can you send me the complete log please. Maybe for whatever reason a compiler switch is different on my system (or something else).

mtelka: @fritzkink I would expect the list of installed packages differs.  On jenkins build server there are almost all packages installed, which is probably not the case on your machine.

PR #9471, opened 2022-10-03
python/importlib_metadata: rename from importlib-metadata; update to …

mtelka:
…4.10.1; rebuild to get packages for python 2.7 and 3.5 obsoleted


AndWac:
Why did you change the name? According to https://pypi.org/project/importlib-metadata/ the hyphen was correct.

components/python/importlib_metadata/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		importlib_metadata
+COMPONENT_VERSION =		4.10.1
+COMPONENT_SUMMARY =		importlib_metadata - Read metadata from Python packages
+COMPONENT_PROJECT_URL =		https://github.com/python/importlib_metadata
+COMPONENT_ARCHIVE_URL =		\
+	https://files.pythonhosted.org/packages/f8/41/8ffb059708359ea14a3ec74a99a2bf0cd44a0c983a0c480d9eb7a69438bb/importlib_metadata-4.10.1.tar.gz
+COMPONENT_ARCHIVE_HASH =	\
```
AndWac, line 27: With the correct COMPONENT_ARCHIVE we can use COMPONENT_ARCHIVE_URL=  $(call pypi_url) for most python packages.

mtelka: Yes, I know. It is also the default now: https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/setup.py-defaults.mk#L22

The `$(call pypi_url)` expands to something like `pypi:///importlib_metadata==4.10.1` which is something recognized by `userland-fetch`, but otherwise I doubt it is generally unknown.  The problem is that the `COMPONENT_ARCHIVE_URL` value could appear in man pages (see any man page generated from oi-userland, for example `man gtar`) and if we put something like `pypi:///importlib_metadata==4.10.1` there we could confuse users.  I would instead prefer to see real download url there.  OTOH, I do not know if we generate any man page from python projects, so maybe this is completely pointless.

Also, the `COMPONENT_ARCHIVE_URL` line in the `Makefile` is automatically generated so to have the full download path there causes zero additional maintenance costs.

Anyway, if you prefer the url in form of `pypi:///...` instead of `https://...` then it would be easy to achieve by modifying the `python-integrate-project` script.

mtelka:
> Why did you change the name? According to https://pypi.org/project/importlib-metadata/ the hyphen was correct.

Unfortunately, the situation is a bit confusing here.  Both `importlib-metadata` and `importlib_metadata` are equally understood by `pypi.org`, see `diff <(curl -s https://pypi.org/pypi/importlib-metadata/json) <(curl -s https://pypi.org/pypi/importlib_metadata/json)`.

When you look at download files at https://pypi.org/project/importlib-metadata/#files, you'll find that tarballs are named `importlib_metadata-VERSION.tar.gz`.  When downloaded and unpacked you'll get top level directory named `importlib_metadata-VERSION`.  Next, the previous `COMPONENT_NAME` was `importlib_metadata`, even everything was in the `components/python/importlib-metadata` directory.  And finally, other packages refer to this one as `importlib_metadata`, for example https://github.com/mkdocs/mkdocs/blob/master/setup.py#L74

Because all of the above, I decided to rename `importlib-metadata` to `importlib_metadata`.

components/python/importlib_metadata/Makefile
AndWac: I haven't seen any package yet with such generated man pages. You are right that this would be irritating. On the other hand using pypi typically doesn't break download url's when a project moves. I don't care which one to use but using pypi looks more convenient.

AndWac:
I understand your reasons but won't such renames break existing code when the package is being used?

mtelka:
> I understand your reasons but won't such renames break existing code when the package is being used?

No.  Old packages (those that require `importlib-metadata` now) will cause the renamed `importlib-metadata` will stay installed.  Once all such packages are rebuilt (and possibly fixed), they will start to require new `importlib_metadata`.  After that the renamed `importlib-metadata` will be automatically uninstalled.

components/python/importlib_metadata/Makefile
mtelka: Okay, that makes sense.  I'll update the `python-integrate-project` script later to move to `pypi:///...` (or `$(call pypi_url)`, to be precise) for future python integrations/rebuilds.  And in a case we will find `pypi:///...` in a man page we will solve that when needed.

PR #9472, opened 2022-10-03
procmail: Update 3.24

ghost:
Update the procmmail package to the latest version 3.24.
I've seen that we've patches most (or all?) vulnerabilities bu I think updating to the proper release is the cleaner solution and removes the need for all those patches.


components/mail/procmail/Makefile
```diff
@@ -21,25 +21,29 @@
...
-COMPONENT_VERSION=	3.22
-COMPONENT_REVISION=	4
-COMPONENT_PROJECT_URL=	http://www.procmail.org
+COMPONENT_VERSION=	3.24
+#COMPONENT_REVISION=	4
+COMPONENT_PROJECT_URL=	https://github.com/BuGlessRB/procmail
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
-COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE=	v$(COMPONENT_VERSION).tar.gz
 COMPONENT_ARCHIVE_HASH=	\
```
AndWac, line 35: I am unhappy with this COMPONENT_ARCHIVE because it will end in the userland archive folder with this name. You should try to keep the archive's file name intact.

PR #9474, opened 2022-10-03
tcpdump: Update to 4.99.1

ghost:
Update the tcpdump package to the latest version 4.99.1.
I ran the tests of the package and they were successful.


components/sysutils/tcpdump/Makefile
```diff
@@ -40,18 +41,13 @@ COMPONENT_LICENSE_FILE=	LICENSE
...
+include $(WS_MAKE_RULES)/common.mk
 
 # fix my alignment issues instead of falling over on sparc
 studio_ALIGN.sparc.64=-xmemalign=16i
 CONFIGURE_OPTIONS.sparc += ac_cv___attribute__=no
 CONFIGURE_OPTIONS += $(CONFIGURE_OPTIONS.$(MACH))
 
-# Install binaries to /usr/sbin
-CONFIGURE_SBINDIR.64 = $(CONFIGURE_SBINDIR.32)
-
```
AndWac, line 54: Why do you change the location from sbin to bin?

ghost: This Configure option doesn't have any effect on the updated version.
If it is set, the binaries are installed to /usr/bin anyway.

PR #9495, opened 2022-10-05
desktop/hexchat: fix for perl536

erwinlem:
fix for newer perl.
libnotify & libproxy do not seem to be used. Or do they need to be enabled by configure?


components/desktop/hexchat/Makefile
```diff
@@ -56,18 +56,15 @@ CONFIGURE_OPTIONS += -Dwith-python=python-3.9
 CONFIGURE_OPTIONS += -Dwith-lua=false
 CONFIGURE_ENV += AR="/usr/gnu/bin/ar"
```
AndWac, line 59: Whenever you change a package you'll either need to change/increment COMPONENT_VERSION or in case the package's version didn't change you'll need to introduce/increment COMPONENT_REVISION. Here you'll need to add COMPONENT_REVISION=1 to the Makefile.

components/desktop/hexchat/Makefile
```diff
@@ -56,18 +57,15 @@ CONFIGURE_OPTIONS += -Dwith-python=python-3.9
...
+PERL_REQUIRED_PACKAGES += runtime/perl
 REQUIRED_PACKAGES += library/desktop/gdk-pixbuf
 REQUIRED_PACKAGES += library/desktop/gtk2
 REQUIRED_PACKAGES += library/desktop/pango
 REQUIRED_PACKAGES += library/desktop/xdg/libcanberra
 REQUIRED_PACKAGES += library/glib2
-REQUIRED_PACKAGES += library/libnotify
-REQUIRED_PACKAGES += library/libproxy
 REQUIRED_PACKAGES += library/security/openssl-11
+REQUIRED_PACKAGES += runtime/python-39
```
mtelka: Was this generated using the latest oi-userland source tree?  If not, then please rebase and run `gmake REQUIRED_PACKAGES` again.  Thanks.

mtelka (changes requested):

components/desktop/hexchat/Makefile
erwinlem: You are right, using `PYTHON_REQUIRED_PACKAGES += runtime/python` now. That's even better.

mtelka: Thanks.  You should run `gmake publish` too to update the `pkg5` file.  Sorry for not mentioning this earlier, I forgot.

mtelka (approved):
seems ok

PR #9498, opened 2022-10-05
bass-o-matic: support job servers

dbh201:
Fixes warning messages when doing `gmake -jx clean`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,6 @@ download setup prep build install pre-publish publish validate clean clobber \
 test component-hook unpack patch: $(SUBDIRS)
 
 $(SUBDIRS):	FORCE
-	@+echo "$(TARGET) $@" ; $(GMAKE) -C $@ $(TARGET)
+	@+echo "$(TARGET) $@" ; $(GMAKE) -s -C $@ $(TARGET)
 
 FORCE:
--- a/components/Makefile
+++ b/components/Makefile
@@ -180,7 +180,7 @@ ifdef CANONICAL_REPO
 endif
 
 $(COMPONENT_DIRS.nosetup):	$(WS_LOGS) FORCE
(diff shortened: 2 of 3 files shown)


Toasterson:
Can you elaborate what a Jobserver is usefull for and how it could be tested?

dbh201:
Technical details are available here: https://www.gnu.org/software/make/manual/html_node/POSIX-Jobserver.html 

The jobserver is basically a blocking (synchronous) unix pipe that has one byte for every allowed subprocess/thread. gmake reads one byte from the pipe to ask the jobserver for permission to run something. gmake then writes the same byte to the pipe when it's done.

Jobservers are built into automake and don't really need to be tested. However, when you invoke `subprocess.Popen()` in python, from within a Makefile target, some of the jobserver information does not get passed along. 

You can see this by running `gmake -jx clean` on COMPONENT_DIR. You'll find that a bunch of warning messages happen - this is because `bass-o-matic` isn't passing all of the jobserver information to its subprocesses. Thus, it's creating a new jobserver for every subprocess - which causes the warnings and slows things down.

gmake needs both environment variables _and_ file descriptors to be sent to its subprocesses in order for jobservers to work; `Popen()`, by default, does not share any of the python interpreter's file descriptors. This PR sends the necessary file descriptors to the gmake invocation and fixes the problem.

You can do a simple test by running `gmake -jx clean`, where x > 1, in COMPONENT_DIR and seeing if the error messages come up. Using this PR, I can confirm that `gmake -jx clean` results in a jobserver with x tasks running at once, and no error messages. I also added -s to the toplevel Makefile to stop the unnecessary "Entering/Exiting directory" messages.

Toasterson:
Fix tested and work. Thank you for the contribution and explanation, very valuable :)

PR #9500, opened 2022-10-05
components/python/*: remove various obsolete Makefile variables

dbh201:


mtelka:
Please do not do this.  It will cause rebuild and republish of all touched components and I'm afraid many will fail (did you tried to build them all?).  I'm going through all python components to refresh, rebuild, update, obsolete, etc. them.  Unfortunately, this will take some time to do it properly, so please be patient.

Thank you.

dbh201:
I haven't rebuilt any yet. These packages failed to clean due to recursive variables so I've removed the references. They clean properly with these commits, so I'm going to build them now as I'm rebuilding the codebase. Should I put this as draft or close it?

mtelka:
Please close.  All (or almost all) future changes in `components/python/*` will be done using the `python-integrate-project` tool to improve maintainability of these components (to keep them up-to-date, make sure their tests pass, old pythons are properly obsoleted, components are rebuilt for new pythons, etc.).

The aim is to have similar situation as we already have with `components/perl/*` where all perl modules are up-to-date and everything is automatically checked daily (on my build machine) for updates.

Thank you.

alarcher:
The macros specifying default and all versions were in place to ease the transition from one major version to another. Unless there is a compelling reason perhaps it would be good not to remove something that we’ll need to re-add when the Python 3 to Python 4 transition becomes a reality. As a general rule we tried to put in place macros that can survive change in the long run and not just adapt to the menu of the day. Some things are in place as helpers to be used when required not at all times.
Another point is that some Python modules are not supported for all versions, so during transition to a minor version you want the ability to add/drop specific version for a given component.

mtelka:
Yes, @alarcher.  All of this will be handled by the `python-integrate-project` tool.

AndWac:
closing it as requested by mtelka

PR #9529, opened 2022-10-10
Add tox-current-env python project

mtelka:


AndWac:
Are you sure that this isn't just a tool that happens to be implemented in python (like eg. ansible)? Or, do we really need this for different python versions?

mtelka:
This is tox plugin only (it delivers no `/usr/bin/...`).  Yes, currently we do not need other versions of `tox-current-env` than -39 because `tox` is built for python 3.9 only, but to have `tox-current-env-37` too is harmless and (currently) easy to maintain, so I strongly prefer to leave it as it is proposed here.

BTW, this tox plugin is needed to properly implement tox based testing for our python packages.

And, maybe, once I implement the tox based testing, I'll find that `tox-current-env-37` is really needed too because it is possible that we will need `tox` for python 3.7 too.  But I'll know for sure only after the tox based testing is fully implemented.

PR #9533, opened 2022-10-12
Add pytest-perf python project

mtelka:


components/python/pytest-perf/Makefile
```diff
@@ -0,0 +1,32 @@
...
+#   $WS_TOOLS/python-integrate-project pytest-perf
+#
+
+BUILD_STYLE = setup.py
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		pytest-perf
+COMPONENT_VERSION =		0.12.0
+COMPONENT_SUMMARY =		pytest-perf - pytest-perf
```
AndWac, line 23: This summary is not sufficient. What about
COMPONENT_SUMMARY= Run performance tests against the mainline code
from the pypi project page?

mtelka: You are right.  Fixed.

PR #9539, opened 2022-10-12
add pkgconf package. 

vtorri:
pkgconf is a package compiler and linker metadata toolkit

### DIFF
--- a/components/developer/pkgconf/Makefile
+++ b/components/developer/pkgconf/Makefile
@@ -1,0 +1,41 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 Vincent Torri
+#
+BUILD_BITS=64
+BUILD_STYLE=configure
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=           pkgconf
+COMPONENT_VERSION=        1.9.3
+COMPONENT_SUMMARY=        Package compiler and linker metadata toolkit
+COMPONENT_PROJECT_URL=    http://pkgconf.org/
+COMPONENT_SRC=            $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=        $(COMPONENT_SRC).tar.xz
+COMPONENT_ARCHIVE_URL=    https://distfiles.dereferenced.org/pkgconf/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=   sha256:5fb355b487d54fb6d341e4f18d4e2f7e813a6622cf03a9e87affa6a40565699d
+COMPONENT_FMRI=           developer/build/pkgconf
+COMPONENT_CLASSIFICATION= Development/System
+COMPONENT_LICENSE=        ISC
+
+include $(WS_MAKE_RULES)/common.mk
+
+# do not build static library
+CONFIGURE_OPTIONS += --disable-static
+
+COMPONENT_BUILD_ENV += CC=$(CC)
+COMPONENT_BUILD_ARGS += CC=$(CC)
+COMPONENT_INSTALL_ARGS += PREFIX=/usr DESTDIR=$(PROTO_DIR)
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += system/library
--- a/components/developer/pkgconf/Makefile
+++ b/components/developer/pkgconf/Makefile
@@ -28,6 +28,8 @@ COMPONENT_FMRI=           developer/build/pkgconf
 COMPONENT_CLASSIFICATION= Development/System
 COMPONENT_LICENSE=        ISC
 
+TEST_TARGET = $(NO_TESTS)
+
 include $(WS_MAKE_RULES)/common.mk
 
 # do not build static library
--- a/components/developer/pkgconf/pkgconf.p5m
+++ b/components/developer/pkgconf/pkgconf.p5m
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2022 <contributor>
+# Copyright 2022 Vincent Torri
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
--- a/components/developer/pkgconf/pkgconf.p5m
+++ b/components/developer/pkgconf/pkgconf.p5m
@@ -1,0 +1,43 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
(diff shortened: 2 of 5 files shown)


vtorri:
@alarcher @Toasterson here is the PR :-)

components/developer/pkgconf/pkgconf.p5m
```diff
@@ -0,0 +1,43 @@
...
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2022 <contributor>
+#
```
AndWac, line 14: You should replace <contributor> with your name here.

vtorri:
@AndWac done

vtorri:
btw, is it preferable  to use configure, or meson ?

components/developer/pkgconf/Makefile
```diff
@@ -0,0 +1,41 @@
...
+COMPONENT_SUMMARY=        Package compiler and linker metadata toolkit
+COMPONENT_PROJECT_URL=    http://pkgconf.org/
+COMPONENT_SRC=            $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=        $(COMPONENT_SRC).tar.xz
+COMPONENT_ARCHIVE_URL=    https://distfiles.dereferenced.org/pkgconf/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=   sha256:5fb355b487d54fb6d341e4f18d4e2f7e813a6622cf03a9e87affa6a40565699d
+COMPONENT_FMRI=           developer/build/pkgconf
+COMPONENT_CLASSIFICATION= Development/System
+COMPONENT_LICENSE=        ISC
+
```
AndWac, line 30: We are missing the kyua test framework. Thus, you should disable tests here by adding:
TEST_TARGET = $(NO_TESTS)

AndWac:
Regarding meson vs. configure: We support both and have no preference. Use what works best.

vtorri:
@AndWac should I add the kuya test framework ?

AndWac:
I am not sure how common this test framework is. And you should make sure that there aren't even more dependencies. If you'll add it, create a separate PR for it to make it possible to merge it first.

vtorri:
@AndWac i don't think it is common but it is packaged by at least ubuntu, fedora, freebsd, openbsd and mac os x. I can add it if you want

AndWac:
That depends on whether it is sufficient and if it works properly (that is, how good the test results are you can generate with it).

vtorri:
i'll try to install it and test pkgconf with it in a temporary directory. I'll report the results here

vtorri:
kyua framework requires another lib (lutok) and postgresql. I compiled and installed lutok, run make check in pkgconf, fixed a bunch of problem in a shell script, then the kyua program crashed. So let's disable tests...

mtelka:
@vtorri, the `developer/build/pkgconf` package is not installable:
```
# pkg install -q developer/build/pkgconf

pkg install: The following packages all deliver file actions to usr/share/aclocal/pkg.m4:

  pkg://openindiana.org/developer/build/pkg-config@0.29.2,5.11-2022.0.0.4:20220718T153400Z
  pkg://openindiana.org/developer/build/pkgconf@1.9.3,5.11-2022.0.0.0:20221016T092104Z

#
```
Please fix it.  Thanks.

vtorri:
@mtelka do you have an advice to fix this ?

alarcher:
Debian uses a script to rename the conflicting files:

https://sources.debian.org/src/pkgconf/1.8.0-6/debian/pkgconf.preinst/

Maybe it is just a matter of renaming the files in the manifest to match their strategy, although to be honest the choice of "real" suffix is not very clear... Do not forget to bump the component revision in the PR providing the fix.

Toasterson:
In ips land that would be a mediated symlink or better since pkg-config and pkgconfig are drop in replacements we should not allow them to be installable together but instead adapt dependencies (build-essential) to allow installataion of either but defaulting to pkg-config. I would prefer that rather than trying to make two competing packages work together. Both softwares are not intended to be installed together. Going against that original developer decision is a bad idea.

vtorri:
@Toasterson i agree. but what should I do, then ?

vtorri:
note that these days, pkg-config code didn't change a lot since 2019 (2 commits this year) and previous ones were in 2017. pkgconf replaced pkg-config in most Unix distributions. Also, pkgconf does not requires glib to be compiled, contrary to pkg-config.

mtelka:
If they are really mutually exclusive and drop-in replacements of each other then the conflict should be solved via mediators.  We do have many precedents doing so.

But, I'm afraid they are not drop-in replacements of each other because the only overlapping file is `usr/share/aclocal/pkg.m4`.  Content of `/usr/bin/` is completely different.  OTOH, this does not prevent to fix the conflict using mediators.

In any case, we need to make both packages installable together (and also usable together if building of some other component will require working `pkgconf`), otherwise we could have problems on build machines where we need to install (almost) all packages to be able to build everything.  For example, how would we build `component A` that requires `pkg-config` and `component B` requiring `pkgconf` using single `gmake publish` in top level directory?

Thanks.

PS: Please note that I know there are few other packages that cannot be installed together with others, but we should not add one more to the list.

mtelka:
The other possible option (if we are brave enough and confident we no longer need `pkg-config`) is simply to obsolete `pkg-config` (or rename to `pkgconf`, to be precise).

vtorri:
@mtelka what other distros do are adding a pkg-config simlink to pkgconf. For example:

archlinux: https://github.com/archlinux/svntogit-packages/blob/packages/pkgconf/trunk/PKGBUILD#L52
ubuntu: http://archive.ubuntu.com/ubuntu/pool/universe/p/pkgconf/pkgconf_1.6.3-5.debian.tar.xz (the link file in this archive)
fedora: https://fedoraproject.org/wiki/Changes/pkgconf_as_system_pkg-config_implementation
freebsd: https://cgit.freebsd.org/ports/tree/devel/pkgconf/Makefile#n31

etc...

mtelka:
> @mtelka what other distros do are adding a pkg-config simlink to pkgconf.

@vtorri, I'm not sure what should I do with this information.  Do you expect some reply from me related to this? (I'm asking just to make sure you do not wait for an reply from me before you start to work to solve the problem.)

vtorri:
@mtelka @alarcher or @Toasterson and yourself are proposing different solutions. I am just a user of OpenIndiana, and I try to help with adding some packages (and fixing some projects upstream when they don't compile on OI). But I have no experience at all about what should be done wrt pkgconf.

So I tell you (yourself and other people) what the other Unix distributions are doing, so that you, the devs, tell me what to do.

mtelka:
@vtorri, I do not mind what solution you implement.  I have only few requirements:
* the `developer/build/pkgconf` package must be installable together with other packages, and
* when the `developer/build/pkgconf` package is installed, it should not break the system.

Sorry, but the ball is on your side to propose a solution.

Provided the above two requirements are met, I'm okay with mediators, I'm okay with obsoleting/renaming `developer/build/pkg-config`, I'm okay with somehow modifying `developer/build/pkgconf` to be installable together with `developer/build/pkg-config`, and, finally, I'm even okay with obsoleting `developer/build/pkgconf`.  So, basically, I'm okay with any solution you propose.

Thanks.

PR #9545, opened 2022-10-13
Updating bash 5.2 to patchlevel2

Toasterson:
This updates the Patchlevel of our bash to 5.2.p2 which also fixes the case where the shell segfaults if given an invalid LC_CTYPE Env var. Tested on the buildsystem. @jclulow @richlowe This is the fix for my issues you helped me diagnose in IRC thanks :)

### DIFF
--- a/components/shell/bash/patches/bash52-001.patchelevel1.patch
+++ b/components/shell/bash/patches/bash52-001.patchelevel1.patch
@@ -1,0 +1,46 @@
+			     BASH PATCH REPORT
+			     =================
+
+Bash-Release:	5.2
+Patch-ID:	bash52-001
+
+Bug-Reported-by:	Emanuele Torre <torreemanuele6@gmail.com>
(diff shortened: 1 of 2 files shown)


components/shell/bash/Makefile
```diff
@@ -31,6 +31,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		bash
 COMPONENT_VERSION=	5.2
+COMPONENT_REVISION=	1
 COMPONENT_FMRI=		shell/bash
```
AndWac: You don't need to set COMPONENT_REVISION here. Later the patch level will be automatically considered when IPS_COMPONENT_VERSION is being set.

Toasterson: Ah, thanks. Updated

AndWac:
Ok, with the removal of COMPONENT_REVISION there were no changes in the Makefile left. Thus, our jenkins didn't publish it. We should have touched anything.

Toasterson:
Yeah was curious if our master build also looks for changed Makefiles only

PR #9547, opened 2022-10-14
python/tox: update to 3.26.0

mtelka:
This also returns back the versioned tox.  Yes, we need it for running tox tests for non-default (currently 3.7) python versions.


components/python/tox/Makefile
```diff
@@ -1,85 +1,37 @@
...
-REQUIRED_PACKAGES += library/python/pytest-mock-39
-# Our pytest-randomly doesn't seem to work
-#REQUIRED_PACKAGES += library/python/pytest-randomly-39
-REQUIRED_PACKAGES += library/python/pytest-reporter-39
-REQUIRED_PACKAGES += library/python/virtualenv-39
-
 # Auto-generated dependencies
-REQUIRED_PACKAGES += library/python/setuptools-39
-REQUIRED_PACKAGES += runtime/python-39
+PYTHON_REQUIRED_PACKAGES += library/python/importlib_metadata
```
AndWac, line 35: What are your plans regarding metadata use from setup.py and setup.cfg? With your changes we will loose these manually added dependencies.
With your changes we now hav 0 tests are being run, formerly we had
 18 failed, 613 passed, 5 skipped, 1 warning

mtelka: I plan to implement fully automatic dependency discovery so the final result is something similar as we do have with perl modules now.  But first I'd like to finish support for tox based testing.

Yes, previously tox used `pytest` (as suggested by `tox.ini`).  This PR switches to `setup.py test` which is currently the only supported automatic testing we do have.  As you properly noticed, tox does not support `setup.py test`, so we ended with `Ran 0 tests` result.  This is interim.  Once I implement (fully automatic) tox based testing this (and many other python projects too) will switch to tox testing and we will get real test results back.

PR #9560, opened 2022-10-15
make-rules;tools/Makefile: disable parallelism

dbh201:
Disabling per-component parallelism as some targets are run out-of-order or multiple times due to ordering and dependency issues. 
Fixes an issue when rebuilding the toplevel directory using -jx, the tools/ directory fails to build.

### DIFF
--- a/components/Makefile
+++ b/components/Makefile
@@ -156,7 +156,7 @@ tools:
 	echo "export PATH WS_TOP" >>$@
 
 $(COMPONENT_DIRS):	$(WS_LOGS) setup FORCE
-	@cd $@ && echo "$(TARGET) $@" && \
+	@+cd $@ && echo "$(TARGET) $@" && \
 	 $(BASS_O_MATIC) --make $(TARGET) $(LOG)
 
 incorporation:
--- a/make-rules/common-lisp.mk
+++ b/make-rules/common-lisp.mk
@@ -25,6 +25,8 @@
 # Since common lisp libraries are all source code, there is no need for
 # BUILD_64 and INSTALL_64.
 
+.NOTPARALLEL:
+
 INSTALL_DIR =	$(PROTOCLDIR)/source/$(COMPONENT_NAME)
 
 $(BUILD_DIR)/%/.built:	$(SOURCE_DIR)/.prep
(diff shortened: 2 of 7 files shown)


AndWac:
I would like to test this. Alas this has to wait until end of next week.

mtelka:
This seems to be needed for some python components, so I'd like to see it integrated, but I have one question:  Will there be a way to enable parallelism on per-component level for some components that are well known to build properly in parallel?  This could make sense for components with very long build time (for example gcc) on machines with a lot of CPUs.

Thanks.

mtelka:
> Will there be a way to enable parallelism on per-component level for some components that are well known to build properly in parallel? This could make sense for components with very long build time (for example gcc) on machines with a lot of CPUs.

Am I right to assume the way to enable simultaneous jobs at a component level is simply to add `-j` to `COMPONENT_BUILD_GMAKE_ARGS`?

dbh201:
Yes. The `.NOTPARALLEL` flag is not inherited for subtasks, so when any target in the component's Makefile runs gmake, that gmake invocation has parallelism enabled again.

dbh201:
> > Will there be a way to enable parallelism on per-component level for some components that are well known to build properly in parallel? This could make sense for components with very long build time (for example gcc) on machines with a lot of CPUs.
> 
> Am I right to assume the way to enable simultaneous jobs at a component level is simply to add `-j` to `COMPONENT_BUILD_GMAKE_ARGS`?

This PR *only* disallows the parallelism in the COMPONENT_DIR Makefile.

The SOURCE_DIR Makefile is responsible for handling parallelism once the build is started. As long as the component's upstream Makefile isn't broken, gmake *should* take as many job slots as it needs from the current jobserver. So if parallelism is supported upstream, it's enabled automatically.

If a SOURCE_DIR Makefile doesn't handle its job slots appropriately, you could put `-j1` in COMPONENT_BUILD_GMAKE_ARGS (which will cause a warning but otherwise work as expected) or patch the SOURCE_DIR Makefile with `.NOTPARALLEL:`.

mtelka:
Makes sense.  Thanks for explanation.

PR #9568, opened 2022-10-17
Switching gdb from python 3.7 to python 3.9

Toasterson:
Note that I had to disable libpython lookup as it broke for 3.9 while it works for 3.7 I did not find out why. I manually added the dependency. The Sample-manifest changed but we do not deliver the pyc files

### DIFF
--- a/components/developer/gdb/Makefile
+++ b/components/developer/gdb/Makefile
@@ -35,7 +35,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		gdb
 COMPONENT_VERSION=	8.0
-# COMPONENT_REVISION=	0
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	GDB: The GNU Project Debugger
 COMPONENT_CLASSIFICATION=Development/System
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
@@ -46,9 +46,6 @@ COMPONENT_ARCHIVE_URL=	https://ftp.gnu.org/gnu/gdb/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL=	https://www.gnu.org/software/gdb/
 
 CONFIGURE_DEFAULT_DIRS=no
-
-# Set python version used by this product
-PYTHON_VERSION=3.7
 
 include $(WS_MAKE_RULES)/common.mk
 
@@ -193,18 +190,15 @@ COMPONENT_POST_INSTALL_ACTION = \
 # altering the results of the debugging experiment.
 ASLR_MODE=$(ASLR_DISABLE)
 
+GENERATE_EXTRA_CMD += | $(GNU_GREP) -v -E '/.*\.pyc$$'
+
 # Auto-generated dependencies
-REQUIRED_PACKAGES += SUNWcs
+PYTHON_REQUIRED_PACKAGES += runtime/python
 REQUIRED_PACKAGES += compress/xz
 REQUIRED_PACKAGES += developer/babeltrace
 REQUIRED_PACKAGES += library/expat
 REQUIRED_PACKAGES += library/ncurses
-REQUIRED_PACKAGES += library/readline
 REQUIRED_PACKAGES += library/zlib
-REQUIRED_PACKAGES += runtime/python-37
 REQUIRED_PACKAGES += shell/ksh93
-REQUIRED_PACKAGES += system/library/g++-7-runtime
-REQUIRED_PACKAGES += system/library/gcc-7-runtime
 REQUIRED_PACKAGES += system/library
 REQUIRED_PACKAGES += system/library/math
-REQUIRED_PACKAGES += text/texinfo
--- a/components/developer/gdb/gdb.p5m
+++ b/components/developer/gdb/gdb.p5m
@@ -21,11 +21,13 @@
 # Copyright (c) 2011, 2017, Oracle and/or its affiliates. All rights reserved.
 # Copyright 2019, Andreas Wacknitz
 # Copyright 2021 Gary Mills
+# Copyright 2022 Till Wegmueller
 #
 
 <transform file path=usr.*/man/.+ -> default mangler.man.stability uncommitted>
 set name=pkg.fmri \
     value=pkg:/developer/debug/gdb@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
 set name=pkg.description \
     value="GDB, the GNU Debugger, is a source-level debugger for Ada, C, C++, Objective-C, Pascal and many other languages.  GDB allows you to see what is going on inside another program while it executes, or what another program was doing at the moment it crashed."
@@ -34,9 +36,14 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
 file path=usr/bin/gdb
 file path=usr/bin/gdbtui mode=0555
+file path=usr/include/ansidecl.h
+file path=usr/include/dis-asm.h
 file path=usr/include/gdb/jit-reader.h
+file path=usr/include/plugin-api.h
+file path=usr/include/symcat.h
 file path=usr/share/gdb/python/gdb/FrameDecorator.py
 file path=usr/share/gdb/python/gdb/FrameIterator.py
 file path=usr/share/gdb/python/gdb/__init__.py
--- a/components/developer/gdb/manifests/sample-manifest.p5m
+++ b/components/developer/gdb/manifests/sample-manifest.p5m
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2021 <contributor>
+# Copyright 2022 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
 set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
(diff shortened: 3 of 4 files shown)


AndWac:
Why is this switch necessary or desirable?

components/developer/gdb/Makefile
```diff
@@ -48,7 +48,7 @@ COMPONENT_PROJECT_URL=	https://www.gnu.org/software/gdb/
 CONFIGURE_DEFAULT_DIRS=no
 
 # Set python version used by this product
-PYTHON_VERSION=3.7
+PYTHON_VERSION=3.9
```
mtelka: Please remove this line. `PYTHON_VERSION` is by default set to `3.9`.

components/developer/gdb/manifests/sample-manifest.p5m
```diff
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2021 <contributor>
+# Copyright 2022 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
```
mtelka, line 17: `pkg.human-version` should be added to `gdb.p5m` too.

components/developer/gdb/gdb.p5m
```diff
@@ -34,7 +37,7 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
-file path=usr/bin/gdb
+file path=usr/bin/gdb pkg.depend.bypass-generate=.*/libpython3.9.so.*
```
mtelka: I'd like to see more details why is this needed.  If it was able to automatically resolve dependency on `libpython3.7m.so.1.0` then it should also work for `libpython3.9.so.1.0` which lives in the same directory.

components/developer/gdb/manifests/sample-manifest.p5m
```diff
@@ -38,24 +39,24 @@ file path=usr/lib/$(MACH64)/libopcodes.a
...
-file path=usr/share/gdb/python/gdb/__pycache__/FrameDecorator.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/FrameIterator.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/__init__.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/frames.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/printing.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/prompt.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/types.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/unwinder.cpython-37.pyc
-file path=usr/share/gdb/python/gdb/__pycache__/xmethod.cpython-37.pyc
+file path=usr/share/gdb/python/gdb/__pycache__/FrameDecorator.cpython-39.pyc
```
mtelka: We now have simple way how to filter out things like these from `sample-manifest.p5m`.  Please look at `components/python/testpath/Makefile` for an example.

mtelka (changes requested):

Toasterson:
> Why is this switch necessary or desirable?

It's mostly desireable to reduce the amount of python versions that we have and gdb is somehow now one of the only few packages that are built against python 3.7. Also it was already a complaint on IRC, so it seems desireable for people to have an up to date python linked against gdb.

mtelka:
> > Why is this switch necessary or desirable?
> 
> It's mostly desireable to reduce the amount of python versions that we have and gdb is somehow now one of the only few packages that are built against python 3.7. Also it was already a complaint on IRC, so it seems desireable for people to have an up to date python linked against gdb.

BTW, the number of python versions is not a problem per se.  The problem with python-3.7 is that it will be EOLed in ca 8 months so it is good to start to rebuild everything against newer python, although it is not very urgent (unless there is some other reason to do the switch ASAP).

Just my €0.02.

components/developer/gdb/Makefile
Toasterson: Done

components/developer/gdb/manifests/sample-manifest.p5m
Toasterson: Done

components/developer/gdb/gdb.p5m
Toasterson: Added some details. Unfortunately I lack the time and experience to hunt down the bug inside pkg but it is definetly a bug that I needed to workaround.

components/developer/gdb/manifests/sample-manifest.p5m
Toasterson: Added thanks for the hint.

Toasterson:
I also fixed some include and texinfo files which where not delivered. In general it was not that important to update gdb but since it had a couple of files missing needed some cleanup et al and somebody poked me on IRC why not.

components/developer/gdb/gdb.p5m
```diff
@@ -34,9 +38,21 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
-file path=usr/bin/gdb
+
+file path=usr/bin/gcore
+# The dependency resolution for python 3.9 does not work as expected. For one the searchpath is different 
+# but even with adapting the searchpath to the correct directories pkg is unable to find the library while ldd
+# and ld.so are able to. This is a bug in pkg so we workaround it for now.
```
mtelka: Could you please share the error message, or log, or whatever showing the problem?  It is very hard to judge without more details.  Thank you.

AndWac:
After the switch I would expect that pkg5 needs also to be updated (eg. by running gmake update-metadata)

components/developer/gdb/gdb.p5m
Toasterson: Not really an error I just get

```
/usr/bin/pkgdepend resolve -e /ws/toasty/oi-userland/components/developer/gdb/build/resolve.deps -m /ws/toasty/oi-userland/components/developer/gdb/build/manifest-i386-gdb.depend /ws/toasty/oi-userland/components/developer/gdb/build/runtime-perl.p5m
/ws/toasty/oi-userland/components/developer/gdb/build/manifest-i386-gdb.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libpython3.9.so.1.0 \
        pkg.debug.depend.reason=usr/bin/gdb pkg.debug.depend.type=elf \
        pkg.debug.depend.path=usr/gnu/lib/amd64 \
gmake: *** [/ws/toasty/oi-userland/make-rules/ips.mk:527: /ws/toasty/oi-userland/components/developer/gdb/build/.resolved-i386] Error 1
```

And can't make it go away by setting pkg's runpath

Toasterson:
> After the switch I would expect that pkg5 needs also to be updated (eg. by running gmake update-metadata)

Yeah.... erm. That command is apparently also broken since the updated metadata still contains python-37...

```json
{
    "dependencies": [
        "SUNWcs",
        "compress/xz",
        "developer/babeltrace",
        "library/expat",
        "library/ncurses",
        "library/readline",
        "library/zlib",
        "runtime/python-37",
        "shell/ksh93",
        "system/library",
        "system/library/g++-7-runtime",
        "system/library/gcc-7-runtime",
        "system/library/math",
        "text/texinfo"
    ],
    "fmris": [
        "developer/debug/gdb"
    ],
    "name": "gdb"
}
```

mtelka:
> > After the switch I would expect that pkg5 needs also to be updated (eg. by running gmake update-metadata)
> 
> Yeah.... erm. That command is apparently also broken since the updated metadata still contains python-37...

It is not broken.  You forgot to run `gmake REQUIRED_PACKAGES`.

components/developer/gdb/gdb.p5m
```diff
@@ -34,9 +38,21 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
...
+file path=usr/bin/gdb pkg.depend.bypass-generate=.*/libpython3.9.so.*
 file path=usr/bin/gdbtui mode=0555
+file path=usr/include/ansidecl.h
+file path=usr/include/bfd.h
+file path=usr/include/bfdlink.h
+file path=usr/include/dis-asm.h
 file path=usr/include/gdb/jit-reader.h
+file path=usr/include/plugin-api.h
+file path=usr/include/symcat.h
+file path=usr/lib/$(MACH64)/charset.alias
```
mtelka: Will this appear in the published package?  It shouldn't.

components/developer/gdb/gdb.p5m
mtelka: Once you confirm you ran `gmake REQUIRED_PACKAGES` and still got python-37 there (instead of python-39) then I'll look at it.

mtelka (changes requested):

components/developer/gdb/gdb.p5m
Toasterson: Ok, removed the .alias file

Toasterson:
Ran `gmake REQUIRED_PACKAGES` and now dependency resolution works... ok learnt something

mtelka (approved):
Now it looks far better :-).  Thank you.

components/developer/gdb/Makefile
```diff
@@ -47,9 +47,6 @@ COMPONENT_PROJECT_URL=	https://www.gnu.org/software/gdb/
 
 CONFIGURE_DEFAULT_DIRS=no
 
-# Set python version used by this product
-PYTHON_VERSION=3.7
-
 include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 51: I recommend to add
PYTHON_VERSIONS= $(PYTHON_VERSION)
for packages that only need or want a single python version. Adding this will eg. change pkg5.

mtelka: Yes, adding that should change the `pkg5` file, but otherwise everything else should be same.  OTOH, for `setup.py` build styles such line would cause shorter build times because only single version will be built instead of all of them.

Maybe we should try to make this default for non-setup.py builds?  And similarly with `PERL_VERSIONS` too?

AndWac: Yes, that would be a nice enhancement.

mtelka: Here it is: PR #9653

Toasterson:
Added PYTHON_VERSIONS= $(PYTHON_VERSION)

mtelka:
> Added PYTHON_VERSIONS= $(PYTHON_VERSION)

Since #9653 is merged the explicit `PYTHON_VERSIONS` is no longer needed.  You just should rebase, remove `PYTHON_VERSIONS`, and re-run all gmake targets to confirm everything is still okay and to regenerate metadata.

Or, we could merge even without that because `PYTHON_VERSIONS` is just not needed, but otherwise harmless.

Toasterson:
Removed PYTHON_VERSIONS= $(PYTHON_VERSION) as requested. Metadata did not change anymore.

components/developer/gdb/gdb.p5m
```diff
@@ -34,9 +35,17 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
-file path=usr/bin/gdb
+
+file path=usr/bin/gcore
```
mtelka: Unfortunately, this will very likely clash with `developer/linker`:
```
# pkg search gcore
INDEX      ACTION   VALUE               PACKAGE
basename   file     usr/bin/i86/gcore   pkg:/developer/linker@0.5.11-2022.0.0.21323
basename   file     usr/bin/amd64/gcore pkg:/developer/linker@0.5.11-2022.0.0.21323
basename   file     usr/bin/i86/gcore   pkg:/developer/linker@0.5.11-2022.0.0.21323
basename   file     usr/bin/amd64/gcore pkg:/developer/linker@0.5.11-2022.0.0.21323
basename   hardlink usr/bin/gcore       pkg:/developer/linker@0.5.11-2022.0.0.21323
#
```

components/developer/gdb/gdb.p5m
```diff
@@ -68,6 +77,7 @@ file path=usr/share/gdb/syscalls/sparc64-solaris.xml variant.arch=sparc
 file path=usr/share/gdb/system-gdbinit/elinos.py pkg.tmp.autopyc=false
 file path=usr/share/gdb/system-gdbinit/wrs-linux.py pkg.tmp.autopyc=false
 file path=usr/share/info/annotate.info
+file path=usr/share/info/bfd.info
```
mtelka: ... and this one with binutils:
```
# pkg search bfd.info
INDEX      ACTION VALUE                   PACKAGE
basename   file   usr/share/info/bfd.info pkg:/developer/gnu-binutils@2.39-2022.0.0.1
#
```

mtelka (changes requested):

components/developer/gdb/gdb.p5m
Toasterson: Well then we can remove it. Which is why it wasn't added before

components/developer/gdb/gdb.p5m
Toasterson: Since this was also not included in the manifest before I removed it again.

components/developer/gdb/gdb.p5m
```diff
@@ -34,9 +36,14 @@ set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
-file path=usr/bin/gdb
+
+file path=usr/bin/gdb
```
mtelka: Please remove trailing space here.

mtelka (changes requested):

components/developer/gdb/gdb.p5m
Toasterson: done

mtelka (approved):

PR #9587, opened 2022-10-20
binutils: disable gnu ld warning about rwx segments

tsoome:
This enables custom gate builds.

### DIFF
--- a/components/developer/binutils/Makefile
+++ b/components/developer/binutils/Makefile
@@ -34,6 +34,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		binutils
 COMPONENT_VERSION=	2.39
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	GNU collection of binary tools like ld, as
 COMPONENT_PROJECT_URL=	https://www.gnu.org/software/binutils/
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
@@ -61,6 +62,12 @@ CONFIGURE_OPTIONS +=	--with-system-zlib
 CONFIGURE_OPTIONS +=    --host=$(GCC_GNU_TRIPLET)
 CONFIGURE_OPTIONS +=    --build=$(GCC_GNU_TRIPLET)
 CONFIGURE_OPTIONS +=    --target=$(GCC_GNU_TRIPLET)
+#
+# This warning will break build of current illumos-gate
+# as warnings will be generated for grub and loader.
+# Once illumos-gate will get appropriate fix, this
+# option should be removed.
+CONFIGURE_OPTIONS +=	--enable-warn-rwx-segments=no
 
 COMPONENT_TEST_ARGS += -k
 


components/developer/binutils/Makefile
```diff
@@ -61,6 +62,7 @@ CONFIGURE_OPTIONS +=	--with-system-zlib
 CONFIGURE_OPTIONS +=    --host=$(GCC_GNU_TRIPLET)
 CONFIGURE_OPTIONS +=    --build=$(GCC_GNU_TRIPLET)
 CONFIGURE_OPTIONS +=    --target=$(GCC_GNU_TRIPLET)
+CONFIGURE_OPTIONS +=	--enable-warn-rwx-segments=no
```
mtelka, line 70: Isn't it possible to fix [bug #15028](https://www.illumos.org/issues/15028) in illumos-gate instead?  I'd prefer that over disabling these warnings for all and forever.

If not, then please add a comment here explaining why is this needed.

Thanks.

mtelka (changes requested):

components/developer/binutils/Makefile
tsoome: > Isn't it possible to fix [bug #15028](https://www.illumos.org/issues/15028) in illumos-gate instead? I'd prefer that over disabling these warnings for all and forever.
> 
> If not, then please add a comment here explaining why is this needed.
> 
> Thanks.

This change does not have to be there forever. Once the root cause is fixed, we can enable this warning. (in ideal world we should not depend on gnu ld at all, but this will take some time).

1. we have rolling updates, once binutils are updated, gate build is broken. I have had to delay build system update for this reason and since I only do have limited set of systems to use, it means I can not test updates as needed. I'm sure I'm not only one there.
2. illumos-gate fix is complicated - proper fix would be to have linker script update, command line option... does it work with older binutils? 
3. it is good to get update packages, but binutils is currently essential to gate build, and updates must consider the impact. You have to value everyones time there.

mtelka: > > Isn't it possible to fix [bug #15028](https://www.illumos.org/issues/15028) in illumos-gate instead? I'd prefer that over disabling these warnings for all and forever.
> > If not, then please add a comment here explaining why is this needed.
> > Thanks.
> 
> This change does not have to be there forever. Once the root cause is fixed, we can enable this warning. (in ideal world we should not depend on gnu ld at all, but this will take some time).

I'm not against this change (see "If not, ..." in my first comment), but, unfortunately, interim solutions (or workarounds) tend to stay with us far longer than expected and I would like to avoid them.  The problem here is: who will remember that we should remove the `--enable-warn-rwx-segments=no` line here once the problem in illumos-gate is fixed without a comment explaining why it was added?

>     1. we have rolling updates, once binutils are updated, gate build is broken. I have had to delay build system update for this reason and since I only do have limited set of systems to use, it means I can not test updates as needed. I'm sure I'm not only one there.

Isn't it possible to use https://github.com/OpenIndiana/oi-userland/commit/44a04e21aa4e7caa77ad62d78fdcd3735b2a92dd as a workaround for you?

>     2. illumos-gate fix is complicated - proper fix would be to have linker script update, command line option... does it work with older binutils?

If by "command line option" you mean `--no-warn-rwx-segments` then the answer is well known: no.
Where is the linker script that would need to be updated to fix the problem properly?

>     3. it is good to get update packages, but binutils is currently essential to gate build, and updates must consider the impact. You have to value everyones time there.

First, I'm not the guy who updated binutils in OI.  Second, it is not possible to value everyones time.  (Unless everyone does not mean everyone). :-)

tsoome: > > > Isn't it possible to fix [bug #15028](https://www.illumos.org/issues/15028) in illumos-gate instead? I'd prefer that over disabling these warnings for all and forever.
> > > If not, then please add a comment here explaining why is this needed.
> > > Thanks.
> > 
> > 
> > This change does not have to be there forever. Once the root cause is fixed, we can enable this warning. (in ideal world we should not depend on gnu ld at all, but this will take some time).
> 
> I'm not against this change (see "If not, ..." in my first comment), but, unfortunately, interim solutions (or workarounds) tend to stay with us far longer than expected and I would like to avoid them. The problem here is: who will remember that we should remove the `--enable-warn-rwx-segments=no` line here once the problem in illumos-gate is fixed without a comment explaining why it was added?

There is an issue for illumos-gate, it can have a note or link to OI issue to remove this workaround. I do not think we have better way to set up reminder.

> 
> > ```
> > 1. we have rolling updates, once binutils are updated, gate build is broken. I have had to delay build system update for this reason and since I only do have limited set of systems to use, it means I can not test updates as needed. I'm sure I'm not only one there.
> > ```
> 
> Isn't it possible to use [44a04e2](https://github.com/OpenIndiana/oi-userland/commit/44a04e21aa4e7caa77ad62d78fdcd3735b2a92dd) as a workaround for you?
> 

I can use it, but it is not just question about me;) 

> > ```
> > 2. illumos-gate fix is complicated - proper fix would be to have linker script update, command line option... does it work with older binutils?
> > ```
> 
> If by "command line option" you mean `--no-warn-rwx-segments` then the answer is well known: no. Where is the linker script that would need to be updated to fix the problem properly?
> 

loader has one  boot/i386/boot.ldscript, but that one is specific for boot2 bits (gptzfsboot, isoboot), bios loader itself has it in loader directory (ldscript.i386). I haven't checked grub (my motivation to touch grub is very low;)

mtelka: > > > > Isn't it possible to fix [bug #15028](https://www.illumos.org/issues/15028) in illumos-gate instead? I'd prefer that over disabling these warnings for all and forever.
> > > > If not, then please add a comment here explaining why is this needed.
> > > > Thanks.
> > > 
> > > 
> > > This change does not have to be there forever. Once the root cause is fixed, we can enable this warning. (in ideal world we should not depend on gnu ld at all, but this will take some time).
> > 
> > 
> > I'm not against this change (see "If not, ..." in my first comment), but, unfortunately, interim solutions (or workarounds) tend to stay with us far longer than expected and I would like to avoid them. The problem here is: who will remember that we should remove the `--enable-warn-rwx-segments=no` line here once the problem in illumos-gate is fixed without a comment explaining why it was added?
> 
> There is an issue for illumos-gate, it can have a note or link to OI issue to remove this workaround. I do not think we have better way to set up reminder.

I believe we do.  In addition to adding a note to the issue, we can add a note here into the `Makefile` too.

mtelka:
What about to include removal of `components/openindiana/illumos-gate/patches/0004-LOAD-segment-with-RWX-permissions.patch` to this PR?  It should no longer be needed once this PR merges.

tsoome:
> What about to include removal of `components/openindiana/illumos-gate/patches/0004-LOAD-segment-with-RWX-permissions.patch` to this PR? It should no longer be needed once this PR merges.

I think, it should be done as next step? bcause updated binutils need to land to build host... But yes, it is not needed then.

mtelka:
> > What about to include removal of `components/openindiana/illumos-gate/patches/0004-LOAD-segment-with-RWX-permissions.patch` to this PR? It should no longer be needed once this PR merges.
> 
> I think, it should be done as next step? bcause updated binutils need to land to build host... But yes, it is not needed then.

It is not necessary to do it as two steps since illumos-gate is rebuilt once a day only, but other components (like binutils) are rebuilt right after the merge.  So once this PR is merged, the binutils package will be immediately rebuilt and installed on the build machine.  Few hours later (assuming this PR is merged at least few hours before the illumos-gate rebuild) the illumos-gate build should complete properly.

In worst case scenario (if this PR is not merged long enough before the illumos-gate build) we could get one illumos-gate build failure, which is both not critical and unlikely.  It is unlikely because when the illumos-gate build is triggered (ca midnight UTC) all people who do PR merges usually sleep :-).

But if you wants to be 100% safe then yes, two separate PRs will achieve that.

tsoome:
> > > What about to include removal of `components/openindiana/illumos-gate/patches/0004-LOAD-segment-with-RWX-permissions.patch` to this PR? It should no longer be needed once this PR merges.
> > 
> > 
> > I think, it should be done as next step? bcause updated binutils need to land to build host... But yes, it is not needed then.
> 
> It is not necessary to do it as two steps since illumos-gate is rebuilt once a day only, but other components (like binutils) are rebuilt right after the merge. So once this PR is merged, the binutils package will be immediately rebuilt and installed on the build machine. Few hours later (assuming this PR is merged at least few hours before the illumos-gate rebuild) the illumos-gate build should complete properly.
> 
> In worst case scenario (if this PR is not merged long enough before the illumos-gate build) we could get one illumos-gate build failure, which is both not critical and unlikely. It is unlikely because when the illumos-gate build is triggered (ca midnight UTC) all people who do PR merges usually sleep :-).
> 
> But if you wants to be 100% safe then yes, two separate PRs will achieve that.

Lets play safe there, it is small change and easy to add.

mtelka:
> Lets play safe there, it is small change and easy to add.

Ok, np :-).

AndWac:
@mtelka: Your worst case scenario is very unlikely as the jenkins has only one build queue and thus the jobs are strongly serialiased. It would only theoretically possible if a merge happens just a few seconds before the illumos-gate build starts.

AndWac:
@tsoome: Please add a comment mtelka requested before with an explanation for disabling the warnings.

tsoome:
> @tsoome: Please add a comment mtelka requested before with an explanation for disabling the warnings.

done.

PR #9596, opened 2022-10-21
barman: update to 3.2.0 and drop O_DIRECTORY patch

AndWac:


AndWac:
Question: Should we remove 01-O_DIRECTORY.patch as illumos supports O_DIRECTORY for some time?

components/python/barman/Makefile
```diff
@@ -20,23 +20,27 @@ BUILD_STYLE=setup.py
...
+COMPONENT_ARCHIVE_HASH=	sha256:2603776a805a5595eb466d493c4529ad8f7581a7266d5e131ee97ac06c0de8e2
 COMPONENT_ARCHIVE_URL=	https://github.com/EnterpriseDB/barman/releases/download/release/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL=	https://pgbarman.org/
 COMPONENT_CLASSIFICATION=	System/Databases
 COMPONENT_FMRI=		database/postgres/barman
 COMPONENT_LICENSE=	GPLv3
 
-PYTHON_VERSIONS=	3.9
-PYTHON_VERSION=		3.9
+PYTHON_VERSIONS= 3.9
```
mtelka: `PYTHON_VERSIONS = $(PYTHON_VERSION)` should do the same without hardcoding one single version.

mtelka:
> Question: Should we remove 01-O_DIRECTORY.patch as illumos supports O_DIRECTORY for some time?

I see no reason to keep it.

components/python/barman/Makefile
AndWac: Done

mtelka (approved):
Looks good.  Thanks.

AndWac:
I moved the package to the database folder because barman is a database tool that happens to be implemented in python.

mtelka:
> I moved the package to the database folder because barman is a database tool that happens to be implemented in python.

This could make future automatic updates of barman harder.  I plan to convert all Python projects to use `python-integrate-project`.  To have them automatically updated (I do so for perl modules now, daily) I'd need a simple way to find such projects.  For perl I just simply try to upgrade everything in `components/perl` every night.  If such python projects are scattered through all directory structure I'd need to scan all components to find those that are automatically upgradable.  It is possible to do so, but it would be slower.  The good side effect of `python-integrate-project`/`perl-integrate-module` is simply "integrate once and forget" and we have everything up-to-date (and fully tested) all the time.

So I'd prefer not to move barman out of `components/python`.

AndWac:
How about checking BUILD_STYLE instead?
We already have packages like ansible which are already outside the python folder. glances would be the next.
The repository's folder structure is loosely structered by fmri's and classifications. As I stated barman is a database tool and not a python library.

AndWac:
Ups, I noticed that ansible is in the python folder. My memory fooled me. But that's a permanent hassle for me. I would expect it in sysutils.

mtelka:
It is not a problem to find such components.  The problem is that `python-integrate-project` expects to have them in `components/python` (this path does not imply library!).  So to support `python-integrate-project` for non-`components/python` path the tool would need to be modified (just to support any location in the `components` directory).  As I said, it is possible, but I'd prefer not to do so.  Everything managed by `python-integrate-project`/`perl-integrate-module` is just _integrate once and forget_ so you would never look at it again (unless something breaks, and you would need to manually fix it, ofc).

But if you have strong preference to move barman out of `components/python` then I have no better argument against doing so.

OTOH, you could move barman out of `components/python` anytime, but moving it out now, and then back in a month just to get it automatically updated (remember: _integrate once and forget_) and never touched again would create two unnecessary moves. :-)

AndWac:
Ok, I'll move it back. But please write some notes (doc folder) about this because we need to remember the reasons for having such packages in an unexpected folder.

mtelka:
> ... having such packages in an unexpected folder.

:-).  It is not unexpected, because it appeared there somehow, so somebody expected to have it there when he put it there :-).

But yes, now and for somebody it could look unexpected.

PR #9599, opened 2022-10-22
library/zlib: update to 1.2.13

dbh201:


components/library/zlib/Makefile
```diff
@@ -28,18 +28,17 @@
 BUILD_BITS= 32_and_64
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 31: Please add
USE_COMMON_TEST_MASTER= no
here and update the test results in the test folder. You can find the new results in build/test/{i86|amd64}/results-{32|64}.snapshot after running gmake test.

PR #9252, opened 2022-09-16
userland-fetch: improvements for downloading and hashing

dbh201:
HASH_DIR support: search for an envvar HASH_DIR, a hashes/ directory in the directory of --file, or a hashes/ in the current directory. If found, look inside the files in that directory for checksums in the format outputted by sha256sum. Checksum files will be used if --hash is not provided.

partial downloads: any cancelled download leaves a .part file, which can be continued later if the server supports it or will be overwritten if not.

retrying: if there is a retryable network error, userland-fetch will try to continue the download once or twice before quitting.

security tests: fail on unsecured, unhashed download with no signature.

Removed deprecated splithost and splittype commands.

PR #9600, opened 2022-10-22
assorted download fixes

dbh201:
These commits fix a few errors I found when testing the new userland-fetch (#9252).


AndWac:
Alexander taught me that with every change we need to set/increment the COMPONENT_REVISION. I am not convinced that this is true because the build times are part of the fmri's. I can't remember the exact reason but IIRC some tools/scripts could be erratic if we don't have the needed unambiguousness.
Thus, I am not sure whether I should merge this as is or whether I should ask you to set/increment all COMPONENT_REVISIONS here.

mtelka:
> Alexander taught me that with every change we need to set/increment the COMPONENT_REVISION. I am not convinced that this is true because the build times are part of the fmri's. I can't remember the exact reason but IIRC some tools/scripts could be erratic if we don't have the needed unambiguousness. Thus, I am not sure whether I should merge this as is or whether I should ask you to set/increment all COMPONENT_REVISIONS here.

The rule of thumb is: every (expected) rebuild of a package must change significant part of package's version.  This is everything between `@` and `:` (IOW, the full version minus timestamp).  The change in timestamp only is considered just a republish of same package.  There should be no functional (or other) changes if only timestamp changes and both such packages must be otherwise equivalent.

In an ideal world if only download url changes we should get the same package as before (so change in timestamp only would be okay), but that's not always true for many reasons.  For example, no package lives in vacuum so with changing other parts of ecosystem (build framework changes, other packages updated, etc) the package could change with every rebuild.

To make it predictable and simpler and safer, etc. we should make sure that with every rebuild the significant part of package's version changes.  There are currently basically two ways to achieve that:
* either update `COMPONENT_REVISION` with every change in `Makefile` in a case the `COMPONENT_VERSION` stays same (this is most defensive approach, which should work always), or
* in a case the previously built package is very old, IOW its `$(RELEASE_MAJOR).$(RELEASE_MINOR).$(UPDATENUM)` tuple is not up-to-date, we could rely on fact that with the rebuild the tuple will get updated (so the version will increase) anyway even without the `COMPONENT_REVISION` change.  Since this approach is a bit risky it is better not to rely on it.

Anyway, there is other problem with updates like this (fix download url only): all touched packages will be rebuilt and republished by jenkins build server (because their `Makefile` changes) so you, @dbh201, should make sure that such rebuilds really works.  So please do `gmake sample-manifest && gmake REQUIRED_PACKAGES && gmake test && gmake publish` for every such component to make sure there is no other change required there to have them rebuilt successfully.  Basically, after four `gmake`s above `git status .` should be clean.

Thanks.

AndWac:
You can remove the commits for firefox and thunderbird as they have already been updated.

dbh201:
Bumped COMPONENT_REVISION and did the 4 `gmake` recipes for each. Libreoffice is failing some tests so I'm going to take a look at that and put any fixes in another PR.

PR #9601, opened 2022-10-22
psycopg2: update to 2.9.4 and drop 32 bit

AndWac:


components/python/psycopg2/pkg5
```diff
@@ -1,16 +1,16 @@
...
         "runtime/python-39",
         "shell/ksh93",
         "system/library"
     ],
     "fmris": [
-        "library/python/psycopg2-37",
-        "library/python/psycopg2-39",
+        "library/python/psycopg2",
+        "library/python/psycopg2",
         "library/python/psycopg2"
```
AndWac: @mtelka: Can you check why this happens?

AndWac: Probably a side effect of GENERATE_EXTRA_CMD=

AndWac: No, without that line the resulting pkg5 doesn't change.

mtelka: This is because you confused the build framework by asking for both: versioned python packages, and unversioned python packages.

components/python/psycopg2/psycopg2-PYVER.p5m
```diff
@@ -9,10 +9,10 @@
 
 #
 # Copyright 2016 Alexander Pyhalov
-# Copyright 2021 Andreas Wacknitz
+# Copyright 2021, 2022 Andreas Wacknitz
 #
 
-set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)-$(PYV)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
```
mtelka: Here is the direct result of the confusion: the file is named `...-PYVER.p5m`, but the fmri lacks `-$(PYV)`.

components/python/psycopg2/Makefile
```diff
@@ -14,40 +14,40 @@
...
-PYTHON_VERSIONS=	3.7 3.9
-
-TEST_TARGET=$(NO_TESTS)
+TEST_TARGET=	$(NO_TESTS)
 include $(WS_MAKE_RULES)/common.mk
 
+# Remove the default post install actions for python packages:
+COMPONENT_POST_INSTALL_ACTION=
+# Suppress the automatic addition of $(PYV) to the package fmri:
+GENERATE_EXTRA_CMD=
```
mtelka: To solve the problem simply remove both `COMPONENT_POST_INSTALL_ACTION` and `GENERATE_EXTRA_CMD` (or fully convert to unversioned python package, which is very likely not what you want).

components/python/psycopg2/manifests/sample-manifest.p5m
```diff
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2021 <contributor>
+# Copyright 2022 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
```
mtelka, line 17: This line should appear in the real manifest too.

mtelka (changes requested):

components/python/psycopg2/Makefile
AndWac: Duh, stupid unreflcted copy and paste error. Sorry for the hassles.

components/python/psycopg2/manifests/sample-manifest.p5m
AndWac: fixed

components/python/psycopg2/psycopg2-PYVER.p5m
```diff
@@ -37,12 +46,5 @@ file path=usr/lib/python$(PYVER)/vendor-packages/psycopg2/extras.py
...
-#file path=usr/lib/python$(PYVER)/vendor-packages/psycopg2/_psycopg.so
-
-# force a dependency on the Python $(PYVER) runtime
-depend type=require fmri=__TBD pkg.debug.depend.file=python$(PYVER) \
-    pkg.debug.depend.path=usr/bin
-
-# force a dependency on the psycopg2 package
-depend type=require \
-    fmri=$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+#file path=usr/lib/python$(PYVER)/vendor-packages/psycopg2/_psycopg.cpython-39.so
```
mtelka, line 45: >> Is this (and |_psycopg.cpython-37m.so| for 3.7) really not needed to make the package working correctly? If so, why is it built?
>
> Github seems to be confused. I cannot see your last messages in the browser but only got emails. So I miss the context and can only guess. I have commented these two lines and manually added the last line because we have transforms/python-3-soabi which manipulates suffixes of .so files.

Ok, thanks.  Makes sense, I think.

components/python/psycopg2/psycopg2-PYVER.p5m
```diff
@@ -9,26 +9,35 @@
...
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+# force a dependency on the Python $(PYVER) runtime
+depend type=require fmri=__TBD pkg.debug.depend.file=python$(PYVER) \
+    pkg.debug.depend.path=usr/bin
+
+# force a dependency on the psycopg2 package
+depend type=require \
```
mtelka: Please drop this. It would cause some problems (and possibly end user questions) during old python version obsoletion.

mtelka (changes requested):

components/python/psycopg2/psycopg2-PYVER.p5m
AndWac: Ok, I removed the dependency on psycopg2.

mtelka:
Hehe, I found the cause for problems with github: if you try to comment in a recent commit (not sure how recent it needs to be, but it seems it is minutes) github is confused.  To make sure github won't be confused, you need to wait with comments until it stops to show "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." and green "Review changes" button appears for the commit instead.  Stupid github.

mtelka:
> Hehe, I found the cause for problems with github: if you try to comment in a recent commit (not sure how recent it needs to be, but it seems it is minutes) github is confused. To make sure github won't be confused, you need to wait with comments until it stops to show "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository." and green "Review changes" button appears for the commit instead. Stupid github.

No, the problem is different.  If you click on [5337108](https://github.com/OpenIndiana/oi-userland/commit/5337108445bdb06a84a9c5bf0d1931911d5365b0) in the `AndWac force-pushed the psycopg2 branch from cae7b55 to 5337108 21 minutes ago` line you'll confuse github (and others), but if you click on [5337108](https://github.com/OpenIndiana/oi-userland/pull/9601/commits/5337108445bdb06a84a9c5bf0d1931911d5365b0) in the `psycopg2: update to 2.9.4 and drop 32 bit 5337108` line everything works well.  Yes, the url you click on differs, but still, it is confusing.

mtelka (approved):

PR #9623, opened 2022-10-23
shared-macros.mk: import python shebang fix functions from solaris us…

AndWac:
…erland

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -717,6 +717,35 @@ PYTHON_DIR= /usr/lib/python$(PYTHON_VERSION)
 PYTHON_LIB= /usr/lib/python$(PYTHON_VERSION)/vendor-packages
 PYTHON_DATA= $(PYTHON_LIB)
 
+# If the component has python scripts then the first line should probably
+# point at the python version currently set by the $(PYTHON) variable so
+# as not to be influenced by the ips python mediator.
+# In the component's Makefile define PYTHON_SCRIPTS with a list of files
+# to be edited.
+
+# Edit the leading #!/usr/bin/python line in python scripts to use the
+# BUILD's $(PYTHON). The source file must be recompiled after that, as
+# the corresponding .pyc file is outdated now.
+PYTHON_SCRIPT_SHEBANG_FIX_FUNC = \
+    $(GSED) -i \
+        -e '1s@/usr/bin/python$$@$(PYTHON)@' \
+        -e '1s@/usr/bin/python\ @$(PYTHON) @' \
+        -e '1s@/usr/bin/env\ $(PYTHON)@$(PYTHON)@' \
+        -e '1s@/usr/bin/env\ python[23]@$(PYTHON)@' \
+        -e '1s@/usr/bin/env\ python@$(PYTHON)@' $(1); \
+    $(PYTHON) -m compileall $(1);
+
+# PYTHON_SCRIPTS is a list of files from the calling Makefile.
+PYTHON_SCRIPTS_PROCESS= \
+    $(foreach s,$(PYTHON_SCRIPTS), \
+            $(call PYTHON_SCRIPT_SHEBANG_FIX_FUNC,$(s)))
+
+# Finally if PYTHON_SCRIPTS is defined in a Makefile then process them here.
+# If multiple installs in the component then clear
+# COMPONENT_POST_INSTALL_ACTION =
+# and re-add $(PYTHON_SCRIPTS_PROCESS)
+COMPONENT_POST_INSTALL_ACTION += $(PYTHON_SCRIPTS_PROCESS) ;
+
 JAVA8_HOME =	/usr/jdk/instances/openjdk1.8.0
 JAVA11_HOME =	/usr/jdk/instances/openjdk11.0.10
 JAVA18_HOME =	/usr/jdk/instances/openjdk18.0.1


AndWac:
This can be used like this (from hplip):

\# Fix python script shebang
COMPONENT_POST_INSTALL_ACTION += \\
    echo "Fixing python shebang..."; \\
    ggrep -rlI -m1 '^\\#!.*python' $(PROTO_DIR) | while read file; do \\
       \$\(call\ PYTHON\\_SCRIPT\\_SHEBANG\\_FIX\\_FUNC, \"\\$\$file\"\) \\
    done;

make-rules/shared-macros.mk
```diff
@@ -717,6 +717,35 @@ PYTHON_DIR= /usr/lib/python$(PYTHON_VERSION)
 PYTHON_LIB= /usr/lib/python$(PYTHON_VERSION)/vendor-packages
 PYTHON_DATA= $(PYTHON_LIB)
 
+# If the component has python scripts then the first line should probably
+# point at the userland default build python so as not to be influenced
```
mtelka: Below you replace `/usr/bin/python` by `$(PYTHON)` which is something like `/usr/bin/python3.7` or `/usr/bin/python3.9` depending what version is currently built for multiversioned components.  This is not "userland default build python", IMHO.  Maybe the wording should be changed a bit here.  Or, could you please explain what you mean by "userland default build python" because it is easily possible that I miss something here.  Thanks.

make-rules/shared-macros.mk
```diff
@@ -717,6 +717,35 @@ PYTHON_DIR= /usr/lib/python$(PYTHON_VERSION)
...
+# by the ips python mediator.
+# In the component's Makefile define PYTHON_SCRIPTS with a list of files
```
mtelka, line 723: I believe it should be possible to automatically detect list of scripts.  Maybe the `file` tool could help with that?

make-rules/shared-macros.mk
```diff
@@ -717,6 +717,35 @@ PYTHON_DIR= /usr/lib/python$(PYTHON_VERSION)
...
+# PYTHON_SCRIPTS is a list of files from the calling Makefile.
+PYTHON_SCRIPTS_PROCESS= \
+    $(foreach s,$(PYTHON_SCRIPTS), \
+            $(call PYTHON_SCRIPT_SHEBANG_FIX_FUNC,$(s)))
+
+# Finally if PYTHON_SCRIPTS is defined in a Makefile then process them here.
+# If multiple installs in the component then clear
+# COMPONENT_POST_INSTALL_ACTION =
+# and re-add $(PYTHON_SCRIPTS_PROCESS)
+COMPONENT_POST_INSTALL_ACTION += $(PYTHON_SCRIPTS_PROCESS)
```
mtelka: I think this needs a semicolon at the end of your addition so it works properly with other similar post install actions, like the one in `setup.py.mk`.

mtelka (changes requested):

make-rules/shared-macros.mk
mtelka: ... or just use the `ggrep` way you found in `hplip`.

mtelka:
Hmmm.  Is this really needed?  If so, why do we have things like these?

```
# head -n1 /usr/bin/activate-global-python-argcomplete-*
==> /usr/bin/activate-global-python-argcomplete-2.7 <==
#!/usr/bin/python2.7

==> /usr/bin/activate-global-python-argcomplete-3.5 <==
#!/usr/bin/python3.5

==> /usr/bin/activate-global-python-argcomplete-3.7 <==
#!/usr/bin/python3.7

==> /usr/bin/activate-global-python-argcomplete-3.9 <==
#!/usr/bin/python3.9
#
```

make-rules/shared-macros.mk
AndWac: That's a verbatim copy from solaris-userland. Maybe we should replace "the userland default build python" with "currently active python".

make-rules/shared-macros.mk
AndWac: Another example:
PYTHON_SCRIPTS += \\$\(PROTO\_DIR\)/\$\(PYTHON\_VENDOR\_PACKAGES_BASE\)/numpy/*.py

AndWac: Escaping such examples is hell.

make-rules/shared-macros.mk
AndWac: I am not sure about the ;
We have ; at the end of multi-line statements but also omitting it in single line actions, like in common.mk.

mtelka: The problem I'm talking about is this:
```
$ cat Makefile 
A += echo a1
A += echo a2
B += echo b1 ;
B += echo b2 ;

a:
        $(A)
        $(B)
$ gmake a
echo a1 echo a2
a1 echo a2
echo b1 ; echo b2 ;
b1
b2
$
```

make-rules/shared-macros.mk
mtelka: What exactly is "currently active python"?  Do you mean that one set by the python mediator?  Or the one set by  the `$(PYTHON)` variable?

AndWac: The latter.

mtelka: Then please say so in the comment, so it is not confusing.  Thanks.

PR #9631, opened 2022-10-23
stack-spur: add resuls-all.master file

cstes:
This is an experimental attempt to add a results-all.master file for the Stack Spur VM for OpenSmalltalk.

The strategy works for any Squeak vm, and in fact the current Makefile is incorrect because it uses whatever squeak VM is installed

The idea is to run all tests of the SUnit test framework, including sound/audio/video (GUI) tests.

The results are then saved in a file results-all.master

### DIFF
--- a/components/runtime/smalltalk/stack-spur/test/results-all.master
+++ b/components/runtime/smalltalk/stack-spur/test/results-all.master
@@ -1,0 +1,26 @@
+SUnit Results
+Squeak5.3
+solaris2.11
+Open Smalltalk Stack VM [StackInterpreterPrimitives VMMaker.oscog-eem.3254]
+Failed Tests
+'AllocationTest>>#testOneGigAllocation'
+'BasicBehaviorClassMetaclassTest>>#testClassDescriptionAllSubInstances'
+'BitBltTest>>#testAllAlphasRgbMul'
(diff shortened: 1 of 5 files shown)


components/runtime/smalltalk/stack-spur/test/results-all.master
```diff
@@ -0,0 +1,24 @@
+SUnit Results
+Squeak5.3
+solaris2.11
+Open Smalltalk Stack VM [StackInterpreterPrimitives VMMaker.oscog-eem.3254]
+4735 run in 0:00:07:34.717, 4676 passes, 42 expected failures, 4 failures, 13 errors, 0 unexpected passes
+Failed Tests
```
AndWac, line 5: You should filter out timings (and dates if available) fom the expected test results by means of appropriate COMPONENT_TEST_TRANSFORMS.

mtelka: Yes, exactly.  And in ideal world all tests should pass.

AndWac:
Please make sure that either the COMPONENT_VERSION or COMPONENT_REVISION is being changed (incremented) with any PR. In this case you'll need to set COMPONENT_REVISION=1.

PR #9649, opened 2022-10-24
obsolete library/fftw-2

AndWac:
We don't have any dependent packages anymore.


components/meta-packages/history/history
```diff
@@ -618,6 +618,7 @@ library/desktop/search/tracker/tracker-firefox@0.6.96-2015.0.2.0
 library/desktop/webkitgtk@1.2.7,5.11-2017.0.0.6
 library/e/elementary@1.12.3,5.11-2017.0.0.1
 library/e/evas_generic_loaders@1.12.0,5.11-2017.0.0.2
+library/fftw-2@2.1.5,5.11-2020.0.1.2 library/fftw-3
```
mtelka: Since `fftw-3` is incompatible with `fftw-2` I think it would be better simply obsolete `fftw-2` instead of renaming.  But this my preference is not strong :-).

mtelka (approved):
I'm glad to see this!

The project's homepage says:
> If you are still using FFTW 2.x, please note that FFTW 2.x was last updated in 1999 and it is obsolete.

And since no other package depends on it, the removal of `fftw-2` is the only reasonable way to go.

Thank you.

components/meta-packages/history/history
AndWac: Ok, done. My idea was that fftw-3 is the canonical replacement for fftw-2.

PR #9650, opened 2022-10-24
Python integrate project

mebenn:
python-integrate-project is really the way to go; especially since Glances has more dependencies, i.e., for client-server operation; so I have a few more Python projects to prepare. All caused by pip automatically installing things on my machine ages ago.  I'll have to set up a new bare OI installation on a VM to see what really needs to be packaged; but python-integrate-project is really great.

As I was obtaining errors when running for defusedxml; I took a look at the script and fixed enough so that it ran to the end; only obtaining ERRORS for test targets for Python versions. This PR fixes the script until it completes running on my machine.
  
As soon as I get some time, I'll continue working on getting python-integrate-project to produce working for defusedxml. It is now producing a Makefile, ./manifests/sample-manifests.p5m; but a defusedxml-PYVER.p5m doesn't look right, I look into this.


mebenn:
Oops, I didn't do a git pull; and I'm just after seeing the the commits from today, and the new defused packages...sorry

mtelka:
I'm sure this PR is just a draft only, so please mark it so (or simply close).

BTW, any changes in the `python-integrate-project` tool should be carefully weighted and discussed.  I didn't look at your changes in `python-integrate-project` in detail, but what I saw is mostly not needed.

Thanks.

mtelka:
> As soon as I get some time, I'll continue working on getting python-integrate-project to produce working for defusedxml. It is now producing a Makefile, ./manifests/sample-manifests.p5m; but a defusedxml-PYVER.p5m doesn't look right, I look into this.

The usual workflow while working with `python-integrate-project` and something does not work as expected is simply to create `python-integrate-project.conf` file, or change something else in the build framework, but that really depends on errors reported by the `python-integrate-project` tool.

Maybe you want to look at [draft documentation](https://github.com/mtelka/oi-userland/blob/pim-doc/doc/perl-modules.md) for `perl-integrate-module`, a more mature brother of `python-integrate-project`.  The `perl-integrate-module` tool is currently used to handle all Perl modules.  The aim of `python-integrate-project` is to do the same for all Python projects.

HTH.

PR #9653, opened 2022-10-25
make-rules: support for single perl/python version builds

mtelka:
This introduces new `SINGLE_PERL_VERSION`/`SINGLE_PYTHON_VERSION` variables (I'm not sure the names are best, so if there is some suggestion I'm open to change them) defaulting to `no` for `makemaker`/`modulebuild`/`setup.py` builds respectively and `yes` otherwise.  The defaults could be overriden in component's `Makefile` if needed.

I added guards to `setup.py.mk` to allow immediate use of it (for components like `barman`).  I didn't add similar guards for perl builds (`makemaker`/`modulebuild`) because I'm not aware about any component that would use it.  If there is a need for them we can add later.

I tested this with `barman` with this change:
```
diff --git a/components/python/barman/Makefile b/components/python/barman/Makefile
index 72b47afe5..505b6e8d6 100644
--- a/components/python/barman/Makefile
+++ b/components/python/barman/Makefile
@@ -31,18 +31,13 @@ COMPONENT_CLASSIFICATION=   System/Databases
 COMPONENT_FMRI=                database/postgres/barman
 COMPONENT_LICENSE=     GPLv3
 
-PYTHON_VERSIONS= $(PYTHON_VERSION)
+SINGLE_PYTHON_VERSION = yes
 
 PKG_MACROS += PYVER=$(PYTHON_VERSION)
 
 TEST_TARGET=   $(NO_TESTS)
 include $(WS_TOP)/make-rules/common.mk
 
-# Remove the default post install actions for python packages:
-COMPONENT_POST_INSTALL_ACTION=
-# Suppress the automatic addition of $(PYV) to the package fmri:
-GENERATE_EXTRA_CMD=
-
 # Install example configuration files
 DOC_DIR=$(PROTO_DIR)/usr/share/barman
 
```
and confirmed that nothing changed.  Please note that I had to set `SINGLE_PYTHON_VERSION` to `yes` because `barman` is `setup.py` build defaulting to `SINGLE_PYTHON_VERSION = no`.

### DIFF
--- a/make-rules/common.mk
+++ b/make-rules/common.mk
@@ -35,6 +35,26 @@ BUILD_STYLE ?= configure
 # Some build styles might want to set some defaults before prep.mk is included.
 -include $(WS_MAKE_RULES)/$(strip $(BUILD_STYLE))-defaults.mk
 
+# The SINGLE_PERL_VERSION variable is used to select building a component for
+# single or multiple Perl versions.  By default we build for single Perl
+# version only unless the component is a Perl module.  If the default value of
+# SINGLE_PERL_VERSION is not what is needed for a component it could be
+# overridden in component's Makefile.
+SINGLE_PERL_VERSION ?= yes
+ifeq ($(strip $(SINGLE_PERL_VERSION)),yes)
+PERL_VERSIONS = $(PERL_VERSION)
+endif
+
+# The SINGLE_PYTHON_VERSION variable is used to select building a component for
+# single or multiple Python versions.  By default we build for single Python
+# version only unless the component is a Python project.  If the default value
+# of SINGLE_PYTHON_VERSION is not what is needed for a component it could be
+# overridden in component's Makefile.
+SINGLE_PYTHON_VERSION ?= yes
+ifeq ($(strip $(SINGLE_PYTHON_VERSION)),yes)
+PYTHON_VERSIONS = $(PYTHON_VERSION)
+endif
+
 include $(WS_MAKE_RULES)/prep.mk
 
 # Override this to limit builds and publication to a single architecture.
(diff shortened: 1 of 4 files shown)


make-rules/common.mk
```diff
@@ -35,6 +35,16 @@ BUILD_STYLE ?= configure
 # Some build styles might want to set some defaults before prep.mk is included.
 -include $(WS_MAKE_RULES)/$(strip $(BUILD_STYLE))-defaults.mk
 
+SINGLE_PERL_VERSION ?= yes
```
AndWac, line 43: Maybe add some explanation about its purpose?

mtelka: Any suggestion about wording?  It is so obvious for me that it does not deserve (IMHO) any comment.  But if something is suggested I'll add it to make it clear(er) for others too.  Thanks.

AndWac: Something like:
Use only a single python/perl runtime for packages that don't provide python/perl libraries (but only use it as a runtime).

mtelka: Hmmmmm.  I think this is not entirely true.  For example `barman` provides python libraries (see `/usr/lib/python3.9/vendor-packages/barman/`).

Maybe something like:  _Build for single perl/python version only when you do not need more._  But this is something that is obvious when you read next three lines of code.

AndWac: The comment should be for anybody with only little package creation experiences.
The idea of a single perl/python runtime is: My package doesn't need more than one version. Either because its only using it as a runtime and/or I want to provide or maintain a single variant only.

mtelka: Yes, I agree.  The problem (or, actually, this is not a problem at all) is that somebody with little package creation experiences will probably never need to change `SINGLE_PERL_VERSION`/`SINGLE_PYTHON_VERSION` because defaults in most cases is what they need.  For example for gdb (PR #9568) nothing needs to be do and single python build will be selected automagically.

Anyway, I'll try to add some comment there...

PR #9740, opened 2022-10-31
Mail::SPF - remove spfquery script as it conflicts with libspf2 binary

fritzkink:
Comparing both spfquery script/binary it makes more sense to keep the binary which comes with libspf2. It is a binary (usually faster) and slightly younger.


AndWac:
Please add Mail-SPF to the commit message and the PR's title to give more context when browsing the git history.

PR #9757, opened 2022-11-02
Exclude components/python/build from .gitignore

mtelka:


mtelka:
The `components/python/build` directory is main dir for a component, not a build directory for `python` component.

PR #9775, opened 2022-11-05
python/setuptools: update to 65.5.1; rebuild to get packages for python 2.7 and 3.5 obsoleted

mtelka:


AndWac:
The following package depend on setuptools-27:
library/python/argcomplete-27
library/python/chardet-27
library/python/coverage-27
library/python/cryptography-27
library/python/nose-27
library/python/pip-27
library/python/tempora-27

The situation is similar for setuptools-35.

mtelka:
Yes, I know.  It is not a (big) problem because the obsoleted `setuptools-27` and `setuptools-35` are (will be) `noincorporate`d so the previous (non-obsoleted) versions will stay installed (or could be installed when needed) even with `userland-incorporation` in place.  We already have similar situation with other python packages (for example `asn1crypto-27`, `asn1crypto-35`, `backport_abc-27`, ...) and this is (and will be) usual case during obsoletion of EOLed python (or perl) versions.

The only issue that could happen with this is possible repo cleanup on IPS server when old packages could disappear, but they shouldn't because they are still needed (we had one such case about a month ago with `importlib-metadata-35`).  So, please, until `python-27` and `python-35` packages are still around, do not do package cleanup on the ips server.  In a case you need to do the cleanup, please let me know and I'll provide you list of obsoleted packages that still needs their non-obsoleted versions (currently the list contains 9 packages, but it could change with every new obsoletion of -27 and/or -35 packages)

Thank you.

PS1: Some of packages you listed above are already obsoleted, for example `argcomplete-27` or `chardet-27`.  The `argcomplete-27` package could (and should) be uninstalled because it is not needed by any other package.  The `chardet-27` is needed by (not yet obsoleted) `requests-27`, so it stays installed (and it needs to stay in the IPS repo too).

PS2: If you disagree with obsoletion of `setuptools-27` and `setuptools-35` now, I can remove it and keep both packages live a bit longer, but I currently see no reason doing so.

PR #9779, opened 2022-11-05
Update gstreamer1 1.20.4

ghost:
Update gstreamer1 and all packages related to it to the latest version 1.20.4.


AndWac:
What about gstreamer1-ffmpeg? It looks like a new package.

PR #9780, opened 2022-11-05
Adding FS-UAE Amiga emulator

Toasterson:
Dependency resolution is broken on my machine, but I think this is an issue in my build environment.


Toasterson:
I managed to fix the dependency issue, but it still depends on a library from hipster-encumbered. Do we need to move it to encumbered?

AndWac:
Yes, evem if a package depends on a single package from encumbered it also has to stay there.

components/encumbered/fs-uae/Makefile
```diff
@@ -0,0 +1,54 @@
...
+COMPONENT_CLASSIFICATION=Applications/System Utilities
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.xz
+COMPONENT_ARCHIVE_URL=	https://fs-uae.net/files/FS-UAE/Stable/$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH=	sha256:606e1868b500413d69bd33bb469f8fd08d6c08988801f17b7dd022f3fbe23832
+COMPONENT_LICENSE=	GPLv2
+COMPONENT_LICENSE_FILE=	fs-uae.txt
+
+TEST_TARGET=$(NO_TESTS) # if no testsuite enabled
+include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 34: You also need to add
include $(WS_MAKE_RULES)/encumbered.mk
before you include common.mk

Toasterson: Thanks! Added.

AndWac:
pkg5 needs an update

PR #9788, opened 2022-11-05
prep-svn.mk : create WS_TOP/archives

cstes:
squeak uses subversion and prep-svn.mk

the latest months I observe the following:   WS_TOP/archives is not automaitcally created on svn export

this PR tries to do the same thing as prep-download.mk  and to MKDIR WS_TOP/archives before svn export

### DIFF
--- a/make-rules/prep-svn.mk
+++ b/make-rules/prep-svn.mk
@@ -48,6 +48,7 @@ CLEAN_PATHS += $$(COMPONENT_SRC$(1))
 CLOBBER_PATHS += COMPONENT_ARCHIVE$(1)
 SOURCE_DIR$(1) = $$(COMPONENT_DIR)/$$(COMPONENT_SRC$(1))
 
+download::      $$(USERLAND_ARCHIVES)
 download::	$$(USERLAND_ARCHIVES)$$(COMPONENT_ARCHIVE$(1))
 
 # First attempt to download a cached archive of the SCM repo at the proper


make-rules/prep-svn.mk
```diff
@@ -48,6 +48,7 @@ CLEAN_PATHS += $$(COMPONENT_SRC$(1))
 CLOBBER_PATHS += COMPONENT_ARCHIVE$(1)
 SOURCE_DIR$(1) = $$(COMPONENT_DIR)/$$(COMPONENT_SRC$(1))
 
+download::      $$(USERLAND_ARCHIVES)
 download::	$$(USERLAND_ARCHIVES)$$(COMPONENT_ARCHIVE$(1))
```
AndWac, line 52: This is already done in prep-download.mk which is included from prep.mk which in turn is included from common.mk.
So in my eyes this looks superfluous.

cstes:
I am not sure why I have to copy the line or even whether my addition is correct.

But in any case it is easy to reproduce : git clone oi-userland in a new directory

Verify that in that directory under the top level archives (the directory archives) does not exist

Go immediately then to runtime/smalltalk/squeak and run "gmake download" or "gmake prep" or even "gmake test".

The problem as I observe it is that for this SVN compoment using subversion the "archives" directory is not created.

When I use the line frop this commit in prep-svn.mk it works for me.
By it works I mean to say htat it then automaticaly does a "mkdir archives" before svn export.

This is not a big problem as I can also manually mkdir archives anyway, but I'm just alerting to the fact that I think this has chnaged a few months ago because back a year or so ago, the "archives" forlder was in the git clone tree because it had a stub empty file in it.

mtelka:
First, I just tried what @cstes described above and I can confirm that the `gmake prep` failed.
Second, @AndWac is right that the same rule is in `prep-download.mk` too, but this rule is executed only in a case there are both `COMPONENT_ARCHIVE` and `COMPONENT_ARCHIVE_URL` defined for the component.  The problem is that the `COMPONENT_ARCHIVE` is not defined by `squeak` itself but instead in `prep-svn.mk`, which is included from `prep.mk` after `prep-download.mk`, so when `prep-download.mk` is evaluated the `COMPONENT_ARCHIVE` is undefined, so the rule @AndWac is referring to is not instantiated.

So to sum: Yes, there is a problem, but it maybe needs different solution than proposed here (the solution proposed here looks more like workaround than long-term solution).  Maybe include `prep-svn.mk` before `prep-download.mk`?  But I didn't check if this could cause some other problem.

mtelka:
... and it looks like my 4841f66917298ee3d9bf96d5e585bd0584246a08 caused this.

mtelka:
I looked at the `prep.mk` file and all files it includes and I believe this is the proper fix for the problem.  In addition, the same fix should be done in `prep-hg.mk` and `prep-git.mk` files too.  @cstes, would you mind to add the same line to `prep-hg.mk` and `prep-git.mk` files too?  Preferably as a part of this PR.  Thank you.

cstes:
I can have a look at this in a few days but I also mentioned that it  the past "archives" was actually a directory with a file :gitkeep in it and it was part of the oi-userland repo so "git clone" of oi-userland would simply have an empty (or almost empty short of the .gitkeep file) directory by default.  That is the way it used to work a year ago or os.

cstes:
This can be closed if the alternative of pull request #9865 is used

cstes:
I'm closing this PR because I think squeak is about the only package that is impacted as there are almost no subversion svn packages in oi-userland short of squeak.   Also I can live with the workaround of manually creating an archives directory.   Finally another reason is that I have a Pull Request to restore the .gitkeep file in the empty archives directory.

PR #9801, opened 2022-11-06
openjdk-17: fix COMPONENT_REVISION

AndWac:


AndWac:
This will be fixed in a different way.

PR #9810, opened 2022-11-07
pyproject.mk: make bootstrap transforms overridable by components

mtelka:


make-rules/pyproject.mk
```diff
@@ -66,7 +66,8 @@ $(BUILD_DIR)/%-3.9/.installed:	COMPONENT_INSTALL_ARGS +=	--destdir $(PROTO_DIR)
 $(BUILD_DIR)/%-3.9/.installed:	COMPONENT_POST_INSTALL_ACTION +=	$(PYTHON) -m compileall $(PROTO_DIR)/$(PYTHON_LIB) ;
 
 # Special transforms needed only until we drop support for Python 3.7
-PUBLISH_TRANSFORMS +=	$(WS_TOP)/transforms/python-bootstrap
+BOOTSTRAP_TRANSFORMS =	$(WS_TOP)/transforms/python-bootstrap
+PUBLISH_TRANSFORMS +=	$(BOOTSTRAP_TRANSFORMS)
 else
```
AndWac, line 71: How does this make the bootstrap transforms overrideable by components?

mtelka: By setting `BOOTSTRAP_TRANSFORMS` to some other value.

AndWac: But shouldn't it be ?= then?

mtelka: Yes, it could be `?=` too, but, actually, it does not matter much because `BOOTSTRAP_TRANSFORMS` will be set by components after the `common.mk` include.  (Please remember that all of these bootstrap workarounds will be removed in ca 8 months once we start to obsolete Python 3.7)

AndWac: Ah ok, I didn't realise that it will be changed later.

PR #9836, opened 2022-11-09
nvidia: update to 520.56.06

AndWac:


mtelka:
Hold on with this please.

AndWac:
This one works but has the same problems like the 515 version. It crashes in certain situations that can be provoked by Mate's window manager but I have successfully used entligthenment and lots of apps with it, eg. atril, emacs, ...

mtelka:
My idea with the `nvidia` driver was to keep it at the **Latest Production Branch Version** (see https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/openindiana/nvidia/Makefile#L22) which means *stable*.  The `520.56.06` version is **Latest New Feature Branch Version** which means *unstable*.  I'd prefer to have `nvidia` to always point to the **Latest Production Branch Version**.  If we really want to track the *unstable* driver I suggest to create new package for it.  For example `nvidia-latest`, or `nvidia-unstable`, or `nvidia-new-feature`, or something.

AndWac:
I missed that 520 is not a stable branch. Its release notes state
"
    Fixed a bug in the Vulkan driver which could lead to corruption in geometry and tessellation control shaders.
    Fixed a regression in 515.76 that caused blank screens and hangs when starting an X server on RTX 30 series GPUs in some configurations where the boot display is connected via HDMI. 
"
So the feature branch fixed a bug of the stable branch. Where can I find the information about what type a release belongs to? I haven't found anything at https://www.nvidia.com/en-us/drivers/unix/solaris-display-archive/

mtelka:
BTW, you apparently didn't run `gmake sample-manifest` because you are changing `nvidia.p5m`, but not the `sample-manifest.p5m`.  But, OTOH, if you would try to run `gmake sample-manifest` then you would find that it fails because `GENERATE_EXTRA_CMD` is broken in the `Makefile` (there is missing `|` at the beginning of the command).

mtelka:
You should always look at the https://www.nvidia.com/en-us/drivers/unix/ for branch info.  The page you linked is just archive of all releases without information about their stability, EOL, etc.  Just pure history.

BTW, I monitor the NVIDIA website and once they release new driver in any driver line we support I'll update it in OI.

mtelka:
> BTW, I monitor the NVIDIA website and once they release new driver in any driver line we support I'll update it in OI.

FYI, I monitor it automatically once a day.

AndWac:
Ok, the | seems to be missing in all nvidia Makefiles.

mtelka:
> Ok, the | seems to be missing in all nvidia Makefiles.

Yep, of course.  :-)

But it was not worth fixing because it would trigger rebuild to get the exactly same package.

mtelka:
> But it was not worth fixing because it would trigger rebuild to get the exactly same package.

I expect they will release new version in all their branches in few months so it will be fixed during update.

mtelka:
Small clarification:  The **New Feature Branch** driver is not *unstable* in sense of driver quality and/or software stability when used, but instead it is *unstable* in sense that it is targeted to early adopters who want to evaluate new features.  It is short lived branch driver where new features are introduced and made available between **Production Branch** releases.

PR #9844, opened 2022-11-09
python/setuptools_scm: rebuild to fix package dependencies

mtelka:


mtelka:
I see the build failed.  I think I know what is the reason.  I should have a fix soon.

mtelka:
The fix will have three parts.  The first one is #9846.  Please merge it now.  It won't solve the `setuptools_scm` build failure, but I need it in git to avoid merge conflicts with following commits.  Thanks.

mtelka:
Please merge #9847 first, then #9848.  Thank you.

PR #9857, opened 2022-11-11
glib: update 2.73.3

cgrzemba:
Update glib to version 2.73.3


components/library/glib/Makefile
```diff
@@ -16,34 +16,28 @@
 
 BUILD_BITS = 32_and_64
 BUILD_STYLE = meson
+USE_COMMON_TEST_MASTER= no
```
mtelka: Why do you need this when both `results-32.master` and `results-64.master` are same?

components/library/glib/Makefile
```diff
@@ -71,6 +64,26 @@ COMPONENT_POST_CONFIGURE_ACTION= \
...
+# We need to install libraries and set correct LD_LIBRARY_PATH before test
+$(TEST_TARGET): $(INSTALL_TARGET)
+
+# dbus-daemon is in /usr/lib
+COMPONENT_TEST_ENV.32 = LD_LIBRARY_PATH=$(PROTO_DIR)/usr/lib PATH=$(PATH):/usr/lib
+COMPONENT_TEST_ENV.64 = LD_LIBRARY_PATH=$(PROTO_DIR)/usr/lib/$(MACH64) PATH=$(PATH):/usr/lib/$(MACH64)
+COMPONENT_TEST_ENV = $(COMPONENT_TEST_ENV.$(BITS))
+COMPONENT_TEST_ENV += LC_ALL=C.UTF-8
+
+# COMPONENT_TEST_MASTER = $(COMPONENT_TEST_RESULTS_DIR)/results-all.master
```
mtelka: I think this comment is useless.  Please remove it.

components/library/glib/patches/04-xopen-version.patch
```diff
@@ -1,6 +1,11 @@
---- glib-2.62.3/meson.build.1	2019-12-18 21:58:02.109823287 +0000
-+++ glib-2.62.3/meson.build	2019-12-18 21:59:14.389494095 +0000
-@@ -2011,8 +2011,6 @@
+Remove Unix standard settings for Solaris 11.4, so we don't exclude all
```
mtelka, line 1: Solaris 11.4?  Really? :-)

components/library/glib/patches/17-python.patch
```diff
@@ -7,8 +7,8 @@
 -python = import('python').find_installation('python3')
 -# used for '#!/usr/bin/env <name>'
 -python_name = 'python3'
-+python = import('python').find_installation('python3.5')
-+python_name = '/usr/bin/python3.5'
++python = import('python').find_installation('python3.9')
```
mtelka, line 12: Why we need to explicitly set this to python3.9?  Why is python3 not sufficient?

mtelka (changes requested):
Thank you for this work.  Really appreciated!

components/library/glib/patches/04-xopen-version.patch
cgrzemba: The patch is borrowed from solaris-userland, so I will keep the comment.

components/library/glib/patches/17-python.patch
cgrzemba: Without this 3 tests more failing.

components/library/glib/patches/26-iconv-U+FFFF.patch
```diff
@@ -1,8 +1,8 @@
-See https://www.illumos.org/issues/14934
```
mtelka, line 1: Why are you dropping this comment?

components/library/glib/patches/17-python.patch
mtelka: It would be great to have this documented.  Could you please add a comment to this patch stating what tests are failing without forcing python3.9?  Thanks.

mtelka (changes requested):

components/library/glib/patches/17-python.patch
mtelka: BTW, when I tested the previous glib version (2.62.6) I found that some tests are unstable.  They usually pass, but sometimes fails.  Maybe this is the case here too.  IOW, maybe additional failed tests are not caused by `python3` here.

components/library/glib/patches/26-iconv-U+FFFF.patch
cgrzemba: sorry, readded ...

components/library/glib/patches/17-python.patch
AndWac: I can confirm that without the patch the number of failing tests is always higher than with it. I have tested it several time on two machines.

mtelka (approved):

PR #9912, opened 2022-11-14
tools/license-detector: add Apache-2.0

erwinlem:
Bleach uses the short version of the Apache license.

### DIFF
--- a/tools/license-headers/Apache-2.0
+++ b/tools/license-headers/Apache-2.0
@@ -1,0 +1,15 @@
+# sed -E -n
+# -e '/^Licensed under the Apache License/,/^limitations under the License./p'
+Copyright <year> <copyright holders>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
--- a/tools/license-detector
+++ b/tools/license-detector
@@ -84,7 +84,7 @@ fi
 TMPFILE=$(mktemp -q)
 [[ -z "$TMPFILE" ]] && printf "ERROR: Temporary file creation failed\n" >&2 && exit 1
 
-for l in "$WS_TOOLS"/licenses/* ; do
+for l in "$WS_TOOLS"/{licenses,license-headers}/* ; do
 	[[ -f "$l" ]] || continue
 
 	LICCMD="$(grep '^# ' "$l" | sed -e 's/^# //' | tr '\n' ' ')"


tools/licenses/Apache-2.0
```diff
@@ -0,0 +1,15 @@
+# sed -E -n
+# -e 's/^(Copyright )\(c\)/\1/'
+Copyright <year> <copyright holders>
```
mtelka: Sorry, this is not full text of Apache-2.0.  Please see https://spdx.org/licenses/Apache-2.0.html

mtelka (changes requested):

tools/licenses/Apache-2.0
erwinlem: This is indeed not the full text, but the text you are supposed to put in your license file. That is what is stated on apache.org, it's not that clear on spdx.org. 

But for now, we could also grep for "http://www.apache.org/licenses/LICENSE-2.0" or some other unique string but i don't think that wil be reliable enough. Or am i missing something here?

The license file i'm looking at is this one : https://github.com/mozilla/bleach/blob/main/LICENSE

mtelka: Yes, this is something called "Standard License Header" (see spdx).  The `license-detector` have no direct support for this yet. Files in the `licenses` directory must match the full license text, otherwise `license-detector` won't work as expected and designed.

BTW, you could add grep for any text directly to the `license-detector` file to handle special cases like this one (please note there is already basic handling for `Apache-2.0`).  Or, the other possible option is to add support for "Standard License Header" into the `license-detector`.  For example you could create directory `license-headers` with such headers, or put headers into the `licenses` directory, but name such headers differently to not clash with full-text licenses (for example `header-Apache-2.0`).  I think the first option (`license-headers` directory) is better.

erwinlem: I added the license-headers directory, but I think we should clarify somehow this is meant for the "Standard License Header" files. Or maybe it's clear enough like this.

I also misunderstood how the sed worked, fixed that too.

tools/license-headers/Apache-2.0
```diff
@@ -0,0 +1,15 @@
+# sed -E -n
+# -e '/^Licensed under the Apache License/,/^limitations under the License./p'
+Copyright <year> <copyright holders>
```
mtelka, line 3: Please copy the header text from https://spdx.org/licenses/Apache-2.0.html.

tools/license-headers/Apache-2.0
```diff
@@ -0,0 +1,15 @@
+# sed -E -n
+# -e '/^Licensed under the Apache License/,/^limitations under the License./p'
```
mtelka, line 2: The Copyright line in the header is not omittable, only partly replaceable, so you should not skip it.

The anchoring of `limitations under the License.`to the beginning of line won't produce the results you expect.  You should either anchor this to the end of line (with `.` quoted), or search  for `Unless required by applicable` (this would be better, I think).  Please note the `license-detector` normalizes the text before it is handed over to the embedded processor.

mtelka (changes requested):

tools/license-headers/Apache-2.0
```diff
@@ -0,0 +1,15 @@
+# sed -E -n
```
mtelka, line 1: With so simple sed script you do not need `-E` here.

mtelka (changes requested):

tools/licenses/Apache-2.0
mtelka: > I added the license-headers directory, but I think we should clarify somehow this is meant for the "Standard License Header" files. Or maybe it's clear enough like this.

Feel free to propose how to clarify that.  One possible option is to add a comment to the `license-detector` script, but I'm open for other solutions too.

BTW, I think it would be good if the commit message says you add header only, not the full license text.

mtelka:
This could be closed because PR #10030 implementing the same functionality was merged.

Thank you.

PR #9917, opened 2022-11-14
pgbadger: update to 12.0 and switch to default perl version

AndWac:


components/database/pgbadger/pgbadger.p5m
```diff
@@ -24,5 +24,5 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license LICENSE license=PostgreSQL
 
-file usr/perl5/5.34/bin/pgbadger path=usr/bin/pgbadger mode=0555
-file usr/perl5/5.34/man/man1/pgbadger.1p path=usr/share/man/man1/pgbadger.1
+file usr/perl5/$(PERL_VERSION)/bin/pgbadger path=usr/bin/pgbadger mode=0555
+file usr/perl5/$(PERL_VERSION)/man/man1/pgbadger.1p path=usr/share/man/man1/pgbadger.1
```
mtelka, line 28: I suspect `usr/bin/pgbadger` and `usr/share/man/man1/pgbadger.1` will be created automatically as mediated symlinks. Please check created packages.  Thanks.

mtelka (changes requested):

components/database/pgbadger/pgbadger.p5m
AndWac: How can I ceck it? At least in my environment they are both ordinary files and the contents are as expected:
╰─➤  pkg contents pgbadger
PATH
usr/bin/pgbadger
usr/share/man/man1/pgbadger.1

mtelka: `pkg contents -mr pgbadger`

AndWac: Still only 2 file actions, 1 depend, 1 license. No link actions.

mtelka: Ok. Thanks.

mtelka: BTW, I think components like this should use `SINGLE_PERL_VERSION = yes` and stuff like file locations should be handled automatically for them (if it is not done so yet).

AndWac: Good idea.

mtelka: Ha, this component already have `SINGLE_PERL_VERSION = yes` because it is default for `justmake`.  Maybe I should implement automatic file relocations for such cases so you do not need to relocate them manually in `.p5m`.

PR #9932, opened 2022-11-16
upowerd: update illumos code for use newer glib version

cgrzemba:
it is necessary for use newer glib since PR#9857. The change is similar to https://www.illumos.org/issues/15025


AndWac:
Please run "gmake sample-manifest" and check if it is still valid (hint: it isn't here and would cause problems on the build server).

PR #9935, opened 2022-11-16
Add curio python project

mtelka:


components/python/curio/Makefile
```diff
@@ -0,0 +1,37 @@
...
+#
+
+BUILD_STYLE = setup.py
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		curio
+COMPONENT_VERSION =		1.6
+COMPONENT_SUMMARY =		curio - Curio
+COMPONENT_PROJECT_URL =		https://github.com/dabeaz/curio
```
AndWac, line 24: The summary is nonsense.

mtelka: Yes, it is, but it is what curio's authors wants to see as a description for their project: see https://pypi.org/project/curio/ and https://github.com/dabeaz/curio/blob/master/setup.py#L15

But maybe it is worth changing...

mtelka: Fixed

AndWac (approved):

PR #9979, opened 2022-11-20
pari: update to 2.15.1

cstes:
Replace test-all by dobench

test-all also seems to measure differences between statically linked PARI and dynamically linked PARI

Anyway tests of this package can be done in the Vagrantfile image with all optional packages installed.

The dobench quick test seems sufficient for just checking the sanity of the build


AndWac:
From the differences between old and new test result I conclude that a lot less tests are running now. I am not able to qualify the tests but from the names most now missing tests don't measure speed.

PR #10014, opened 2022-11-23
nvidia: update to 515.86.01

mtelka:


components/openindiana/nvidia/Makefile
```diff
@@ -59,7 +60,7 @@ $(INSTALL_$(MK_BITS)):	$(BUILD_$(MK_BITS))
 	$(CP) -a $(SOURCE_DIR)/NVDAgraphics/reloc $(PROTO_DIR)/usr
 	$(TOUCH) $@
```
AndWac, line 62: What about adding
PKG_MACROS += HUMAN_VERSION="$(HUMAN_VERSION)"
here and make use of it in the manifest file?

mtelka: That would need to make it generated into the sample-manifest too because long term plan is to make both the sample-manifest and the real manifest to match (as much as possible).  Now it is simpler the same diff as generated for the sample-manifest just apply to the real manifest too.

PR #10208, opened 2022-12-11
Add - iaxmodem version 1.3.3 (fax modem driver for asterisk)

fritzkink:


mtelka:
After the package installation the service is enabled and goes to maintenance:
```
# svcs -xv
svc:/network/iaxmodem:default (IAX modem)
 State: maintenance since December 13, 2022 at 08:25:06 AM CET
Reason: Restarting too quickly.
   See: http://illumos.org/msg/SMF-8000-L5
   See: man -M /usr/share/man -s 1M iaxmodem
   See: /var/svc/log/network-iaxmodem:default.log
Impact: This service is not running.
#
```
The reason for maintenance is apparently this:
```
# ls -la /var/cores/
total 43138
drwxr-xr-x   2 root     sys            6 Dec 13 08:25 .
drwxr-xr-x  64 root     sys           64 Oct 22 15:40 ..
-rw-------   1 root     root     5368115 Dec 13 08:24 core.iaxmodem.5.1670916270
-rw-------   1 root     root     5384499 Dec 13 08:24 core.iaxmodem.5.1670916299
-rw-------   1 root     root     5384499 Dec 13 08:25 core.iaxmodem.5.1670916301
-rw-------   1 root     root     5384499 Dec 13 08:25 core.iaxmodem.5.1670916303
# mdb -e '::status;::stack' /var/cores/core.iaxmodem.5.1670916303
debugging core file of iaxmodem (64-bit) from il-bld
file: /usr/sbin/iaxmodem
initial argv: /usr/sbin/iaxmodem config.ttyIAX
threading model: native threads
status: process terminated by SIGSEGV (Segmentation Fault), addr=431000
libc.so.1`memset+0xbde()
iaxmodem+0xa90()
main+0x10d()
_start_crt+0x87()
_start+0x18()
#
```
@fritzkink, please change the service so `iaxmodem` is disabled by default after the installation.  Also, root cause and fix of the segfault would be great :-).  Thank you.

PR #10263, opened 2022-12-17
libxml2 - update to version 2.10.3

fritzkink:
to be carefully tested


components/library/libxml2/libxml2-39.p5m
```diff
@@ -20,20 +20,74 @@
...
 set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
-file path=usr/lib/python3.9/vendor-packages/drv_libxml2.py
-file path=usr/lib/python3.9/vendor-packages/libxml2.py
-# Use implicit transform for so ABI rename
-file usr/lib/python3.9/vendor-packages/libxml2mod.so path=usr/lib/python3.9/vendor-packages/libxml2mod.so
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/lib/python3.9/site-packages/drv_libxml2.py
+file path=usr/lib/python3.9/site-packages/libxml2.py
```
AndWac: We want our files in vendor-packages and not in site-packages.

components/library/libxml2/patches/15.lib-run-path-fix.patch
```diff
@@ -0,0 +1,24 @@
...
+-            shift
+-            libs="@XML_LIBS@"
+-        else
+-            libs="@XML_LIBS@ @XML_PRIVATE_LIBS@ @MODULE_PLATFORM_LIBS@ @LIBS@"
+-        fi
+-
+-        if [ "@XML_LIBDIR@" != "-L/usr/lib" -a "@XML_LIBDIR@" != "-L/usr/lib64" ]; then
+-            libs="@XML_LIBDIR@ $libs"
+-        fi
++        libs="@XML_LIBS@"
```
hrosenfeld, line 21: This patch breaks other packages' configure scripts that may want to run "xml2-config --libs --dynamic". While it's fine to ignore --dynamic if it doesn't make sense anymore, it shouldn't be treated as an error if it was specified.

PR #10364, opened 2022-12-28
gcc-10: add convenience package with links into /usr/lib

AndWac:


mtelka:
The `gcc-10-compat-links` and `gcc-11-compat-links` packages are not installable together.  They also clash with `system/library/gcc-4-runtime` and `system/library/g++-4-runtime`.  Shouldn't they use mediators?

AndWac:
I have thought about that but using a mediator would have been a worse solution. The intended use of the new packages are to replace the old gcc-4-runtime and g++-4-runtime packages which were needed for rustup (which is being used to build 3 packages atm: firefox, thunderbird, and clamav). Before these new packages it was mandatory to install gcc-4-runtime and g++-4-runtime to build these packages. Now I have successfully built clamav with the new gcc-10-compat-links package instead. Alas neither firefox nor thunderbird are buildable at the moment. They both fail at similar points but that happens also with gcc-4-runtime and g++-4-runtime in place.

mtelka:
So what exactly is the short term plan?  Obsolete `gcc-4-runtime` and `g++-4-runtime`?   Or obsolete `gcc-10-compat-links` and/or `gcc-11-compat-links`?  Or leave the conflict as is?

I believe we should not introduce new conflicting packages.

BTW, why would a mediator (likely some new one) be a worse solution than conflicting packages?

AndWac:
In order to be able to use a mediator too much work would be needed as I don't want to touch gcc-49 anymore and gcc-7 doesn't build at the moment. The new packages have been introduced by me mainly to be able to check whether rustup works with newer libstdc++ versions and thus we can get rid of older gcc versions. We can make use of the gcc mediator if gcc-49 and its runtimes are gone (or somebody takes the burden to make them behave nicely) and gcc-7 is also gone or fixed.

PR #10475, opened 2023-01-08
valgrind: update to 3.20.0

cstes:


AndWac:
Did you run any tests with the new version?

cstes:
I have managed a few times in the past to run Squeak under Valgrind but not with this new version.   My focus is on Squeak and OpenSmalltalk so I'll close this PR (pull request) and when I rebuild Squeak I'll give valgrind a try again (to see whether it works with the latest valgrind version).    Note that I think also with older valgrind versions there were some issues.

PR #10502, opened 2023-01-10
gst-plugins-bad1: Disable spandsp module because it breaks building on the server

ghost:


components/encumbered/gst-plugins-bad1/Makefile
```diff
@@ -57,6 +57,7 @@ CONFIGURE_OPTIONS += -Dexamples=disabled
 CONFIGURE_OPTIONS += -Dopenexr=enabled
```
AndWac, line 57: You also need to set/increment COMPONENT_REVISION

PR #10558, opened 2023-01-14
libtorrent-rasterbar: Update to 2.0.8

ghost:
Update the libtorrent-rasterbar package to the latest version as this is required for the latest QBittorrent,that I want to push after the Qt6 problem is solved.
The library itself doesn't depend on Qt6,so I guess it's safe to update it already.


ghost:
After thinking about it again,it's probably better to wait with merging until we can also update QBittorrent,as updating the library will likely break the current build.

PR #10575, opened 2023-01-15
smartypants: Add package

ghost:
Add a package for the smartypants python module, which is required for libsoup3.

### DIFF
--- a/components/python/smartypants/Makefile
+++ b/components/python/smartypants/Makefile
@@ -1,0 +1,39 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2023 Niklas Poslovski
+#
+
+BUILD_STYLE = setup.py
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		smartypants
+HUMAN_VERSION =			2.0.1
+COMPONENT_SUMMARY =		smartypants is a Python fork of SmartyPants
+COMPONENT_PROJECT_URL =		https://github.com/leohemsted/smartypants.py
+COMPONENT_SRC =			smartypants.py-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE =		$(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.gz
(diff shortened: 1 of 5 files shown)


mtelka:
@nipos, do you plan to maintain this integration, or is it just one shot?

ghost:
I often check outdated packages in the whole repository and update them,and I recheck those that I've already touched more often,but I don't have a list of packages that I always maintain and check for new versions daily,if you mean something like that.

mtelka:
Python packages are special because they need a rebuild at least once a year (the plan is to integrate yearly every new Python major release).  To make this as easy as possible we need some automation.  This means that all packages (especially new ones) should be integrated using the `python-integrate-project` script.  Please use `python-integrate-project` to add all new Python projects.  If you won't do that then somebody else will need to do that anyway.  I believe you do not want to create more work for somebody else (likely me) than it is needed.

Thank you.

ghost:
I tried to add all python packages using python-integrate-project, but that didn't work for different reasons.
In case of smartypants and typogrify, it failed because it couldn't parse some Python files for dependencies because there were backslashes where the script didn't expect them.
I checked the source of the file in one case and there was a regex which looks totally valid to me.
That means I had to add a pkg.depend.bypass-generate command to the manifest.
As python-integrate-project recreates all files at every run, I took what it had already created and completed it by hand.
In case of gi-docgen, it creates a binary at /usr/bin and installs some manpages, and building in two different versions creates conflicts on these files.
As gi-docgen is a standalone tool and no module that gets imported in other Python projects,there's also no need for building two versions.
So again I took what python-integrate-project had already created and added a PYTHON_VERSIONS = 3.9 line to the Manifest and then completed the package by hand.
Another non-standard thing here with smartypants is that it doesn't have a source archive listed at PyPI, so python-integrate-project fails very early.
I looked at the PyPI page and found the Git repository of the project,which contains tagged releases with the source code.
I hope you can better understand now what I've been doing.
When python-integrate-project gets updated to be able to workaround those problems, especially with dependency detection, I will happily use it.
It could save me a lot of time, but for some projects it unfortunately doesn't work yet.

mtelka:
Please look at `python-integrate-project.conf` and `SINGLE_PYTHON_VERSION`.

Please note that `python-integrate-project` does not support `SINGLE_PYTHON_VERSION` yet so it is okay to integrate such projects without `python-integrate-project`.

Thank you.

ghost:
I've updated the typogrify and markdown pull requests to work with python-integrate-project.
It doesn't work with smartypants,however.
There are just too many non-standard things here.
The archive is from GitHub and the license isn't called LICENSE but COPYING.
No matter what I try,python-integrate-project will leave some variables at TODO or have duplicated variables,once with the value I manually set in python-integrate-project.conf and the other with the automatically generated TODO as value,which leads gmake publish to fail in the end.
Please accept smartypants as it is,at least for now.

PR #10622, opened 2023-01-19
marco: prevent 1.26.1 from crashing related uninitialized pointer

cgrzemba:
Marco crashes in XResQueryClientIds with this pointer uninitialized


mvwieringen:
I myself have frozen marco at 1.26.0 as it indeed crash happy with sunrays. But I think there are more problems with 1.26.1
that upstream is trying to address. e.g, see

https://github.com/mate-desktop/marco/commit/0f7a8c6d19874336040a11cf62a7ecb3c1dd9849
https://github.com/mate-desktop/marco/commit/81ca7129c159bdae1d9ae9a0cd4e196a8354e7ce
https://github.com/mate-desktop/marco/commit/dcd5d21e4c53f1670b33cec78c96bdd3da87037d

probably can be rolled into one patch for the time being. Its really nasty what they have coded assuming
everyone has xres extension without even testing that is available.

mvwieringen:
b.t.w. those have also landed in https://github.com/mate-desktop/marco/commits/1.26

PR #10625, opened 2023-01-19
firefox: update to 102.7.0 and update dependencies

AndWac:


mtelka:
Currently it is not possible to update the `firefox` package to version 102.7.0-2023.0.0.0 when there are `gcc-4-runtime` and `g++-4-runtime` packages already installed, because firefox now requires `gcc-10-compat-links` (see https://github.com/OpenIndiana/oi-userland/pull/10364#issuecomment-1367570778).

It is time to either obsolete `gcc49` or make `gcc-10-compat-links` installable together with `gcc-4-runtime` and `g++-4-runtime`.

PR #10628, opened 2023-01-20
Fix geany-plugins to be build using autoconf-2.71

klausz65:
This fixes the following error during build after configure has taken place:
libtool: Version mismatch error.  This is libtool 2.4.6 Debian-2.4.6-15, but the
libtool: definition of this LT_INIT comes from libtool 2.4.7.
libtool: You should recreate aclocal.m4 with macros from libtool 2.4.6 Debian-2.4.6-15
libtool: and run autoconf again.

### DIFF
--- a/components/editor/geany-plugins/Makefile
+++ b/components/editor/geany-plugins/Makefile
@@ -19,7 +19,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         geany-plugins
 COMPONENT_VERSION=      1.38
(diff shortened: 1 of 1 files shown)


components/editor/geany-plugins/Makefile
```diff
@@ -33,7 +33,7 @@ COMPONENT_LICENSE=      GPLv2
```
AndWac, line 33: Please bump the COMPONENT_REVISION

PR #10631, opened 2023-01-20
Update sane-backends to version 1.1.1

klausz65:
This updates Scanner support in OI to latest avail


AndWac:
Please also update pkg5

PR #10666, opened 2023-01-22
vlc: Update to 3.0.18

ghost:
Update VLC to the latest version 3.0.18


components/encumbered/vlc/Makefile
```diff
@@ -224,7 +224,6 @@ REQUIRED_PACKAGES += library/desktop/gdk-pixbuf
 REQUIRED_PACKAGES += library/desktop/gtk3
 REQUIRED_PACKAGES += library/fribidi
 REQUIRED_PACKAGES += library/glib2
-REQUIRED_PACKAGES += library/libarchive
 REQUIRED_PACKAGES += library/libebml
```
AndWac: Please update your environment and re-run gmake REQUIRED_PACKAGES. The libarchive package have been fixed recently. It had an erroneous dependency on a non-existing iconv package that have never been able to be fulfilled and thus libarchive hasn't been used.

PR #10673, opened 2023-01-23
Make ngrep 64bit only

klausz65:
Just a small one

### DIFF
--- a/components/sysutils/ngrep/Makefile
+++ b/components/sysutils/ngrep/Makefile
@@ -10,13 +10,16 @@
 
 #
 # Copyright 2016 Adam Stevko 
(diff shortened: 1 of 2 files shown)


components/sysutils/ngrep/Makefile
```diff
@@ -10,13 +10,16 @@
 
 #
 # Copyright 2016 Adam Stevko 
+# Copyright 2023 Klaus Ziegler
 #
 
+BUILD_BITS= 64
```
erwinlem, line 16: No need to specify this, 64bit is the default (see 26233532a668304e19972ecf3be66aa929a97e13)

PR #10723, opened 2023-01-29
python/zc.lockfile: rename package to normalize it; obsolete packages for python 2.7 and 3.5

mtelka:


AndWac:
Are you aware that the obsoletion wasn't successful?
 library/python/zc.lockfile-39                     2.0-2020.0.1.1             i-r

mtelka:
Why do you think it was not successful?  It is expected that some renamed packages are still installed during the rebuild/rename process.  Similar situation is with other packages, for example `importlib_metadata-39` or `typing_extensions-39`.

The `library/python/zc.lockfile-39` will get uninstalled automatically once I rebuild `cherrypy`.

AndWac:
Ok, my wording wasn't correct. I should have written "isn't fully effective yet".

PR #10725, opened 2023-01-29
Newer autoconf puts manpages (correctly) in section 4 instead of 7.

erwinlem:
Illumos (and nearly all other oses) put devices in section 4. Section 7 was used by older Solaris/System V for devices.


erwinlem:
The relevant section in the configure script contains this comment :

```
# Solaris 2.0 - 11.3 use SysV man page section numbers, so we
# check for a man page file found in later versions that use
# traditional section numbers instead
```

PR #10728, opened 2023-01-29
Switch meson option b_pie from false to true.

klausz65:
This should turn on gcc options fpic/fPIC for component build done by meson.


alarcher:
@klausz65 Did you really mean PIC or PIE? There used to be issues with some components when built with PIE.

PR #10768, opened 2023-01-30
binutils: create a fake 2.40 version because the real one is broken f…

AndWac:
…or us


AndWac:
The problem showed up in qt5 binaries. It looks like gstrip is broken for us on 2.40.

PR #10885, opened 2023-02-06
re-add 32 bit libs to aspell needed for enchant.

klausz65:
without the 32 bit libraries it's impossible to build enchant.


AndWac:
This PR has 2 commits!

PR #10932, opened 2023-02-09
nvidia: update to 525.89.02

mtelka:


hoewweken:
Hello!
My System with supported Chipset M1200 won't boot with this release. Your system is working? I use legacy boot btw.. (?)
(sorry for hijacking, but i believe this is of common interest)
Regards,
Stephan

PR #10938, opened 2023-02-09
mariadb-106: add back missing include files

mtelka:
These files were removed in #10919.  I do not know why.  I just know that without them it is impossible to compile any software that requires mariadb/mysql.


components/database/mariadb-106/Makefile
```diff
@@ -27,6 +26,7 @@ COMPONENT_NAME=		mariadb
 COMPONENT_MAJOR_VERSION=10.6
 COMPONENT_MINOR_VERSION=12
 COMPONENT_VERSION=	$(COMPONENT_MAJOR_VERSION).$(COMPONENT_MINOR_VERSION)
+COMPONENT_REVISION=	1
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
```
AndWac, line 30: You don't need to set COMPONENT_REVISION as the version hasn't been published.

mtelka: > You don't need to set COMPONENT_REVISION as the version hasn't been published.

Why do you think so?

https://hipster.openindiana.org/jenkins/job/oi-userland/8795/

AndWac: I confused it with php-8_1. That was one of the failing builds after the mariadb upgrade.

PR #10956, opened 2023-02-10
python/compizconfig-python: rebuild for python 3.7 and 3.9 and to get packages for python 2.7 and 3.5 obsoleted

mtelka:


AndWac:
desktop/compiz/ccsm depends on the obsoleted python-3.5 version of this.
It was installed on the build server and let the jenkins job fail during the pkg update phase.

mtelka:
Ooops, sorry.  Should be fixed in #10965.

PR #10962, opened 2023-02-10
netpbm - new package version 10.73.42

fritzkink:


mtelka:
@fritzkink, the `pkg.fmri` in `netpbm-bins.p5m` and `netpbm-libs.p5m` is wrong and causes that all three `.p5m` files produce packages with the same name.

Also, it looks like `CC = gcc` (from `config.mk`) is used during build instead of userland's `CC="/usr/gcc/10/bin/gcc"`.

Please fix both issues.  Thank you.

PR #10968, opened 2023-02-11
del inkscape dependency for lasso not needed.

klausz65:


components/library/lasso/Makefile
```diff
@@ -54,10 +54,6 @@ CONFIGURE_OPTIONS += LDFLAGS=-L$(OPENSSL_LIBDIR)
 
 unexport SHELLOPTS
 
-# Manually added dependencies
-# inkscape is needed for png file creation:
-REQUIRED_PACKAGES += image/editor/inkscape
-
 # Auto-generated dependencies
```
AndWac, line 57: The comment is right: inkscape is needed for the png file creation. If it is not installed the png files were not created.

PR #10998, opened 2023-02-12
gst-editing-services1: Update to 1.22.0

ghost:
Update the gst-editing-services1 package to the latest version 1.22.0.
Needs #10989 and #10990 merged first.


AndWac:
There is a missing dependency on gst-plugins-bad1:

Run-time dependency gstreamer-plugins-bad-1.0 found: NO (tried pkgconfig and cmake)

tests/check/meson.build:42:2: ERROR: No default provided for dependency <NotFoundDependency gstreamer-plugins-bad-1.0: False>, which is not pkg-config, cmake, or config-tool based.

PR #11040, opened 2023-02-16
Obsolete net-snmp-27, Python3 into snmp-base.

klausz65:
mainly the history file was wrong, based on advise by Marcel Telka also integrated the Python3 FMRI into the main package for Net-SNMP.


components/sysutils/net-snmp/Makefile
```diff
@@ -48,7 +48,6 @@ include $(WS_MAKE_RULES)/common.mk
 # Our build server doesn't like parallel builds of this package:
 COMPONENT_BUILD_ARGS=-j1
 
-PYV=          $(shell echo $(PYTHON_VERSION) | tr -d .)
 PKG_MACROS += PYV=$(PYV)
 PKG_MACROS += PYVER=$(PYTHON_VERSION)
```
mtelka: Isn't this (both PYV and PYVER) already added via `make-rules/ips.mk`?

AndWac: I think it's only set there if either a -PYVER manifest file exists or SINGLE_PYTHON_VERSION=yes is set.

mtelka: Exactly.  SINGLE_PYTHON_VERSION=yes should be default in this case, I think.

mtelka (approved):

PR #11059, opened 2023-02-17
Incorporate obsoleted illumos packages for Python 2.7

mtelka:


AndWac:
You should fix the commit message (packages misspelled).

mtelka:
Fixed via #11060.

PR #11062, opened 2023-02-17
curl: update to 7.88.0

AndWac:
sample-manifest didn't change


3eka:
Not sure if this is proper place to comment, but anyway:

* It is possible to re-compile curl, so it supports WS and WSS (WebSocket and WebSocketSecure protocols) `--enable-websockets`
* OI provides `c-ares` library already: `library/libcares 1.18.1`, so maybe to add it there too: `--enable-ares`

Regards

PR #11074, opened 2023-02-19
libgnt: get rid of python2 - migrate to python3.

klausz65:


AndWac:
Please also run "gmake REQUIRED_PACKAGES" and update the sample-manifest.
Plus, set name=pkg.human-version value=$(HUMAN_VERSION) is missing from the manifest, too.

AndWac:
sample-manifest and pkg5 still need to be updated

PR #11078, opened 2023-02-19
Python-3.9 rebuild after libgdbm-1.23 update

klausz65:


mtelka:
I suggest to use different wording here.  Without the referred PR number.  For example something like _Python-3.9 rebuild after libgdbm update_.  Thanks.

AndWac:
Changing the title of the PR is not sufficient. Your commit message needs to be fixed, too. This is what we will have in our git commit history...

PR #11084, opened 2023-02-19
Ruby-23 rebuild after libgdbm-1.23 update

klausz65:


AndWac:
This one needs more work

PR #11099, opened 2023-02-19
Mail::SpamAssassin - new perl package for mail spam detection

fritzkink:


mtelka:
1. Why there is spamassassin.xml file when it is not referenced in any `p5m` file?
2. Please do not manually edit automatically generated files, like `Mail-SpamAssassin-PERLVER.p5m`.

fritzkink:
it is very hard to use perl-integrate-module for this kind of package. Mail-SpamAssassin-GENFRAG.p5m was generated automatically (no manual interaction all defined in perl-integrate-module.conf). Mail-SpamAssassin-PERLVER.p5m was just missed to add to commit which I now manually did.

mtelka:
Hmm, my question was not addressed: Why there is `spamassassin.xml` file when it is not referenced in any `p5m` file (or otherwise used)?

PR #11106, opened 2023-02-20
mysql_fdw: fix mariadb / postgress dependencies.

klausz65:


components/database/postgresql-14-mysql_fdw/mysql_fdw.p5m
```diff
@@ -22,7 +23,7 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
-depend type=require fmri=database/mariadb-101/library
+depend type=require fmri=database/mariadb-103/library
```
AndWac, line 27: mariadb-103 doesn't seem to be right. Our default is mariadb-10.6 nowadays and it's referenced in the Makefile and in pkg5.

PR #11138, opened 2023-02-22
Add Regexp::Common perl module

fritzkink:


mtelka (changes requested):
This module would benefit if [perl-version-convert](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/tools/perl-version-convert) is updated to convert versions like 2017060201 to 2017.6.2.1.  Strictly speaking, it is not needed, but if it is done then version numbers would be far easier to read (and remember).

Thank you.

mtelka (approved):
> This module would benefit if [perl-version-convert](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/tools/perl-version-convert) is updated to convert versions like 2017060201 to 2017.6.2.1. Strictly speaking, it is not needed, but if it is done then version numbers would be far easier to read (and remember).

It was implemented by #11131, so this one is ready for merge.

PR #11157, opened 2023-02-23
Fix libevent to be build without python2.

klausz65:


components/library/libevent/manifests/sample-manifest.p5m
```diff
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2022 <contributor>
+# Copyright 2023 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
```
AndWac, line 18: Please add this line also in the real manifest file.

klausz65:
This line is already in the main manifest:
klausz@oibuild % grep HUMAN libevent.p5m

PR #11167, opened 2023-02-24
Correct pkglint path names for Perl in userland.py

klausz65:


mtelka:
Interesting.  Could you please describe what problem is this going to fix?  Thank you.

klausz65:
I don't know what is fixed by this change, it's just plain wrong since we don't have any 32 bit perl runtime any more.

mtelka:
If you do not know what is going to be fixed by this change, how could you be sure you are not introducing some regression?  I do not like changes with unclear impact and justification...

The only justification I can give you is the fact that this change is in use on SPARC since:
Jan 11  2022 illumos-gate has been build inside oi-userland about 4-5 times using it and that I've also produced 264 Perl modules with it. In fact 99.9% of all SPARC packages have been renewed since then and therefore have all passed pkglint.

fritzkink:
This change is a must and pretty simple to understand. Just execute perl -V and check the native @INC pathes:
  @INC:
    /usr/perl5/site_perl/5.36/i86pc-solaris-thread-multi-64
    /usr/perl5/site_perl/5.36
    /usr/perl5/vendor_perl/5.36/i86pc-solaris-thread-multi-64
    /usr/perl5/vendor_perl/5.36
    /usr/perl5/5.36/lib/i86pc-solaris-thread-multi-64
    /usr/perl5/5.36/lib
as you can see the expression *solaris-64* will never match.

mtelka:
I'm sorry, I probably miss something obvious.  I understand that the `perl -V` output does not match lines in the `userland.py` file, but I do not understand how this implies that we must change the `userland.py` file.  Please do not get me wrong: I'm not trying to say we should keep the `userland.py` file as it is now.  I just try to understand _why_ we need the change and _what consequences_ such a change would cause.

Let me give an example:  if you look at https://src.illumos.org/source/search?project=illumos-gate&full=interupt you'll find several occurrences of word _interupt_, which is obviously a typo (missing _r_), but this does not mean that all occurrences of word _interupt_ in the illumos-gate should be replaced by _interrupt_.  There is at least one _interupt_ that must stay as it is.

IOW, if there is something that looks wrong it does not mean it must be changed without further justification and/or explanation.

Disclaimer:  Feel free to ignore me if this is really obvious.  I do not need to know and understand everything.  I'm just curios why nobody complained earlier if this change is a must.  Is this code ever executed?

Thanks.

PR #11181, opened 2023-02-25
python/python-rapidjson: rename package to normalize it; update to 1.9

mtelka:


components/python/python-rapidjson/pkg5
```diff
@@ -5,13 +5,13 @@
...
+        "system/library/gcc-10-runtime"
     ],
     "fmris": [
-        "library/python/rapidjson-37",
-        "library/python/rapidjson-39",
-        "library/python/rapidjson"
+        "library/python/python-rapidjson-37",
+        "library/python/python-rapidjson-39",
+        "library/python/python-rapidjson"
     ],
```
AndWac, line 15: Do we really need this kind of normalisation? There is already python in the fmri. Does it need to be there twice?

mtelka: > Do we really need this kind of normalisation? There is already python in the fmri. Does it need to be there twice?

Yes we do.  The package name must match project name at pypi.org otherwise automatic detection of dependencies won't work.

In this particular case there is no other package (except pkg) depending on `python-rapidjson` project, so strictly speaking we do not need it now, but:
1. it is easier to normalize than not,
1. in future, if there is a project that needs `python-rapidjson` we would need to normalize anyway.

mtelka: Also, there are other packages with double python in the fmri:
```
$ pkg list 'library/python/python*'
library/python/python-compizconfig                0.8.18-2023.0.0.2          i--
library/python/python-compizconfig-37             0.8.18-2023.0.0.2          i--
library/python/python-compizconfig-39             0.8.18-2023.0.0.2          i--
library/python/python-dateutil                    2.8.2-2023.0.0.0           i--
library/python/python-dateutil-37                 2.8.2-2023.0.0.0           i--
library/python/python-dateutil-39                 2.8.2-2023.0.0.0           i--
library/python/python-dbus                        1.2.18-2022.0.0.2          i--
library/python/python-dbus-35                     1.2.18-2022.0.0.2          i--
library/python/python-dbus-37                     1.2.18-2022.0.0.2          i--
library/python/python-dbus-39                     1.2.18-2022.0.0.2          i--
library/python/python-dotenv                      0.21.1-2023.0.0.0          i--
library/python/python-dotenv-37                   0.21.1-2023.0.0.0          i--
library/python/python-dotenv-39                   0.21.1-2023.0.0.0          i--
library/python/python-gst                         1.22.0-2023.0.0.0          i--
library/python/python-gst-39                      1.22.0-2023.0.0.0          i--
library/python/python-magic                       0.4.27-2023.0.0.0          i--
library/python/python-magic-37                    0.4.27-2023.0.0.0          i--
library/python/python-magic-39                    0.4.27-2023.0.0.0          i--
library/python/python-xlib                        0.33-2023.0.0.0            i--
library/python/python-xlib-37                     0.33-2023.0.0.0            i--
library/python/python-xlib-39                     0.33-2023.0.0.0            i--
$
```

AndWac: And what are you going to do about pkg's dependencies? We already have a dangling dependency on library/python/jsonrpclib-39 which is dangerous.

mtelka: > And what are you going to do about pkg's dependencies? We already have a dangling dependency on library/python/jsonrpclib-39 which is dangerous.

Why do you think it is dangerous?

I plan to rebuild pkg to remove dependency on renamed packages.  But this is low priority now, because there is nothing wrong on dependency on renamed packages.  Also, I'm glad I postponed such rebuild because otherwise I would need to do it twice :-).

AndWac: Cleaning the repository is imminent. The publish times begin to annoy.

mtelka: > Cleaning the repository is imminent. The publish times begin to annoy.

Renamed packages are not a problem for repo cleanup, because repo cleanup does not remove them.  The problem for repo cleanup are only packages that were obsoleted, but some other packages still needs them.  Currently we do have three such packages:
```
library/python/notify2-35                         0.3.1-2020.0.1.1           i--
library/python/setuptools-27                      44.1.1-2020.0.1.0          i--
library/python/setuptools-35                      50.3.2-2020.0.1.1          i--
```

PR #11236, opened 2023-03-03
libreoffice: update to 7.5.1.2

AndWac:
Work mostly done by Geoff Weiss


iigs:
I had tested enable-python=no to CONFIGURE_OPTIONS += --with-system-python and noticed libreoffice writer was prone to crashing.

The test I recommend to verify the python bindings are stable is to test librelogo.  See https://help.libreoffice.org/7.5/en-US/text/swriter/librelogo/LibreLogo.html

librelogo is baked in libreoffice if python is enabled during compiling.

It will take me a day or two to test this PR to see if this is stable or not for me.

64-bit libreoffice 7.3 and 7.4 were unusable for me when I used --with-system-python.  I don't recall testing 32-bit to compare.  With 7.5, I started without enabling python and it was stable.

iigs:
I find this version unstable.  This is the test that I use to show the python bindings cause a crash:
* Start LibreOffice Writer
* Select the View menu, Toolbar hierarchic menu, and  make sure Logo is checked.
* There is a Logo command lineedit box in the toolbar
* Enter this in the logo command line:
   LT 90 FD 100 RT 100 FD 190
* Press enter

If things worked, there be shape that looks like an L .
Instead, libreoffice crashes.

If I was to make a guess, something may be broken  with UNO. That would explain the crash in the test and this issue may be a problem with python uno.  This is the relevant lines from the test:
/lib/amd64/libc.so.1'__sighndlr+0x6 [0x7fffaf3502d6]
/lib/amd64/libc.so.1'call_user_handler+0x1db [0x7fffaf34319b]
[0xe6d4090]
$(@D)/instdir/program/libuno_cppu.so.3'_ZN4cppu17_copyConstructAnyEP8_uno_AnyPvP33_typelib_TypeDescriptionReferenceP24_typelib_TypeDescriptionPFvS2_EP12_uno_Mapping+0x815 [0x7fffa42583f5]

I am not familiar enough to read  this, but it looks like uno_cpp caused a fatal exception signal.  Could this be caused by a problem in one of the files in files/gcc3_solaris_x86-64/ that was contributed to provide 64-bit support for OI?
I am pretty sure I had python binding support working for 32-bit LibreOffice on OI (I no longer have a BE to verify this).

This was the reason why I decided to not include python support in 7.5.1.2.  It would be nice to have both python and java bindings enabled to have better compatibility with other platforms and support more third party extensions.

AndWac:
Ok, I have disabled python for now. We should investigate this further if time permits.

PR #11260, opened 2023-03-05
libmad: Cleanup and modernize the Makefile

ghost:
Cleanup and modernize the Makefile for libmad,replace IPS_COMPONENT_VERSION with HUMAN_VERSION and switch to HTTPS for the website.
The library itself doesn't change,there is no newer version available.


AndWac:
While you are at it: please also update the sample-manifest and the manifest (HUMAN_VERSION is missing) and re-run gmake REQUIRED_PACKAGES

PR #11301, opened 2023-03-09
APR: distribute only 64 bit libraries and binaries.

klausz65:
3 more #PRs to follow that one to make the Apache web engine 32 bit clean.
Next ones have to be apr-util, subversion followed by Apache itself.


AndWac:
Please provide also an updated sample-manifest.

PR #11303, opened 2023-03-10
gcc-7: install-strip should install libobjc

mtelka:
This PR intentionally won't trigger rebuild of gcc-7 for two reasons:
1. We do not need the rebuild/republish of gcc-7 now because this PR just fixes the build and otherwise changes nothing in the final package.
2. gcc-7 is currently not buildable using (the default) gcc-10.  Possible solutions (untested):

- add `GCC_VERSION = 7`, or
- switch gcc-7 to 64-bit, or
- force to compile with `-m32` flag.


AndWac:
GCC_VERSION = 7 is not sufficient as I already tried that.

PR #11216, opened 2023-02-27
python/py3c: move from library/py3c; rename package to normalize it; update to 1.4

mtelka:

PR #11311, opened 2023-03-10
subversion: rebuild 64bit drop 32bit dependencies.

klausz65:
Please do not merge this commit before #PR11310 has been merged.

### DIFF
--- a/components/developer/subversion/Makefile
+++ b/components/developer/subversion/Makefile
@@ -21,7 +21,7 @@
 
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
-# Copyright (c) 2021 Klaus Ziegler
+# Copyright (c) 2023 Klaus Ziegler
 #
 
 include ../../../make-rules/shared-macros.mk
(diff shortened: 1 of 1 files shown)


components/developer/subversion/Makefile
```diff
@@ -151,7 +149,7 @@ COMPONENT_TEST_MASTER = \
 	$(COMPONENT_TEST_RESULTS_DIR)/results-$(MACH64).master
 
 # Build dependencies
-REQUIRED_PACKAGES += library/py3c
+REQUIRED_PACKAGES += library/python/py3c-39
```
AndWac, line 153: I think this line should be replaced with
PYTHON_REQUIRED_PACKAGE += library/python/pyc3c

components/developer/subversion/Makefile
```diff
@@ -151,7 +149,7 @@ COMPONENT_TEST_MASTER = \
 	$(COMPONENT_TEST_RESULTS_DIR)/results-$(MACH64).master
 
 # Build dependencies
-REQUIRED_PACKAGES += library/py3c
+PYTHON_REQUIRED_PACKAGE += library/python/py3c-39
```
AndWac, line 153: No, the -39 needs to be removed here because the python version will be automatically determined and added.

AndWac:
This doesn't build here because of #11216:
/export/home/andreas/oi-userland/components/developer/subversion/subversion-1.14.2/subversion/bindings/swig/python/libsvn_swig_py/swigutil_py3c.h:47:10: fatal error: py3c.h: No such file or directory
   47 | #include <py3c.h>
      |          ^~~~~~~~

PR #11313, opened 2023-03-10
pan: add system/library/math needed on SPARC

klausz65:


components/desktop/pan/Makefile
```diff
@@ -56,3 +52,4 @@ REQUIRED_PACKAGES += library/gnutls-3
 REQUIRED_PACKAGES += library/spell-checking/enchant
 REQUIRED_PACKAGES += library/zlib
 REQUIRED_PACKAGES += system/library
+REQUIRED_PACKAGES += system/library/math
```
AndWac: You should move the new line before the generated code and add an appropriate comment. Otherwise the next time gmake REQUIRED_PACKAGES will be used on x86 the line gets removed again.

PR #11328, opened 2023-03-11
 Hyperscan - a high-performance multiple regex matching library (newp…

fritzkink:
…ackage v5.4.1)


AndWac:
Doesn't build here:
-- Build date: 2023-03-11
CMake Error at CMakeLists.txt:97 (message):
  Ragel state machine compiler not found


-- Configuring incomplete, errors occurred!

fritzkink:
ah, need to add required package ragel manually

PR #11329, opened 2023-03-11
asterisk - update to version 20.2.0

fritzkink:


AndWac:
Did you run gmake REQUIRED_PACKAGES?
If I do it here, I'll get a new dependency:

+REQUIRED_PACKAGES += system/library/libupnp

fritzkink:
this package/option gets implicitly configured once the package is already on the system. I installed libupnp before doing again a make REQUIRED_PACKAGE and now it is also added to required packages.

fritzkink:
I'll update accordingly.

PR #11354, opened 2023-03-13
pgtcl: drop 32bit add SPARC vis instruction option

klausz65:
Please do not merge this change before PR #11351 has been merged.


AndWac:
Please also provide an updated pkg5 file.

PR #11364, opened 2023-03-14
openexr: make architecture independent.

klausz65:


components/library/openexr/Makefile
```diff
@@ -39,6 +40,7 @@ COMPONENT_FMRI=         library/openexr
 COMPONENT_CLASSIFICATION=System/Libraries
 COMPONENT_LICENSE=      BSD
 COMPONENT_LICENSE_FILE= LICENSE.md
+COMPONENT_BUILD_ARGS= -j10
```
AndWac: Please remove that line. You don't need to set the value in the Makefile anyways. You can set this in your environment.

PR #11389, opened 2023-03-17
gcc-11: solaris (and illumos) have no /proc/%d/task/

mtelka:


Toasterson:
Pinging @citrus-it  Since I didn't see this patch in Omnios, If it's there and I am blind sorry.  Can you mention where you get GCC patches from? We sync them over the distro border. If you have already done so plese not it in the PR. Thanks for the contribution.

mtelka:
See #11304 for more details.

Toasterson (approved):
LGTM in principle. I want to give some people in LLVM project time to respond and merge this once people have seen it.

mtelka:
You mean something like https://github.com/llvm/llvm-project/issues/61320?  It is referenced in the patch file...

Toasterson:
Yes I was refering to that issue. We can merge simply if it gets irgnored. But having some people look into it would be nice.

citrus-it:
> Pinging @citrus-it Since I didn't see this patch in Omnios, If it's there and I am blind sorry. Can you mention where you get GCC patches from? We sync them over the distro border. If you have already done so plese not it in the PR. Thanks for the contribution.

As far as I can tell, this code is not used on Solaris - it's only called from libsanitizer/sanitizer_common/sanitizer_stoptheworld_linux_libcdep.cpp which is linux-specific.

If we wanted to add an illumos stop-the-world implementation, we could do so using `/proc/<pid>/lwp`, which is the illumos parallel to the Linux task list.

OmniOS patches come from branches on https://github.com/illumos/gcc wherever possible (these are the compiler branches endorsed by the illumos project), and from https://github.com/omniosorg/gcc otherwise.

Toasterson:
Thanks for all the comments on all the channels.

PR #11393, opened 2023-03-17
libX11: don't use hardcoded arch. names in path.

klausz65:


components/x11/libX11/Makefile
```diff
@@ -19,7 +19,7 @@ include $(WS_MAKE_RULES)/x11.mk
 
 COMPONENT_NAME=     libX11
 COMPONENT_VERSION=  1.7.5
-COMPONENT_REVISION =	1
+COMPONENT_REVISION =	2
```
mtelka: The change in libx11.p5m is basically no-op (for intel platform) and does not need package rebuild because resulting package will be exactly same as rev. 1.  Because of that you do not need to force package rebuild by changing the `Makefile`.

Thanks.

mtelka (changes requested):

components/x11/libX11/Makefile
klausz65: But as it is it does never ever publish on SPARC again, and Andreas always want's to have a uprev if you change anything.

mtelka: > But as it is it does never ever publish on SPARC again,

Why not?  Just run `gmake publish` in `components/x11/libX11` and it will be published.  I do not know what infrastructure you use for building SPARC packages, but if it is similar to official OpenIndiana one, then if the previous build by jenkins failed, you do not need to bump revision to get the package rebuilt.

> and Andreas always want's to have a uprev if you change anything.

You need to uprev only in a case the package is going to be rebuild.  In this case we do not need to rebuild the package so we do not need to uprev.

klausz65: Please ask Andreas - I don't make rules. And according to infrastructure: I use a v240/T5440/T2000 and quite some more SPARCs to test on, and oi-userland are totally the same. Except that I don't have a Jenkins Infrastructure, and of course that previous Jenkins job does not fail as it is x86/amd64 based - but if you publish on a SPARC and have amd64 in one of the manifests and or Makefiles set it fails - of course!

PR #11440, opened 2023-03-23
pyproject.mk: pyproject_installer should compile site-packages too

mtelka:


mtelka:
This is needed to make sure the `compileall` does not depend on order of `COMPONENT_POST_INSTALL_ACTION` lines in the `pyproject.mk` file.  Currently the order is okay, but once we obsolete Python 3.7 the order will (very likely) change and we would see boostrapping process (i.e. building of first handful packages) to fail.

PR #11465, opened 2023-03-26
scientific/gdl: rebuild to fix metadata.

klausz65:


components/scientific/gdl/Makefile
```diff
@@ -44,6 +44,7 @@ CMAKE_OPTIONS += -DUDUNITS2=OFF
 CMAKE_OPTIONS += -DGRIB=OFF
 CMAKE_OPTIONS += -DGLPK=OFF
 CMAKE_OPTIONS += -DSHAPELIB=OFF
+CMAKE_OPTIONS += -DQHULL=OFF
```
AndWac: qhull is available so why do you disable it?

components/scientific/gdl/Makefile
```diff
@@ -44,6 +44,7 @@ CMAKE_OPTIONS += -DUDUNITS2=OFF
 CMAKE_OPTIONS += -DGRIB=OFF
```
AndWac: Please set COMPONENT_REVISION

PR #11495, opened 2023-03-30
transforms/generate-cleanup: transform $(PYVER) for usr/bin files from Python packages only

mtelka:


AndWac:
Something went wrong, building (latest) vim got into trouble:
pkgmogrify: File /export2/home/andreas/oi-userland/transforms/generate-cleanup, line 88: transform (<transform file path=usr/bin/[^/]+-3\.\d+$ -> default tmp.fmri >) has 'default' operation syntax error
if [ "X" != "X" ]; \
then sed  /export2/home/andreas/oi-userland/components/editor/vim/manifests/sample-manifest.p5m \

mtelka:
Yay.  This is because vim does not define `COMPONENT_FMRI`.  I mistakenly assumed it is always defined.  I'm sorry.  I'll fix the issue soon.

PR #11619, opened 2023-04-12
Firefox: update to 112.0

iigs:
h264 video playback broken with most recent OI gstreamer (and/or ffmpeg) packages--it works fine with a BE from March 5.

Regression with bitmap pcf fonts.  Seen on html5test.com . xorg core fonts do not render any more.  Fedora Linux, where the page works fine with Firefox, doesn't list pcf fonts with fc-list ":" even with the xorg core 100dpi fonts installed.  This appears to be an issue with Solaris and OI.  I'll provide a work-around with an update to fontconfig.  I don't think I have the best solution, but it is the only solution I could find with the time I had available.


AndWac:
Am 12.04.23 um 07:43 schrieb iigs:
>
> h264 video playback broken with most recent OI gstreamer (and/or
> ffmpeg) packages--it works fine with a BE from March 5.
>
I think that firefox only needed to be rebuild. h264 is working again
and also did before I updated gstreamer again.

PR #11644, opened 2023-04-16
rsyslog - update to version 8.2302.0 (plus a lot more plugins)

fritzkink:


AndWac:
Doesn't build here:
 CC       omfile_hardened_la-omfile-hardened.lo
/export/home/andreas/oi-userland/components/sysutils/rsyslog/rsyslog-8.2302.0/contrib/omfile-hardened/omfile-hardened.c:64:10: fatal error: omfile.h: No such file or directory
   64 | #include "omfile.h"
      |          ^~~~~~~~~~
compilation terminated.

fritzkink:
sorry forgot to add necessary patch.

PR #11704, opened 2023-04-21
python/ansible: skip collection dependency on ansible itself

mtelka:
This will add `community/skydive` and `cyberark/conjur` to the package.

### DIFF
--- a/components/python/ansible/Makefile
+++ b/components/python/ansible/Makefile
@@ -20,6 +20,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME =		ansible
 HUMAN_VERSION =			7.4.0
+COMPONENT_REVISION =		1
 COMPONENT_SUMMARY =		ansible - Radically simple IT automation
 COMPONENT_PROJECT_URL =		https://ansible.com/
 COMPONENT_ARCHIVE_URL =		\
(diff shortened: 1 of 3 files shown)


AndWac:
I have a question regarding system/management/ansible
Both, ansible (ansible-meta.p5m) and ansible-core (ansible.p5m) define this fmri.
Is that intentional? The one from ansible-core is hidden by the one from ansible and during a repo cleanup the one from ansible-core will be deleted.

mtelka:
> I have a question regarding system/management/ansible Both, ansible (ansible-meta.p5m) and ansible-core (ansible.p5m) define this fmri. Is that intentional?

Yes. This was needed for proper (and easy) switch from monolitic ansible package to ansible-core + ansible as we have now.  The  `ansible-core/ansible.p5m` file will not be used again and will be automatically removed by the `python-integrate-project` script once the script is run again.  I was thinking about removing the `ansible-core/ansible.p5m` file in a separate commit, but it is not worth doing (the file is harmless anyway).

> The one from ansible-core is hidden by the one from ansible and during a repo cleanup the one from ansible-core will be deleted.

That's okay.

So in short: do not worry.  The situation will sort out automatically on next ansible-core rebuild/update which should happen shortly (they release new ansible/ansible-core version every month or so).

PR #11855, opened 2023-04-27
perl-536: update to 5.36.1

mtelka:


mtelka:
I see builds [#9655](https://hipster.openindiana.org/jenkins/job/oi-userland/9655/) and [#9656](https://hipster.openindiana.org/jenkins/job/oi-userland/9656/) failed.  Is there something like `gmake clobber` or `gmake clean` called before the `publish`?  If it is not, then it probably should.  Thank you.

PR #11944, opened 2023-05-03
clamav - update to version 1.1.0

fritzkink:


AndWac:
Verifications failed here:
    validating signature... using https://www.clamav.net/downloads/production/clamav-1.1.0.tar.gz.sig... invalid
gmake: *** [/export/home/andreas/oi-userland/make-rules/prep-download.mk:82: /Daten/ungesichert/OpenIndiana/userland-sources/clamav-1.1.0.tar.gz] Error 1

PR #12048, opened 2023-05-11
codeblocks_svn-source_rev_13046

hoewweken:
codeblocks-source from svn revision 13046
should be a temporary solution until there is a new release (codeblocks::forum: "after the summer holidays")


hoewweken:
Some more Information to be 'almost' complete...

i tried newer svn versions as well, with these i stumble across an error


```c
projectmanagerui.h:248:29: error: 'wxFileSystemWatcher' was not declared in this scope; did you mean 'FileSystemWatcher'?
  248 |             std::unique_ptr<wxFileSystemWatcher> watcher;

``` 
...where somewhere the right file system watcher candidate can not be resolved.
Candidate is the 'generic' wxPollingFileSystemWatcher (or wxInotifyFileSystemWatcher if we had inotify support..)

Something goes wrong with these newer releases, i haven't found a solution so far (lack of C knowledge).

So for now we'll stick with svn revision 13046..

AndWac:
This doesn't build here:
/bin/bash: /export2/home/andreas/oi-userland/components/editor/codeblocks/codeblocks-13046/configure: No such file or directory
gmake: *** [/export2/home/andreas/oi-userland/make-rules/configure.mk:167: /export2/home/andreas/oi-userland/components/editor/codeblocks/build/amd64/.configured] Error 127

The Makefille is missing a bootstrap action, eg. something like
COMPONENT_PREP_ACTION += ( cd $(@D) ; NOCONFIGURE=1 $(CONFIG_SHELL) bootstrap );

hoewweken:
?
The line in the Makefile should be (as in my local copy)
COMPONENT_PRE_CONFIGURE_ACTION = ( cd $(SOURCE_DIR); bash bootstrap )

i'll update the PR.

PR #12071, opened 2023-05-12
firefox: update to113.0.1

iigs:


AndWac:
You are a little late, see #12069

PR #12086, opened 2023-05-13
Make install types arch. independent / add aarch64

klausz65:
and sparcv9


klausz65:
want to do just one more test to create a iso with these new types...

PR #12095, opened 2023-05-14
shared-macros.mk: fix unset BITS for simple BUILD_BITS

AndWac:
This fixes a problem where BITS is not correctly set (eg. for thunderbird).

We still have a problem with BUILD_BITS either 32_and_64 or 64_and_32 where BITS is not available (but needed) in configure phase.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -134,6 +134,11 @@ LINKER =		gcc
 # Possible values are: '32', '64', '32_and_64', '64_and_32', and 'NO_ARCH' (the
 # orderings specify build preference).
 BUILD_BITS ?= 64
+ifeq ($(strip $(BUILD_BITS)),64)
+BITS ?=         64
+else
+BITS ?=         32
+endif
 
 # Based on BUILD_BITS, determine which binaries are preferred for a build.
 # This macro is for the convenience of other make-rules files and should not be


mtelka:
What exactly is the problem with thunderbird?  I suspect there might be something wrong in the thunderbird Makefile, not in `shared-macros.mk`.

mtelka:
I think the problem in thunderbird could be this:
```
--- a/components/mail/thunderbird/Makefile
+++ b/components/mail/thunderbird/Makefile
@@ -145,7 +145,7 @@ $(LANG_FILES):
 
 download prep:: $(LANG_FILES)
 
-CONFIGURE_OPTIONS = CC=$(CC) CXX=$(CXX)
+CONFIGURE_OPTIONS += CC=$(CC) CXX=$(CXX)
 
 ENV +=  CFLAGS="$(CFLAGS)"
 ENV +=  CXXFLAGS="$(CXXFLAGS)"
```

AndWac:
No, even with your proposed change the build is running into an error when BITS is not explicitly set.
With my proposed fix thunderbird builds fine without BITS being set.
Nevertheless, I don't have a fix for my real problem: When I try to build avari-0.8 with qt5 enabled configure fails because it cannot find qt5 libs because the paths depend on BITS which is unset.

AndWac:
You can check the difference in the thunderbird package folder.
Without this change:╰─➤  gmake print-BITS
BITS= (origin: undefined, flavor: undefined)

With this fix:
╰─➤  gmake print-BITS
BITS=64 (origin: file, flavor: recursive)

AndWac:
BITS is set too late in shared-macos.mk for use in the configure phase of a package build.

mtelka:
Without this change I see:
```
$ cd components/mail/thunderbird/
$ gmake print-BITS
BITS=64 (origin: file, flavor: recursive)
$
```
Because `BITS` is set directly in thunderbird's Makefile.

OTOH, `BITS` is designed to be set only during real build (or configure phase, etc.).  See https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/configure.mk#L152 and https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/make-rules/shared-macros.mk#L430.  If it is set out of that scope it could break other components or features of build framework.

It is not set too late for configure.  It is set directly for configure in `configure.mk`!

If you feel you need to do something with `BITS` then in almost all cases there is a problem in component's Makefile that needs to be root caused and fixed.

If you show me what is the real problem with thunderbird (or other component) then I can try to look at it.  Please note that unset `BITS` is not the real problem.

mtelka:
> When I try to build avari-0.8 with qt5 enabled configure fails because it cannot find qt5 libs because the paths depend on BITS which is unset.

How can I reproduce this problem?

AndWac:
When you remove BITS=64 (which has been introduced as a workaround for this problem) from thunderbird its build will fail even with your proposed fix. With this patch the build succeeds even with BITS=64 removed and without your proposed patch.

AndWac:
I have created PR #12097 to show my actual problem I want to solve with qt5 being used for avahi.

AndWac:
Regarding setting BITS in configure.mk:
It is exactly what I wrote: the assignment is too late as it is done after the fact (when configure is done).
Thus, BITS is unset during configure.

mtelka:
Please wait.  I'm working on it...

mtelka:
The problem is caused by this line in thunderbird Makefile:
```
ENV := ulimit -n 16384; $(ENV)
```
The `:=` assignment is causing the `ENV` evaluation is done earlier than designed.

In addition to the problem the line is causing, it does not increase the limit, because the default limit is 64k.  So the line seems to be plain evil.

I'll propose a PR with the fix soon.

PR #12103, opened 2023-05-15
Add initial SPARC support for mariadb.

klausz65:


AndWac:
Please update the package dependencies. We have changed the default for openssl from 1.0 to 1.1 and this impacts a new mariadb-103 build:

/export/home/andreas/oi-userland/components/database/mariadb-103/build/manifest-i386-mariadb-103.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libcrypto.so.1.1 \
        pkg.debug.depend.reason=usr/mariadb/10.3/bin/aria_chk \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=usr/gcc/10/lib/amd64 \

klausz65:
updated

PR #12123, opened 2023-05-18
Add openssl-3.1

Toasterson:
This is a draft As I also need to make a follow up PR to rebuild the dependants.


mtelka:
Could you please elaborate what are dependants of newly added package?  Thank you.

components/library/openssl/openssl-3.1/Makefile
```diff
@@ -0,0 +1,140 @@
...
+BUILD_BITS= 32_and_64
+include ../../../../make-rules/shared-macros.mk
+
+COMPONENT_VERSION_SHORT= 3.1
+COMPONENT_NAME=	openssl-$(COMPONENT_VERSION_SHORT)
+
+# When new version of OpenSSL comes in, you must update both COMPONENT_VERSION
+# and IPS_COMPONENT_VERSION.
+COMPONENT_VERSION=	3.1.0
+HUMAN_VERSION=		3.1.0
```
mtelka: When `HUMAN_VERSION == COMPONENT_VERSION` you do not need to specify both.  `COMPONENT_VERSION` is enough.

components/library/openssl/openssl-3.1/Makefile
```diff
@@ -0,0 +1,140 @@
...
+#
+
+BUILD_BITS= 32_and_64
+include ../../../../make-rules/shared-macros.mk
+
+COMPONENT_VERSION_SHORT= 3.1
+COMPONENT_NAME=	openssl-$(COMPONENT_VERSION_SHORT)
+
+# When new version of OpenSSL comes in, you must update both COMPONENT_VERSION
+# and IPS_COMPONENT_VERSION.
```
mtelka: This comment deserves update, or removal.

components/library/openssl/openssl-3.1/patches/043-x86_asm_illegal_subtraction.patch
```diff
@@ -0,0 +1,32 @@
...
+# are fine.
+#
+# There is an upstream bug that matches this:
+#   https://github.com/openssl/openssl/issues/1715
+#
+# However it was closed with advice to use newer/better (i.e. GNU) assembler.
+#
+# This is tracked as enhancement for Oracle Developer Studio:
+#
+#     30005149 treat undefined subtraction operand in leal instruction as relocation entry
```
mtelka: I think we do not need this patch since it looks Studio specific.

components/library/openssl/openssl-3.1/patches/042-bn_sqr_mont.patch
```diff
@@ -0,0 +1,13 @@
+See ../../common/patches/042-bn_sqr_mont.patch for details.
```
mtelka: I think we do not have such file.

components/library/openssl/openssl-3.1/patches/051-conf-ISA-expand.patch
```diff
@@ -0,0 +1,16 @@
+--- openssl-3.0.8/crypto/conf/conf_api.c	2023-02-07 13:43:33.000000000 +0000
```
mtelka, line 4: Why is this patch needed?

components/library/openssl/openssl-3.1/patches/libs.patch
```diff
@@ -0,0 +1,63 @@
+--- openssl-3.1.0/Configurations/10-main.conf.~1~	2023-03-14 14:59:07.000000000 +0200
```
mtelka, line 21: Why is this patch needed?

mtelka (changes requested):

components/library/openssl/openssl-3.1/patches/libs.patch
Toasterson: Added comment

components/library/openssl/openssl-3.1/patches/051-conf-ISA-expand.patch
Toasterson: Added comment

components/library/openssl/openssl-3.1/patches/042-bn_sqr_mont.patch
Toasterson: copied original comment from solaris-userland

components/library/openssl/openssl-3.1/patches/043-x86_asm_illegal_subtraction.patch
Toasterson: Correct this might be an issue on sparc but I am not looking into too many things here. Removed for now.

components/library/openssl/openssl-3.1/Makefile
Toasterson: Removed

components/library/openssl/openssl-3.1/Makefile
Toasterson: Fixed

Toasterson:
ANdreas wants to merge as fast as possible so I haven't looked into packages that depend on openssl. As I am still using API=1.1.1 everything should build fine. We can look into solaris userland for packages that need upgrades. And then look at updating simple packlages faster as we go thanks to our mediators.

mtelka:
Thank you.  I think it is okay now.

PR #12177, opened 2023-05-22
openssl-3.1: add SPARC support, fix pkgdepend.

klausz65:
Just one line up in the manifest was the pkgdepend hassel with libz.so.1


Toasterson:
Ah, so much to the path was in the output of pkgdepend.... thanks for the fix. LGTM

Toasterson (approved):

PR #12178, opened 2023-05-22
rapidfuzz: SPARC support / exceptions in manifest

klausz65:


components/python/rapidfuzz/rapidfuzz-PYVER.p5m
```diff
@@ -74,12 +74,12 @@ file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/distance/_initialize_
 file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/distance/_initialize_py.py
 file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/distance/metrics_cpp.pyi
 file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/distance/metrics_cpp.so
-file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/distance/metrics_cpp_avx2.so
+file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/distance/metrics_cpp_avx2.so variant.arch=i386
```
mtelka, line 77: Please do not edit automatically generated files.  Such changes will be lost.

components/python/rapidfuzz/rapidfuzz-PYVER.p5m
```diff
@@ -36,7 +36,7 @@ file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/__pyinstaller/hook-ra
 file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/__pyinstaller/test_rapidfuzz_packaging.py
 file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/_common_py.py
 file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/_feature_detector.py
-file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/_feature_detector_cpp.so
+file path=usr/lib/python$(PYVER)/vendor-packages/rapidfuzz/_feature_detector_cpp.so variant.arch=i386
```
mtelka, line 39: Are you sure rapidfuzz works properly without this file?

mtelka (changes requested):

components/python/rapidfuzz/rapidfuzz-PYVER.p5m
klausz65: How would one differentiate between architectures then, the sample-manifest is automatically created by: gmake sample-manifest and the purpose of the real manifest is there to exactly do such things - to differentiate, we used to do this for quite much non-python packages.

components/python/rapidfuzz/rapidfuzz-PYVER.p5m
klausz65: At least I was able to build python/cleo which depends on rapidfuzz.

mtelka: That proves nothing because I believe that even empty rapidfuzz package would allow you to successfully build cleo.  rapidfuzz is needed to run cleo, not to build cleo.

components/python/rapidfuzz/rapidfuzz-PYVER.p5m
mtelka: Yes, it works like that (unfortunately) for non-python (and non-perl) packages, but not for python packages integrated using `python-integrate-project`.  I'm sorry, but such change will be lost during next rapidfuzz rebuild - when either new rapidfuzz is released or we add support for new Python major version.

To do tweaks like this properly you should change `python-integrate-project.conf` accordingly.  In this case you should look at `%hook-manifest%`.

PR #12234, opened 2023-05-29
libmicrohttpd: removed unneeded old workarounds

Karlson2k:
Several changes:
* Dropped unneeded rebuild of autotools files
* Removed enforcement of `-lnsl -lsocket` libs, they are detected automatically by configure
* Removed forced fallback to socketpair instead of EventFD. EventFD has long been fixed in the illumos kernel, fallback is not needed anymore.
* Removed disable of epoll functionality. epoll has long been fixed in the illumos kernel, disabling is not needed anymore.
* Added configure option to ensure that HTTPS is detected and built.
* Removed not needed libgcrypt library. It was required for proper initialisation of GnuTLS in early 2.x days.

Tests results file was updated, but it has no difference. This is an additional proof that workarounds are not needed.

### DIFF
--- a/components/library/libmicrohttpd/Makefile
+++ b/components/library/libmicrohttpd/Makefile
@@ -11,6 +11,7 @@
 #
 # Copyright 2014 Josef 'Jeff' Sipek <jeffpc@josefsipek.net>
 # Copyright 2019 Michal Nowak
+# Copyright 2023 Karlson2k (Evgeny Grin) <k2k@narod.ru>
 #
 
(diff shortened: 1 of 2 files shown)


AndWac:
You PR contains two unrelated deletions in other packages! Please remove them.

Karlson2k:
> You PR contains two unrelated deletions in other packages! Please remove them.

Fixed.
It was unnoticed leftovers of git-for-windows processing (the problem is `aux` used as a directory name).

PR #12288, opened 2023-06-04
libxml2: correct vendor_packages install bug.

klausz65:
Move old python-libxml versions to history file. Remove --with-icu conigure option to get rid of C++ dependencies, causes too much trouble with other software which is needed by libxml2. Oracle also doesn't use this option at all.


AndWac:
I am about to rebuild libxml2 after an icu upgrade. This will collide with it. Please wait until the icu upgrade is done.

PR #12291, opened 2023-06-04
libxml2: correct vendor_packages install bug.

klausz65:
move old python-libxml versions to history file, get rif of --with-icu option to get rid of C++ dependencies.

### DIFF
--- a/components/library/libxml2/Makefile
+++ b/components/library/libxml2/Makefile
@@ -31,6 +31,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libxml2
 COMPONENT_VERSION=	2.10.4
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	Libxml2 is the XML C parser and toolkit developed for the Gnome project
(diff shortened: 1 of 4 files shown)


components/library/libxml2/Makefile
```diff
@@ -63,18 +64,18 @@ CONFIGURE_OPTIONS += --sysconfdir=$(ETCDIR)
 CONFIGURE_OPTIONS += --localstatedir="$(VARDIR)"
 CONFIGURE_OPTIONS += --disable-static
 CONFIGURE_OPTIONS += --with-pic
-CONFIGURE_OPTIONS += --with-icu
+# dont't re-enable this option again without good reasons,
+# it looks like it is breaking the packaging system: pkg
+# CONFIGURE_OPTIONS += --with-icu
 CONFIGURE_OPTIONS += --with-ftp
```
AndWac, line 73: Please change the warning. The dependency doesn't break pkg per se but only during upgrades of icu.
Furthermore, I would expect that the dependency will vanish. In your Makefile and pkg5 file the dependency is still there.

PR #12325, opened 2023-06-09
openmpi - fix duplicate man pages

fritzkink:
remove man pages which are already provided by the /library/mpich/gcc| package.


alarcher:
@fritzkink Thanks for doing this, I will test the new version on the lab's illumos compute nodes.

PR #12332, opened 2023-06-10
sendmail - rebuild because of updated libicu

fritzkink:


AndWac:
This has already been rebuild in #12319 . Do we need to rebuild twice?

fritzkink:
hmm don't know did not work for me after update. Was still linked to the old version.

AndWac:
This has already been rebuild. If it fails then there must be something else which we have to investigate.

PR #12394, opened 2023-06-14
a2ps: Update to 4.15.4

ghost:
Update a2ps to the latest version 4.15.4 and drop 32bit.
Please re-run tests before merging, for me most fail but that looks like a issue specific to my computer (memory fault, segfault).


AndWac:
No, alas I have also lots of core dumps running the tests. This happens since around 4.15.1.

PR #12593, opened 2023-07-03
runtime/perl: rebuild

mtelka:
Something went wrong.  There is no package with `COMPONENT_REVISION = 3` in the repo so we need to force a rebuild.

### DIFF
--- a/components/meta-packages/perl/Makefile
+++ b/components/meta-packages/perl/Makefile
@@ -18,7 +18,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME =		perl
 COMPONENT_VERSION =		$(PERL_VERSION)
-COMPONENT_REVISION =		3
+COMPONENT_REVISION =		4
 COMPONENT_SUMMARY =		Perl - Highly capable, feature-rich programming language
 COMPONENT_PROJECT_URL =		https://www.perl.org/
 COMPONENT_FMRI =		runtime/perl


mtelka:
I'm afraid [this](https://hipster.openindiana.org/jenkins/job/oi-userland/9690/) again failed to rebuild the `runtime/perl` package (with `COMPONENT_REVISION = 4`) and simply republished the old one with `COMPONENT_REVISION = 2`.  Apparently, the build server does something weird.  This needs to be investigated and fixed.  In the meantime, could you please manually rebuild the package with `COMPONENT_REVISION = 4`?  Thank you.

mtelka:
Is `gmake clobber` or `gmake clean` ran when the jenkins build/publish job is started (or at the end of the previous job run)?  If not, then this likely causes the problem.  Please make sure the cleanup is done at the beginning of the job.  Thank you.

PS: [Build #9691](https://hipster.openindiana.org/logs/oi-userland/9691/meta-packages.perl.publish.log) again produced `runtime/perl@5.36,5.11-2023.0.0.2` instead of `runtime/perl@5.36,5.11-2023.0.0.4` since the cleanup step is missing.

mtelka:
OTOH, it is strange.  I just tried this in `components/meta-packages/perl`:
```
env - gmake publish
sed -i -e 's/\(COMPONENT_REVISION =.*\)4/\15/' Makefile
env - gmake publish
```
and I've got both `runtime/perl@5.36-2023.0.0.4` and `runtime/perl@5.36-2023.0.0.5` so the cleanup/clobber step is not needed to have this package rebuilt properly.  Something else must cause the problem :-(.

AndWac:
I ran manually gmake clean before I ran the jenkins job.Von meinem iPad gesendetAm 04.07.2023 um 07:40 schrieb Marcel Telka ***@***.***>:﻿
Is gmake clobber or gmake clean ran when the jenkins build/publish job is started (or at the end of the previous job run)?  If not, then this likely causes the problem.  Please make sure the cleanup is done at the beginning of the job.  Thank you.
PS: Build #9691 again produced ***@***.***,5.11-2023.0.0.2 instead of ***@***.***,5.11-2023.0.0.4 since the cleanup step is missing.

mtelka:
I think I reproduced the problem.  It happens when `gmake publish` is started in the top level directory.  When `gmake publish` is executed in component's dir everything works.

mtelka:
Weird.  It looks like the problem is caused by combination of missing clobber/cleanup step and running `gmake publish` in top level directory.  Really strange.

mtelka:
... and since this is in the `components/Makefile`:
```
$(COMPONENT_DIRS):      $(WS_LOGS) setup FORCE
        @+cd $@ && echo "$(TARGET) $@" && \
         $(BASS_O_MATIC) --make $(TARGET) $(LOG)
```
I assume the problem is in the `bass-o-matic` script.

mtelka:
Yes, the outdated files in the `components/meta-packages/perl/build` directory are causing the problem.  Are you sure that after the `gmake clean` you ran manually the `components/meta-packages/perl/build` dir disappeared?

Here are steps how to reproduce the problem:
```
cd components/meta-packages/perl
env - gmake clean
env - gmake publish
sed -i -e 's/\(COMPONENT_REVISION =.*\)./\15/' Makefile
env - gmake publish
sed -i -e 's/\(COMPONENT_REVISION =.*\)./\16/' Makefile
env - gmake publish
```
The 3rd `gmake publish` creates `runtime/perl@5.36,5.11-2023.0.0.5` instead of `runtime/perl@5.36,5.11-2023.0.0.6`.

PR #12787, opened 2023-07-12
rebuild libebml with gcc-10

hoewweken:
one tiny lib left with gcc-7 dependencies..

before
$ ldd /usr/lib/libebml.so.5.0.0 
        libstdc++.so.6 =>        /usr/gcc/7/lib/libstdc++.so.6
        libgcc_s.so.1 =>         /usr/gcc/7/lib/libgcc_s.so.1

after
$ ldd /usr/lib/libebml.so.5.0.0 
        libstdc++.so.6 =>        /usr/gcc/10/lib/libstdc++.so.6
        libgcc_s.so.1 =>         /usr/gcc/10/lib/libgcc_s.so.1

### DIFF
--- a/components/library/libebml/Makefile
+++ b/components/library/libebml/Makefile
@@ -20,6 +20,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         libebml
 COMPONENT_VERSION=      1.4.4
+COMPONENT_REVISION=     1
 COMPONENT_SUMMARY=      Extensible Binary Markup Language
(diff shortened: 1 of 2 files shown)


AndWac:
This PR added a malformed patch to illumos-gate. Alas I didn't recognize it that the patch wasn't belonging to libebml.

hoewweken:
Ouch :-(
i am am very sorry about this!
i will be more exact and prudently cautious with my PRs.
Sorry!

PR #12788, opened 2023-07-12
rebuild libofa with gcc-10

hoewweken:
one more leftover tiny lib with gcc-7 dependencies

after:
$ ldd /usr/lib/libofa.so.0.0.0 
        libfftw3.so.3 =>         /usr/lib/libfftw3.so.3
        libstdc++.so.6 =>        /usr/gcc/10/lib/libstdc++.so.6
        libgcc_s.so.1 =>         /usr/gcc/10/lib/libgcc_s.so.1


AndWac:
I would expect that the pkg5 file also needs to be updated. It should be dirty after either gmake publish or gmake update-metadata. Please add this to your PR's.

PR #12789, opened 2023-07-12
rebuild protobuf

hoewweken:
one more lib with gcc-7-dependencies


AndWac:
Please also update pkg5

PR #12791, opened 2023-07-12
rebuild glu part of mesa

hoewweken:


AndWac:
This PR contains two commits. One of them doesn't belong here.

PR #12805, opened 2023-07-12
perl/Module-Build: rebuild for Perl 5.38

mtelka:


AndWac:
Something is wrong:
  /library/perl-5/module-build-538

        /library/perl-5/module-build-538
pkg install returned 1
gmake: *** [/jenkins/jobs/oi-userland/workspace/make-rules/environment.mk:51: component-environment-prep] Error 1

mtelka:
Fixed in #12816

PR #12806, opened 2023-07-12
rebuild glw part of mesa

hoewweken:
removed motif as we don't have the source anymore in oi-userland

### DIFF
--- a/components/x11/glw/Makefile
+++ b/components/x11/glw/Makefile
@@ -20,6 +20,7 @@ include $(WS_MAKE_RULES)/x11.mk
 
 COMPONENT_NAME= glw
 COMPONENT_VERSION= 8.0.0
+COMPONENT_REVISION= 1
 COMPONENT_SUMMARY= libglw part of the Mesa 3-D Graphics Library
 COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE= $(COMPONENT_NAME)-$(COMPONENT_VERSION).tar.bz2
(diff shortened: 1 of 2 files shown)


components/x11/glw/Makefile
```diff
@@ -35,10 +36,12 @@ PATCH_LEVEL=0
...
 
-CONFIGURE_OPTIONS += --enable-motif
+#CONFIGURE_OPTIONS += --enable-motif
+REQUIRED_PACKAGES += developer/build/autoconf/xorg-macros
+REQUIRED_PACKAGES += x11/header/x11-protocols
 
 # Auto-generated dependencies
-REQUIRED_PACKAGES += library/motif
+# REQUIRED_PACKAGES += library/motif
 REQUIRED_PACKAGES += system/library
```
AndWac, line 45: Did you really run gmake REQUIRED_PACKAGES to create this?

hoewweken: that was manually.
i now did use gmake REQUIRED_PACKAGES

components/x11/glw/Makefile
```diff
@@ -35,10 +36,11 @@ PATCH_LEVEL=0
 
 LDFLAGS= -L$(X11_SERVERLIBS_DIR) -R$(X11_SERVERLIBS_DIR)
 
-CONFIGURE_OPTIONS += --enable-motif
+#CONFIGURE_OPTIONS += --enable-motif
+REQUIRED_PACKAGES += developer/build/autoconf/xorg-macros
```
AndWac, line 41: When you comment out something you should leave a comment why you did it. Otherwise others have to guess.
Especially when someone updates the version such relics aren't helpful. Thus, either through it away or leave it like that with a comment. In any way you should explain why you did it in the commit. Otherwise this looks like an arbitrary change.
Furthermore, when you change the configuration for a package make sure it still builds all files that are in the manifest. You should run gmake sample-manifest for packages you change the configuration for. And also if a package hasn't been updated for a long time. Our environment changes (we add, remove or modify other packages) and the build system might detect more or less packages than earlier.
You will lessen the burden of doing this for me if you write a comment in the PR what you already did. If there is no comment I will need to do that and this is a lot of work for a single person. Alas it is quite seldom that somebody else tests and comments PR's.

hoewweken:
Added a Makefile comment for the deactivation of motif

hoewweken:
sample-manifest didnt change after gmake sample-manifest:

$ diff -u old-manifest.p5m manifests/sample-manifest.p5m 
--- old-manifest.p5m    Fri Jul 14 07:09:52 2023
+++ manifests/sample-manifest.p5m       Fri Jul 14 07:10:10 2023
@@ -10,10 +10,11 @@
 #
 
 #
+# Copyright 2023 <contributor>
 #
 
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
 set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)

PR #12840, opened 2023-07-14
rebuild libidn2

hoewweken:
sample-manifest didnt change.


AndWac:
Already rebuilt with #12819
Note also that this PR contains another unrelated commit which has also already been merged

PR #12874, opened 2023-07-15
gcc-11: add gccgo for SPARC

klausz65:
And remove --enable-float128 option.

### DIFF
--- a/components/developer/gcc-11/Makefile
+++ b/components/developer/gcc-11/Makefile
@@ -18,6 +18,7 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_VERSION= 11.4.0
+COMPONENT_REVISION= 1
 COMPONENT_SRC= gcc-releases-gcc-$(COMPONENT_VERSION)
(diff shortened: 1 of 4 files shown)


components/developer/gcc-11/Makefile
```diff
@@ -34,6 +35,8 @@ GMP_ARCHIVE_HASH=	sha256:eae9326beb4158c386e39a356818031bd28f3124cf915f8c5b1dc4c
 
 include $(WS_MAKE_RULES)/gcc-component.mk
 
+COMPONENT_BUILD_ARGS=	-j4
```
mtelka: Please do not do this.  If it is safe to use parallel build for this component then please set `USE_PARALLEL_BUILD = yes` before the `shared-macros.mk` include.

mtelka (changes requested):

components/developer/gcc-11/Makefile
klausz65: removed

PR #13075, opened 2023-08-01
rebuild sysutils for gcc-10-deps

hoewweken:
this is an automated approach.

i ran a script doing

    find packages that need a rebuild in componentes/library/
    by looking at pkg5 and find "gcc-7"

    patch Makefile by adding or increasing COMPONENT_REVISION

    IF gmake publish && gmake REQUIRED_PACKAGES or ok, then
    git checkout -b , git add, git commit, etc..

like this:
gmake publish && gmake REQUIRED_PACKAGES && git add Makefile && git add pkg5 && git commit -m "$COMMITMESSAGE"

git rebase -i


components/sysutils/wireshark/Makefile
```diff
@@ -34,7 +34,7 @@ MICRO_VERSION=8
 
 COMPONENT_NAME=		wireshark
 COMPONENT_VERSION=	$(MAJOR_VERSION).$(MICRO_VERSION)
-#COMPONENT_REVISION =     1
+COMPONENT_REVISION=     2
 COMPONENT_PROJECT_URL=	https://www.wireshark.org/
```
AndWac, line 38: You have been fooled by the old commented out COMPONENT_REVISION that somebody left himself as a reminder. Please set COMPONENT_REVSION=1 as it is 0 at the moment.

hoewweken: Ah didn't noticed that. Makefile corrected, Re-publish was successful.

PR #13083, opened 2023-08-02
rebuild libraries for gcc-10-deps part ii

hoewweken:
this is an automated approach.

i ran a script doing

    find packages that need a rebuild in componentes/library/
    by looking at pkg5 and find "gcc-7"

    patch Makefile by adding or increasing COMPONENT_REVISION

    IF gmake publish && gmake REQUIRED_PACKAGES or ok, then
    git checkout -b , git add, git commit, etc..

like this:
gmake publish && gmake REQUIRED_PACKAGES && git add Makefile && git add pkg5 && git commit -m "$COMMITMESSAGE"

git rebase -i

### DIFF
--- a/components/library/libcrossguid/Makefile
+++ b/components/library/libcrossguid/Makefile
@@ -19,6 +19,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         libcrossguid
 COMPONENT_VERSION=      1.0.2
+COMPONENT_REVISION= 1
 COMPONENT_FMRI=         library/libcrossguid
 COMPONENT_CLASSIFICATION=System/Libraries
 COMPONENT_PROJECT_URL=  https://github.com/graeme-hill/crossguid/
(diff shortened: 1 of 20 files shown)


hoewweken:
Are these dependencies ok for OpenColorIO: ?
REQUIRED_PACKAGES += library/pystring
REQUIRED_PACKAGES += library/yaml-cpp

components/library/libcrossguid/Makefile
```diff
@@ -51,3 +52,7 @@ COMPONENT_POST_INSTALL_ACTION= \
 REQUIRED_PACKAGES += system/library
 REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
 REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
+# Auto-generated dependencies
```
AndWac, line 52: You'll need to remove the old lines as they are duplicated now.

AndWac:
The new dependencies should be ok if gmake REQUIRED_PACKAGES detects them. Our environment changes permanently, either by changes in illumos-gate or by changed or updated userland packages. Thus, you should run gmake sample-manifest, gmake REQUIRED_PACKAGES and gmake test every time you touch a package. If the sample-manifest changes you'll need to check and update the real manifest probably also. Note also, that your build environment should be as near as possible to the official build server. That includes not only to be on the latest updates but also have almost all packages installed. The build systems (eg. cmake, meson, ...) also change frequently so there might be changes from that direction, too.

PR #13264, opened 2023-08-08
firefox: add patch to support binutils >= 2.41

iigs:
vp8 and vp9 videos still play


mtelka (changes requested):
You should also enable the component back in the `components.ignore` file (see #13248).

PR #13333, opened 2023-08-11
Disable python/mypy

mtelka:
They changed the version git tag and so the tarball changed.  Disable for now.


mtelka:
Build log: https://hipster.openindiana.org/logs/oi-userland/9948/python.mypy.publish.log

PR #13370, opened 2023-08-15
Switching mate-install to allow older installs to update

Toasterson:
and make nvidia the default driver.

This got reported by @hrosenfeld today as it was blocking upgrading.

### DIFF
--- a/components/meta-packages/install-types/includes/server_desktop_amd64
+++ b/components/meta-packages/install-types/includes/server_desktop_amd64
@@ -10,7 +10,7 @@ depend type=require fmri=diagnostic/top
 depend type=require fmri=driver/graphics/agpgart
 depend type=require fmri=driver/graphics/atiatom
 depend type=require fmri=driver/graphics/drm
-depend type=require fmri=driver/graphics/nvidia-470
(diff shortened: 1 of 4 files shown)


components/meta-packages/install-types/includes/server_desktop_amd64
```diff
@@ -10,7 +10,7 @@ depend type=require fmri=diagnostic/top
 depend type=require fmri=driver/graphics/agpgart
 depend type=require fmri=driver/graphics/atiatom
 depend type=require fmri=driver/graphics/drm
-depend type=require fmri=driver/graphics/nvidia-470
+depend type=require-any fmri=driver/graphics/nvidia fmri=driver/graphics/nvidia-515 fmri=driver/graphics/nvidia-510 fmri=driver/graphics/nvidia-495 fmri=driver/graphics/nvidia-470 fmri=driver/graphics/nvidia-460 fmri=driver/graphics/nvidia-390 fmri=driver/graphics/nvidia-340
```
mtelka: Why do we need renamed `nvidia-460`, `nvidia-495`, and `nvidia-510` here?

Toasterson: Oh, I did not notice they where renamed. Removed them and fixed aarch64 in the progress aswell.

mtelka:
I think you should bump `COMPONENT_REVISION` in `Makefile` too.

Toasterson:
You're right thanks

AndWac (approved):

PR #13500, opened 2023-08-23
samqfs: update 5.64.5

cgrzemba:
- all userland services now build in 64bit
	- read LTO loadtimeout from drive to support LTO9
	- keep RPC API compatible with OHSM
	- use GCC10 for build
	- add version to kernel modules


components/openindiana/samqfs/Makefile
```diff
@@ -35,9 +35,9 @@ TEST_TARGET = $(NO_TESTS)
 
 include $(WS_MAKE_RULES)/common.mk
 
-COMPONENT_BUILD_ARGS = DEBUG=no MYSQL_VERSION=mariadb/10.3 SAMQFS_VERSION=$(COMPONENT_VERSION)$(BRANCH_VERSION)
+COMPONENT_BUILD_ARGS = AMD64=yes DEBUG=no MYSQL_VERSION=mariadb/10.3 SAMQFS_VERSION=$(COMPONENT_VERSION)$(BRANCH_VERSION) NO_BUILD_ACSLS=-D_NoACSLS_
```
mtelka: Could you please rebuild this with mariadb 10.6?  Thank you.

mtelka (changes requested):

components/openindiana/samqfs/Makefile
mtelka: With a bit of luck `MYSQL_VERSION=$(MYSQL_IMPLEM)/$(MYSQL_VERSION)` should work, I think.

cgrzemba: works, Thanks!

cgrzemba:
works, Thanks!

PR #13566, opened 2023-08-30
berkeleydb: rebuild with gcc-10

alarcher:
Tested in build zone


AndWac (commented):
Please also update pkg5

PR #13575, opened 2023-08-31
Upgrade fuse to version 1.5

jpandre:
For some reason it is still compiled both as 32 and 64 bits, but only the 64 bit variant is now installed


jpandre:
Closing for squashing and resubmitting

PR #13628, opened 2023-09-06
perl/Mail-SpamAssassin: rebuild for Perl 5.38

mtelka:


mtelka:
This also fixes all issues mentioned in #11099.

PR #13629, opened 2023-09-06
librsvg - update to version 2.55.3

fritzkink:


AndWac:
Running gmake test fails. Thus, results-all.master is empty:
error: the option `Z` is only accepted on the nightly compiler
error: test failed, to rerun pass `--lib`
make[4]: *** [Makefile:1535: check-local] Error 101

PR #13635, opened 2023-09-07
gcc12: add to OI

cgrzemba:
This brings GCC12 to OI userland. Patches are inspired by OmniosCE.
There are an addtional patch 0039 for building GO runtime. This requires to run 'autogen Makefile.def', which generates a new Makefile.in.
Unfortunately this overwrites some changes done before by patches 0020 and 0025.
Please review.


AndWac:
Unlike OmniOS we don't support the series file for patches. Usually we create an order by prefixing the patches with numbers. Thus, you can remove the series file from your PR.

Toasterson (approved):

PR #13758, opened 2023-09-15
gcc12: add some missing files to package

cgrzemba:
There was some header and go files missing.


AndWac:
For every change in package manifests you'll need to change either the COMPONENT_VERSION or COMPONENT_REVISION in the Makefile. Here, you'll need to add COMPONENT_REVISION= 1

cgrzemba:
sorry, forgotten ...

PR #13763, opened 2023-09-15
runtime/perl: switch to Perl 5.38

mtelka:
This is second step of switching to Perl 5.38.  I suggest to merge this as first PR after the illumos-gate is built with #13762 in place.


mtelka:
Everything looks okay.  We are ready to merge this one.

mtelka:
Everything works okay after this.  Thank you!

PR #13830, opened 2023-09-20
qbittorrent: Rebuild after libtorrent-rasterbar update

ghost:
Rebuild the qBittorrent package after updating its dependency libtorrent-rasterbar.
Needs #13829 merged first.


AndWac:
This isn't necessary because the original update hasn't been published yet.

PR #13868, opened 2023-09-21
mpg123: enable large file support for 32-bit libraries

mtelka:


mtelka:
This fixed the following problem detected by new `libsndfile`: https://github.com/libsndfile/libsndfile/issues/966

PR #13910, opened 2023-09-26
firefox: update to 118.0

iigs:
Mozilla added support for binutils 2.41.  removed firefox patch for binutils 2.41


hoewweken:
we already had 118.0.1 (?) maybe its tool late for me this eve ;-)

PR #13961, opened 2023-10-01
developer/objfw: Add build-essential to deps

Midar:


AndWac:
If you want to change something like this you'll need to touch the Makefile otherwise our build server doesn't publish.
Furthermore, you'll need to either change the COMPONENT_VERSION or (like in this case) set or increment COMPONENT_REVISION (which needs to be delelted or reset if you change the COMPONENT_VERSION).

Midar:
I see. Done!

PR #13965, opened 2023-10-01
elinks: Update to 0.16.1.1

ghost:
Update the elinks browser to the latest version 0.16.1.1.


components/web/elinks/Makefile
```diff
@@ -20,37 +20,35 @@
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
 # Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2023, Niklas Poslovski
 #
-
-BUILD_BITS=64
-
+BUILD_STYLE=meson
+USE_OPENSSL30=yes
```
AndWac: There is no USE_OPENSSL30 variable in our make rules.
Use OPENSSL_VERSION=3.1 if you want to be explicit. OpenSSL-3.1 is out momentary default, though.

PR #13974, opened 2023-10-01
clang-17: Remove 05-no-default-libgcc_s-linking.patch

Midar:
libgcc_s must always come before libgcc, so that symbols that are in both are taken from libgcc_s. The only case where libgcc_s must not be included is when creating a purely static binary that uses no shared libraries at all. As soon as any shared library is involved, libgcc_s *must* be used. Not doing so causes symbols to be taken from libgcc, which leads to crashes. In particular, with this patch, unwinding is broken and crashes. Hence, remove it.


Midar:
clang-13 and clang-15 also have the same problem. Does it still make sense to fix those?

mtelka:
Did you run tests?  What are the results?

In any case, this sounds reasonable.

Midar:
`gmake test` is still running, but I already wanted to send the commit for review since I don't know how many hours it will take.

I didn't manage to get the package installed yet because for whatever reason `pkg install -g ~/oi-userland/i386/repo developer/clang-17` keeps installing the one from the repos and not the one I have built. But I'm fairly certain that this is the problem, given what adding `-v` during linking prints.

mtelka:
> clang-13 and clang-15 also have the same problem. Does it still make sense to fix those?

Hopefully all new and/or updated components should use clang-17 for builds.  If they won't or cannot and they will stick with either clang-13 or clang-15 then changing the current (even broken) behavior for them could cause more trouble than be of any help.

Midar:
OK, actually, it only looked like it installed the version from the repos again. With this patch, the updated clang-17 package now works without requiring any hacks for libgcc_s.

mtelka:
> `gmake test` is still running, but I already wanted to send the commit for review since I don't know how many hours it will take.

Then please mark this PR as a draft until you have test results.

> 
> I didn't manage to get the package installed yet because for whatever reason `pkg install -g ~/oi-userland/i386/repo developer/clang-17` keeps installing the one from the repos and not the one I have built. But I'm fairly certain that this is the problem, given what adding `-v` during linking prints.

Maybe `userland-incorporation` is preventing the install of newly built clang package.  If so, you can simply uninstall it.

Midar:
Is there a way to run the tests on CI? Otherwise I'll need to let my notebook run over night, this looks like it could take a while.

mtelka:
> Is there a way to run the tests on CI?

AFAIK, no.

Midar:
Unfortunately I was unable to run the full test suite due to limited resources (as in, I had to cancel because it took forever, trying to run them with gmake -j8 resulted in it rebuilding clang entirely, and then I ran out of disk space, so I gave up).  But some manual tests all look fine and it looks like the test suite only tests LLD, not how the linker is being called if it's not LLD. I'm therefore marking as ready for review.

AndWac:
We should report OmniOS about this problem. The patch has been imported from their build receipt.

PR #13975, opened 2023-10-01
objfw: Remove hacks now that clang is fixed

Midar:
Pending on #13974

### DIFF
--- a/components/library/objfw/Makefile
+++ b/components/library/objfw/Makefile
@@ -30,7 +30,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME = objfw
 COMPONENT_VERSION = 1.0.3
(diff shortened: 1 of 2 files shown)


Midar:
Merged with #13974 so it can be merged in one go.

mtelka:
> Merged with #13974 so it can be merged in one go.

Sorry, this won't work because both components would be compiled concurrently so objfw will be compiled with previous clang-17, not the one without the patch (#13974).

You need the clang-17 patch removal to be merged first, then you need to wait until new clang-17 is built, published and updated on the build server and then finally you can add the workaround removal.

Midar:
Ah, I see. Ok, will remove.

Midar:
Reopening.

Midar:
clang update is live now, marking as ready for review.

PR #14009, opened 2023-10-03
fmt: Update to 10.1.1

ghost:
Update the fmt package to the latest version 10.1.1.

### DIFF
--- a/components/developer/fmtlib/Makefile
+++ b/components/developer/fmtlib/Makefile
@@ -10,21 +10,20 @@
 
 #
 # Copyright 2023 Friedrich Kink
+# Copyright 2023 Niklas Poslovski
 #
 
-BUILD_BITS= 64
 BUILD_STYLE= cmake
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         fmt
-COMPONENT_VERSION=      9.1.0
-COMPONENT_REVISION=	2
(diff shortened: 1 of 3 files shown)


components/developer/fmtlib/test/results-all.master
```diff
@@ -31,13 +31,20 @@
...
-100% tests passed, 0 tests failed out of 20
+90% tests passed, 2 tests failed out of 20
 
+
+The following tests FAILED:
+	  3 - chrono-test (Failed
+	 17 - xchar-test (Failed
+Errors while running CTest
+Output from these tests are in: $(@D)/Testing/Temporary/LastTest.log
+Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```
AndWac: You have two newly failing tests. Did you check why they are failing now?

ghost: Yes,they seem to fail because they expect my time zone to be +01:00 but it's +02:00 because of summer time.

AndWac: Then can you try to add somehing like that to the Makefile and recheck:
COMPONENT_TEST_ENV += TZ=UTC

ghost: That helped, all tests pass now.

PR #14058, opened 2023-10-05
p11-kit: autoreconf needs PATH

mtelka:
BTW, all tests pass here.  Not sure if it is because something changed in past three months in illumos or userland, or because I do have all packages installed, or because I run tests using `env - gmake test`, or there is some other reason.


AndWac:
More likely the reason is this:
andreas::::type=normal;defaultpriv=basic,dtrace_user,file_dac_read,file_dac_write,proc_clock_highres;profiles=Software Installation,CPU Power Management,Service Management,ZFS File System Management,Network Management

PR #14078, opened 2023-10-06
fix ruby-32 fmri

AndWac:


mtelka:
You need to update `pkg5` file too.  And something should be done with this:
```
# pkg list -afv runtime/ruby-26
FMRI                                                                         IFO
pkg://openindiana.org/runtime/ruby-26@3.2.2-2023.0.0.0:20231005T205238Z      i--
pkg://openindiana.org/runtime/ruby-26@2.6.10-2023.0.0.1:20230219T204316Z     ---
#
```

AndWac:
I am working on this. Alas the repo server is quite old and doesn't support pkgrepo remove.

mtelka:
Removal of pkg is not a solution for machines already updated to `runtime/ruby-26@3.2.2`.

mtelka:
I think the proper (and best) solution is simply to obsolete `runtime/ruby-26` because it is EOLed since 2022-04-12 anyway.

AndWac:
graphviz-ruby depends on it and graphviz doesn't build for me at the moment.

AndWac:
User who have already updated can force the old version.

mtelka:
> graphviz-ruby depends on it ...

And few other components lists ruby-26 as build dependency too (for example `developer/dtracetoolkit`).

> ... and graphviz doesn't build for me at the moment.

:-(

mtelka:
> User who have already updated can force the old version.

But since the new package is already incorporated one would need to either uninstall the `userland-incorporation` or do some other `pkg`-fu magic.  If we are unable to remove `runtime/ruby-26@3.2.2` from the repo or obsolete `runtime/ruby-26` soon then some heads-up needs to be sent to users so they are aware.

Anyway, I'm looking at packages requiring `ruby-26` so we can obsolete it as soon as possible.

mtelka:
> Anyway, I'm looking at packages requiring `ruby-26` so we can obsolete it as soon as possible.

Hm, `graphviz` needs a bit more work so as an immediate solution I just propose to rebuild `ruby-26`: #14080

AndWac:
I have planned to remove ruby-26@3.2.2 and the two offending userland-incorporations.

mtelka:
> I have planned to remove ruby-26@3.2.2 and the two offending userland-incorporations.

Okay, that would solve the problem too.

PR #14092, opened 2023-10-06
erlang: update 24.2.2

cgrzemba:
did a test to build ejabberd with the versin 24.2.2 and it wents fine.
Patches are inspired by Oracle solaris-userland.


AndWac:
Please also check the other transforms. It looks like there are more changes needed because of changed paths.

cgrzemba:
Thanks you found this. I replaced the transforms in a upgrade resistant kind.

PR #14150, opened 2023-10-08
python/setuptools-scm: silence a warning

mtelka:


mtelka:
See also: https://github.com/pypa/setuptools_scm/issues/953

PR #14169, opened 2023-10-08
history: switch to pkg build style

mtelka:
Please note that the publish time stays similar.


AndWac:
What are the advantages of this change?

mtelka:
- Unification and simplification of userland build framework
- Removal of `history.mk` -> less code -> less bugs :-)
- `pkg5` will no longer contain obsoleted and removed packages (this is anomaly seen only in history metapackages)

PR #14186, opened 2023-10-10
firefox: update to 118.0.2

iigs:


mtelka (changes requested):
I think you should remove `developer/clang-15` from `REQUIRED_PACKAGES`.

iigs:
Thank you for catching that the clang version was not updated.

PR #14244, opened 2023-10-15
samba - update to version 4.19.1

fritzkink:


components/network/samba/Makefile
```diff
@@ -27,16 +27,16 @@
 
 BUILD_BITS =			64_and_32
 BUILD_STYLE =			waf
-USE_OPENSSL11 =			yes
+OPENSSL_VERSION =		3.1
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 30: Are you sure that OpenSSL is really being used? I don't see a reference in the Makefile or in pkg5.

fritzkink: you are right. samba uses gnutls. reference removed

PR #14278, opened 2023-10-17
trousers: fix license and add PATH for autoreconf

mtelka:


components/library/trousers/trousers.p5m
```diff
@@ -78,16 +76,14 @@ file path=usr/include/tss/tss_error.h
...
 file path=usr/lib/$(MACH64)/libtspi.so.1.2.0
-file path=usr/lib/$(MACH64)/tcsd mode=0555 owner=root
-#file path=usr/lib/libtddl.a
+file path=usr/lib/$(MACH64)/tcsd
 link path=usr/lib/libtspi.so target=libtspi.so.1.2.0
 link path=usr/lib/libtspi.so.1 target=libtspi.so.1.2.0
 file path=usr/lib/libtspi.so.1.2.0
-file path=usr/lib/tcsd mode=0555 owner=root
+file path=usr/lib/tcsd
 file path=usr/share/man/man3/Tspi_ChangeAuth.3
```
AndWac, line 87: Are you sure that this works without the original mode= and owner= attributes?

mtelka: Yes.

Quick check:
```
$ pkg contents -g i386/repo/ -mr library/security/trousers@0.3.11.2-2023.0.0.5:20231017T152234Z | grep 'path=.*tcsd ' | awk '{print $1 " " $7 " " $8 " " $9 " " $10}'
file group=bin mode=0555 owner=root path=usr/lib/amd64/tcsd
file group=bin mode=0555 owner=root path=usr/lib/tcsd
$
```
Why?
Here is owner:
```
$ grep transform.*owner transforms/*
transforms/defaults:<transform dir file -> default owner root>
transforms/generate-cleanup:<transform dir file link hardlink license -> delete owner .*>
$
```
And here is mode (this transform is newly added):
```
$ grep transform components/library/trousers/trousers.p5m 
<transform file path=usr/lib/.*tcsd$ -> default mode 0555>
$
```

PR #14323, opened 2023-10-20
ntp: add patch to use a mutex

AndWac:
- Work by Bill Sommerfeld (https://www.illumos.org/issues/15984)


hart-NTP:
Please see my note in the bug tracker, I don't think this is the right fix.

https://www.illumos.org/issues/15984

PR #14329, opened 2023-10-20
gcc-12: fix go crash, gcc bug 106813

cgrzemba:
Fix for crash gccgo and go.

### DIFF
--- a/components/developer/gcc-12/patches/Bug106813.patch
+++ b/components/developer/gcc-12/patches/Bug106813.patch
@@ -1,0 +1,18 @@
+
+https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106813
+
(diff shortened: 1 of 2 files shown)


components/developer/gcc-12/patches/Bug106813.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/libgo/runtime/go-signal.c b/libgo/runtime/go-signal.c
```
mtelka, line 4: Please add url here pointing to the bug report (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106813).

mtelka (approved):

mtelka:
@cgrzemba, I assume gcc-10 is affected similarly.  Do you plan to fix it too?  Thank you.

cgrzemba:
yes, for test gcc-13 and gcc-10 builds run in the moment.

PR #14351, opened 2023-10-22
cairo - update to version 1.18.0

fritzkink:
also 32-bit support removed


components/library/cairo/patches/03-libsocket.patch
```diff
@@ -0,0 +1,11 @@
...
++++ cairo-1.18.0/meson.build.new	2023-10-22 17:45:17.714076324 +0200
+@@ -21,7 +21,7 @@
+ 
+ # The libtool shared library version stuff.
+ # Try and maintain compatibility with the previous library versioning.
+-cairo_version_sonum = cairo_version_major + 1
++cairo_version_sonum = cairo_version_major + 2
+ cairo_version =  cairo_version_major * 10000 + cairo_version_minor * 100 + cairo_version_micro
+ 
+ if cairo_version_minor % 2 == 1
```
AndWac: Why is it necessary to change the sonum?
The name of the patch file is irritating, too.

fritzkink: resolved by removing superfluous library calls

PR #14630, opened 2023-11-03
Update sysutils/nut to v2.8.1

jimklimov:
* Updated NUT baseline configuration, making direct use of many now configurable tweaks and thus removing earlier recipe patches
* Trying to deliver both python 2.7 and 3.9 binding and UI app (2.7 version checked interactively to still work)
* Revised package structure to depend drivers on respective libs, etc.


jimklimov:
Thinking of it, one point not addressed now (but can experiment in a later PR or by another person): registering USB IDs for UPSes to go via ugen. Perhaps upstream or this recipe should generate something similar to `udev` rules already present in NUT.

For earlier notes, see https://networkupstools.org/docs/solaris-usb.html

Can an horde of such rules be added as a single file (not ad-hoc hacky `cat` into `/etc/...` or a hundred of `rem_drv/add_drv` calls)?

UPDATE: `samqfs.p5m` has `driver` actions that seem like a relevant example, e.g.:
````
driver name=samioc perms="* 0666 root sys"
````

Thus I'd like to generate a series of "driver" actions, or maybe one long action like `driver name=ugen perms="* 0664 root nut" alias=usb463,ffff.* alias=usb665,5161.* alias=... variant.opensolaris.zone=global` (according to https://docs.oracle.com/cd/E26502_01/html/E21383/pkgterms.html example) to register the OS handler for USB UPSes which should be usable by NUT drivers via libusb. Ideally this is something to convert from the list in NUT sources being built, and hot-patched into `build/*.p5m` files (for usb, modbus driver groups), not hard-coded for the particular package version and build.

**Comment by jimklimov** [comment]
Also, got a question on SMF, in case I missed something in the past year or two: did it by chance grow any way to execute installable late-shutdown logic? (like `rc.6/*` scripts allowed, or `systemd-shutdown` hooks)? Namely, I want to integrate the NUT ability to call the UPS driver when the system is otherwise quiesced, to tell the device to cut the power (when shutdown conditions call for that). This needs a way to either let certain daemons to survive past the enforced `killall` phase circa https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/svc/startd/graph.c#L3522 or for the shutdown scripts to  call the `nutshutdown` script perhaps after they remounted R/O and maybe lost most of the filesystem along the way.

One idea I've long mulled about but never got to try, was adding a kernel module for that effect (if they may call userspace programs some way or another) - at least that might be shipped as part of NUT, unlike direct changes to `svc.startd` :)

**Comment by jimklimov** [comment]
NOTE: Can't get the package installation to auto-register the SMF services. Does it not suffice to land a file (or symlink as here) into e.g. `var/svc/manifest/application/nut-server.xml` (should it be specifically `/lib/svc/...`)? Subsequent manual `svccfg import ...` gets them just fine...

UPDATE: changing to `/lib/...` did not help. Using `restart_fmri=svc:/system/manifest-import:default` neither :\

I don't think the problem should be in that I install it from the global zone (`pkg -R /zones/... install ...`) -- it should still have and register the service actions?

UPDATE: a manual `svcadm restart svc:/system/manifest-import:default` also registered the services, so filenames delivered by package do satisfy it. Package installation itself did note `Services to change: 1` and `Services: restart_fmri: svc:/system/manifest-import:default`, according to `restart_fmri=svc:/system/manifest-import:default` in the p5m file. So not sure why it did not fire :\

Per `tail -f /var/svc/log/system-manifest-import\:default.log &` it did not run during package installation (driven from NGZ):
````
root@jimoi:/etc/zones# pkg -R /zones/testnut/root install -v pkg://oi-userland-jim/system/management/nut@2.8.1
           Packages to install:        12
            Services to change:         1
     Estimated space available:  31.48 GB
Estimated space to be consumed: 508.44 MB
          Rebuild boot archive:        No

Changed packages:
oi-userland-jim
  library/python/pynut-39
    None -> 2.8.1-2023.0.0.0
  system/management/nut
    None -> 2.8.1-2023.0.0.0
  system/management/nut/clients
    None -> 2.8.1-2023.0.0.0
  system/management/nut/common
    None -> 2.8.1-2023.0.0.0
  system/management/nut/drivers
    None -> 2.8.1-2023.0.0.0
  system/management/nut/drivers/modbus
    None -> 2.8.1-2023.0.0.0
  system/management/nut/drivers/serial
    None -> 2.8.1-2023.0.0.0
  system/management/nut/drivers/snmp
    None -> 2.8.1-2023.0.0.0
  system/management/nut/drivers/software
    None -> 2.8.1-2023.0.0.0
  system/management/nut/drivers/usb
    None -> 2.8.1-2023.0.0.0
  system/management/nut/libs
    None -> 2.8.1-2023.0.0.0
  system/management/nut/server
    None -> 2.8.1-2023.0.0.0

Services:
  restart_fmri:
    svc:/system/manifest-import:default

Editable files to change:
  Install:
    etc/nut/ups.conf
    etc/nut/upsd.conf
    etc/nut/upsd.users
    etc/nut/upsmon.conf
    etc/nut/upssched.conf
    usr/bin/upssched-cmd
    usr/lib/nut/bin/notifyme

Completed                              12/12       271/271      5.1/5.1      --

Installing new actions                       521/521
Updating package cache                           6/6
````

Probably this succumbs to the documented (https://docs.oracle.com/cd/E26502_01/html/E21383/smfdelivery.html) behavior, due to my "alternate BE" use-case:
> If the package is added to a live system, this action is performed once all packages have been added to the system during that packaging operation. If the package is added to an alternate boot environment, this action is performed during the first boot of that boot environment.

File: components/sysutils/nut/Makefile [hunk 1]
```diff
@@ -101,26 +168,97 @@ CONFIGURE_OPTIONS +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-$(BITS)
...
+endif
+#REQUIRED_PACKAGES += library/python/notify2
 
 # Auto-generated dependencies
 REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
 REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
 REQUIRED_PACKAGES += library/gd
 REQUIRED_PACKAGES += library/libtool/libltdl
+REQUIRED_PACKAGES += library/libmodbus
 REQUIRED_PACKAGES += library/neon
```
AndWac, line 213: Did you really run gmake REQUIRED_PACKAGES? If I do it here, I'll get some differences:
 # Auto-generated dependencies
+PYTHON_REQUIRED_PACKAGES += library/python/pygobject
+PYTHON_REQUIRED_PACKAGES += library/python/pygtk2
+PYTHON_REQUIRED_PACKAGES += runtime/python
 REQUIRED_PACKAGES += library/gd
-REQUIRED_PACKAGES += library/libtool/libltdl
 REQUIRED_PACKAGES += library/libmodbus
+REQUIRED_PACKAGES += library/libtool/libltdl
+REQUIRED_PACKAGES += library/libusb-1
 REQUIRED_PACKAGES += library/neon
+REQUIRED_PACKAGES += library/python/pyqt5
+REQUIRED_PACKAGES += library/security/openssl-31
-REQUIRED_PACKAGES += SUNWcs
 REQUIRED_PACKAGES += system/management/freeipmi
-#REQUIRED_PACKAGES += system/management/nut/libs
 REQUIRED_PACKAGES += system/management/powerman
 REQUIRED_PACKAGES += system/management/snmp/net-snmp

You should avoid editing behind the #Auto-generated dependencies marker. It will be changed by the next gmake REQUIRED_PACKAGES run. If you need to manually add packages then use the area above.

jimklimov: Did at some point, then it was failing (some chicken-and-egg, without proper stuff in the Makefile it could not complete processing of p5m files), so possibly missed on the last loop. Will recheck.

jimklimov: Thinking of it, possibly it was not run as a standalone rule (lately at least) but as part of `gmake -j 20 clean publish sample-manifest REQUIRED_PACKAGES` sort of incantation. Is there something that might require it to be a standalone step?

jimklimov: Revised, some manually added stuff above dropped.

jimklimov: Huh, the smart magic misfired somewhat:
````
  /library/python/pygobject-39
  /library/python/pygtk2-39
````

These are only needed for the 2.x version of UI; the 3.x one is with QT5.

There *might be* no harm adding these to the build environment, but hopefully it would not pull python3 into users of older releases who might just download and run oi-userland to build a new NUT for themselves... if they existed for 3.x.

Comment (jimklimov) [reply components/sysutils/nut/Makefile:213]: Curiously, it left `REQUIRED_PACKAGES += library/python/pyqt5` alone, without magic auto-versioning.

**Comment by jimklimov** [comment]
On the upside, found that something is broken with Jenkins recipe (or rather helper scripts): it reported an invalid build situation, built nothing, and passed green: https://jenkins.openindiana.aurora-opencloud.org/blue/organizations/jenkins/OpenIndiana%20Organisation%2Foi-userland/detail/PR-14630/6/pipeline/20
````

  /library/python/pygobject-39
  /library/python/pygtk2-39

 Startup: Refreshing catalog 'hipster-encumbered' ... Done
 Startup: Refreshing catalog 'openindiana.org' ... Done
 Startup: Refreshing catalog 'userland' ... Done
 Startup: Caching catalogs ... Done

	/library/python/pygtk2-39
	/library/python/pygobject-39

pkg install returned 1

gmake: *** [/ws/jenkins/workspace/rganisation_oi-userland_PR-14630/make-rules/environment.mk:51: component-environment-prep] Error 1

+ rc=2
+ cd /ws/jenkins/workspace/rganisation_oi-userland_PR-14630
+ echo 'jenkinshelper: done with components/sysutils/nut return code 2'
jenkinshelper: done with components/sysutils/nut return code 2
+ exit 0
````

I guess `echo` is the killer of joy (returns `0` on its own).

UPDATE: followed up in #14690 and not related to NUT PR per se.

**Comment by jimklimov** [comment]
Anyhow, that `REQUIRED_PACKAGES` automagic torpedoed the PR :)

As of last push, it fails (and also locally) with `pkgdepend` issues the earlier iteration somehow danced around. If only I could also see why, seems to still mention all the same package names (plus python runtime here, could it be the problem? or the order of these lines?)

* https://jenkins.openindiana.aurora-opencloud.org/blue/organizations/jenkins/OpenIndiana%20Organisation%2Foi-userland/detail/PR-14630/7/pipeline/20
````
/ws/jenkins/workspace/rganisation_oi-userland_PR-14630/components/sysutils/nut/build/manifest-i386-nut-monitor-gui-py2gtk.depend has unresolved dependency '

        pkg.debug.depend.reason=usr/share/nut/nut-monitor/app/NUT-Monitor-py2gtk2 \
        pkg.debug.depend.type=python \
        pkg.debug.depend.file=64/glade.so \
        pkg.debug.depend.file=64/glademodule.so \
        pkg.debug.depend.file=glade.py \
        pkg.debug.depend.file=glade.pyc \
        pkg.debug.depend.file=glade.pyo \
        pkg.debug.depend.file=glade.so \
        pkg.debug.depend.file=glade/__init__.py \
        pkg.debug.depend.file=glademodule.so \
        pkg.debug.depend.path=usr/lib/python2.7/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/lib-dynload/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/lib-old/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/lib-tk/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/plat-sunos5/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/site-packages/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/vendor-packages/gtk \
        pkg.debug.depend.path=usr/lib/python2.7/vendor-packages/gtk-2.0/gtk \
        pkg.debug.depend.path=usr/lib/python27.zip/gtk \
        pkg.debug.depend.path=usr/share/nut/nut-monitor/app/gtk'.

/ws/jenkins/workspace/rganisation_oi-userland_PR-14630/components/sysutils/nut/build/manifest-i386-nut-monitor-gui-py2gtk.depend has unresolved dependency '

        pkg.debug.depend.reason=usr/share/nut/nut-monitor/app/NUT-Monitor-py2gtk2 \
        pkg.debug.depend.type=python \
        pkg.debug.depend.file=64/gtk.so \
        pkg.debug.depend.file=64/gtkmodule.so \
        pkg.debug.depend.file=gtk.py \
        pkg.debug.depend.file=gtk.pyc \
        pkg.debug.depend.file=gtk.pyo \
        pkg.debug.depend.file=gtk.so \
        pkg.debug.depend.file=gtk/__init__.py \
        pkg.debug.depend.file=gtkmodule.so \
        pkg.debug.depend.path=usr/lib/python2.7 \
        pkg.debug.depend.path=usr/lib/python2.7/lib-dynload \
        pkg.debug.depend.path=usr/lib/python2.7/lib-old \
        pkg.debug.depend.path=usr/lib/python2.7/lib-tk \
        pkg.debug.depend.path=usr/lib/python2.7/plat-sunos5 \
        pkg.debug.depend.path=usr/lib/python2.7/site-packages \
        pkg.debug.depend.path=usr/lib/python2.7/vendor-packages \
        pkg.debug.depend.path=usr/lib/python2.7/vendor-packages/gtk-2.0 \
        pkg.debug.depend.path=usr/lib/python27.zip \
        pkg.debug.depend.path=usr/share/nut/nut-monitor/app'.

/ws/jenkins/workspace/rganisation_oi-userland_PR-14630/components/sysutils/nut/build/manifest-i386-nut-monitor-gui-py2gtk.depend has unresolved dependency '

    depend type=require fmri=__TBD pkg.debug.depend.file=gtk/__init__.py \
        pkg.debug.depend.reason=usr/share/nut/nut-monitor/app/NUT-Monitor-py2gtk2 \
        pkg.debug.depend.type=python \
        pkg.debug.depend.path=usr/lib/python2.7 \
        pkg.debug.depend.path=usr/lib/python2.7/lib-dynload \
        pkg.debug.depend.path=usr/lib/python2.7/lib-old \
        pkg.debug.depend.path=usr/lib/python2.7/lib-tk \
        pkg.debug.depend.path=usr/lib/python2.7/plat-sunos5 \
        pkg.debug.depend.path=usr/lib/python2.7/site-packages \
        pkg.debug.depend.path=usr/lib/python2.7/vendor-packages \
        pkg.debug.depend.path=usr/lib/python2.7/vendor-packages/gtk-2.0 \
        pkg.debug.depend.path=usr/lib/python27.zip \
        pkg.debug.depend.path=usr/share/nut/nut-monitor/app'.

gmake: *** [/ws/jenkins/workspace/rganisation_oi-userland_PR-14630/make-rules/ips.mk:517: /ws/jenkins/workspace/rganisation_oi-userland_PR-14630/components/sysutils/nut/build/.resolved-i386] Error 1
````

**Comment by jimklimov** [comment]
Hm, so this time it passed (at least locally) with addition of un-versioned `library/python/pygtk2`. Previously it seemed not needed (was commented away and stuff still worked).

**Comment by jimklimov** [comment]
Also passed Jenkins. Leaving this PR as is - and noting that `gmake REQUIRED_PACKAGES` is a bullet, not always silver :)

Checked the minimal and expected difference if I run it again:
````
...
*** Please edit your Makefile and verify the new or updated content at the end ***

:; git diff
diff --git a/components/sysutils/nut/Makefile b/components/sysutils/nut/Makefile
index 6cddc1dbc1..5405410c2c 100644
--- a/components/sysutils/nut/Makefile
+++ b/components/sysutils/nut/Makefile
@@ -251,9 +251,8 @@ endif
 endif

 # Auto-generated dependencies
-# These two are "discovered" but do not in fact exist for 3.x:
-#PYTHON_REQUIRED_PACKAGES += library/python/pygobject
-#PYTHON_REQUIRED_PACKAGES += library/python/pygtk2
+PYTHON_REQUIRED_PACKAGES += library/python/pygobject
+PYTHON_REQUIRED_PACKAGES += library/python/pygtk2
 PYTHON_REQUIRED_PACKAGES += runtime/python
````

File: components/sysutils/nut/Makefile [hunk 2]
```diff
@@ -10,44 +10,88 @@
...
 
 #
-# Copyright 2016 Jim Klimov
+# Copyright 2016-2023 Jim Klimov
 #
 
 BUILD_BITS= 64
+
+# More details on python versions handled here are commented below
+SINGLE_PYTHON_VERSION=	no
```
mtelka: Why is this needed?  I believe the main problem with this `Makefile` is that you try to fiddle with Python versions.  This is almost never needed and is usually counterproductive.  Also, introducing dependency on Python 2.7 is wrong.  Please do not try to do it.  There is only one package that is allowed to depend on Python 2.7: `gimp`.  Python 2.7 was [EOLed almost 4 years ago](https://devguide.python.org/versions/#unsupported-versions).

jimklimov: I think this got "lost" in discussion on IRC, so rehashing here :)

On one hand, NUT tries to be as backwards compatible as feasible. Literally systems from turn of millennium like Solaris 8 can still build it. As long as they run (and someone goes through the hassle of setting up a build environment to test it), they may be power-protected.

The python bindings are one aspect of that, with the same PyNUT.py module and its test script re-coded in a few places to survive with both python 2.6/2.7 series and the many 3.x series (half a dozen tried).

On another, there is also an interactive client GUI, NUT-Monitor, which was "forked" and exists as either a Py2+GTK2 or a Py3+QT5 app, and a wrapper script to pick the preferred one by default. From what I remember about the conversion discussion, the Py3 version was less functional about something, can't pinpoint what, so if the user's system has python 2.x - that UI version is the default.

The dependency here is "soft", at least for the end-user installations: *IFF* py2.7 etc. is present (not all installations are new every time), the appropriate version of the app and module would be installed. I suppose (but could not check fully due to the mix of publishers locally) that if the user asks to explicitly install the UI client variant "py2gtk", that would pull in python 2.7 and some of its modules (gtk etc.) but this is not default behaviour.

components/sysutils/nut/Makefile
mtelka: > Thinking of it, possibly it was not run as a standalone rule (lately at least) but as part of `gmake -j 20 clean publish sample-manifest REQUIRED_PACKAGES` sort of incantation. Is there something that might require it to be a standalone step?

Please do not use `-j` when invoking `gmake`.  Instead, you can use something like `gmake PARALLEL_JOBS=20 ...` if default `8` is not enough for you.

Yes, you should make sure that `publish`, `sample-manifest`, `REQUIRED_PACKAGES` (and also `test`) targets pass and no files got changed in your component directory after they are run.  If all these targets pass and no files are changed then their order is not so important, but if that's not the case (usually during the component integration development) then the recommended order is this (before going to next step the current step should pass; you can return to any previous step if needed):

1. First make sure `sample-manifest` produces what you want (edit `Makefile` and/or other files then `gmake clobber` then `env - gmake sample-manifest`).
2. `env - gmake test`.
3. Create/update your `*.p5m` file(s) based on `sample-manifest.p5m`.
4. `env - gmake REQUIRED_PACKAGES`.
5. `gmake clobber` then `env - gmake publish`.

components/sysutils/nut/Makefile
mtelka: Please behave here as if there is no Python 2.7 available.  If there is some missing functionality (or some other issue) in NUT when compiled with Python 3.x, then the upstream should fix it.  All Pythons older than 3.8 are dead horses (or dead snakes, to be precise :-)).

jimklimov: Oh well, at least that was an interesting challenge :D

jimklimov:
Preparing an additional commit to remove python-2.x rules (and packages). I would like to keep this PR as such two commits (modernization to 2.8.1 and its flexibility, and dumbing down to distro of the day) in case "someone for whatever reason" would want to restore that variant for their local build from git history.

mtelka:
If you need two commits in the git repo, then you need two PRs.  All commits in single PR are squashed by default during merge.

jimklimov:
Then we can merge this one and post the removal of python 2.7 as the follow-up :)

mtelka:
> Then we can merge this one and post the removal of python 2.7 as the follow-up :)

Sorry, it makes no sense to introduce (possibly broken) Python 2.7 support and then immediately remove it.  I fail to see any benefit from it.  Please squash and forget about Python 2.7.

jimklimov:
Broken, it is not ;) was tested.

jimklimov:
The alleged benefit is that I know some systems stuck with older OI revisions. For whatever (time-constraint) reasons they won't go into a full reinstall, and can't just upgrade in place. Borked something over the years. With this commit in git history, they can check out oi-userland of the older revision that fits the deployments, and cherry-pick the recipe with older python, and build the package locally. Hope this makes sense :)

In the meanwhile, current distro users indeed would not be confused by compatibility with obsolete components.

jimklimov:
Ok, squashed. I realized that I can keep the github source branch with that prequel PR present, as a sort of bookmark to whoever *would* need that recipe variant.

components/sysutils/nut/Makefile
```diff
@@ -10,44 +10,54 @@
 #
 
 #
-# Copyright 2016 Jim Klimov
+# Copyright 2016-2023 Jim Klimov
 #
 
 BUILD_BITS= 64
+
```
mtelka: `BUILD_BITS = 64` is default, please remove.

components/sysutils/nut/Makefile
```diff
@@ -10,44 +10,54 @@
...
 COMPONENT_LICENSE_FILE=	COPYING
 
+# NUT-Monitor UI has two similar implementations, for py2gtk and py3qt5.
+# The PyNUT.py (general binding) module is written to work with both 2.x
+# and 3.x interpreters as the same source file (patched for shebang and
+# installed to different vendor/site locations several times in practice).
+# If for some reason you are interested in getting the python-2.7 variant
+# installed, please check git history shortly before this recipe change
+# (Nov 2023) for details. Current recipe aims for default Python version
+# supported by OI du-jour (assuming 3.x, eventually newer).
```
mtelka: Please update this comment to reflect reality.

components/sysutils/nut/Makefile
```diff
@@ -101,27 +134,85 @@ CONFIGURE_OPTIONS +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-$(BITS)
...
+# to environment (busy ports, congested hsots and timeouts):
+#CONFIGURE_OPTIONS +=	--enable-check-NIT
+# No CPPCHECK here yet:
+CONFIGURE_OPTIONS +=	--disable-cppcheck
 
 CONFIGURE_OPTIONS +=	$(CONFIGURE_OPTIONS.$(BITS))
 
+# Fiddle with packaging:
+# * move man1m pages to target directory (see also patches/07-man.patch)
+# * move python stuff to vendor-packages
```
mtelka, line 173: Why is this needed?

mtelka: Hm, `site-packages` vs. `vendor-packages`.  There are probably better ways how to do it, but okay.

components/sysutils/nut/Makefile
```diff
@@ -101,27 +134,85 @@ CONFIGURE_OPTIONS +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-$(BITS)
...
+        case "$${PYVER}" in \
+        2.*) python$${PYVER} -c "import pygtk;pygtk.require('2.0');import pynotify" >/dev/null || { \
+                echo "Neutering import of pynotify in NUT-Monitor-py2gtk2 for this build with python$${PYVER}" >&2 ; \
+                $(GSED) -i "$(PROTOUSRSHAREDIR)/nut/nut-monitor/app/NUT-Monitor-py2gtk2" \
+                    -e 's,\(\s\)\(import pynotify\),\1return; \#\#\#NOT_HERE_\2,' || exit ; \
+            } ; \
+        esac ; \
+    done;
+
+test: $(TEST_64)
```
mtelka: Is this really needed?

components/sysutils/nut/Makefile
```diff
@@ -101,27 +134,85 @@ CONFIGURE_OPTIONS +=	--with-net-snmp-config=$(USRBINDIR)/net-snmp-config-$(BITS)
...
+
 # Manually added dependencies:
 REQUIRED_PACKAGES += text/asciidoc
 REQUIRED_PACKAGES += library/augeas-tools
 
+### Variants for someone building on older platforms:
```
mtelka: Please remove this comment block.

components/sysutils/nut/library-python-pynut-PYVER.p5m
```diff
@@ -0,0 +1,35 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2023 Jim Klimov
+#
+
+# This file should cover several versions of python (e.g. 2.7 and 3.x series)
```
mtelka: Please remove 2.7 here.

components/sysutils/nut/Makefile
jimklimov: Old recipe versions (by Alex P., I believe) mentioned a failure to install `*.1m` man pages into `man1m`, such comment was in the patch file as well.

Move from site-packages to vendor-packages (and pre-compilation) was picked up from a few existing components, probably `meld`, as I revised existing best practices about python package installations. There are a few hits to `git grep vendor-packages | grep -iw mv`. If "vendor" means OS distro, this makes sense...

components/sysutils/nut/nut-monitor-gui-common.p5m
```diff
@@ -0,0 +1,78 @@
...
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform dir path=usr/share/nut/nut-monitor -> set group bin>
+<transform dir path=usr/share/nut/nut-monitor/(.*) -> set group bin>
+
+### NUT GUI clients, require python2+GTK or python3+QT5
```
mtelka: No.  python2 is not required.

components/sysutils/nut/nut-monitor-gui-py3qt5.p5m
```diff
@@ -0,0 +1,49 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+<transform dir path=usr/share/nut/nut-monitor/app -> set group bin>
+<transform dir path=usr/share/nut/nut-monitor/app/ui -> set group bin>
+
+depend fmri=$(COMPONENT_FMRI)/nut-monitor-gui-common@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+
+### NUT GUI clients, require python2+GTK or python3+QT5
```
mtelka: I think this comment block should be removed.

components/sysutils/nut/Makefile
jimklimov: Not quite: I don't think it runs tests despite the presence (or absence) of this line. But generally, it would be nice to confirm that the in-distro build is valid. Not too hard a requirement for NUT though - its CI farm includes OI and OO.

components/sysutils/nut/nut-monitor-gui-py3qt5.p5m
jimklimov: I went by PYV.3X, py2gtk and such patterns, so missed the non-matched text - sorry for the review stress :)

components/sysutils/nut/Makefile
mtelka: What is OO?

If the line is not needed, then please remove it.  If it is needed, then please add a comment above it why it is needed.  Thank you.

components/sysutils/nut/Makefile
```diff
@@ -10,44 +10,53 @@
...
 COMPONENT_FMRI=	system/management/$(COMPONENT_NAME)
 COMPONENT_CLASSIFICATION=	System/Administration and Configuration
 COMPONENT_LICENSE=	GPLv2+
 COMPONENT_LICENSE_FILE=	COPYING
 
+# NUT-Monitor UI has two similar implementations, for py2gtk and py3qt5.
+# The PyNUT.py (general binding) module is written to work with both 2.x
+# and 3.x interpreters as the same source file (patched for shebang and
+# installed to different vendor/site locations several times in practice).
+# If for some reason you are interested in getting the python-2.7 variant
```
mtelka: No, we are not interested in python-2.7 variant.

components/sysutils/nut/Makefile
```diff
@@ -10,44 +10,53 @@
...
+# locally built and installed, please check commits parked in
+#   https://github.com/OpenIndiana/oi-userland/pull/14692
```
mtelka: Please remove this.  It is useless.

components/sysutils/nut/nut-drivers-modbus.p5m
```diff
@@ -0,0 +1,55 @@
...
+<transform file path=usr/share/man/man1/(.+) -> set action.hash usr/share/man/man1/%<1> >
+<transform file path=usr/share/man/man1m/(.+) -> set action.hash usr/share/man/man1m/%<1> >
+
+dir path=usr/lib/nut/bin group=bin
+
+# The server package also includes some baseline tools, services
+# and configuration files for drivers
+depend fmri=$(COMPONENT_FMRI)/libs@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+#depend type=require fmri=library/libmodbus
```
mtelka: Why this line?

components/sysutils/nut/nut-drivers-neon.p5m
```diff
@@ -0,0 +1,45 @@
...
+<transform file path=usr/share/man/man1/(.+) -> set action.hash usr/share/man/man1/%<1> >
+<transform file path=usr/share/man/man1m/(.+) -> set action.hash usr/share/man/man1m/%<1> >
+
+dir path=usr/lib/nut/bin group=bin
+
+# The server package also includes some baseline tools, services
+# and configuration files for drivers
+depend fmri=$(COMPONENT_FMRI)/libs@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+#depend type=require fmri=library/neon
```
mtelka: Why this?

components/sysutils/nut/nut-drivers-powerman.p5m
```diff
@@ -0,0 +1,48 @@
...
+<transform file path=usr/share/man/man1/(.+) -> set action.hash usr/share/man/man1/%<1> >
+<transform file path=usr/share/man/man1m/(.+) -> set action.hash usr/share/man/man1m/%<1> >
+
+dir path=usr/lib/nut/bin group=bin
+
+# The server package also includes some baseline tools, services
+# and configuration files for drivers
+depend fmri=$(COMPONENT_FMRI)/libs@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+#depend type=require fmri=system/management/powerman
```
mtelka: Why this?

components/sysutils/nut/nut-drivers-snmp.p5m
```diff
@@ -0,0 +1,45 @@
...
+<transform file path=usr/share/man/man1/(.+) -> set action.hash usr/share/man/man1/%<1> >
+<transform file path=usr/share/man/man1m/(.+) -> set action.hash usr/share/man/man1m/%<1> >
+
+dir path=usr/lib/nut/bin group=bin
+
+# The server package also includes some baseline tools, services
+# and configuration files for drivers
+depend fmri=$(COMPONENT_FMRI)/libs@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+#depend type=require fmri=system/management/snmp/net-snmp
```
mtelka: Why this?

components/sysutils/nut/nut-drivers-software.p5m
```diff
@@ -0,0 +1,58 @@
...
+<transform file path=usr/share/man/man1/(.+) -> set action.hash usr/share/man/man1/%<1> >
+<transform file path=usr/share/man/man1m/(.+) -> set action.hash usr/share/man/man1m/%<1> >
+
+dir path=usr/lib/nut/bin group=bin
+
+# The server package also includes some baseline tools, services
+# and configuration files for drivers
+#depend fmri=$(COMPONENT_FMRI)/libs@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+#depend fmri=library/security/openssl fmri=library/security/openssl-31 type=require-any
```
mtelka: Why?

jimklimov:
For those series of "why this line" in driver deps - at first those lines were needed during iterations for pkgdepend to accept the p5m recipes. Later it claimed the dependencies are defined twice, but I did not track down why - due to prerequisites for conditional dependency from `nut.p5m` or due to `REQUIRED_PACKAGES` update. Left them commented to figure out and they stayed there :\

jimklimov:
Posted for next round of diligent comments... If we cross ways in Prague, beer should be on me, right? ;)

jimklimov:
Also, a reminder: integrating NUT with last seconds of uptime during a power-fail shutdown is still an open question for me with SMF systems (to tell an UPS "Now is the time to cut the power!" when it is safe for the protected host). But this may well be a separate feature PR. I have a nagging feeling it would have to go through illumos-gate to extend SMF abilities. Would love to be proven wrong and shown that a vehicle exists already though :)

AndWac:
@mtelka Are you satisfied with this or is there something still not adressed?

jimklimov:
On my side, currently fiddling locally with ideas to generate and package `driver` actions, so USB devices would be seen out of the box as commented above. This can well wait for a follow-up PR because I've no idea when/if this would succeed actually :)

UPDATE: `driver` actions presumed fixed, pushed.

As also commented, would be great to find a way to (again after SYSV init-scripts) handle late-shutdown. But I'm afraid this needs a bright future tech far away today.

mtelka:
> @mtelka Are you satisfied with this or is there something still not adressed?

I think it is mostly okay.  Only two final suggestions:
1. please rename `library-python-pynut-PYVER.p5m` to `library-python-pynut.p5m` because we do not need per-version Python packages (we build for single Python version only).   You could also rename the package to, for example, `system/management/nut/python`, because it does not need to be named `library/python/pynut`, but if you'll keep it `library/python/pynut` then it is okay too.
2. please shorten the commit message; something like 'Update sysutils/nut to v2.8.1' provides the exactly same information with far less words.

Thank you.

jimklimov:
Sounds ok, but tomorrow then :)

As for pynut - it is a client library, conceptually same as a http or telnet client, just happens to be for NUT networked protocol, and happens to be used by NUT-Monitor among potentially other things people can write. So I'd say the python/library pattern is more correct.

Do you anticipate globally supporting more than one python version at the same time in the future?

mtelka:
> As for pynut - it is a client library, conceptually same as a http or telnet client, just happens to be for NUT networked protocol, and happens to be used by NUT-Monitor among potentially other things people can write. So I'd say the python/library pattern is more correct.

Ok, sure.  But then ... is this your `pynut` the same as https://pypi.org/project/pynut/ ?

If not (I assume so), then you should use different name for your package, because `library/python/XXX` package names are automatically assigned to (and reserved for) `XXX` PyPI projects.  So if we will try to package https://pypi.org/project/pynut/ in future then we will have a problem with package name clash.  In addition, the automatic Python dependency detection won't work properly if you hijack `library/python/XXX` namespace without backing such project at PyPI.

If yes, then we should package directly from PyPI if you want to use the `library/python/pynut` package name.

> Do you anticipate globally supporting more than one python version at the same time in the future?

Yes, of course.

jimklimov:
Oh, imposters... No good idea what that project is about, something about various-protocol connections but no relation to NUT :\

I get your point on packaging names, but their history starts from Mar 23, 2022 and PyNUT's across a couple of source file renames - from 2009 when it was contributed into NUT codebase from a third-party project :\

mtelka:
> I get your point on packaging names, but their history starts from Mar 23, 2022 and PyNUT's across a couple of source file renames - from 2009 when it was contributed into NUT codebase from a third-party project :\

Unfortunately, this is not about the history length.  It is purely about naming and namespace.  Similarly, if we would want to upload `pkg` to PyPI we won't be able to use name `pkg` for it there.

jimklimov:
> > Do you anticipate globally supporting more than one python version at the same time in the future?

> Yes, of course.

My understanding of the `filename-PYVER.p5m` is that it is specifically a vehicle for such code, to automatically deliver and name it for particular version(s) (and an umbrella to dispatch dependencies for the installed interpreter versions). So why nuke it then?

Also, "we build for single Python version only" - so which one answer is that correct one? ;)

jimklimov:
> > I get your point on packaging names, but their history starts from Mar 23, 2022 and PyNUT's across a couple of source file renames - from 2009 when it was contributed into NUT codebase from a third-party project :\
> 
> Unfortunately, this is not about the history length. It is purely about naming and namespace. Similarly, if we would want to upload `pkg` to PyPI we won't be able to use name `pkg` for it there.

Yeah, I get that. I suppose for ~15 years nobody bothered to upload it there and pick a unique name from that naming authority :)

https://pypi.org/project/nut2/ claims to be an "API overhaul" of NUT's module, with some renamed methods, last bumped on github some 4 years ago.

Looking at https://packaging.python.org/en/latest/tutorials/packaging-projects/ - probably something could be done about uploading it, maybe as "PyNUTClient" (at least the name is available and matches the main class name). So far I made an account and maybe some year would have time for next steps, to make sense of the doc and see how to wrangle the code stored in NUT codebase into the structure they want :)

UPDATE: Well, took a day of detour, but hopefully got PyPI packages rolling. No idea how it would handle the mismatch of "PyNUT" (per module source) and "PyNUTClient" (pip/metadata, and the primary class), though.

mtelka:
> > > Do you anticipate globally supporting more than one python version at the same time in the future?
> 
> > Yes, of course.
> 
> My understanding of the `filename-PYVER.p5m` is that it is specifically a vehicle for such code, to automatically deliver and name it for particular version(s) (and an umbrella to dispatch dependencies for the installed interpreter versions).

Yes, it is.

> So why nuke it then?

Because nut is built for single Python version, so you'll never have more than one such package.  If you really want to support all Python versions (not only the recent one) then a standalone Python project (integrated using `python-integrate-project`) would be the best option.  You will also need `SINGLE_PYTHON_VERSION=no` too (which is default for standalone Python projects).  I think we can postpone this discussion (and decision) once there is at least one non-NUT package that would need pynut.  If pynut is only internal dependency (i.e. needed by other NUT packages only) then the actual package name does not matter much and also you do not need to support multiple Pythons.  So for now I'd take the easiest path possible: `SINGLE_PYTHON_VERSION=yes` (which you already have because it is default) and unversioned Python package.

> Also, "we build for single Python version only" - so which one answer is that correct one? ;)

We build nut for single Python version only.  PyPI projects are built for all Pythons (which is currently 3.9 only, but that will change in future).

jimklimov:
Well... just to end these clarifications at some point - that module is a project-provided binding, to benefit whoever might want to write a NUT client in Python. As such, its further users are not really known, nor a concern of the upstream project. Sometimes various projects that integrate with NUT make themselves known and get listed to end up on the nut-website, but I can't remember any specific python ones.

So one known, but potentially not exclusively the only one, consumer of this module is the (two implementations of) NUT-Monitor GUI provided nowadays also in NUT codebase. Some 10+ years ago it might have been a separate third-party project, but was since contributed for assimilation by the upstream.

jimklimov:
Hopefully solved the puzzle for `driver` action generation and injection. At least, passes locally.

mtelka:
@jimklimov, I'm not sure what message is used for the merged commit.  I suspect it is PR title instead of included commit message.  Just to be safe I suggest to update the PR title too.  Thank you.

mtelka:
I'm not sure about other lines in the commit message, but at least the following two should be removed.  They describe something that is not visible in the final commit.
```
* ...and adjust again to forfeit python-2.x (squashed)
* apply review comments to clean up the recipe
```
Thank you.

jimklimov:
Updated the title and commit message details.

mtelka:
Most of the [packaging issues](https://hipster.openindiana.org/jenkins/job/oi-userland/10677/consoleText) could be solved by adding this `components/sysutils/nut/history` file:
```
system/management/nut@2.7.4,5.11-2022.0.0.6 system/management/nut/common
system/management/nut-common@2.7.4,5.11-2022.0.0.6 system/management/nut/common
system/management/nut/bins@2.7.4,5.11-2022.0.0.6 system/management/nut/common
```
The only issue that needs closer look is this one:
```
The following packages all deliver driver actions to ugen:

  pkg://openindiana.org/system/management/nut/drivers/usb@2.8.1,5.11-2023.0.0.0:20231109T123723Z
  pkg://openindiana.org/driver/usb/ugen@0.5.11,5.11-2023.0.0.21862:20231107T011116Z
```

mtelka:
Sorry, the following line should **NOT** go to the `components/sysutils/nut/history` file:
> system/management/nut@2.7.4,5.11-2022.0.0.6 system/management/nut/common

mtelka:
@jimklimov could you please fix packaging issues mentioned above?

hoewweken:
i made a test, uninstalled ugen and installed nut, there was an error message:
```

driver (ugen) install failed with return code 1
command run was: /usr/sbin/add_drv -u -i "usb1,0.*" "usb3f0,1.*" "usb3f0,1f01.*" "usb3f0,1f02.*" "usb3f0,1f06.*" "usb3f0,1f08.*" "usb3f0,1f09.*" "usb3f0,1f0a.*" "usb3f0,1fe0.*" "usb3f0,1fe1.*" "usb3f0,1fe2.*" "usb3f0,1fe3.*" "usb3f0,1fe5.*" "usb3f0,1fe6.*" "usb3f0,1fe7.*" "usb3f0,1fe8.*" "usb463,1.*" "usb463,ffff.*" "usb47c,ffff.*" "usb483,35.*" "usb483,a113.*" "usb4b3,1.*" "usb4b4,5500.*" "usb4d8,d004.*" "usb4d8,d005.*" "usb50d,375.*" "usb50d,551.*" "usb50d,750.*" "usb50d,751.*" "usb50d,900.*" "usb50d,910.*" "usb50d,912.*" "usb50d,980.*" "usb50d,f51.*" "usb50d,1100.*" "usb51d,0.*" "usb51d,2.*" "usb51d,3.*" "usb51d,4.*" "usb592,2.*" "usb592,4.*" "usb5b8,0.*" "usb5dd,41b.*" "usb5dd,a011.*" "usb5dd,a0a0.*" "usb665,5161.*" "usb6da,2.*" "usb6da,3.*" "usb6da,4.*" "usb6da,5.*" "usb6da,201.*" "usb6da,601.*" "usb6da,ffff.*" "usb75d,300.*" "usb764,5.*" "usb764,501.*" "usb764,601.*" "usb925,1234.*" "usb9ae,1.*" "usb9ae,1003.*" "usb9ae,1007.*" "usb9ae,1008.*" "usb9ae,1009.*" "usb9ae,1010.*" "usb9ae,1330.*" "usb9ae,2005.*" "usb9ae,2007.*" "usb9ae,2008.*" "usb9ae,2009.*" "usb9ae,2010.*" "usb9ae,2011.*" "usb9ae,2012.*" "usb9ae,2013.*" "usb9ae,2014.*" "usb9ae,3008.*" "usb9ae,3009.*" "usb9ae,3010.*" "usb9ae,3011.*" "usb9ae,3012.*" "usb9ae,3013.*" "usb9ae,3014.*" "usb9ae,3015.*" "usb9ae,3016.*" "usb9ae,3024.*" "usb9ae,4001.*" "usb9ae,4002.*" "usb9ae,4003.*" "usb9ae,4004.*" "usb9ae,4005.*" "usb9ae,4006.*" "usb9ae,4007.*" "usb9ae,4008.*" "usbd9f,1.*" "usbd9f,4.*" "usbd9f,a2.*" "usbd9f,a3.*" "usbd9f,a4.*" "usbd9f,a5.*" "usbd9f,a6.*" "usbf03,1.*" "usb10af,1.*" "usb10af,4.*" "usb10af,8.*" "usb14f0,c9.*" "usb1cb0,32.*" "usb1cb0,35.*" "usb1cb0,38.*" "usb2341,36.*" "usb2341,8036.*" "usb2a03,36.*" "usb2a03,40.*" "usb2a03,8036.*" "usb2a03,8040.*" "usb2b2d,ffff.*" "usb2e51,0.*" "usb2e51,ffff.*" "usb2e66,101.*" "usb2e66,201.*" "usb2e66,202.*" "usb2e66,203.*" "usb2e66,300.*" "usb2e66,302.*" "usb4234,2.*" "usbffff,0.*" -m * 0664 root nut ugen
command output was:
------------------------------------------------------------
b'Cannot find module (ugen).\n' ------------------------------------------------------------

``` 


but, nut found my UPS device.
``` 

# nut-scanner 
Scanning USB bus.
No start IP, skipping SNMP
Scanning XML/HTTP bus.
No start IP, skipping NUT bus (old connect method)
Scanning NUT bus (avahi method).
Failed to create Avahi client: Daemon not running
[nutdev1]
        driver = "apc_modbus"
        port = "auto"
        vendorid = "051D"
        productid = "0003"
        product = "Smart-UPS 750 FW:UPS 09.3 / ID=18"
        serial = "AS1827134173"
        vendor = "American Power Conversion"
        bus = "160"
        device = "003"
        busport = "008"
        ###NOTMATCHED-YET###bcdDevice = "0106"
```

jclulow:
FWIW, I don't expect you're going to be able to deliver aliases to the ugen driver in the package: you're not delivering the driver, after all!

hoewweken:
@jimklimov: I am willing to help testing.. 
i don't seem to get it to work. Do you also have an UPS connected via USB ?
nut-scanner does see my UPS, but driver:UPS1 keeps in maintanace state..
```
$ svcs|grep nut
online          9:54:06 svc:/system/power/nut:default
online         10:01:19 svc:/system/power/nut-driver-enumerator:default
maintenance    10:01:19 svc:/system/power/nut-driver:UPS1
maintenance    10:01:19 svc:/system/power/nut-server:default
maintenance    10:01:24 svc:/system/power/nut-monitor:default
```

jimklimov:
@jclulow : thanks for the comments... given that (according to earlier reading and experience) the correct way to get a NUT run-time user permissions to access the USB device filesystem node is via `add_drv`, as I and some posters before me logged in https://networkupstools.org/docs/solaris-usb.html

On the other hand, IPS driver actions seem to be the way to relay Solaris/illumos driver configuration information in packaging, per https://docs.oracle.com/cd/E26502_01/html/E21383/pkgterms.html#gluea - although indeed I found no clear statements whether multiple packages may tweak it this way. An alternative might be a post-install script (or a run-once method), I guess.

@hoewweken : what did you mean by "uninstalled ugen and installed nut" - did you remove `ugen` as a package? or did you just run `rem_drv` to unload it (with later `add_drv` to re-load it with new entries; with original hope that the driver actions ensure that)? To be clear: `ugen` *IS* needed as the low-level OS handler; it however has to be told to allow NUT to access certain USB devices.

Regarding `apc_modbus` in particular, it needs an USB-capable version of libmodbus (which I think is in upstream PR stage so far - probably not packaged in OI yet). There was a fix for `nut-scanner` after NUT v2.8.1 release to prefer `usbhid-ups` and `nutdrv_qx` for devices whose IDs may be handled by several drivers (but then mileage may vary for vendors who use same IDs for unrelated devices). APC support for USB HID is lame since ~2010, but so far `usbhid-ups` may be better than the one which relies on absent library features :\

What does `/var/svc/log/*UPS1*.log` say?

jclulow:
> @jclulow : thanks for the comments... given that (according to earlier reading and experience) the correct way to get a NUT run-time user permissions to access the USB device filesystem node is via `add_drv`, as I and some posters before me logged in https://networkupstools.org/docs/solaris-usb.html

I've skimmed this document a little and I don't believe it comes to the correct conclusion.  The only time you should need to override the binding with an explicit **ugen(4D)** alias is when the device identifies as some other class than what it should, _and_ that natural binding is to a driver that doesn't expose **ugen(4D)** nodes.

In the case of HID devices, for example, it's no longer necessary to change the binding: our HID driver should expose ugen nodes without requiring reconfiguration.  The fallback driver that binds to nodes with no other driver, which I believe is **usb_mid(4D)**, should also expose ugen nodes.

In the cases where that's not happening, it _may_ be appropriate to rebind to **ugen(4D)**, but only as a workaround for what is really an OS bug.  The nodes should be exposed where necessary by whatever driver is bound there.  I would definitely not ship a bulk, unconsidered set of aliases (no matter the mechanism) that overrides many USB devices.  It's more likely to break things that would have worked already than to help.

If you do want to ship a tool, I would instead write one that takes the list of devices supported by the software and looks at the state of the system to see if any are attached using libdevinfo.  It could print a list of detected devices, and note which drivers are attached to those devices if any.  It could be geared towards easily gathering the information to know whether a device is already expected to work, or if an OS ticket needs to be filed to get things fixed for the attached device.

jimklimov:
@jclulow : Thanks for the detailed answer; commuting now so it may take a while to wrap my head around its nuances :)

Taking a step back to the practical requirements, because maybe the answer is different or I'm even implicitly asking a wrong question up there, so:

An unprivileged daemon running as `nut:nut` should be allowed to read and write device nodes for USB devices (handled by `ugen` on the OS side) with any of those UPS-related 120+ VID:PID combos known to be supported by NUT drivers; not all are HID devices though (some use effectively serial-over-USB shims, etc. - but seemed to work in this arrangement anyway). What is the better way to tell the OS to permit this user/group account such access, persistently (across USB resets, device re-plugs and system reboots)? Is there some dynamic subsystem like `udev`, `upower`, `devd` etc. as used on some other systems?

jclulow:
> An unprivileged daemon running as `nut:nut` should be allowed to read and write device nodes for USB devices (handled by `ugen` on the OS side) with any of those UPS-related 120+ VID:PID combos known to be supported by NUT drivers; not all are HID devices though (some use effectively serial-over-USB shims, etc. - but seemed to work in this arrangement anyway). What is the better way to tell the OS to permit this user/group account such access, persistently (across USB resets, device re-plugs and system reboots)? Is there some dynamic subsystem like `udev`, `upower`, `devd` etc. as used on some other systems?

So, it's not _directly_ applicable here, but the [logindevperm(5)](https://illumos.org/man/5/logindevperm) file is an example of one mechanism that provides a measure of policy based permissions reconfiguration.  That particular mechanism applies at login time, and is not hot plug aware, so it's not exactly useful to you here.

There are several options I can think of initially:

- Rework the program to use a privilege separation architecture where the initial program is not totally unprivileged, but forks an unprivileged child to do most of the work.  The privileged parent process would be responsible for enumerating and opening the device and passing the file descriptors down to the unprivileged child.
- Ship a separate, privileged daemon alongside the existing unprivileged software whose sole job it is to listen to device hot plug sysevents and poll devinfo looking for relevant device nodes as they show up, and then use chown and chmod calls to fix the permissions for your application.
- Ship a small program that would perform the appropriate chown/chmod on the device nodes for a specific permitted USB device, and also an RBAC profile that allows the nut user to invoke it with sufficient privileges as required.  Have the software run this program using pfexec to gain access to a particular device node.

It's conceivable we could provide an appropriate facility in the OS once the requirements are clear, but there's basically no reason you can't prototype the facility you need first in one of the ways above.

jimklimov:
Gimme a couple of years till next big distraction from dayjob and other projects :)

At least currently, some constraints on NUT side would be that it does not actually know which pathname to USB device it is looking at. It asks libusb to enumerate, and takes those descriptors wherever they point (in driver configs a `port` keyword is required, but ignored for USB drivers as of now so is typically `port=auto`). So watching for hotplug events and then (re-)starting potentially impacted drivers might be doable, with opportunistic hope that the watcher changes the same device(s) which NUT driver(s) would then match, but really seems something like the OS should do.

Regarding starting as `root` and dropping privileges: that could be doable, some other NUT daemons do the switcheroo, but then (almost) why not just keep running as that - options like `user=root` are long there and often used as a workaround for permissions problem in general. In particular, drivers try to support reconnection on the fly if the device went AWOL and re-appeared (USB reset, etc.) - this usually involves the OS resetting permissions to `root:root` or whatever default it has, unless something like `udev` (further) switches it to a particular owner and other subsequent reactions. A driver running as `nut` after the privilege drop would not be able to reconnect to a root-owned device then - so would have to know to exit rather than try in such cases (so the root-owned parent, if it lingers, or the SMF service instance would restart it from scratch).

Sounds like a lot of cumbersome private workarounds for what should have been an OS facility...

jimklimov:
Bummer: re-read the `add_drv` manuals and ran some experiments -- at least the simplistic approach taken here with `*` in the `-m` permission option ends up as a blanket `ugen:* 0664 root nut` in `/etc/minor_perm` so indeed not likely what we want generally. And not sure how to list the minors (what values?) to match those supported UPSes...

jimklimov:
@jclulow : as for watching for hotplug and devinfo events - is that a different stream from FS events? My earlier notes bookmarked https://github.com/emcrisostomo/fswatch as a possible way to react to file changes (for `nut-driver-enumerator` service to optionally run as soon as someone saves `ups.conf` - similar to systemd path units). Would it be helpful here for hotplug? Are there any similar facilities in SMF proper by now?

hoewweken:
Hello!

I uninstalled pkg ugen as the pkg install did not allow if ugen is there " ... can not be installed together ..."
Following [jclulow](https://github.com/jclulow) comments it should "work" with usb_hid driver without installed ugen (?)

This is my
```
$ cat /var/svc/log/*UPS1*.log
[ Nov 13 10:01:19 Enabled. ]
[ Nov 13 10:01:19 Executing start method ("/sbin/sh -c 'NUTDEV="`/usr/lib/nut/bin/nut-driver-enumerator.sh --get-device-for-service UPS1`" && [ -n "$NUTDEV" ] || { echo "FATAL: Could not find a NUT device section for service unit UPS1" >&2 ; exit 1 ; } ; SMF_FMRI=svc:/system/power/nut-server:default /usr/share/nut/solaris-smf/method/svc-nut-server create_run_dir && /usr/sbin/upsdrvctl start "$NUTDEV"'"). ]
/sbin/sh: /usr/share/nut/solaris-smf/method/svc-nut-server: cannot execute [Permission denied]
[ Nov 13 10:01:19 Method "start" exited with status 126. ]
... repeat ...
```

jimklimov:
Oh wow, getting quickly back to the working desk then. Somehow worked for me locally a few iterations ago when I started the PR, but there were so many hot-fixes and force-pushes to get it completed, something must have slipped with exec permissions. And the idea with `driver` actions to gain device permissions was apparently incorrect :\

jimklimov:
@hoewweken : sorry about the confusion made here, hoping the revised recipe will fare better (and you would do well to install `ugen` back when eventually installing the updated package).

In the meanwhile, can you please check if `chmod +x /usr/share/nut/solaris-smf/method/*` fixes the service-starting problem and is the only such issue to address here? (I suppose an USB-based driver would have problems starting without `ugen` installed at the moment, but that is "predictable")

hoewweken:
Hello!

It works better now:
```
$ svcs -xv
svc:/system/power/nut-driver:UPS1 (NUT power device driver instance wrapper)
 State: maintenance since 14 November 2023 at 13:39:38 CET
Reason: Start method failed repeatedly, last exited with status 1.
   See: http://illumos.org/msg/SMF-8000-KS
   See: man -M /usr/share/man -s 8 upsdrvctl
   See: man -M /usr/share/man -s 5 ups.conf
   See: /var/svc/log/system-power-nut-driver:UPS1.log
Impact: This service is not running.
```

```
$ svcs -a|grep nut
disabled       13:38:32 svc:/system/power/nut-driver-enumerator:daemon
online         13:39:07 svc:/system/power/nut-driver-enumerator:default
online         13:39:38 svc:/system/power/nut-server:default
online         13:39:38 svc:/system/power/nut-monitor:default
online         13:39:38 svc:/system/power/nut:default
maintenance    13:39:38 svc:/system/power/nut-driver:UPS1
```
maybe i have an USB configuration error in nups.conf..

```
[UPS1]
        driver = "usbhid-ups"
        port = "auto"
        vendorid = "051D"
        productid = "0003"
        product = "Smart-UPS 750 FW:UPS 09.3 / ID=18"
        serial = "AS1827134173"
        vendor = "American Power Conversion"
        bus = "160"
        device = "003"
        busport = "008"

```

jimklimov:
@hoewweken : the config does not look incorrect; what does `/var/log/svc/system-power-nut-driver:UPS1.log` say?

Per earlier discussion, I expect that FS permissions would be a problem, so adding `user="root"` to the UPS1 config section can help...

Or even the lack of `ugen`... If there's no way to force installation of two conflicting packages, that could be solved by the #14822 changes, which drop the `driver=...` actions from NUT package.

hoewweken:
Hi!

```
# cat /var/svc/log/system-power-nut-driver:UPS1.log
...
Network UPS Tools - Generic HID driver 0.52 (2.8.1)
USB communication driver (libusb 1.0) 0.46
libusb1: Could not open any HID devices: insufficient permissions on everything
No matching HID UPS found
upsnotify: failed to notify about state 4: no notification tech defined, will not spam more about it
Driver failed to start (exit status=1)
...
```
after adding user = "root" in /etc/nut/ups.conf i get this:

```[ Nov 15 09:22:50 Enabled. ]
[ Nov 15 09:22:50 Executing start method ("/sbin/sh -c 'NUTDEV="`/usr/lib/nut/bin/nut-driver-enumerator.sh --get-device-for-service UPS1`" && [ -n "$NUTDEV" ] || { echo "FATAL: Could not find a NUT device section for service unit UPS1" >&2 ; exit 1 ; } ; SMF_FMRI=svc:/system/power/nut-server:default /usr/share/nut/solaris-smf/method/svc-nut-server create_run_dir && /usr/sbin/upsdrvctl start "$NUTDEV"'"). ]
Using subdriver: APC HID 0.100
Network UPS Tools - Generic HID driver 0.52 (2.8.1)
USB communication driver (libusb 1.0) 0.46
WARNING: Needed to fix group access to filesystem socket of this driver, but failed; run the driver with more debugging to see how exactly.
Consumers of the socket, such as upsd data server, can fail to interact with the driver and represent the device: /var/db/nut/usbhid-ups-UPS1
Network UPS Tools - UPS driver controller 2.8.1
[ Nov 15 09:22:50 Method "start" exited with status 0. ]
```
All services are running now, however

What about 
```
NUT-Monitor is a graphical application to monitor and manage UPSes connected to a NUT server. This application is written in Python and PyGTK, and uses the [PyNUT](https://networkupstools.org/projects.html#PyNUT) class.
```
.. is it included, i saw the dependency to python (?)
i tried and modified ```NUT_USER=monuser NUT_PASS=pass ~/test_nutclient.py  ``` this seems to work :-)

jimklimov:
@hoewweken : Thanks for the report! (And note that now the updated package recipe was merged, perhaps you can get the non-conflicting set of `nut` and `ugen` again)

Regarding this message:
````
WARNING: Needed to fix group access to filesystem socket of this driver, but failed; run the driver with more debugging to see how exactly.
Consumers of the socket, such as upsd data server, can fail to interact with the driver and represent the device: /var/db/nut/usbhid-ups-UPS1
````

You can use `debug_min=6` in `ups.conf` to bump debugging verbosity - see more at https://github.com/networkupstools/nut/wiki/Changing-NUT-daemon-debug-verbosity

Or try `ls -la /var/db/nut /var/run/nut` for these files in particular. On my test container with the package and a dummy driver, I see the directory owned as `root:nut` with RWX permissions to both user and group, so there should not be any problems about it; more so if the driver runs as root (hm, would it set socket group to nut then?):
````
root@testnut:~# ls -la /var/*/nut
/var/db/nut:
total 3
drwxrwx---   2 root     nut            3 Nov 14 05:35 .
drwxr-xr-x   3 root     sys            3 Nov 14 05:10 ..
srw-rw----   1 nut      nut            0 Nov 14 05:35 dummy-ups-dummy

/var/run/nut:
total 40
drwxrwx---   2 root     nut          325 Nov 14 05:37 .
drwxr-xr-x   5 root     sys         1187 Nov 14 05:13 ..
-rw-r--r--   1 nut      nut            5 Nov 14 05:35 dummy-ups-dummy.pid
-rw-r--r--   1 nut      nut            5 Nov 14 05:35 upsd.pid
-rw-r--r--   1 root     root           5 Nov 14 06:04 upsmon.pid
````

If I add the `user=root` to the device section, I see these permissions in effect:
````
/var/db/nut:
total 3
drwxrwx---   2 root     nut            3 Nov 15 03:35 .
drwxr-xr-x   3 root     sys            3 Nov 14 05:10 ..
srw-rw----   1 root     root           0 Nov 15 03:35 dummy-ups-dummy

/var/run/nut:
total 40
drwxrwx---   2 root     nut          325 Nov 15 03:35 .
drwxr-xr-x   5 root     sys         1187 Nov 14 05:13 ..
-rw-r--r--   1 root     root           6 Nov 15 03:35 dummy-ups-dummy.pid
-rw-r--r--   1 nut      nut            5 Nov 14 05:35 upsd.pid
-rw-r--r--   1 root     root           5 Nov 14 06:04 upsmon.pid
````
...however `upsmon` finds it and `upsc dummy` lists the data points.

**Comment by jimklimov** [comment]
The `test_nutclient.py` is part of PyNUTClient module. The `NUT-Monitor` (UI app) should get installed automatically with NUT if you already have the prerequisite packages (PyNUTClient, python3, its qt5 module etc.), or should pull them in if you ask to install it explicitly. I am not sure it will pull in X11 (at least, does not do so in NUT recipe - might do from Qt5?) but that's what the UI app needs to be practically useful :)

More details at https://github.com/networkupstools/nut/blob/master/scripts/python/app/README.adoc

**Comment by jimklimov** [comment]
For the socket warning, can reproduce here :(
````
   1.008268     [D2] dstate_init: sock /var/db/nut/dummy-ups-dummy open on fd 6
   1.008394     [D1] Group and/or user account for this driver was customized ('root:nut') compared to built-in defaults. Fixing socket '/var/db/nut/dummy-ups-dummy' ownership/access.
   1.008438     [D1] WARNING: opening socket file for stat/chown failed: Operation not supported on transport endpoint
   1.008456     WARNING: Needed to fix group access to filesystem socket of this driver, but failed; run the driver with more debugging to see how exactly.
Consumers of the socket, such as upsd data server, can fail to interact with the driver and represent the device: /var/db/nut/dummy-ups-dummy
````

Seems `fstat()` faults should be handled more flexibly...

Now I wonder how/why `upsd` running as `nut` can still see that socket and talk over it :\
````
   2.009352     sock_connect: enabling asynchronous mode (auto)
   2.009392     [D3] new connection on fd 7
   2.009427     [D6] Driver on /var/db/nut/dummy-ups-dummy is now handling DUMPALL with 1 args
````

**Comment by jimklimov** [comment]
Finally, regarding `upsnotify: failed to notify about state 2: no notification tech defined, will not spam more about it`: with some service frameworks (systemd, kubernetes et al from what I'm told) there are hooks to let the programs tell the framework that they completed initialization and are ready to serve requests - rather than the framework assuming that "forking happened, means server is ready". Similarly there are watchdog heartbeats, notification of planned graceful exit, etc.

Currently NUT has an extensible `upsnotify` method to handle these, with one implementation for systemd. Since the lack of "tech" may be a build error (people forgot libsystemd-dev), or some run-time inability to send a message can be an admin error (not running under systemd properly), the method notifies once that it can not interact with such messaging. On some platforms this is noise, on others something people can fix by reconfiguration or rebuild, and benefit from tighter service integration. On yet others, such framework facilities exist but NUT codebase does not know about them - and this can tickle someone to post a PR to add it :)

@jclulow @mtelka : any idea if such facilities appeared in SMF?

**Comment by hoewweken** [comment]
Hi!

Just to be complete, with my pkg from yesterday and user=root i see the device parameters, so USB **does** work.
Iwill have to learn about the nut framework and check if the autmotic shutdown at x% will do..

Thanks! so much for your work on this!! :-)

```
# upsc UPS1@localhost 
battery.charge: 100
battery.charge.low: 10
battery.charge.warning: 50
battery.runtime: 6180
battery.runtime.low: 120
battery.type: PbAc
battery.voltage: 27.1
battery.voltage.nominal: 24.0
device.mfr: American Power Conversion 
device.model: Smart-UPS 750
device.serial: AS1827134173  
device.type: ups
driver.debug: 0
driver.flag.allow_killpower: 0
driver.name: usbhid-ups
driver.parameter.bus: 160
driver.parameter.busport: 008
driver.parameter.device: 003
driver.parameter.pollfreq: 30
driver.parameter.pollinterval: 2
driver.parameter.port: auto
driver.parameter.product: Smart-UPS 750 FW:UPS 09.3 / ID=18
driver.parameter.productid: 0003
driver.parameter.serial: AS1827134173
driver.parameter.synchronous: auto
driver.parameter.vendor: American Power Conversion
driver.parameter.vendorid: 051D
driver.state: quiet
driver.version: 2.8.1
driver.version.data: APC HID 0.100
driver.version.internal: 0.52
driver.version.usb: libusb-1.0.26 (API: 0x1000109)
ups.beeper.status: enabled
ups.delay.shutdown: 20
ups.firmware: UPS 09.3 / ID=18
ups.mfr: American Power Conversion 
ups.mfr.date: 2018/07/05
ups.model: Smart-UPS 750
ups.productid: 0003
ups.serial: AS1827134173  
ups.status: OL
ups.timer.reboot: -1
ups.timer.shutdown: -1
ups.vendorid: 051d

```

**Comment by mtelka** [comment]
@jimklimov, maybe you are looking for the *Exiting and Exit Status* section in the `smf_method(7)` man page?

**Comment by jimklimov** [comment]
@hoewweken : Shutdowns are a bit of a mixed bag so far: `upsmon` will trigger the shutdown of the system it runs on (whether it is the same one as with the driver and `upsd` or a remote client fed from same UPS); however currently there does not seem to be a *safe* way to power off the UPS itself on Solaris/illumos systems with SMF managing the OS lifecycle.

Historically NUT did this by patching init-script or equivalent script that handled system shutdown (all the way to `poweroff` or `halt`, like `/etc/rc0.d` in the olden days), and provides a `nutshutdown` script that can be used as e.g. a `systemd-shutdown` hook.

Either way, when the system was in its last seconds, filesystems read-only and all, this addition checked if a "killpower" flag-file exists to signify that the shutdown is due to critical power situation, and in this case sent the UPS a command to power off the load (and hopefully return when wall power does, or power-cycle if it already did), or waited in a long sleep (long enough for battery to deplete if the UPS does not shut down/cycle) and then called `reboot` if still alive.

This approach allows racks of servers to complete their shutdowns if they started on this path, so later they would deterministically power up and initialize as they normally do. Aborting mid-shutdown is a bad idea generally (certain sites are welcome do to as they please and their technologies and workloads allow). This also avoids the "power-race condition", that while we were shutting down, the wall power got repaired - which happens fairly often. It is unfortunate to keep the rack of servers idling or halted indefinitely in this case, until someone comes around to push their buttons :)

However ever since Solaris 10+ switched to SMF, and in particular as `svc.startd` tells services to stop, waits for a certain hard-coded time, kills all remaining user-land processes, and then proceeds to shut down the system, there seems to be no place for such hooks anymore.

As a site-specific equivalent, one can extend the NUT shutdown logic (for `upsmon` probably) to at least command the UPS to go down - assuming their model can do so, and especially that it supports a delay between the command and poweroff (since you tell it while the OS is quite alive and just its services race to shut down). Maybe it is possible to add SMF services that would only "start" as part of shutdown, and so guarantee that they are late enough in the game to be relatively safe (and if they are fortunate to do so before `svc.startd`'s hard-coded timeout to `killall`) - not sure.

If you would get to tinker with such workarounds (or say even re-design `svc.startd` a bit), keep in mind that as part of system shutdown the driver daemons are expected to stop, and that patch/hook starts a new copy of the program to talk to the UPS just for the killpower handling. Since NUT 2.8.1 however there is a feature to allow a daemonized driver to kill power (need to flip a safety switch first), so avoiding re-initialization of the device and need for the filesystem backing the driver program and the libraries it needs. I think this mode was only proposed and merged, but there are no integrations built around it yet.

**Comment by jimklimov** [comment]
> maybe you are looking for the Exiting and Exit Status section in the smf_method(7) man page?

>> The required behavior of a start method is to delay exiting until the service instance is ready to answer requests or is otherwise functional.

Not quite. This is about the exit status (of the method script).

Probably the script could be taught to wait for some touch-file, and the "notification tech" would be creation of one when the driver gets initialized, but it is not really interaction with the framework.

Also ideally there would be no forking ("child/wait" service type and "foregrounded" mode for NUT daemons), and so no exit code until the daemon exits. However mid-way through its lifetime (well, some 5-10 seconds after start, usually) the daemon would yell "I am ready!" and the service would become "online" only then.

And also something about watchdogs (e.g. tickled every time the driver passes its polling loop), so that SMF can kill drivers stuck with I/O error (device unplugged or lagging), for example - they look healthy and alive for the PID-centric monitoring, but are not service-wise.

**Comment by hoewweken** [comment]
Hi!
I will think of the shutdown strategy, and thanks for your hints !!

One side note. On a different system "nut-client-only" i tried to install the python GUI with this:
``` # pkg install pkg:/system/management/nut/nut-monitor-gui-py3qt5 ``` 
and tested with
``` 
$ /usr/share/nut/nut-monitor/app/NUT-Monitor-py3qt5 
  File "/usr/share/nut/nut-monitor/app/NUT-Monitor-py3qt5", line 35, in <module>
    from PyQt5.QtCore import *
ModuleNotFoundError: No module named 'PyQt5.sip'
``` 
so there is a dependency missing on that route "gui-client-only": ``` pkg:/library/python/pyqt5-sip ``` 
With this package the GUI works :-)
P.S. Not on the latest version yet - connection to pkg.openindiana.org is sooo slow the last days on my side ...

jimklimov:
@hoewweken : Seems the `pkg` recipes outsmarted ourselves and themselves. Over the many iterations in this recipe, at some point I had a lot of explicit dependencies that they re-detected via python and claimed as duplicates, that dropping them proved too much. I guess this is a fallout of that clean-up iteration.

PR #14643, opened 2023-11-04
openjade: add test results; still needs gcc-7

AndWac:


components/developer/openjade/Makefile
```diff
@@ -17,40 +17,41 @@ include ../../../make-rules/shared-macros.mk
...
+	cd  $(@D) &&\
+	rm -f aclocal.m4 missing &&\
+	cp config/configure.in . &&\
+	PATH="$(PATH)" libtoolize --force &&\
+	aclocal -I config &&\
+	autoconf --force )
+
+
+# openjade segfaults in libostyle.so when compiled using gcc 10 or later
+GCC_VERSION= 7
```
mtelka: This should be before `common.mk` include.  Otherwise `USERLAND_REQUIRED_PACKAGES` won't be correct.

mtelka (changes requested):

components/developer/openjade/Makefile
AndWac: Sorry, fixed.

PR #14655, opened 2023-11-04
Obsolete perl/DBD-mysql

mtelka:


mtelka:
For reference:
https://openindiana.org/pipermail/openindiana-discuss/2023-October/026461.html
https://openindiana.org/pipermail/oi-dev/2023-October/033471.html
```
After 10 years of DBD::mysql Perl module available in OpenIndiana we
need to obsolete it.  If you need similar functionality you should
switch to DBD::MariaDB.  The DBD::MariaDB Perl module is available in
the library/perl-5/dbd-mariadb package.

Starting with version 5.001 of DBD::mysql released on 2023-10-04 it is
no longer possible to build it on OpenIndiana because MySQL 8.x is
required for building.  We do not have MySQL packaged for OpenIndiana.

If DBD::mysql is a must for you and you cannot switch to DBD::MariaDB
then you should contribute and package MySQL for OpenIndiana.
```

3eka:
Many thanks for explanation!

PR #14670, opened 2023-11-05
add libcupsfilters 2.0.0

AndWac:


AndWac:
The cups-filters package has been split into several packages. This is a former part of cups-filters.

PR #14688, opened 2023-11-06
pkg: requires "six"

jimklimov:
See #14686

### DIFF
--- a/components/openindiana/pkg/Makefile
+++ b/components/openindiana/pkg/Makefile
@@ -122,3 +122,4 @@ REQUIRED_PACKAGES += library/python/cherrypy-39
 REQUIRED_PACKAGES += library/python/python-rapidjson-39
 REQUIRED_PACKAGES += library/python/jsonschema-39
 REQUIRED_PACKAGES += library/python/jsonrpclib-pelix-39
+REQUIRED_PACKAGES += library/python/six-39


mtelka:
This is not enough to fix the problem.  You should do something in the [pkg5 repo](https://github.com/OpenIndiana/pkg5) too.  Probably the best way would be to simply drop the `six` dependency.

3eka:
Happy New Year!

Is there any ETA to fix this?
I just have created zone, but cannot do anything, because of error
```
:; pfexec zlogin solarium
[Connected to zone 'solarium' pts/14]
Last login: Wed Jan  3 00:55:56 on pts/14
The illumos Project     illumos-8361acf58a      January 2024
root@solarium:~# pkg install pkg:/package/pkg/zones-proxy
  File "/usr/bin/pkg", line 74, in <module>
    import six
ModuleNotFoundError: No module named 'six'
```
I have copied pkg **library/python/six-39** structure from GZ manually (from zone itself) as quick and dirty fix:
```
:; scp -r predrag_zecevic@GZ:/usr/lib/python3.9/vendor-packages/six-1.16.0-py3.9.egg-info /usr/lib/python3.9/vendor-packages/six-1.16.0-py3.9.egg-info/
:; scp predrag_zecevic@GZ:/usr/lib/python3.9/vendor-packages/six.py /usr/lib/python3.9/vendor-packages/six.py
:; scp predrag_zecevic@GZ:/usr/lib/python3.9/vendor-packages/__pycache__/six.cpython-39.pyc /usr/lib/python3.9/vendor-packages/__pycache__/six.cpython-39.pyc
```
and then it has worked:
```
root@solarium:~# pkg install pkg:/package/pkg/zones-proxy
No updates necessary for this image.
```

Best regards

PR #14720, opened 2023-11-07
firefox: update to 119.0.1

iigs:
Even though this is built using gcc 13, the system/library/gcc-10-compat-links package is still being used.  I was unable to install system/library/gcc-13-compat-links due to conflicts with gcc-10-compat-links.  I was unable to uninstall gcc-10-compat-links due to 10 packages using it.

Added a security patch early that will be inegrated in firefox 120.


iigs:
Replaced patches that had gotten mangled by github's web editor.

PR #14822, opened 2023-11-14
Update nut-v2.8.1 packaging

jimklimov:
Lessons learned from #14630 :
* avoid delivering "driver" actions after all - they do not let NUT user get permissions to just the devices with wanted USB VID:PID, and it conflicts with ugen package overall :\ End-users with USB UPSes would probably have to set `user=root` in `ups.conf` sections, at least for now.
* make sure SMF method scripts are executable

### DIFF
--- a/components/sysutils/nut/history
+++ b/components/sysutils/nut/history
@@ -1,0 +1,1 @@
+system/management/nut-common@2.7.4,5.11-2022.0.0.6 system/management/nut/common
--- a/components/sysutils/nut/nut-bins.p5m
+++ b/components/sysutils/nut/nut-bins.p5m
@@ -1,0 +1,27 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2016-2023 Jim Klimov
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)/bins@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY) - Grouping of NUT command-line and daemon client and server sides"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend fmri=$(COMPONENT_FMRI)/clients@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
--- a/components/sysutils/nut/Makefile
+++ b/components/sysutils/nut/Makefile
@@ -20,6 +20,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=	nut
 COMPONENT_VERSION=	2.8.1
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	Network UPS Tools (NUT)
 COMPONENT_DESCRIPTION=	Network UPS Tools (NUT) is a versatile power-device monitoring toolkit to facilitate safe shutdowns
 COMPONENT_SRC=	$(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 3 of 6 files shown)


mtelka (changes requested):
Please add the `history` file as suggested in https://github.com/OpenIndiana/oi-userland/pull/14630#issuecomment-1803888806 and https://github.com/OpenIndiana/oi-userland/pull/14630#issuecomment-1804193303 to allow smooth upgrades.

jimklimov:
Ok, got any suggested precedents to look at? :)
Or a description of that syntax, what does it mean? :)

For example, `nut-bins` of the old is superseded by two or more packages now (the nut-server and nut-clients baselines, possibly some driver commons). Can this be listed here?

* The singular `nut-drivers` was split into many packages for sets of daemons grouped by third-party dependency footprint. But the name remains used by an umbrella package which spells the optionally-required dependencies.

jimklimov:
Also, stumbled on one more issue: services use `/var/run/nut` for pid files and such, which appears after package install but is gone (tmpfs) after a reboot. What is the IPS best practice to ensure such locations are present - a small custom service to depend on which would `mkdir + chown + chmod`? Or is there an OS facility like `systemd-tmpfiles` in the competitiors' land, so the package just needs to drop-in a config file?

UPDATE: method scripts for nut-server and nut-monitor take care of this.

mtelka:
Syntax of `history` files is simple:
```
OLD-PKG [NEW-PKG] [noincorporate]
```
If there is no `NEW-PKG` then the `OLD-PKG` will be obsoleted.  If there is `NEW-PKG` then the `OLD-PKG` will be renamed to `NEW-PKG`.  If there is `noincorporate` at the end then the resulting package will not be incorporated into `userland-incorporation`.  I believe you do not need that, so you should avoid `noincorporate`.  Renames are 1:1 mapping, so if you have two superseded packages, you need to choose one.

Examples are in `history` files scattered around the git repo.

If this is not sufficient for you then you could create own transitional package(s) instead of doing so automatically via the `history` file.  IOW, you could create full `.p5m` file(s) with whatever you need there.

For this particular case we need to do something with `system/management/nut-common` and `system/management/nut/bins`.

components/sysutils/nut/history
```diff
@@ -0,0 +1,3 @@
+system/management/nut-common@2.7.4,5.11-2022.0.0.6 system/management/nut/common
+system/management/nut/bins@2.7.4,5.11-2022.0.0.6 system/management/nut/server system/management/nut/clients
+
```
mtelka: Please remove the empty line here.

jimklimov: Yeah, and also a second package name, that did not work.

jimklimov: I guess it is better to get a new `nut-bins` that just depends on server and client.

mtelka: > I guess it is better to get a new `nut-bins` that just depends on server and client.

Like a meta-package?  Yes, looks so.

mtelka: > Yeah, and also a second package name, that did not work.

Yes, obviously.  FYI, the `history` file is processed by the [generate-history](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/tools/generate-history) awk script.

components/sysutils/nut/nut-bins.p5m
```diff
@@ -0,0 +1,27 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY) - Grouping of NUT command-line and daemon client and server sides"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+depend fmri=$(COMPONENT_FMRI)/clients@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
+depend fmri=$(COMPONENT_FMRI)/server@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION) type=require
```
mtelka, line 27: I would remove `@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)` from both lines here to keep them as simple as possible.  The version info brings ~~almost~~ no value here, I believe.

mtelka (approved):

components/sysutils/nut/nut-bins.p5m
jimklimov: It is similar in other files on one hand, and some time ago I was told this is the way to ensure that specifically these package builds will be brought in. Both less confusion, and less work for `pkg` programs to prepare the tree of possibilities.

mtelka: Maybe it is done that way somewhere else, but one of the [biggest beasts does not do that](https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/meta-packages/build-essential/build-essential.p5m).

For the `pkg` performance, I'm not sure.  Maybe.  But since we usually do have only few package versions in the ips repo I'd say the difference is small.  Everything is anyway constrained by the `userland-incorporation` so this cannot have any significant impact in most cases.

Anyway, since the version is harmless here, I do not mind if it is kept there.

jimklimov:
A few rough edges about services remain (e.g. `nut-driver-enumerator` complains when it finds no configured devices, rather than returning success about no work to do when there are not service instances too - will be fixed in a next NUT release, which may be soon-ish as we collect feedback from different packagers), but at least for a dummy driver in a local zone - works for me.

mtelka:
@jimklimov, FYI: https://github.com/networkupstools/nut/issues/2180

mtelka:
@AndWac please merge this one.  Thank you.

PR #14829, opened 2023-11-14
librsvg - update to version 2.57.0

fritzkink:


fritzkink:
It seems that the build server does not have the newest cairo package installed.

components/library/librsvg/pkg5
```diff
@@ -14,6 +14,7 @@
         "system/library/fontconfig",
         "system/library/freetype-2",
         "system/library/gcc-10-runtime",
+        "system/library/gcc-13-runtime",
```
AndWac, line 16: Why depends this on both gcc-10 and gcc-13 runtimes?

fritzkink: I guess it is a race condition, because the installed librsvg package depends on gcc-10 runtime and if you check on build/prototype/i386/usr/lib/amd64/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so it loads /usr/lib/amd64/librsvg-2.so.2 which is depended on gcc-10. Unfortunately I can't uninstall the old librsvg package due to many dependencies.

AndWac: Add pkg.depend.bypass-generate=.* in the manifest for every affected file. Then rerun gmake REQUIRED_PACKAGES and gmake publish. Here it was necessary to change two lines (I have only checked the first):
file path=usr/bin/rsvg-convert pkg.depend.bypass-generate=.*
file path=usr/lib/$(MACH64)/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-svg.so pkg.depend.bypass-generate=.*

AndWac: I am not sure whether this is a good idea. Maybe it would be better to first merge as is and then re-publish revision 1. If your theory is sound this would fix the dependency problem.

AndWac: No, this doesn't bring the desired result. Even after installation of the new version gmake REQUIRED_PACKAGES still produces the dependency on gcc-10. There must be something else needing it.

mtelka:
I think this is caused by rust.  rust should be recompiled first using gcc-13.

components/library/librsvg/pkg5
mtelka: I think this is caused by rust.  rust should be recompiled first using gcc-13.

PR #14888, opened 2023-11-17
Disable python/PyQt5

mtelka:
The build failed.  I think some build dependencies are missing.  I need to investigate further...


mtelka:
@AndWac, what version of the `library/qt5` package is installed on the build server?

AndWac:
I have build the package successfully and let jenkins only publish it. This happens quite a lot on the build server.

PR #14908, opened 2023-11-18
Fix nut-monitor-gui-py3qt5.p5m recipe (deps)

jimklimov:
Follow-up to https://github.com/OpenIndiana/oi-userland/pull/14630#issuecomment-1812640745 report

### DIFF
--- a/components/sysutils/nut/Makefile
+++ b/components/sysutils/nut/Makefile
@@ -20,7 +20,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=	nut
 COMPONENT_VERSION=	2.8.1
-COMPONENT_REVISION=	1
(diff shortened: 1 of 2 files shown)


jimklimov:
Interesting... while typing up the changes, I had `py3qt5-sip` added in both `pkg5` and `Makefile`; however the subsequent `make REQUIRED_PACKAGES sample-manifest publish` removed these lines. If it is so much a dependency of (I suppose) `py3qt5` itself, why was it not installed for @hoewweken in the original report?

AndWac:
Our dependency detection is quite simplistic and doesn't detect all dependencies. Missing build dependencies should be added in a separate section in the Makefile; runtime dependencies should be added to the manifest file like you did.

PR #14997, opened 2023-11-25
ghostscript - update to version 10.02.1 and drop 32bit support

fritzkink:


AndWac:
I have a similar change lying around for a while. Alas there are many new failing tests that prohibited me to update ghostscript.

PR #15012, opened 2023-11-26
Update libjpeg to version 9

Toasterson:
Move libjpeg-turbo out of default paths


components/library/libjpeg6-ijg/pkg5
```diff
@@ -1,11 +0,0 @@
-{
-    "dependencies": [
-        "SUNWcs",
-        "system/library"
-    ],
-    "fmris": [
-        "image/library/libjpeg6",
-        "image/library/libjpeg6-ijg"
```
mtelka, line 8: These packages are required by other packages, so you cannot simply remove them.

Toasterson: I can also rename them to the new libjpeg package. The project to replace those two packages started 5 years ago. If they still used I consider that a Bug that needs fixing.

mtelka: If you want to remove `components/library/libjpeg6-ijg` then you need either obsolete or rename both `image/library/libjpeg6` and `image/library/libjpeg6-ijg`.  Since both packages are runtime and build dependencies of many other packages (for example `desktop/office/abiword`, `desktop/remote-desktop/freerdp`, `runtime/java/openjdk22`, `library/perl-5/tk-538`, `library/qt6`, etc.) then obsoletion does not look correct.  But since they are still massively used I think it would be far better first to make sure they are no longer used, and then remove (obsolete) them.

Toasterson: Ok, I'll rename them. As for the dependency. Unfortunately this looks like a severe problems as libjpeg-turbo should have been used. Unfortunately I first must change libjpeg6 to libjpeg9 so libjpeg9 gets detected.

AndWac:
╰─➤  git diff pkg5
diff --git a/components/library/libjpeg-turbo/pkg5 b/components/library/libjpeg-turbo/pkg5
index ebb7414af6..6c132646f7 100644
--- a/components/library/libjpeg-turbo/pkg5
+++ b/components/library/libjpeg-turbo/pkg5
@@ -5,6 +5,7 @@
     ],
     "fmris": [
         "image/library/libjpeg-turbo",
+        "image/library/libjpeg6-turbo",
         "image/library/libjpeg7-turbo",
         "image/library/libjpeg8-turbo"
     ],

hoewweken:
Hi!
As libjpeg9 is needed now,
and libjpeg6 has many dependants,
won't it be easier to add a new package libjpeg9 at a first step,
and then after cleaning up the older bits obsolete libjpeg6 (if it is clear that its possible )

just my 2 cents

mtelka:
> won't it be easier to add a new package libjpeg9 at a first step, and then after cleaning up the older bits obsolete libjpeg6 (if it is clear that its possible )

Yes.  This is exactly what I propose above.

Toasterson:
Unfortunately due to the broken detection mechanism it is not possible. Which is why I am preparing three PR's to merge in the same general timeframe to change everything

mtelka:
> Unfortunately due to the broken detection mechanism it is not possible. Which is why I am preparing three PR's to merge in the same general timeframe to change everything

What exactly is broken?  Could you share more info please?

Such PR sets are usually risky and often something goes wrong, so I suggest to do the transition slowly.

Thank you.

Toasterson:
We need to patch a ton of packages and definetly openjdk22, qt6 and poppler refuse to find libjpeg8-turbo. It has gotten so bad, that most of these packages that you see as dependants of libjpeg6 are in fact bugs for libjpeg8-turbo detection. openjdk11 was libjpeg8-turbo and now openjdk22 is libjpeg6. Thats bad and it gets through the checks all the time. We cannot control the versions other libs depend upon and if we are afraid to rebuild our packages we can stop doing this distribution. We have to be able to rebuild packages. If that does not work we end up doing as much work as nixos. And thats not feasable with this small a group. We tried to replace libjpeg6 with libjepeg8-turbo for 5 years. and it's been such trouble that it still hasn't finally happened. I am happy to rebuild everything and then publish it and call it a day.

mtelka:
@Toasterson, okay, I see your frustration, but honestly I do not understand what exactly is the problem.

I must admit that I know almost nothing about `libjpeg`.  I just looked at wikipedia and it looks like the only currently supported versions of `libjpeg` are `6b` (released 25 years ago) and `9e` (release ca 2 years ago).  Also, there is a fork called `libjpeg-turbo` with the latest version `3.0.1` (released one month ago).

It looks like we do have both `libjpeg 6b` and `libjpeg-turbo 3.0.1` packaged, so we are basically (mostly) up to date.  We only miss `libjpeg 9e` and this PR seems to be mostly about it.

So why not to add `libjpeg 9e` as a new package, then rebuild one component after the other using either `libjpeg 9e` or `libjpeg-turbo 3.0.1` and then (if we really want to) obsolete `libjpeg 6b` when it becomes unused?

You claim that some packages refuse to find  `libjpeg8-turbo`.  I looked at `poppler` from your list and it depends on `image/library/libjpeg8-turbo`.  So what is a problem with it?

Thank you.

cstes:
I've been asked by email by Till to update squeak from libjpeg8-turbo to libjpeg9.   This is very low priority for me.   The current version of squeak is 4.20.7 and I did an effort to use libjpeg8-turbo although that Squeak has the worst of all solutions : a copy of an old libjpeg in the Squeak source tree, so using the builtin libjpeg would have been the easiest (and also what Squeak is doing on Windows32).   Anyway my experience with this in https://www.illumos.org/issues/7391  where I understand this Illumos issue as a request to use libjpeg8-turbo or the https://libjpeg-turbo.org/ in general.    Note that Squeak is an atypical example anyway because libjpeg is a C library and Squeak is a Smalltalk programming language so personally I think libjpeg should be removed from Squeak altogether and replace libjpeg with a Smalltalk implementation, like Squeak is doing for Vector Graphics (SVG) using the vep (vectorengine) plugin.    The fact that there still is libjpeg in Squeak is an indication of how slow moving Squeak and Smalltalk is, but what can you expect from a programming language from 1972 or 1983 or so !??   Note that Squeak gets updates from time to time but this is going to take many months or years, and Squeak is certainly not a showstopper for this pull request (related to libjpeg).

Toasterson:
Oh, dear. looks like a rebase mishap. Will open a new one.

PR #15013, opened 2023-11-26
jenkins: Default to Java 21

NotMyFault:
The change proposed updates the Java version to 21 because Jenkins 2.426.1 onwards supports and recommends Java 21, same as the weekly releases.

Additionally, I've updated an outdated website reference in `components/developer/jenkins/Makefile`.


components/developer/jenkins-core-lts/Makefile
```diff
@@ -87,4 +87,4 @@ install: $(INSTALL_32)
 test: $(NO_TESTS)
 
 # Auto-generated dependencies
-REQUIRED_PACKAGES += runtime/java/openjdk17
+REQUIRED_PACKAGES += runtime/java/openjdk21
```
AndWac, line 90: You cannot achieve what you wanted with this change. It is a generated setting (created by running "gmake REQUIRED_PACKAGES"). The system is either using the default openjdk17 at the moment) or something that has been configured explicitly.

PR #15031, opened 2023-11-27
illumos-gate: add patch for 64-bit libnwam

AndWac:


AndWac:
Our nwam-manager needs to become 64 bit. This is a prerequisite.

PR #15032, opened 2023-11-27
clutter - update to version 1.26.4

fritzkink:


AndWac:
I have some collisions in my build environment when I publish this, eg.
pkg install: The following packages all deliver file actions to usr/lib/amd64/girepository-1.0/Cogl-1.0.typelib:

  pkg://openindiana.org/library/desktop/cogl@1.22.8,5.11-2023.0.0.0:20231127T211818Z
  pkg://openindiana.org/library/desktop/clutter@1.6.20,5.11-2022.1.0.4:20221221T111307Z

fritzkink:
this will be solved with the new package clutter 1.26.4

PR #15037, opened 2023-11-28
FIX: 14481 - UIManager.getSystemLookAndFeelClassName() returns Motif not GTK

fronald:
Fix for the issue https://www.illumos.org/issues/14481

The official OpenJDK repository only distinguishes whether it is Gnome or not. But it really doesn't make sense that the Java distribution present in OpenIndiana returns Motif as the look and feel of the system, when running Mate DE. And so a simple contribution follows.

About the change in the JDK8 Makefile: On my first attempt, I was unable to compile due to lack of dependencies, I ran "gmake REQUIRED_PACKAGES" and it added the GCC_RUNTIME_PKG dependency to the Makefile and then compiled normally. Please check whether I should keep this change or not.

### DIFF
--- a/components/runtime/openjdk-11/Makefile
+++ b/components/runtime/openjdk-11/Makefile
@@ -14,6 +14,7 @@
 # Copyright 2019 Michal Nowak
 # Copyright 2021 Andreas Grueninger, Grueninger GmbH, (grueni). All rights reserved.
 # Copyright 2023 Niklas Poslovski
+# Copyright 2023 Franklin Ronald <franklin@wiselabs.com.br>
 #
 
 include ../../../make-rules/shared-macros.mk
@@ -23,6 +24,7 @@ OPENJDK_UPDATE=	0
 OPENJDK_BUILD=	20
 COMPONENT_NAME=		openjdk
 COMPONENT_VERSION=	$(OPENJDK_VERSION).$(OPENJDK_UPDATE).$(OPENJDK_BUILD)
+COMPONENT_REVISION = 1
 COMPONENT_FMRI=	runtime/java/$(COMPONENT_NAME)$(OPENJDK_VERSION)
 COMPONENT_SUMMARY=	Open-source implementation of the Java Platform, Standard Edition
 COMPONENT_SRC=	jdk$(OPENJDK_VERSION)u-jdk-$(COMPONENT_VERSION)-ga
--- a/components/runtime/openjdk-11/patches/src_java.desktop_share_classes_javax_swing_UIManager.java.patch
+++ b/components/runtime/openjdk-11/patches/src_java.desktop_share_classes_javax_swing_UIManager.java.patch
@@ -1,0 +1,16 @@
+--- a/src/java.desktop/share/classes/javax/swing/UIManager.java.orig	Mon Nov 27 19:15:37 2023
++++ b/src/java.desktop/share/classes/javax/swing/UIManager.java	Mon Nov 27 19:45:00 2023
+@@ -657,9 +657,12 @@
+         if (osType == OSInfo.OSType.WINDOWS) {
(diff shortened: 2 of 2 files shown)


fronald:
To help with testing, after building and installing the package, run the following script with the argument corresponding to the JDK version.
test_script
```bash
#!/usr/bin/bash
WKDR="$(realpath $(dirname $0))"
JAVA_VERSION="$1"
if [ "$JAVA_VERSION" == "" ] || [ "$JAVA_VERSION" == "8" ]; then
	JAVA_VERSION="1.8.0"
fi
JAVA_HOME="/usr/jdk/openjdk$JAVA_VERSION"

function compile_test() {
	cat <<EOF > "$WKDR/Main.java"
import javax.swing.UIManager;
public class Main {
    public static void main(String[] args) {
        System.out.println("System look and feel: " + UIManager.getSystemLookAndFeelClassName());
    }
}
EOF
	"$JAVA_HOME/bin/javac" "$WKDR/Main.java"
}

function run_test() {
	echo "Testing $JAVA_VERSION..."
	$JAVA_HOME/bin/java -cp "$WKDR" Main
	rm -f "$WKDR/Main.class"
	rm -f "$WKDR/Main.java"
	echo "Done."
}

if [ -f "$JAVA_HOME/bin/javac" ]; then
	compile_test
	run_test
else
	echo "$JAVA_HOME not found."
fi
```

AndWac:
You'll need to touch the Makefiles of all affected packages in order to trigger our jenkins.
Furthermore, you'll either need to change the COMPONENT_VERSION or set/increment the COMPONENT_REVISION of every package in order to ensure that installed packages will be updated.

fronald:
> You'll need to touch the Makefiles of all affected packages in order to trigger our jenkins. Furthermore, you'll either need to change the COMPONENT_VERSION or set/increment the COMPONENT_REVISION of every package in order to ensure that installed packages will be updated.

Thanks for the answer. I have some doubts.

Regarding the version increment, these packages are concatenating the JDK version, what is the recommendation? Insert a "-1" at the end of the name?

image

The second point is Jenkins. When building OpenJDK 8 it threw the error "No configurations found for [...] openjdk-8/build/amd64/! Please run configure to create a configuration." What I can't know is the correlation with my changes, since the previous package was building normally and there were no structural changes (as you can see). In my development environment, gmake publish runs normally.

image

With these questions resolved, I will update this PR.

AndWac:
openjdk-8 seems to be broken at the moment. It hasn't been touched for a while and the environment has changed a lot since its last successful build. Maybe you should concentrate on the newer versions and create PR's for every single JDK. This way we can merge the PR's that are ok and won't block everything.

mtelka:
> Regarding the version increment, these packages are concatenating the JDK version, what is the recommendation? Insert a "-1" at the end of the name?

If there is no `COMPONENT_REVISION` in the `Makefile` then add `COMPONENT_REVISION = 1` there.  If it is there, then increase its value.

fronald:
> > Regarding the version increment, these packages are concatenating the JDK version, what is the recommendation? Insert a "-1" at the end of the name?
> 
> If there is no `COMPONENT_REVISION` in the `Makefile` then add `COMPONENT_REVISION = 1` there. If it is there, then increase its value.

Perfect! I will.

fronald:
> openjdk-8 seems to be broken at the moment. It hasn't been touched for a while and the environment has changed a lot since its last successful build. Maybe you should concentrate on the newer versions and create PR's for every single JDK. This way we can merge the PR's that are ok and won't block everything.

Done. This PR now refers only to OpenJDK 11.

AndWac:
Alas we cannot build OpenJDK-11 on our build server because its processors are missing SIMD extensions which are needed by the bootstrap JDK that is being used here. Later versions are ok, though.

PR #15053, opened 2023-11-29
New package: Multitail

fronald:
MultiTail lets you view one or multiple files like the original tail program. The difference is that it creates multiple windows on your console (with ncurses).

More details: https://vanheusden.com/multitail/

image

### DIFF
--- a/components/sysutils/multitail/manifests/sample-manifest.p5m
+++ b/components/sysutils/multitail/manifests/sample-manifest.p5m
@@ -1,0 +1,32 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2023 <contributor>
+#
+
+set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=etc/multitail.conf.sample
+file path=usr/bin/multitail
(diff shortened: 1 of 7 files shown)


components/sysutils/multitail/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,36 @@
...
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/multitail
+file path=usr/etc/multitail.conf.new
+file path=usr/etc/multitail/conversion-scripts/colors-example.pl
```
AndWac: usr/etc is wrong.

fronald: But this is the sample (generated by gmake). The effective manifest is multitail.p5m.

AndWac: Ok, but if the manifest uses a wrong folder for the configuration it's quite likely that the real application is also looking at the wrong path for its configuration. You should use settings in the Makefile to fix this.

fronald: Nice. Done :-D

fronald:
sample-manifest fixed.

AndWac:
You shall not edit the sample-manifest. It's not possible to fix this issue with that. You'll need to change the Makefile (or the sources by patches) to make a gmake sample-manifest generate the expected results.

fronald:
> You shall not edit the sample-manifest. It's not possible to fix this issue with that. You'll need to change the Makefile (or the sources by patches) to make a gmake sample-manifest generate the expected results.

Oh sorry. I was always thinking "what is this sample-manifest for?", now I understand the purpose... It is to validate the installation prefixes of the construction stage.

I'll fix it. Thanks for the answer.

AndWac:
The sample-manifest is the basis to build the real manifest from. Sometimes you can use the sample-manifest by just copy it , sometimes it needs some specialities for IPS.
Furthermore, with the sample-manifest you can check for differences when you rebuild or update to a newer version. You can see the differences easier than comparing a new sample-manifest against the adopted manifest.

fronald:
> The sample-manifest is the basis to build the real manifest from. Sometimes you can use the sample-manifest by just copy it , sometimes it needs some specialities for IPS. Furthermore, with the sample-manifest you can check for differences when you rebuild or update to a newer version. You can see the differences easier than comparing a new sample-manifest against the adopted manifest.

Thank you for the explanation.

I have already changed the Makefile to prepare the prototype directory with the correct Multitail installation artifacts.

I have already checked that multitail is seeing the configuration file correctly after it is installed (/etc/multitail.conf), as can be seen in CMakeLists.txt in the value of the "CONFIG_FILE" property.

Done \o/

AndWac:
Please check the sample-manifest again. mine differs.

fronald:
> Please check the sample-manifest again. mine differs.

Done.

PR #15091, opened 2023-11-30
fix perl-536 manifest

AndWac:


mtelka:
Sorry for this.  When 5.36.1 appeared in the manifest it looked suspicious to me (it is obviously wrong), but I somehow ignored it.

PR #15117, opened 2023-12-01
illumos-gate: ignore runtime/perl during package lint

mtelka:


mtelka:
Upstream bug report: https://www.illumos.org/issues/16108

PR #15150, opened 2023-12-04
add warning about 32bit build requirements for SunRay

cgrzemba:
add warnings that this package should still build in 32bit too, because GDM tool gdm-session-worker have to run in 32bit and requires this libraries.


mtelka:
To get better support for SunRay I propose to create a metapackage that would list all dependencies required directly by SunRay.  This would allow SunRay users to simply install such metapackage to make sure they do have everything needed to run SunRay.  This would also allow oi-userland maintainers to easily see when something breaks.

@cgrzemba, could you please try to create such metapackage?  Thank you.

PR #15190, opened 2023-12-07
illumos-gate: made 64-bit libnwam compatible for 32-bit nwamd

cgrzemba:
The data structures used for the doorfs calls between nwam-manager (64-bit) and nwamd (32-bit) have to be compatible.


mtelka:
Please make sure this and the related patch (#15031) goes to upstream illumos-gate too.

Thank you.

AndWac:
We already have another patch starting with 003-
Please choose another prefix.

mtelka:
Will building `nwamd` as 64-bit solve the problem too?

cgrzemba:
building nwamd as 64-bit would raise new problems.

PR #15341, opened 2023-12-21
Obsolete python/Pyro4

mtelka:


mtelka:
Reasons for obsoletion:
- there is no known consumer for this
- development halted, users are encouraged to switch to Pyro5 and the repository has been archived - see also: https://github.com/irmen/Pyro4

PR #15364, opened 2023-12-23
squeak: update to 4.20.6

cstes:
Update to 4.20.6

This has an update to the VectorEngine plugin and to the OpenSSL module for SSL 3.x support

However we have set USE_OPENSSL10=yes because even if the module supports SSL 3.x the Smalltalk images that do the tests (SUnit framework) assume OpenSSL 1.0.2 I think.

In any case "gmake test" works for me but I have used Perl 5.36 and GCC 7.   I think it will work with Perl 5.38 and GCC 13 as well.


components/runtime/smalltalk/squeak/Makefile
```diff
@@ -43,6 +43,10 @@
 BUILD_BITS=32_and_64
 USE_COMMON_TEST_MASTER=no
```
AndWac, line 54: You are still building 32 bit which we phase out at the moment.
Your build won't work anymore because we don't have all needed 32 bit libraries anymore.

cstes:
Andreas,

The Smalltalk image https://files.squeak.org/4.6/  is 32 bit.

Also the historical images that we can run with this VM such as Squeak 1.1 or Saueqk 2 are 32 bit

For example to run historical images like the one from https://files.squeak.org/1.1/
there is a need for a 32 bit interpreter.

As far as I know, the COG and Stack VM from OpenSmalltalk is not able to run those old images.

Application software like Squeak can be useful even if it is 32 bit only ...

AndWac:
Supporting 32 bit is a burden. We don't have enough maintainers to continue support for 32 bit.
At the moment we are only making one exception by keeping 32 bit for GDM (which is needed by SunRay software).
For the affected packages somebody created WARNING files in order to inform maintainers about the reason to keep 32 bit. At least you'll need to do a similar thing for all packages that are needed for squeak then.

AndWac:
I just removed 32 bit from freetype-2 because harfbuzz is 64 bit only for some time.

cstes:
The squeak 32bit interpreter could be removed if the 64bit version of Squeak or OpenSmalltalk would support old 32bit (binary) smalltalk images.   I am pessmistic as to whether the Squeak upstream developers will do that.   They will most likely respond to use their packages  for their limited list of supported operating systems, instead of IPS or an unsupported operating system like OpenIndiana.      So I will do a request to support 32bit libraries ...

cstes:
There eixsts a method to load 32bit images into 64bit Smalltalk systems described in 
[](https://www.illumos.org/issues/16134)

However methods to load those images for example using JavaScript and so on, may exist, but may pose different problems.

So if possible at all I'd like to ask to still provide support for 32bit libraries, but as indicated in the RFE , there may be a method to load the Smalltalk 32bit image into OpenSmalltalk 64bit and then OpenIndiana could drop Squeak 32bit but it would be useful to still provide Squeak 32bit for a while so that I can compare the two methods on OpenIndiana, perhaps on a new machine as I am thinking of installing and testing OpenIndiana on a new workstation.

cstes:
The link is https://www.illumos.org/issues/16134

AndWac:
Am 25.12.23 um 14:06 schrieb David Stes:
>
> There eixsts a method to load 32bit images into 64bit Smalltalk
> systems described in
>
> However methods to load those images for example using JavaScript and
> so on, may exist, but may pose different problems.
>
> So if possible at all I'd like to ask to still provide support for
> 32bit libraries, but as indicated in the RFE , there may be a method
> to load the Smalltalk 32bit image into OpenSmalltalk 64bit and then
> OpenIndiana could drop Squeak 32bit but it would be useful to still
> provide Squeak 32bit for a while so that I can compare the two methods
> on OpenIndiana, perhaps on a new machine as I am thinking of
> installing and testing OpenIndiana on a new workstation.
>
>
Reverting the already dropped 32 bit versions alone is a lot of work. We
started to drop 32 bit roughly two months ago. You should have noticed
it by yourself.
Keeping 32 bit, even for some libraries, is additional work I would like
to avoid. Alas the engagement in OI is still declining and we are behind
in many important areas.
I

cstes:
I have found a workaround where if required I can build only the 64bit binary of Squeak and make it open the 32bt images by dropping support for the 68000 format  ... 
So if it is really required to drop the 32bit executable (which I would regret) then I could use that workaround because there are few 68000 images and thousands of 6500 images (32bit) so reconfiguring the 64bit executable to deal with 32images is a workaround or solution ...

I have this working now and it passes the tests so I'll submit that update in a few days (next weekend or so).
Thanks for the feedback in any case (and for supporting Squeak in OpenIndiana).

PR #15375, opened 2023-12-24
Request support of 32bit for some libraries needed by Squeak

cstes:
WARNING:  this will add WARNING files to critical libraries and I have no idea whether this will force a J JENKINGS rebuild or rebuild of critical system libraries.

Basically I am just requesting support for 332bit for the Squeak 32bit interpreter.
This interpreter is required to run 32bit Smalltalk images.

Given the enormous amount of work that solaris-userland and oi-userland did in the past to support both 32bit and 64bit, it would be nice to continue to support 32bit if possible.

I understand that if some of the upstream components like CAIRO or PANGO or FREETYPE would drop support for 32bit then obviously OpenIndiana cannot continue to support 32bit versions.

### DIFF
--- a/components/library/libffi/WARNING
+++ b/components/library/libffi/WARNING
@@ -1,3 +1,7 @@
 Attention
 
 GDM for SunRay (http://pkg.toc.de/sunray) depends on this package. There are 32-bit and 64-bit builds required.
+
+The package runtime/smalltalk/squeak depends on this library.
+A 32bit and 63bit version of this library is required for Squeak.
+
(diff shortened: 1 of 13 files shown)


components/library/libffi/WARNING
```diff
@@ -1,3 +1,4 @@
-Attention
 
-GDM for SunRay (http://pkg.toc.de/sunray) depends on this package. There are 32-bit and 64-bit builds required.
+The package runtime/smalltalk/squeak depends on this library.
+A 32bit and 63bit version of this library is required for Squeak.
+
```
AndWac: Please don't delete the warning for SunRay!

AndWac:
There are a lot more packages involved because of the transitive dependencies.
Are you willing to help keeping them 32 bit?
As I already wrote, keeping 32 bit creates a lot of additional work; some package receipts need tricky things to work properly. Dropping 32 bit will ease maintainers work. Plus, reviving already dropped 32 bit packages is also a painful thing because we have cyclic dependencies. Eg. harfbuzz depends on freetype-2 and vice versa.

PR #15400, opened 2023-12-25
memcached - update to version 1.6.22

fritzkink:


components/network/memcached/Makefile
```diff
@@ -21,36 +21,51 @@
 
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2023, Friedrich Kink. All rights reserved.
 #
 
-BUILD_BITS=64
+BUILD_BITS=		64
+OPENSSL_VERSION=	3.1
+
```
AndWac: There doesn't seem to be a depedency on openssl at all.

fritzkink: There would be if `#CONFIGURE_OPTIONS +=   --enable-tls` would be enabled. For the time being it does not work. So it is just a short cut/help/reminder in case being enabled again.

AndWac: How essential is TLS for memcached?

components/network/memcached/Makefile
```diff
@@ -21,36 +21,51 @@
...
-CFLAGS += -D__ATTRIBUTE_DISABLED
+CFLAGS += -I/usr/include/libevent2 -D__ATTRIBUTE_DISABLED
 
-CONFIGURE_BINDIR.64 =	$(CONFIGURE_PREFIX)/lib
-CONFIGURE_OPTIONS.64 +=	--with-libevent=/usr/lib/${MACH64}
+CONFIGURE_OPTIONS +=	--with-libevent=/usr/lib/${MACH64}
 CONFIGURE_OPTIONS +=	--enable-dtrace
+CONFIGURE_OPTIONS +=	--disable-sasl
+#CONFIGURE_OPTIONS +=	--enable-asan
+#CONFIGURE_OPTIONS +=	--enable-werror
```
AndWac: Why these formerly enabled options have been disabled? Please write at least some explanation so we don't have to guess.

components/network/memcached/Makefile
mtelka: Both `BUILD_BITS=64` and `OPENSSL_VERSION=3.1` are defaults.  Setting them here is a call for more work in future.  So please remove both.  Thank you.

PR #15469, opened 2023-12-31
python/attrs: do not incorporate obsoleted attrs-37

mtelka:


mtelka:
This fixes the jenkins job failure: https://hipster.openindiana.org/jenkins/job/oi-userland/11006/

PR #15551, opened 2024-01-08
Obsolete x11/compatibility

mtelka:
I propose obsoletion of seven X11 compatibility packages.

These packages were added more than 5 years ago via #4564 by @alarcher and I doubt they are still needed for anything.  Inside `oi-userland` there is only `mate_install` that needs `compatibility/packages/SUNWxwplt` but such dependency is actually no-op because all packages needed by `compatibility/packages/SUNWxwplt` are already needed by `mate_install` too.  There is no other known dependency around.

Since it is more than 10 years since our ways parted with (Open)Solaris and since similar compatibility packages were already cleaned up via #9137 we probably should obsolete these X11 compatibility packages as well.

@alarcher, do you have any suggestion here?

Thanks.


alarcher:
> I propose obsoletion of seven X11 compatibility packages.
> 
> These packages were added more than 5 years ago via #4564 by @alarcher and I doubt they are still needed for anything. Inside `oi-userland` there is only `mate_install` that needs `compatibility/packages/SUNWxwplt` but such dependency is actually no-op because all packages needed by `compatibility/packages/SUNWxwplt` are already needed by `mate_install` too. There is no other known dependency around.
> 
> Since it is more than 10 years since our ways parted with (Open)Solaris and since similar compatibility packages were already cleaned up via #9137 we probably should obsolete these X11 compatibility packages as well.
> 
> @alarcher, do you have any suggestion here?
> 
> Thanks.

It was likely because these components were not integrated to userland for a while and some packages relied on them for backward compatibility.
Also when we integrated the dependency resolution for components these showed as unresolved leaves since they were not republished to the OI repository.
I am not sure about the consumers but some binary-only Solaris SVR4 packages may have needed these dependencies back then.
You can probably ignore them now.

mtelka:
Thanks @alarcher for sharing these details.  I believe there is nothing left preventing the merge of this PR.  Thank you.

PR #15564, opened 2024-01-09
transforms/defaults: make sure that files in usr/libexec are executable

AndWac:


transforms/defaults
```diff
@@ -98,6 +98,9 @@
 <transform file path=.*bin/.* -> default mode 0555>
 <transform file path=.*lib/inet/.* -> default mode 0555>
 
+# files in libexec should hace mode 0555
```
mtelka: have

AndWac: fixed, thanks.

PR #15571, opened 2024-01-09
Obsolete deprecated x11/network/rstart

mtelka:
See https://gitlab.freedesktop.org/xorg/app/rstart

PR #15576, opened 2024-01-09
X-incorporation: do not incorporate ancient packages

mtelka:


mtelka:
This fixes issue introduced by #15571.

PR #15596, opened 2024-01-11
icu: add upstream maintenance patches

mtelka:
This also removes few old patches that are not needed.

Please note that since #14456 there are some failing tests.  Details are here: https://unicode-org.atlassian.net/browse/ICU-22629.  Maybe we should switch back to gcc 10 (or gcc 12 if we want as new gcc as possible), but I do not know how it will affect other packages that depend on `icu`.  @AndWac, what do you think?


AndWac:
I haven't seen a problem yet with the actual situation and also haven't heard from any user. Thus, I would only switch back to an older gcc version if some errors will be reported.

PR #15611, opened 2024-01-11
openconnect: update 9.12

cgrzemba:


components/network/openconnect/openconnect.p5m
```diff
@@ -10,8 +10,7 @@
 #
 
 #
-# Copyright 2016 Jim Klimov
-# Copyright 2019 Michal Nowak
```
mtelka, line 14: You should not remove previous Copyrights.

PR #15650, opened 2024-01-14
alpine: add 2.26

iigs:
Makefile patch was not needed since this now is the same as other distros:  use --without-tcl

### DIFF
--- a/components/mail/alpine/alpine.p5m
+++ b/components/mail/alpine/alpine.p5m
@@ -1,0 +1,35 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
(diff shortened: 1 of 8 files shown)


components/mail/alpine/alpine.p5m
```diff
@@ -0,0 +1,35 @@
...
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2024 <contributor>
+#
```
AndWac, line 14: You should replace <contributor>

iigs: updated alpine.p5m

PR #15657, opened 2024-01-15
library/objfw: update to 1.0.6

Midar:


Midar:
Hm, this is odd. Why does -lgnutls not work on the build system, when this is coming from `pkg-config`?

mtelka:
Please do not forget to run all commands below to make sure everything is okay:
```
env - gmake sample-manifest
env - gmake publish
env - gmake test
env - gmake REQUIRED_PACKAGES
```
I'm sure you forgot at least one of them.

Midar:
I have indeed not done any of those and just increased the version on GitHub‘s editor, as I currently don’t have access to an OI system. However, all this release changes is a single bug fix that doesn’t even touch GnuTLS, so nothing has changed. That means that whatever was wrong already was wrong before. I’ll need to check since once I have access to an OI system again.

mtelka:
Sorry, but oi-userland does not work like that.  The build framework is always evolving, packages are updated all the time, so it is pretty common that at least metadata needs to be regenerated whenever you touch a component.  Often REQUIRED_PACKAGES needs update too.  And to make sure the build is okay you should test.

If you do not do your due diligence then it could happen that you will break the build server locking everybody else and somebody would need to fix your work.  It is far better if we could prevent that as much as possible.

Thank you.

mtelka:
And re the `continuous-integration/jenkins/pr-merge` check: you can often safely ignore it, because the machine running the check is not configured properly.

AndWac:
The problem is that we have removed 32 bit support for gnutls.

Midar:
> If you do not do your due diligence then it could happen that you will break the build server locking everybody else and somebody would need to fix your work. It is far better if we could prevent that as much as possible.

Even when just submitting a draft PR? That sounds … dangerous?

In any case, I'll try it locally first. Even though that complicates my release process quite a bit, given the number of distros / OSes I need to update.

AndWac:
This package would not create such drastic problems but others would. A rolling release model has some advantages and some disadvantages.

mtelka:
> > If you do not do your due diligence then it could happen that you will break the build server locking everybody else and somebody would need to fix your work. It is far better if we could prevent that as much as possible.
> 
> Even when just submitting a draft PR? That sounds … dangerous?

Of course, not.  You asked, so I just wrote my observations :-).

mtelka:
> This package would not create such drastic problems but others would. A rolling release model has some advantages and some disadvantages.

Any build failure on the official build server (the one that builds the release packages) causes problems that somebody (usually you @AndWac) have to look at and then solve.  Because it blocks all future builds until solved.

Midar:
Just for my understanding: But the CI failing will prevent it being merged to begin with (not to mention that the PR was draft), so it would not have broken the build?

In any case, I dropped 32 bit support now. My first reaction was to switch to OpenSSL, but I couldn't get it to work: It would always link against OpenSSL 1 instead of 3.

AndWac:
Our openssl mediator (look at the output of "pkg mediator") is set to 1.0 while our default openssl version in the build system is 3.1 (look at make-rules/shared-macros.mk). If a package doesnt't use the version of our build system it is typically a sign of poor following rules (eg. pkgconfig settings). Most build recipes can be fixed by proper use of OPENSSL_INCDIR, OPENSSL_LIBDIR and OPENSSL_PREFIX for CFLAGS, CPPFLAGS, LDFLAGS, and some others. Search for examples in our repository (eg. "cd oi-userland/components && find . -name Makefile -exec grep OPENSSL_PREFIX {} \; -print").

Our CI is not yet reliable, if it fails I usually have closer look of the possible causes. Most PR's will be checked by me by manually running a sequence of
gmake sample-manifest
gmake REQUIRED_PACKAGES
gmake test
gmake publish
and checking for failures or dirty files.

It is recommended to update your build environment to the latest released bits before creating a PR. Otherwise it is not unlikely that we will have differences or problems.

Midar:
I tried setting `OPENSSL_VERSION=3.1` before including `shared-macros.mk`, as well as adding `CONFIGURE_OPTIONS += "LDFLAGS=-L/usr/openssl/$(OPENSSL_VERSION)/lib/$(BITS)"` (not like this, but merging with the one already there, as otherwise the second would override the first). I could even see in the `config.log` that `-L/usr/openssl/3.1/lib/32` (for the 32 bit build) was before `-lssl` - and yet the linker failed resolving `SSL_set1_host`, indicating it was trying to link against OpenSSL 1. This was why this package used GnuTLS from the start and not OpenSSL. However, with GnuTLS dropping 32 bit support, I think this package should probably just drop 32 bit support as well, as this PR is doing now. Unless there is a general preference of OpenSSL over GnuTLS in OpenIndiana and it should be changed anyway?

AndWac:
Former maintainers had a preference of openssl over gnutls but I don't know the reasons.

PR #15658, opened 2024-01-15
icu: update to 74.2; preserve ICU 72 libraries

mtelka:
The most important thing this PR does is removal of the `README` file that described armageddon when the major ICU version is changed.  Such armageddon should no longer happen since we do not remove the previous version of essentially needed libraries.  Technically, we just add new libraries for ICU version 74.2 and keep the previous libraries (version 72.1).  The only real update here is in the `developer/icu` package.  But since no other package needs it during runtime such update is of very low risk.

Once this PR is merged, successfully built and installed we will start to support multiple versions of ICU.  This should be okay because ICU is designed with this in mind and uses versioned API.  So in a case when two versions of single library appears in the same address space everything should work properly because no API function should overlap, clash and cause problems.

Also, after this, we will no longer need to rebuild all ICU consumers as soon as possible to make them working again.  We will have plenty of time to do so lazily when such a component really needs a rebuild.  There will be no hard deadline for doing so.

The other thing implemented here is a split of the library package for (the new) ICU 74 to two: the first one contains 64-bit libraries, the other one 32-bit libraries.  This is to easily track what package really needs 32-bit ICU.  It will allow us to remove 32-bit support here without breaking anything.

Practical tips for merging this:
- please make sure this is built alone on the build server (with no other changesets)
- please do not start another build job until all possible issues are solved
- please be prepared for the armageddon scenario as described in the `README` file (so please backup old libraries, etc)
- if something won't go smooth please let me know what exactly you had to do to recover the build server

The last, but not least:  In the `README` there is stated that both `pkg` and `sshd` depends on ICU.  I looked a bit at both but I failed to see such dependency.  I probably didn't look hard enough. @AndWac, could you please enlighten me how to see such dependency?  Without clarifying this last point I do not recommend to merge this, because it looks like I miss something important here and my work could cause some unexpected harm.

Thank you.


AndWac:
Sorry, I don't remember the details of the problems I had when I updated ICU last time. For pkg it is most probably a python package that will be broken when the old ICU libraries are missing. For ssh I can't imagine a cause.

mtelka:
Okay, thanks. I'm satisfied with this explanation.  Please merge once you are ready.

PR #15669, opened 2024-01-16
pkg: add pkg.human-version and userland.info.component to all packages

mtelka:

PR #15676, opened 2024-01-17
pkg: clobber in pkg subdir to force package publish

mtelka:
There are two distinct, but related problems with package publish:

1. I noticed that the package publish is sometimes not done at all when there is nothing changed.
2. In some other cases the publish is done but [silently fails](https://github.com/OpenIndiana/oi-userland/pull/15669#issuecomment-1895792424).

To solve both cases (the first one for sure, the second one hopefully) I propose to clobber packages before we try to publish them.

### DIFF
--- a/components/openindiana/pkg/Makefile
+++ b/components/openindiana/pkg/Makefile
@@ -56,6 +56,7 @@ $(BUILD_DIR)/$(MACH)/.built: $(SOURCE_DIR)/.downloaded
 	cd $(SOURCE_DIR)/src && \
 	PATH=$(PATH) $(MAKE) && PATH=$(PATH) $(MAKE) install && \
 	cd pkg && \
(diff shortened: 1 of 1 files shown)


mtelka:
It is still weird because in the [log](https://hipster.openindiana.org/logs/oi-userland/11094/openindiana.pkg.publish.log) from the build with this PR merged there is now completely missing the log for publish.  Only republish is there.  The publish step definitely happened because the repo directory was clobbered here:

```
hipster.openindiana.org --> 1 job
rm -rf pkgtmp
hipster.openindiana.org --> 1 job
rm -rf /jenkins/jobs/oi-userland/workspace/components/openindiana/pkg/pkg/src/pkg/../../packages/i386 Makefile.link
```

So without the actual publish the republish later would fail.

PR #15684, opened 2024-01-18
history: remove library/audio/libmikmod

mtelka:
... because it is not obsolete.  It is build in `library/libmikmod`.


mtelka:
But after looking closer the `library/audio/libmikmod` is unused in OpenIndiana and the latest upstream was released 6 years ago.  So maybe we should simply (properly) obsolete it?  OTOH, the project itself is not dead yet (commits in their git repo in September 2023) so for now I've no problem if it stays.

PR #15685, opened 2024-01-18
mtx: add build recipe version 1.3.12

cgrzemba:


components/sysutils/mtx/Makefile
```diff
@@ -0,0 +1,36 @@
...
+COMPONENT_ARCHIVE_URL=  https://downloads.sourceforge.net/$(COMPONENT_NAME)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH= sha256:0261c5e90b98b6138cd23dadecbc7bc6e2830235145ed2740290e1f35672d843
+COMPONENT_LICENSE=		GPLv2
+COMPONENT_LICENSE_FILE= LICENSE
+COMPONENT_SUMMARY=	Media Changer Tools
+COMPONENT_DESCRIPTION=  These tools are used to view information about, and to control, Media Changer devices such as Tape and DVD/CD libraries
+
+include $(WS_MAKE_RULES)/common.mk
+
+REQUIRED_PACKAGES += system/library
```
mtelka, line 33: Please run `gmake REQUIRED_PACKAGES`.

components/sysutils/mtx/Makefile
```diff
@@ -0,0 +1,36 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2024 Carsten Grzemba
+#
+BUILD_BITS= 64
+BUILD_STYLE= configure
```
mtelka: These two are default so I suggest to remove them to make things simpler.

components/sysutils/mtx/Makefile
```diff
@@ -0,0 +1,36 @@
...
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		mtx
+COMPONENT_VERSION=	1.3.12
+COMPONENT_REVISION=	0
```
mtelka: This is not needed and we (usually) do not do that.  But since it is harmless otherwise then it could stay if preferred so.

components/sysutils/mtx/Makefile
```diff
@@ -0,0 +1,36 @@
...
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_PROJECT_URL=	https://sourceforge.net/projects/mtx
+COMPONENT_ARCHIVE=  $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=  https://downloads.sourceforge.net/$(COMPONENT_NAME)/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH= sha256:0261c5e90b98b6138cd23dadecbc7bc6e2830235145ed2740290e1f35672d843
+COMPONENT_LICENSE=		GPLv2
+COMPONENT_LICENSE_FILE= LICENSE
+COMPONENT_SUMMARY=	Media Changer Tools
+COMPONENT_DESCRIPTION=  These tools are used to view information about, and to control, Media Changer devices such as Tape and DVD/CD libraries
```
mtelka: This is unused, so please remove.

components/sysutils/mtx/patches/02-install-man.patch
```diff
@@ -0,0 +1,17 @@
+It is not necessary to install the manpages. IPS finds them in src path
```
mtelka: But `gmake sample-manifest` does not.  It is better when the `sample-manifest.p5m` match as close as possible the real manifest, so if the build works properly without this patch then please remove it.

mtelka (changes requested):

PR #15702, opened 2024-01-19
mate_install: replace xscreensaver by mate-screensaver

mtelka:
I played a bit with screensavers and I found that `xscreensaver` is hard to deal with.  First, I encountered [black screen during OpenIndiana installation](https://www.illumos.org/issues/16037) (very annoying for newcomers).  Then after the installation I saw few screen lockups where I was unable to get out of them without a hard hammer.  There are also other [more](https://www.illumos.org/issues/16185) or [less](https://www.illumos.org/issues/16199) serious issues with `xscreensaver`.  As a result I simply uninstalled the `desktop/xscreensaver` and installed `desktop/mate/mate-screensaver` and I no longer see any problems with screensavers or screen locking.

So I propose to replace `xscreensaver` by `mate-screensaver` in the default desktop installation.


mtelka:
There is also [this thread](https://openindiana.org/pipermail/openindiana-discuss/2024-January/026698.html) showing how confusing the xscreensaver is.

PR #15753, opened 2024-01-23
postfix: update to 3.8.5

AndWac:
I don't know why postfix depends on library/icu-74/32. It should be 64 bit only.


mtelka:
@AndWac, hm, yes, you are right.  The automatic dependency detection didn't work well.  I'm looking at it.  Please do not merge anything that depends on ICU until this is solved.

mtelka:
It looks like the problem is extra `/usr/lib` here:
```
$ elfdump build/prototype/i386/usr/sbin/postfix | egrep -i 'rpath|runpath'
      [15]  RUNPATH           0x3c39              /usr/gcc/13/lib/amd64:/usr/lib/
      [16]  RPATH             0x3c39              /usr/gcc/13/lib/amd64:/usr/lib/
$
```
Also, it looks like this is `postfix`-specific issue only.  So hopefully other components are not affected (unless they do have similar issue).

mtelka:
Heh, `pkglint` complains about that too:
```
Lint engine setup...
Starting lint run...
WARNING userland.action001.3      32-bit runpath in 64-bit binary, '/data/builds/ul-workspace/components/mail/postfix/build/prototype/i386/usr/lib/postfix/anvil' includes '/usr/lib/'
WARNING userland.action001.3      32-bit runpath in 64-bit binary, '/data/builds/ul-workspace/components/mail/postfix/build/prototype/i386/usr/lib/postfix/bounce' includes '/usr/lib/'
```

mtelka:
`components/mail/postfix/patches/01-libicu_rpath.patch` is the culprit.

mtelka:
@AndWac, what exactly was the reason to add the patch?  I just ran `gmake sample-manifest` successfully without it.

mtelka:
@AndWac, the funny fact is that I completely missed your initial comment here when you created the PR and I spotted the issue with 32-bit ICU by accident only.

PR #15755, opened 2024-01-23
bind: fix link-local addresses scoped with interface names

Bill-Sommerfeld:
If bind's configure can't find if_nametoindex(), it only lets you use interface ids (fe80::1%5) instead of interface names (like fe80::1%wan0); pass LIBS=-lsocket to configure and it will find the function in -lsocket.

### DIFF
--- a/components/network/bind/Makefile
+++ b/components/network/bind/Makefile
@@ -30,6 +30,7 @@ COMPONENT_TEST_DIR = $(@D)/tests
 
 COMPONENT_NAME=		bind
 COMPONENT_VERSION=	9.18.21
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	BIND DNS name server and configuration tools.
 COMPONENT_DESCRIPTION=	BIND is open source software that implements the Domain Name System \
                         (DNS) protocols for the Internet.  This package contains the DNS \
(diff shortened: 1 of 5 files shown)


components/network/bind/Makefile
```diff
@@ -83,6 +83,7 @@ CONFIGURE_OPTIONS += --with-tuning=default
 CONFIGURE_OPTIONS += --disable-linux-caps
 CONFIGURE_OPTIONS += READLINE_CFLAGS=-I/usr/include/readline
 CONFIGURE_OPTIONS += READLINE_LIBS="-lreadline -ltermcap"
+CONFIGURE_OPTIONS += LIBS="-lsocket"
 
 # Install.
 #
```
AndWac: When you change a package you either need to change COMPONENT_VERSION or increment/set COMPONENT_REVISION. So please add
COMPONENT_REVISION= 1

Bill-Sommerfeld: Done.

AndWac: Where? I cannot see it in the Makefile...

mtelka: Sorry, not done.

Please also make sure you ran following make targets at least once, all of them passed and there is nothing changed in the bind component directory: `sample-manifest`, `publish`, `test`, `REQUIRED_PACKAGES`.  Thank you.

Bill-Sommerfeld: Arrgh, miss a `git add` and I look like an idiot.

I'm chasing down a test issue (tests work fine when I cd into the build tree, but when run from the component directory, they all fail with errors resembling:
```
./aes_test: line 43: libtool_execute_magic: unbound variable
FAIL aes_test (exit status: 1)
```
 it looks like it's the result of SHELLOPTS being exported rather than my change; I see many component makefiles have a `unexport SHELLOPTS` and will be adding one to bind's.

mtelka (approved):

PR #15756, opened 2024-01-24
python39: update to 3.9.18

Bill-Sommerfeld:


components/python/python39/Makefile
```diff
@@ -109,6 +109,7 @@ CONFIGURE_ENV += ac_cv_func_getentropy=no
 CONFIGURE_OPTIONS += --infodir=$(CONFIGURE_INFODIR)
 CONFIGURE_OPTIONS += --enable-shared
 CONFIGURE_OPTIONS += --enable-optimizations
+CONFIGURE_OPTIONS += --enable-loadable-sqlite-extensions
 CONFIGURE_OPTIONS += --with-dtrace
 CONFIGURE_OPTIONS += --with-system-expat
 CONFIGURE_OPTIONS += --with-system-ffi
```
AndWac, line 114: 1. When you change a package you either need to change COMPONENT_VERSION or increment/set COMPONENT_REVISION.
2. Our python builds receipts are broken for some time. We cannot publish updates before this has been fixed.

Bill-Sommerfeld: > Our python builds receipts are broken for some time. We cannot publish updates before this has been fixed.

I'd be happy to help with that if possible if I can get a brain dump on what needs to happen (I'm also interested in helping bring newer python versions into OI)

AndWac: When you run gmake sample-manifest you'll find out that the newly generated one is missing the tk library. I haven't found out yet why this doesn't get build anymore.

Bill-Sommerfeld: I think I found it - setup.py isn't looking in /usr/lib/64 for libtcl and libtk which is the only place you'll find it if there aren't 32-bit libraries.

mtelka: > I'm also interested in helping bring newer python versions into OI

Such help is always welcome.  Python 3.12 would be great.  Thanks.

mtelka: @AndWac, FYI, addition of new major Python version needs special handling to make sure it does not break rest of oi-userland build framework or user's machines, so please wait for my review/approval when such PR appears.

@Bill-Sommerfeld, do not worry about the above :-).

AndWac: Sure

mtelka (approved):

PR #15757, opened 2024-01-24
gtk+4: add 4.10.5

iigs:
This is not the latest version of gtk+4 since openindiana does not have glib 2.76.  I had to choose between 4.10.5 or 4.11.3 which was the last to have glib 2.74 support.  4.10.5 was released a month later--it was chosen for this build.

"gmake test" fails somewhere in the gtk4 meson/ninja/Make files (the actual tests of gtk4 complete):

Ok:                 700
Expected Fail:      0
Fail:               29
Unexpected Pass:    0
Skipped:            1
Timeout:            5

Full log written to /foo/oi-userland/components/library/gtk+4/build/amd64/meson-logs/testlog.txt
FAILED: meson-internal__test
/usr/bin/meson test --no-rebuild --print-errorlogs
ninja: build stopped: subcommand failed.

This build appears to be solid (tested well only with nViida GPU).  The demo apps do not appear to break with the current nimbus theme.  Themeing looks better with the latest "solaris-11" theme work from rocketman.

Media support not compiled in this build since it requires encumbered x264 support.


iigs:
Added docutils python module so man pages can build.

Added ENV = /usr/bin/env so tests can build.  I don't know what changed to need this since the end of January.

When running gmake tests, I found having a single terminal window open allows more tests to be ok.

Only 699 tests are ok now.  I don't have a record of what tests were ok in January.

Assorted other cleanups in Makefile--removed cruft that does not affect build.

components/library/gtk+4/Makefile
```diff
@@ -0,0 +1,98 @@
...
+COMPONENT_FMRI=		library/desktop/gtk4
+
+BUILD_STYLE=		meson
+
+include $(WS_MAKE_RULES)/common.mk
+
+PATH=$(PATH.gnu)
+
+gcc_OPT = -O2
+LD_Z_IGNORE=
```
mtelka: Could you please add a comment explaining why we do need these two lines?

components/library/gtk+4/Makefile
```diff
@@ -0,0 +1,98 @@
...
+
+# Uses isinf, etc. only available in c99 and newer
+CFLAGS.gcc += -std=gnu11
+
+CONFIGURE_OPTIONS += -Dwayland-backend=false
+# media-gstreamer requires encumbered x264 support
+CONFIGURE_OPTIONS += -Dmedia-gstreamer=disabled
+CONFIGURE_OPTIONS += -Dman-pages=true
+
+CONFIGURE_OPTIONS += -Dintrospection=disabled
```
mtelka: Why is this needed?

components/library/gtk+4/Makefile
```diff
@@ -0,0 +1,98 @@
...
+
+COMPONENT_BUILD_ENV += CC=$(CC)
+COMPONENT_BUILD_ENV += CFLAGS="$(CFLAGS)"
+
+COMPONENT_POST_INSTALL_ACTION  = \
```
mtelka, line 51: Why do we need these `elfedit`s?

components/library/gtk+4/Makefile
```diff
@@ -0,0 +1,98 @@
...
+
+COMPONENT_TEST_TRANSFORMS += \
+        '-e "s/seed:[0-9]*/seed:/"'
+# get rid of the timestamps from the test results
+COMPONENT_TEST_TRANSFORMS += '-e "s/[0-9]*[0-9].[0-9][0-9]s//" '
+COMPONENT_TEST_TRANSFORMS += '-e "s/OK      [0-9]//" '
+COMPONENT_TEST_TRANSFORMS += \
+        '-n ' \
+        '-e "/OK/p" '
+COMPONENT_TEST_MASTER = $(COMPONENT_TEST_RESULTS_DIR)/results-$(MACH64).master
```
mtelka, line 65: Is this line really needed?

components/library/gtk+4/patches/01-meson.patch
```diff
@@ -0,0 +1,17 @@
+--- gtk-4.10.5/testsuite/gtk/meson.build.~1~	2023-08-05 05:50:38.000000000 -0400
```
mtelka, line 1: Why do we need this patch?

iigs:
removed:
gcc_OPT = -O2
LD_Z_IGNORE=
They must have been added for either prior glib integration or needed for an earlier attempt to build an earlier version of gtk4.

enabled introspection.  It must have been added for either prior glib integration or needed for an earlier attempt to build an earlier version of gtk4.

COMPONENT_POST_INSTALL_ACTION lines are needed for this:
        libjpeg.so.8 =>  (file not found)
        libjpeg.so.8 =>  /usr/lib/libjpeg8-turbo/lib/amd64/libjpeg.so.8

This is what one of  the test lines look like:
708/736 gtk:reftest / reftest rotated-layout.ui                                                     OK              2.95s   1 subtests passed

Will the default transform remove the time or will the Oracle provided COMPONENT_TEST_MASTER work?  I can't test this since gmake test fails before producing the snapshot.

gtk4 does not during compile without 01-meson.patch.  Oracle developed this patch--I may have had to port it for this version of gtk4.

mtelka:
> enabled introspection. It must have been added for either prior glib integration or needed for an earlier attempt to build an earlier version of gtk4.

Do we need an explicit introspection enable?  If so, please add a comment to the `Makefile` saying so (and possibly also explaining why).

> COMPONENT_POST_INSTALL_ACTION lines are needed for this: libjpeg.so.8 => (file not found) libjpeg.so.8 => /usr/lib/libjpeg8-turbo/lib/amd64/libjpeg.so.8

Please add a comment to the `Makefile` explaining the need for `elfedit`.

> Will the default transform remove the time or will the Oracle provided COMPONENT_TEST_MASTER work? I can't test this since gmake test fails before producing the snapshot.

I believe you should make sure that all tests pass.  Or, at least provide test results even with few failing tests.

> gtk4 does not during compile without 01-meson.patch. Oracle developed this patch--I may have had to port it for this version of gtk4.

If it is ported from solaris-userland, then please add a comment to the patch referencing the source.

Thank you.

PR #15765, opened 2024-01-24
bind: Multiple fixes which allow tests to pass.

Bill-Sommerfeld:
1) unexport SHELLOPTS to avoid libtool wrapper failures

2) Clear LD_B_DIRECT; direct binding of bind's shared libraries breaks several of the self tests.

3) Fix how we set the timezone to GMT in one of the tests.

4) When adding test ipv4 addresses to lo0, use a /32 prefix len so that it doesn't default to a /8.

### DIFF
--- a/components/network/bind/test/results-all.master
+++ b/components/network/bind/test/results-all.master
@@ -1,102 +1,109 @@
-FAIL: serve-stale
-PASS: doth
-PASS: acl
-PASS: additional
-PASS: addzone
-PASS: allow-query
-PASS: auth
-PASS: autosign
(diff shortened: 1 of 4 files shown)


components/network/bind/test/results-all.master
Bill-Sommerfeld, line 1: Not sure why there's this much churn in this file, but all the FAILs are gone.

mtelka:
I suggest to squash this with #15755.

Bill-Sommerfeld:
Squashed into #15755.

PR #15808, opened 2024-01-26
libvpx: update to 1.14.0 & drop 32 bit

AndWac:


AndWac:
This version is ABI incompatible with earlier releases according to its own changelog. vlc and opal don't compile with it yet so we need to wait until the dependent projects updated their dependency.

PR #15809, opened 2024-01-26
shared-macros.mk: build, install and test in clean environment

mtelka:
This is something we had to do long time ago.  Actually, in past it was partly like that, but it was [changed](https://github.com/OpenIndiana/oi-userland/commit/61c3737c7c8ec8e7edc610acb73c9b0ec36ea13b) for unclear reasons to the current state with dirty environment passed down to component's `Makefile`.

I _believe_ the polluted environment is causing some strange build failures like [this one about a month ago for perl/IO-Compress-Brotli](https://hipster.openindiana.org/logs/oi-userland/10968/perl.IO-Compress-Brotli.publish.log).  I'm **sure** many test issues are caused by this and there is need to workaround such problems.  [Here](https://github.com/OpenIndiana/oi-userland/blob/927d03bfc81a53d599ef8ca1b3193b9d84d251fc/components/network/bind/Makefile#L115) is an recent example of a workaround.  Today I was bitten by this while running tests for `automake`.

Once this is integrated we will be able to simplify many `Makefile` files and remove workarounds related to polluted environment.

OTOH, it should be clearly stated that there could be rare cases when some components _depend_ on dirty environment so they will fail to build with this change in place.  If that's the case then the fix is often easy and something just needs to be added to the component's environment.  As a last resort there is always possibility to add `ENV = /usr/bin/env` right after the `include shared-macros.mk` line in the component's `Makefile` to return back to the dirty environment for the particular component.

To test this I did build and test of handful components to make sure they work.


AndWac:
Maybe we should add a note in our doc folder about this.

mtelka:
> Maybe we should add a note in our doc folder about this.

Not sure about that.  The impact of this change is almost always positive.  So far there is no known component where it causes regression so I expect such failures are rare, if there is any.  I do not think it needs to be specially documented.  OTOH, if we will find more than usual number of questions related to this I'll consider to document it.

Anyway, I went through the `doc` folder to see if there is anything in contradiction with this change and I found nothing.

AndWac:
Build kvm fails since 27th January. Its most probably either caused by this PR or the integrated illumos-gate PR from the same day (16114 ksh should not be isaexec'd). The error is irritating me. Do you have an idea?

/usr/gcc/7/bin/gcc -Wl,-Bdirect -Wl,-zfatal-warnings -Wl,-zassert-deflib -Wl,-zguidance -m64 -shared -nodefaultlibs -std=gnu89 -Wl,-M/export/home/andreas/oi-userland/components/openindiana/illumos-gate/illumos-gate//usr/src/common/mapfiles/common/map.pagealign -Wl,-M/export/home/andreas/oi-userland/components/openindiana/illumos-gate/illumos-gate//usr/src/common/mapfiles/common/map.noexdata -Wl,-ztext -Wl,-zdefs -Wl,-zignore -Wl,-M/export/home/andreas/oi-userland/components/openindiana/illumos-gate/illumos-gate//usr/src/cmd/mdb/common/modules/conf/mapfile-extern -L/../../../illumos-gate/illumos-gate/proto/root_i386/lib/amd64 -L/../../../illumos-gate/illumos-gate/proto/root_i386/usr/lib/amd64 -o kvm.so kvm_mdb.o -lc
ld: warning: dynamic library found on default search path (/lib/amd64): libc.so
make: *** [Makefile:315: kvm.so] Error 1

mtelka:
Hm, strange.  I kicked the build to see if I'm able to repro locally...

mtelka:
FYI: reproduced.

mtelka:
And yes, the failure is caused by the clean env.

mtelka:
BTW, I'm working on fix.

mtelka:
The root cause is that build depends on the `PWD` environment variable and `env -i` cleaned that.

Okay, this is first component that depended on dirty environment. :-)

PR #15851, opened 2024-01-29
inkscape: only depend on packages for the main python version

AndWac:


mtelka:
I'm curious why this change was needed?

AndWac:
Because some python-37 packages are already missing, eg.
pkg install: No matching version of library/python/cachecontrol-37 can be installed:
  Reject:  pkg://openindiana.org/library/python/cachecontrol-37@0.13.1-2023.0.0.0
  Reason:  All acceptable versions of 'require' dependency on library/python/msgpack-37 are obsolete

mtelka:
Damn, that had to be solved differently.  BTW, what exactly was the command causing that failure?

...and, why inkscape needs Python modules at all?

AndWac:
The jenkins job failed because inkscape was installed on the build server and it wasn't able to update inkscape from 1.2.2 to 1.3.2 because of unsolvable dependencies. The explanation for the introduction of python package dependencies is in the manifest: # inkscape extensions require python module packages

mtelka:
What exactly in the inkscape package requires Python?  That comment is nice, but I failed to find the real need there.

mtelka:
[This](https://openindiana.org/pipermail/openindiana-discuss/2024-January/026765.html) is the better way how to solve the problem.

AndWac:
This is only true if there is only one python version available for all packages. If there are more than one (say eg. we later have python 3.11 packages) then any mixture of package versions will satisfy the generic dependencies.

mtelka:
Sorry, I do not understand what you are trying to say.

Yes, the generic (unversioned) dependencies would be satisfied by any installed Python version.  And no, not _any version mixture_ will satisfy that.  If we do have `runtime/python-39` installed, then `appdirs-39`, `cachecontrol-39`, `lxml-39` etc. satisfy the dependencies.  If we do have also `runtime/python-312` installed, then with unversioned dependencies we will also get `appdirs-312`, `cachecontrol-312`, `lxml-312` etc. installed.  So we will get support for all installed Python versions.  And the support will be complete, not a mixture.

Anyway, I still do not understand what is _This_ (your first word in your first sentence [here](https://github.com/OpenIndiana/oi-userland/pull/15851#issuecomment-1915198288)).  Are you arguing that my solution [here](https://openindiana.org/pipermail/openindiana-discuss/2024-January/026765.html) for your inkscape update problem would not work if we do have other Python versions?

Hmm, I feel I miss something obvious here... :-(

AndWac:
Yes, that was what I meant. But maybe I have a problem to understand package dependencies.
If I look at cachecontrol, eg. by pkg contents -m -r cachecontrol, I see
...
depend fmri=library/python/cachecontrol-37@0.13.1,5.11-2023.0.0.0 predicate=runtime/python-37 type=conditional
depend fmri=library/python/cachecontrol-39@0.13.1,5.11-2023.0.0.0 predicate=runtime/python-39 type=conditional
depend fmri=library/python/cachecontrol-37@0.13.1,5.11-2023.0.0.0 fmri=library/python/cachecontrol-39@0.13.1,5.11-2023.0.0.0 type=require-any
...
So, the last line states that any of the two package will satisfy the dependency constraint.

mtelka:
Yes, if there is only the last line (from your output) then it would behave as you [wrote above](https://github.com/OpenIndiana/oi-userland/pull/15851#issuecomment-1915198288), but since there are also other two lines then it behaves as [I wrote](https://github.com/OpenIndiana/oi-userland/pull/15851#issuecomment-1915241129).  See `man pkg.7` for details.

mtelka:
In short, the dependency pattern (require-any + conditionals) used for Perl and Python (and maybe few others, possibly Ruby??? I'm not sure) will make sure that:

- at least one Python,Perl,etc. version is installed (if there was none previously), and
- for all installed Python,Perl,etc. versions we will get the versioned package variant installed too.

AndWac:
Ok, thanks for clarification. I still wonder how the conditionals have been defined as I haven't seen it in cachecontrol's manifest.

mtelka:
This is done automatically, of course.  Humans would do mistakes if not done automatically :-).  Look for `mkgeneric` in `ips.mk` for implementation details.

mtelka:
BTW and FYI, technically, the problem we are talking about here with the solution [I sent to the mailing list](https://openindiana.org/pipermail/openindiana-discuss/2024-January/026765.html) is not caused by the slow obsoletion of Python 3.7 support, but by the IPS repo cleanup you do from time to time.

And also by the fact that you probably never received this my mail:
```
From: Marcel Telka <REDACTED>                                                                                                                                                                                                                 
To: Andreas Wacknitz <REDACTED@gmx.REDACTED>                                                                                                                                                                                                             
Subject: OI packages                                                                                                                                                                                                                                 
Date: Sat, 28 Oct 2023 22:38:15 +0200                                                                                                                                                                                                                

Hi Andreas,

Please add back to the OI pkg repo these packages:

pkg://openindiana.org/library/python/argcomplete-37@3.1.1-2023.0.0.0:20230612T154623Z
pkg://openindiana.org/library/python/asttokens-37@2.2.1-2023.0.0.0:20230223T170625Z
pkg://openindiana.org/library/python/autopage-37@0.5.1-2023.0.0.0:20230418T094616Z
pkg://openindiana.org/library/python/babel-37@2.12.1-2023.0.0.0:20230425T205155Z
pkg://openindiana.org/library/python/black-37@23.3.0-2023.0.0.0:20230403T161129Z
pkg://openindiana.org/library/python/bleach-37@6.0.0-2023.0.0.0:20230222T064221Z
pkg://openindiana.org/library/python/build-37@0.10.0-2023.0.0.0:20230114T092859Z
pkg://openindiana.org/library/python/cachetools-37@5.3.1-2023.0.0.0:20230528T080645Z
pkg://openindiana.org/library/python/certifi-37@2023.5.7-2023.0.0.0:20230508T082659Z
pkg://openindiana.org/library/python/cffi-37@1.15.1-2023.0.0.0:20230217T223509Z
```
... the list continued with many entries.

PR #15852, opened 2024-01-29
kvm: switch to gcc-10

mtelka:


mtelka:
As a side effect it also fixes [this issue](https://github.com/OpenIndiana/oi-userland/pull/15809#issuecomment-1915059295).

PR #15866, opened 2024-01-31
Add text/dwdiff

mtrower:
Hi,

This adds a build recipe for the dwdiff package, so that the current package (which lacks source) can be removed from the repositories.


components/text/dwdiff/Makefile
```diff
@@ -0,0 +1,50 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2024 Matthew R. Trower
+#
+
+BUILD_BITS=64
```
mtelka: This is default, so please remove it.

components/text/dwdiff/dwdiff.license
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
mtelka: This is copy of the `COPYING` file.  It is better to do not add it to git and instead add `COMPONENT_LICENSE_FILE = COPYING` to `Makefile`.  This will avoid possible future usage of outdated copy of the license.

components/text/dwdiff/Makefile
```diff
@@ -0,0 +1,50 @@
...
+TEST_TARGET=$(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+CONFIGURE_OPTIONS+=	INSTALL=/usr/ucb/install
+CONFIGURE_OPTIONS+=	CFLAGS=-D_POSIX_PTHREAD_SEMANTICS
+
+# Wants to be built in-tree, so copy source tree to build dir
+CONFIGURE_SCRIPT = $(@D)/configure
+COMPONENT_PRE_CONFIGURE_ACTION=	(cp -a $(SOURCE_DIR)/* $(@D) )
```
mtelka: Please use *cloney* for this.

components/text/dwdiff/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+TEST_TARGET=$(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+CONFIGURE_OPTIONS+=	INSTALL=/usr/ucb/install
+CONFIGURE_OPTIONS+=	CFLAGS=-D_POSIX_PTHREAD_SEMANTICS
+
+# Wants to be built in-tree, so copy source tree to build dir
+CONFIGURE_SCRIPT = $(@D)/configure
```
mtelka: Is this really needed?

mtelka (changes requested):
Your copy of oi-userland is at least 3 weeks old.  Please rebase your change on recent oi-userland and re-run all required make targets to get updated version of generated files.

components/text/dwdiff/Makefile
```diff
@@ -0,0 +1,50 @@
...
+COMPONENT_ARCHIVE_HASH= \
+	sha256:df16fec44dcb467d65a4246a43628f93741996c1773e930b90c6dde22dd58e0a
+COMPONENT_LICENSE=      GPLv3
+
+TEST_TARGET=$(NO_TESTS)
+
+include $(WS_MAKE_RULES)/common.mk
+
+CONFIGURE_OPTIONS+=	INSTALL=/usr/ucb/install
+CONFIGURE_OPTIONS+=	CFLAGS=-D_POSIX_PTHREAD_SEMANTICS
```
mtelka, line 40: I recommend to add comment for the above two lines to explain why they are needed.  It is not obvious.  Thank you.

components/text/dwdiff/Makefile
mtrower: Alright.  It is part of templates/configure.mk that I used as a base, so maybe it should be removed from that template file?  Or are those templates not meant to be used as basis for new components?

EDIT: On second thought, probably not; that would hurt discoverability of the variable, and it's a pretty important one.  Maybe it just needs to be commented out, or maybe you can just bring it up in review like you did here.

/shrug

components/text/dwdiff/Makefile
mtrower: dwdiff isn't designed with out-of-tree builds in mind.  It is full of relative references to in-tree files.  It might build without this option, but then it's not an in-tree build --- configure will be simultaneously referencing components from both the build directory (containing a full copy of the source) and the original source directory.

With this option, we are just doing a standard in-tree build, which is semantically what I am trying to achieve in the first place.   If you really want me to remove it though, I can try to do so.

mtrower:
Hi Marcel,

Thank you for your feedback.  I would like to update my pull request to address your items, however I am unfamiliar with the nuances of the github workflow, and more importantly the preferences of this project.

I see that I can just push a second commit to my branch, and it will show up in the pull request; would you like me to do that?  Will you be able to squash that down when you merge the PR, or do you need me to force-push an updated commit to my branch?

Please advise on your preferences.

components/text/dwdiff/Makefile
mtelka: Yes, that (the second thought).  Templates will never cover all possible cases.

components/text/dwdiff/Makefile
mtelka: Okay, makes sense.

mtelka:
Please squash and force-push.  Thanks.

components/text/dwdiff/Makefile
mtrower: I'm not familiar with this tool.  I guess that it is an oi-userland special tool?  I see that it's in the tree, but I don't see any documentation for it.  Why is this tool preferred?

Reading the source, I see a mode that creates a symlink tree (default) and a mode that copies (equivalent to what I did above).  I'm guessing that you want me to use the symlink mode (presumably to save on space and time), like so?

`$(CLONEY) $(SOURCE_DIR) $(@D)`

mtelka: Yes, something like that.  The symlink mode is the default mode.  If it works, then  please use it.

mtrower:
One more question --- because I am providing `COMPONENT_SIG_URL`, `tools/.gnupg/pubring.gpg` becomes modified on `gmake download`.  Should I be including this file in my commit?  If not, is it safe to add it to `.gitignore`?

mtelka:
> One more question --- because I am providing `COMPONENT_SIG_URL`, `tools/.gnupg/pubring.gpg` becomes modified on `gmake download`. Should I be including this file in my commit? If not, is it safe to add it to `.gitignore`?

Not sure about this.  @AndWac probably knows more.

mtrower:
> > One more question --- because I am providing `COMPONENT_SIG_URL`, `tools/.gnupg/pubring.gpg` becomes modified on `gmake download`. Should I be including this file in my commit? If not, is it safe to add it to `.gitignore`?
> 
> Not sure about this. @AndWac probably knows more.

After consulting the variable documentation again, and examining the commit history, it appears I need to include it.

Let me know if this update adequately addresses your concerns.

components/text/dwdiff/Makefile
mtelka: Okay, `getpwuid_r(3C)` explains `_POSIX_PTHREAD_SEMANTICS`, but it is still unclear why we do need `/usr/ucb/install`.  What standards the `/usr/ucb/install` conforms to?  What is wrong with the default `/usr/gnu/bin/install` or `/usr/sbin/install`?

components/text/dwdiff/Makefile
```diff
@@ -0,0 +1,50 @@
...
+
+# Wants to be built in-tree, so copy source tree to build dir
+COMPONENT_PRE_CONFIGURE_ACTION=	( $(CLONEY) $(SOURCE_DIR) $(@D) )
+
+# Build dependencies
+REQUIRED_PACKAGES+= 	text/gnu-gettext
+
+# Auto-generated dependencies
+REQUIRED_PACKAGES += library/icu
```
mtelka: This is strange.  How you achieved the dependency on `library/icu`?  Is your system up-to-date?

mtelka (changes requested):

AndWac:
We don't have a rule for adding .pubring.gpg yet (at least one that I know of). The build server usually adds public keys by itself. Alas sometimes this failed in the past and we chose to don't provide the key and stick to the SHA256 instead. Maybe its time to discuss and make a decision about it.

AndWac:
When I run gmake REQUIRED_PACKAGES and gmake publish here, I'll get the following diffs:
╰─➤  git diff Makefile
diff --git a/components/text/dwdiff/Makefile b/components/text/dwdiff/Makefile
index cfc372df6f..7de7035b6d 100644
--- a/components/text/dwdiff/Makefile
+++ b/components/text/dwdiff/Makefile
@@ -46,5 +46,5 @@ COMPONENT_PRE_CONFIGURE_ACTION=       ( $(CLONEY) $(SOURCE_DIR) $(@D) )
 REQUIRED_PACKAGES+=    text/gnu-gettext

 # Auto-generated dependencies
-REQUIRED_PACKAGES += library/icu
+REQUIRED_PACKAGES += $(ICU_LIBRARY_PKG)
╭─andreas@saturn ~/oi-userland/components/text/dwdiff  ‹mtrower-dwdiff*›
╰─➤  git diff pkg5
diff --git a/components/text/dwdiff/pkg5 b/components/text/dwdiff/pkg5
index c7861361a4..402e10f582 100644
--- a/components/text/dwdiff/pkg5
+++ b/components/text/dwdiff/pkg5
@@ -1,6 +1,6 @@
 {
     "dependencies": [
-        "library/icu",
+        "library/icu-74",
         "system/library",
         "text/gnu-gettext"
     ],

components/text/dwdiff/Makefile
mtrower: `/usr/sbin/install`  uses very different flags from other Unixes.  It does not work for this package.  install(8) recommends /usr/ucb/install for better compatibility with other Unix `install` tools, and it works fine here.

`/usr/gnu/bin/install` would probably work also, but it is not the default for `install` unless `/usr/gnu/bin` happens to be first in your path.  Does the default build environment have such a configuration?

I favor the ucb tool, as it is present in illumos-gate, but I guess it doesn't really matter which is used here, as long as both are present in a base install.

components/text/dwdiff/Makefile
mtrower: Well, that's what `gmake REQUIRED_PACKAGES` produced when I re-ran it as you suggested.

> Is your system up-to-date?

Well, it was at the time.  It looks like I can now go from `0.5.11-2024.0.0.21992 -> 0.5.11-2024.0.0.21994`, if you feel that would be helpful.  What is strange about depending on `library/icu`, though?

mtrower:
> When I run gmake REQUIRED_PACKAGES and gmake publish here, I'll get the following diffs: ╰─➤ git diff Makefile diff --git a/components/text/dwdiff/Makefile b/components/text/dwdiff/Makefile index cfc372df6f..7de7035b6d 100644 --- a/components/text/dwdiff/Makefile +++ b/components/text/dwdiff/Makefile @@ -46,5 +46,5 @@ COMPONENT_PRE_CONFIGURE_ACTION= ( (CLONEY)(SOURCE_DIR) $(@d) ) REQUIRED_PACKAGES+= text/gnu-gettext
> # Auto-generated dependencies
> 
> -REQUIRED_PACKAGES += library/icu +REQUIRED_PACKAGES += $(ICU_LIBRARY_PKG) REQUIRED_PACKAGES += system/library ╭─andreas@saturn ~/oi-userland/components/text/dwdiff ‹mtrower-dwdiff*› ╰─➤ git diff pkg5 diff --git a/components/text/dwdiff/pkg5 b/components/text/dwdiff/pkg5 index c7861361a4..402e10f582 100644 --- a/components/text/dwdiff/pkg5 +++ b/components/text/dwdiff/pkg5 @@ -1,6 +1,6 @@ { "dependencies": [
> 
>     * ```
>          "library/icu",
>       ```
> 
> 
>     * ```
>          "library/icu-74",
>          "system/library",
>          "text/gnu-gettext"
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       ],

Yes, I had that result initially as well.  I forget what changed to resolve it; something between updating my build zone and fast-forwarding the oi-userland repo, I believe.

Depending on `library/icu-74` seems wrong, semantically, as it is overly specific.  Unless there are there breaking changes between `icu` versions, that require the build version and runtime version to be the same?

components/text/dwdiff/Makefile
mtelka: > `/usr/gnu/bin/install` would probably work also, but it is not the default for `install` unless `/usr/gnu/bin` happens to be first in your path. Does the default build environment have such a configuration?

```
PATH = $(PATH.gnu)
```

components/text/dwdiff/Makefile
mtelka: Currently it is not possible to get dependency on `library/icu` using `gmake REQUIRED_PACKAGES` unless your system is outdated, or misconfigured, or there is some bug.  Either of that needs to be fixed.  It is not possible to get dependency on `library/icu` for two weeks now.

mtelka:
> Yes, I had that result initially as well. I forget what changed to resolve it; something between updating my build zone and fast-forwarding the oi-userland repo, I believe.

In both your changesets (published in this PR) you have `library/icu` in `pkg5`.  That shows that your system is outdated (most likely), or misconfigured, or there is some bug somewhere.

> Depending on `library/icu-74` seems wrong, semantically, as it is overly specific. Unless there are there breaking changes between `icu` versions, that require the build version and runtime version to be the same?

Maybe it seems wrong for you, but it is correct.

mtrower:
> > Yes, I had that result initially as well. I forget what changed to resolve it; something between updating my build zone and fast-forwarding the oi-userland repo, I believe.
> 
> In both your changesets (published in this PR) you have `library/icu` in `pkg5`. That shows that your system is outdated (most likely), or misconfigured, or there is some bug somewhere.

Indeed, while I did encounter this behaviour, I did not submit any result with `library/icu-74`.  See below for probable cause.

> > Depending on `library/icu-74` seems wrong, semantically, as it is overly specific. Unless there are there breaking changes between `icu` versions, that require the build version and runtime version to be the same?
> 
> Maybe it seems wrong for you, but it is correct.

Pardon, but this is no explanation.  You may assert that it is correct, but without an explanation addressing my point I will not understand the reason, and that is detrimental to my ability to make correct decisions on such matters.

Honestly, I don't understand at all the situation with `*/icu`, and that's probably why I'm not on the same page as you.  Consider the following:

```
mtrower@oi-build:/export/home/mtrower$ pkg search pkg.fmri:icu*
INDEX      ACTION VALUE                             PACKAGE
pkg.fmri   set    openindiana.org/developer/g++/icu pkg:/developer/g++/icu@4.6-2014.1.3.1
pkg.fmri   set    openindiana.org/developer/icu     pkg:/developer/icu@72.1-2023.0.0.2
pkg.fmri   set    openindiana.org/developer/icu     pkg:/developer/icu@72.1-2024.0.0.3
pkg.fmri   set    openindiana.org/developer/icu     pkg:/developer/icu@74.2-2024.0.0.0
pkg.fmri   set    openindiana.org/library/icu       pkg:/library/icu@72.1-2024.0.0.4
pkg.fmri   set    openindiana.org/library/icu-74    pkg:/library/icu-74@74.2-2024.0.0.0
pkg.fmri   set    openindiana.org/library/icu-74/32 pkg:/library/icu-74/32@74.2-2024.0.0.0
```

it appears that `library/icu` was split off into a new package for version 74.2, however `developer/icu`  remained the same package and only updated the version.  Is this disparity intentional?  It is confusing to me, and seems like it might have some ramifications regarding having the correct thing installed.

Furthermore, `pkg:/developer/icu@72.1-2024.0.0.3` exists, and is newer than `pkg:/developer/icu@74.2-2024.0.0.0`?


Well, in the end, here is what my build zone has installed:

```
mtrower@oi-build:/export/home/mtrower$ pkg list |grep icu
developer/icu                                     72.1-2023.0.0.2            i--
library/icu                                       72.1-2024.0.0.4            i--
library/icu-74                                    74.2-2024.0.0.0            i--
library/icu-74/32                                 74.2-2024.0.0.0            i--
```

...which is probably why `gmake REQUIRED_PACKAGES` wants `library/icu` rather than `library/icu-74`.  `pkg update -nv` didn't seem to want to update it previously, but it does seem to have an update now, so I'll try that and see where we end up.

Aside from that... I cannot find the jenkins report for the failed build on the initial PR commit.  But the failure was due to attempting to install `icu-74`  to meet the `icu` dependency.  Hopefully this does not reappear if I update the dependency to directly specify `icu-74`.

mtrower:
Okay, so, now I have this:
```
mtrower@oi-build:/usr/src/oi-userland/git/components/text/dwdiff$ pkg list |grep icu
developer/icu                                     74.2-2024.0.0.0            i--
library/icu                                       72.1-2024.0.0.4            i--
library/icu-74                                    74.2-2024.0.0.0            i--
library/icu-74/32                                 74.2-2024.0.0.0            i--
```


and after `gmake REQUIRED_PACKAGES publish` I have this diff:
_EDIT: Removed; you can click compare to view it easier_

Is this what you are hoping to see?  I'll push it out and see if it builds on jenkins.

mtrower:
And, now we are back at the original error, where jenkins doesn't have `library/icu-74` installed, and goes down the rabbit hole trying and failing to install it.

https://jenkins.openindiana.aurora-opencloud.org/blue/rest/organizations/jenkins/pipelines/OpenIndiana%20Organisation/pipelines/oi-userland/branches/PR-15866/runs/3/nodes/20/steps/21/log/?start=0

mtelka:
> > > Depending on `library/icu-74` seems wrong, semantically, as it is overly specific. Unless there are there breaking changes between `icu` versions, that require the build version and runtime version to be the same?
> > 
> > 
> > Maybe it seems wrong for you, but it is correct.
> 
> Pardon, but this is no explanation. You may assert that it is correct, but without an explanation addressing my point I will not understand the reason, and that is detrimental to my ability to make correct decisions on such matters.

:-)  I just corrected you that the dependency on `library/icu-74` is (seems) incorrect.  That was not meant as an explanation because you didn't ask for one :-).  I considered your question as an rhetorical question only.  More below.

> Honestly, I don't understand at all the situation with `*/icu`, and that's probably why I'm not on the same page as you. Consider the following:
> 
> ```
> mtrower@oi-build:/export/home/mtrower$ pkg search pkg.fmri:icu*
> INDEX      ACTION VALUE                             PACKAGE
> pkg.fmri   set    openindiana.org/developer/g++/icu pkg:/developer/g++/icu@4.6-2014.1.3.1
> pkg.fmri   set    openindiana.org/developer/icu     pkg:/developer/icu@72.1-2023.0.0.2
> pkg.fmri   set    openindiana.org/developer/icu     pkg:/developer/icu@72.1-2024.0.0.3
> pkg.fmri   set    openindiana.org/developer/icu     pkg:/developer/icu@74.2-2024.0.0.0
> pkg.fmri   set    openindiana.org/library/icu       pkg:/library/icu@72.1-2024.0.0.4
> pkg.fmri   set    openindiana.org/library/icu-74    pkg:/library/icu-74@74.2-2024.0.0.0
> pkg.fmri   set    openindiana.org/library/icu-74/32 pkg:/library/icu-74/32@74.2-2024.0.0.0
> ```
> 
> it appears that `library/icu` was split off into a new package for version 74.2, however `developer/icu` remained the same package and only updated the version. Is this disparity intentional?

Yes.

For simplicity, when you see package named `library/icu` please read it as `library/icu-72`.

> It is confusing to me, and seems like it might have some ramifications regarding having the correct thing installed.

You cannot have incorrect things installed, if you are up-to-date :-).  In addition, if you have `userland-incorporation` installed then you cannot have incorrect things installed, even if you are not up-to-date. :-)

So if you see both `library/icu` and `library/icu-74` installed on your machine, then it is correct.

> Furthermore, `pkg:/developer/icu@72.1-2024.0.0.3` exists, and is newer than `pkg:/developer/icu@74.2-2024.0.0.0`?

Why do you think `pkg:/developer/icu@72.1-2024.0.0.3` is newer than `pkg:/developer/icu@74.2-2024.0.0.0`?
```
$ pkg list -afv developer/icu
FMRI                                                                         IFO
pkg://openindiana.org/developer/icu@74.2-2024.0.0.0:20240117T230804Z         i--
pkg://openindiana.org/developer/icu@72.1-2024.0.0.3:20240111T172224Z         ---
pkg://openindiana.org/developer/icu@72.1-2023.0.0.2:20231029T132109Z         ---
pkg://openindiana.org/developer/icu@72.1-2023.0.0.1:20231010T124105Z         ---
$
```

> 
> Well, in the end, here is what my build zone has installed:
> 
> ```
> mtrower@oi-build:/export/home/mtrower$ pkg list |grep icu
> developer/icu                                     72.1-2023.0.0.2            i--
> library/icu                                       72.1-2024.0.0.4            i--
> library/icu-74                                    74.2-2024.0.0.0            i--
> library/icu-74/32                                 74.2-2024.0.0.0            i--
> ```

That's not correct.  Your system is outdated.  `developer/icu@74.2` is newer than your `developer/icu@72.1`.  And it looks like you do not have `userland-incorporation` installed because with the incorporation it is not possible to have both `library/icu@72.1-2024.0.0.4` and `developer/icu@72.1-2023.0.0.2` installed together.

> 
> ...which is probably why `gmake REQUIRED_PACKAGES` wants `library/icu` rather than `library/icu-74`. `pkg update -nv` didn't seem to want to update it previously, but it does seem to have an update now, so I'll try that and see where we end up.

`developer/icu@72.1-2023.0.0.2` is outdated for three weeks now.

> Aside from that... I cannot find the jenkins report for the failed build on the initial PR commit. But the failure was due to attempting to install `icu-74` to meet the `icu` dependency. Hopefully this does not reappear if I update the dependency to directly specify `icu-74`.

Should you have more questions, feel free to ask.

mtelka:
> Is this what you are hoping to see? I'll push it out and see if it builds on jenkins.

Not hoping.  Expecting.  :-)

mtelka:
> And, now we are back at the original error, where jenkins doesn't have `library/icu-74` installed, and goes down the rabbit hole trying and failing to install it.
> 
> https://jenkins.openindiana.aurora-opencloud.org/blue/rest/organizations/jenkins/pipelines/OpenIndiana%20Organisation/pipelines/oi-userland/branches/PR-15866/runs/3/nodes/20/steps/21/log/?start=0

Do not worry about jenkins.  The machine you are referring to is apparently not configured properly.

components/text/dwdiff/Makefile
mtelka: Please try `PATH = $(PATH.gnu)` instead of `INSTALL=/usr/ucb/install`.  If it works, then please use it.  We prefer to set `PATH` instead of pointing individual tools.  Thank you.

mtelka (approved):

mtrower:
> Why do you think pkg:/developer/icu@72.1-2024.0.0.3 is newer than pkg:/developer/icu@74.2-2024.0.0.0?

Because the version substring (2024.0.0.3 vs 2024.0.0.0) is a greater version.  Is that substring not a Hipster build version string?  If so, then it would seem that `developer/icu@72.1-2024.0.0.3` was built/published at a later date in time than `developer/icu@74.2-2024.0.0.0`.  If not, then how is that substring arrived at?  Is it just a year followed by a package-specific incremental build number?

> And it looks like you do not have userland-incorporation installed because with the incorporation it is not possible to have both library/icu@72.1-2024.0.0.4 and developer/icu@72.1-2023.0.0.2 installed together.

Well, you would be correct about that.  The question is, why?

In the global zone:
```
mtrower@saturn:/etc$ pkg list userland-incorporation
consolidation/userland/userland-incorporation     0.5.11-2024.0.0.22832      i--
```

In the build zone:
```
mtrower@oi-build:/usr/src/oi-userland/git/components/text/dwdiff$ pkg list userland-incorporation
 
  userland-incorporation
```

Is userland-incorporation not part of a default zone-install?

components/text/dwdiff/Makefile
mtrower: > PATH = $(PATH.gnu)

In fact that does not work (the advice is fine given the information I provided, but upon investigation there are more specifics).

If you want the details, the configure script does the following:

1. use `$INSTALL`, if set
2. test `install`, use that if it passes
3. use `./install.sh` if `install` does not pass
* On SunOS specifically, testing `install` is bypassed (because it is known not to work and (quote) "because it is dangerous", and `./install.sh` is used instead.  (Therefore, adjusting `PATH` becomes irrelevant.)

Okay, so with the above logic, `./install.sh` will always be used unless I set `$INSTALL`.  But `./install.sh` is broken for `make install` due to the relative path used, and the fact that the project has sub-directories.  So the options are:

- Patch `./configure`, and possibly other files, which may or may not prove trivial, and may break if these files are ever updated upstream.
- Possibly, we could symlink `./install.sh` into any sub-directories.  This would probably work; it may break with upstream upgrades.
- Submit a bug report upstream and hope it gets fixed (not something I want to get involved in personally).
- Set `$INSTALL` to a compatible install program.

Setting `INSTALL=/usr/ucb/install` seems clean, simple and reasonable to me.  If for some reason you really don't like the ucb install tool (which is provided by illumos-gate for exactly this purpose), I could instead set `INSTALL=/usr/gnu/bin/install` or `INSTALL=ginstall`.

I realize the above may seem confusing; I'm happy to provide more info or clarification on points.

mtrower:
> Do not worry about jenkins. The machine you are referring to is apparently not configured properly.

Okay, I see you approved changes, but the PR is still open.  So where are we at on this, then?  Are we just working to resolve the `INSTALL` situation (see my latest comment if so)?  Are we waiting on the jenkins machine to be fixed up?  Are we waiting for additional sign-offs?  Or is there more work required from me?

Forgive me if the above sounds rude, I am just unfamiliar with the process here and am trying to understand if I need to do more work, or just wait for now.  Better wording is probably possible, but escapes me at the moment.

AndWac:
Look in make-tools/ips-buildinfo.mk for BRANCH_ID.

AndWac:
The PR checker is not reliable. Whenever it fails I have to manually check.

mtrower:
> Look in make-tools/ips-buildinfo.mk for BRANCH_ID.

Thanks, seems my secondary postulation was correct.

> The PR checker is not reliable. Whenever it fails I have to manually check.

Ah okay.  Thank you both!

mtelka:
> > Why do you think pkg:/developer/icu@72.1-2024.0.0.3 is newer than pkg:/developer/icu@74.2-2024.0.0.0?
> 
> Because the version substring (2024.0.0.3 vs 2024.0.0.0) is a greater version. Is that substring not a Hipster build version string? If so, then it would seem that `developer/icu@72.1-2024.0.0.3` was built/published at a later date in time than `developer/icu@74.2-2024.0.0.0`. If not, then how is that substring arrived at? Is it just a year followed by a package-specific incremental build number?

`man pkg.7` is your friend.

PR #15883, opened 2024-02-01
mosh: Current protobuf and abseil require C++17

Bill-Sommerfeld:


components/shell/mosh/patches/cxx17.patch
```diff
@@ -9,7 +9,7 @@ diff -Nru mosh-1.4.0.orig/configure.ac mosh-1.4.0/configure.ac
 +# Protobuf 3.6+ requires C++14.
```
mtelka: I assume this line should be updated too.

Bill-Sommerfeld: Done, but I seem to have confused Jenkins (the revised patch applies cleanly for me but not for Jenkins).

PR #15889, opened 2024-02-02
fail2ban: add 1.0.2

iigs:


components/network/fail2ban/Makefile
```diff
@@ -0,0 +1,71 @@
...
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# This file was automatically generated using the following command:
+#   $WS_TOOLS/python-integrate-project -d multimedia/yt-dlp yt-dlp
```
mtelka, line 14: This file was **not** generated using the above command.

@iigs, it is not a problem to copy a `Makefile` generated using `python-integrate-project`, tweak it, and use it for another component.  But if you do that, then please remove this comment because it is confusing and simply not true.  Even better would be if you really integrate using the `python-integrate-project`.  Also, this file is missing the Copyright notice (I do not mind much, but somebody could).

Thank you.

iigs: Yes, I started with a previous Makefile when I worked on this one.  I did not notice that the header comments were different.  Fixed in a new commit.

fail2ban is not in pypi.org using a web search and running python-integrate-project results in an error.

PR #16000, opened 2024-02-07
firefox: fix CamerasAccessStatus

iigs:
Back port Bug 1877679 to fix 122/122.0.1 crashing problem with ebay's site.


hoewweken:
Thanks a lot, i can confirm it works  :-)

PR #16012, opened 2024-02-09
bhyve-fw-2023-08-01

hoewweken:
Thanks for all the input in my previous (closed) PR.

I integrated a patch from OMNIOS to solve the ASM errors,
and changed otherwise 'only' the source version of the UEFI builds and switched to GCC-13 just to be on the latest release..
This does build and publish, no more checks so far..


hoewweken:
By the way, this build uses python-2.7 because we have no alternative to this yet because of upstream restrictions..

components/openindiana/bhyve-fw/Makefile
```diff
@@ -30,17 +30,19 @@ COMPONENT_ARCHIVE_HASH= \
...
 COMPONENT_ARCHIVE_HASH_1= \
-	sha256:ca26699ec8f8f7e45e01f2d5a23649997d33f0b8437405ca731f6e301d6a71a8
+	sha256:8aaacf4b7c6a875c78492ad74856802abcbce89ee8cf980efbd12c74946b3b79
 COMPONENT_ARCHIVE_URL_1= \
 	https://github.com/illumos/$(COMPONENT_NAME)/releases/download/$(COMPONENT_TAG_1)/$(COMPONENT_ARCHIVE_1)
 
 COMPONENT_PROJECT_URL=	https://github.com/illumos/uefi-edk2
 COMPONENT_LICENSE=	simplified-BSD
 
+GCC_VERSION= 13
```
mtelka, line 44: This is not enough to switch from illumos-gcc to gcc-13.  Below you force `GCCPATH=/opt/gcc/4.4.4` so you effectively do not use gcc-13 I assume.  Or, is it used?

components/openindiana/bhyve-fw/Makefile
```diff
@@ -53,10 +55,11 @@ COMPONENT_BUILD_ENV += GAS=/usr/bin/gas
 COMPONENT_BUILD_ENV += GAR=/usr/bin/gar
 COMPONENT_BUILD_ENV += GLD=/usr/bin/gld
 COMPONENT_BUILD_ENV += GOBJCOPY=/usr/bin/gobjcopy
+COMPONENT_BUILD_ENV += PYTHON=/usr/bin/amd64/python2.7
 
 # Four variants of the bhyve firmware are built;
 # DEBUG and RELEASE, and with and without CSM.
-VARIANTS =	DEBUG DEBUG.csm RELEASE RELEASE.csm
+VARIANTS =	DEBUG RELEASE RELEASE.csm DEBUG.csm
```
mtelka, line 62: Is the order important here?  If yes, then please add a comment explaining it.  If not, then why you change it?

mtelka (commented):
Please rebase and run again all make targets.  Your branch is more than one month old.

Thank you.

components/openindiana/bhyve-fw/Makefile
mtelka: Ah I see.  `GCCPATH` is set to `/opt/gcc/4.4.4` for csm builds only.  For non-csm you use gcc-13.

Then, this line is not needed, because it is default.  Or, is it critical to use exactly gcc-13 for non-csm builds and nothing newer or older?  If basically any gcc could be used, then this line should not be there.

Thank you.

mtelka:
You also need to update `COMPONENT_VERSION` (or `COMPONENT_REVISION`).

hoewweken:
i will create it new..

PR #16020, opened 2024-02-09
bhyve-fw-2023-08-01

hoewweken:
I integrated a patch from OMNIOS to solve the ASM errors,
and changed otherwise 'only' the source version of the UEFI builds.

This does build and publish, no more checks so far..

By the way, this build uses python-2.7 because we have no alternative to this yet because of upstream restrictions..

P.S. i removed the explicit use of GCC-13 as GCC-13 is used anyway.


mtelka (approved):

hoewweken:
Thanks for your help with this Marcel, always appreciated :-)

Toasterson (approved):

PR #16035, opened 2024-02-10
Set libjpeg-turbo8 as default jpeg library

Toasterson:
Rebuild libjpeg commit after broken rebase

### DIFF
--- a/components/library/libjpeg-turbo/Makefile
+++ b/components/library/libjpeg-turbo/Makefile
@@ -18,6 +18,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= libjpeg-turbo
 COMPONENT_VERSION= 3.0.2
+COMPONENT_REVISION= 1
 COMPONENT_SUMMARY= libjpeg -JPEG Turbo libraries
 COMPONENT_PROJECT_URL= https://www.libjpeg-turbo.org/
 COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
--- a/components/library/libjpeg-turbo/libjpeg-turbo.p5m
+++ b/components/library/libjpeg-turbo/libjpeg-turbo.p5m
@@ -29,6 +29,13 @@ depend fmri=image/library/libjpeg8-turbo type=require
 # Dependency on libjpeg6 with removed docs and symlinks
 depend fmri=image/library/libjpeg6@6.0.2-2020.0.1.1 type=optional
 
+# Ship libjpeg-turbo headers in default locations
+link path=usr/include/jconfig.h      target=libjpeg8-turbo/jconfig.h
+link path=usr/include/jerror.h       target=libjpeg8-turbo/jerror.h
+link path=usr/include/jmorecfg.h     target=libjpeg8-turbo/jmorecfg.h
+link path=usr/include/jpeglib.h      target=libjpeg8-turbo/jpeglib.h
+link path=usr/include/turbojpeg.h    target=libjpeg8-turbo/turbojpeg.h
+
 # Only shipped by libjpeg-turbo
 file path=usr/share/doc/libjpeg-turbo/LICENSE.md
 file path=usr/share/doc/libjpeg-turbo/README.ijg
@@ -64,3 +71,19 @@ link path=usr/lib/$(MACH64)/pkgconfig/libjpeg.pc \
 	target=../../libjpeg8-turbo/lib/$(MACH64)/pkgconfig/libjpeg.pc
 link path=usr/lib/$(MACH64)/pkgconfig/libturbojpeg.pc \
 	target=../../libjpeg8-turbo/lib/$(MACH64)/pkgconfig/libturbojpeg.pc
+
+# Add cmake links for cmake detection
+link path=usr/lib/$(MACH64)/cmake/libjpeg-turboConfig.cmake target=../../libjpeg8-turbo/lib/$(MACH64)/cmake/libjpeg-turbo/libjpeg-turboConfig.cmake
+link path=usr/lib/$(MACH64)/cmake/libjpeg-turboConfigVersion.cmake target=../../libjpeg8-turbo/lib/$(MACH64)/cmake/libjpeg-turbo/libjpeg-turboConfigVersion.cmake
+link path=usr/lib/$(MACH64)/cmake/libjpeg-turboTargets-release.cmake target=../../libjpeg8-turbo/lib/$(MACH64)/cmake/libjpeg-turbo/libjpeg-turboTargets-release.cmake
+link path=usr/lib/$(MACH64)/cmake/libjpeg-turboTargets.cmake target=../../libjpeg8-turbo/lib/$(MACH64)/cmake/libjpeg-turbo/libjpeg-turboTargets.cmake
+
+# Add symlinks so libjpeg-turbo can be detected from cmake and non pkgconfig builds
+link path=usr/lib/$(MACH64)/libjpeg.so \
+	target=../libjpeg8-turbo/lib/$(MACH64)/libjpeg.so.8.3.2
+link path=usr/lib/$(MACH64)/libjpeg.so.8 \
+	target=../libjpeg8-turbo/lib/$(MACH64)/libjpeg.so.8.3.2
+link path=usr/lib/$(MACH64)/libturbojpeg.so \
+	target=../libjpeg8-turbo/lib/$(MACH64)/libturbojpeg.so.0.3.0
+link path=usr/lib/$(MACH64)/libturbojpeg.so.0 \
+	target=../libjpeg8-turbo/lib/$(MACH64)/libturbojpeg.so.0.3.0
--- a/components/library/libjpeg6-ijg/Makefile
+++ b/components/library/libjpeg6-ijg/Makefile
@@ -18,7 +18,7 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		libjpeg6-ijg
 COMPONENT_VERSION=	6.0.2
 LIBJPEG_API_VERSION= 6b
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_FMRI= 	image/library/libjpeg6-ijg
 COMPONENT_PROJECT_URL=	http://www.ijg.org/
 COMPONENT_SUMMARY=	libjpeg - Independent JPEG Group library version 6b
--- a/components/library/libjpeg6-ijg/libjpeg6.p5m
+++ b/components/library/libjpeg6-ijg/libjpeg6.p5m
@@ -28,10 +28,6 @@ depend fmri=image/library/libjpeg6-ijg type=require
 depend fmri=image/library/libjpeg-turbo type=require
 
 # Only provide headers and libraries for backward compatibility
-link path=usr/include/jconfig.h      target=$(COMPONENT_NAME)/jconfig.h
-link path=usr/include/jerror.h       target=$(COMPONENT_NAME)/jerror.h
-link path=usr/include/jmorecfg.h     target=$(COMPONENT_NAME)/jmorecfg.h
-link path=usr/include/jpeglib.h      target=$(COMPONENT_NAME)/jpeglib.h
 link path=usr/lib/libjpeg.so         target=$(COMPONENT_NAME)/lib/libjpeg.so.62.0.0
 link path=usr/lib/libjpeg.so.62      target=$(COMPONENT_NAME)/lib/libjpeg.so.62.0.0
 link path=usr/lib/libjpeg.so.62.0.0  target=$(COMPONENT_NAME)/lib/libjpeg.so.62.0.0
--- a/components/library/libjpeg6-ijg/manifests/sample-manifest.p5m
+++ b/components/library/libjpeg6-ijg/manifests/sample-manifest.p5m
@@ -10,10 +10,11 @@
 #
 
 #
-# Copyright 2016 <contributor>
+# Copyright 2024 <contributor>
 #
 
 set name=pkg.fmri value=pkg:/$(COMPONENT_FMRI)@$(IPS_COMPONENT_VERSION),$(BUILD_VERSION)
+set name=pkg.human-version value=$(HUMAN_VERSION)
 set name=pkg.summary value="$(COMPONENT_SUMMARY)"
 set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
 set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
--- a/components/library/libjpeg6-ijg/pkg5
+++ b/components/library/libjpeg6-ijg/pkg5
@@ -1,6 +1,5 @@
 {
     "dependencies": [
-        "SUNWcs",
         "system/library"
     ],
     "fmris": [


mtelka:
Previous PR: #15012

components/library/libjpeg-turbo/Makefile
```diff
@@ -17,7 +17,9 @@ USE_COMMON_TEST_MASTER= no
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= libjpeg-turbo
+COMPONENT_REVISION= 2
 COMPONENT_VERSION= 3.0.2
+COMPONENT_REVISION= 1
```
mtelka, line 21: Duplicate `COMPONENT_REVISION`.

components/library/libjpeg/pkg5
```diff
@@ -0,0 +1,9 @@
+{
+    "dependencies": [
+        "system/library"
+    ],
+    "fmris": [
+        "image/library/libjpeg"
```
mtelka: This package is already provided by `meta-packages/libjpeg` component.

mtelka (changes requested):
Sorry, this change is too intrusive and would break a lot of things in oi-userland.  Please do [this](https://github.com/OpenIndiana/oi-userland/pull/15012#issuecomment-1827195169) instead.  IOW, please simply add new libjpeg9 package without other changes.

Thank you.

components/library/libjpeg/pkg5
Toasterson: Good catch. That is now a duplicate. Removed.

Toasterson:
To address the concerns posted also in the previous PR and here.

We do have libjpeg6 which we need to get out of the standard location, as it will become the resolving target for all cmake libjpeg scripts. Some packages like `desktop/remote-desktop/freerdp` do not want to find libjpeg outside of /usr/lib and cannot be persuaded to look for libjpeg-turbo.

So current the plan is.
1. Update all components in the list below to libjpeg-turbo if possible
2. merge in the upgrade of libjpeg6 to libjpeg9 in the default location 
4. Update components to use libjpeg9
5. retire libjpeg-turbo

Currently our list of components still depending on the old libjpeg6 is:
- desktop/office/abiword
- desktop/remote-desktop/freerdp
- library/graphics/wxwidgets
- library/graphics/wxwidgets-3
- library/perl-5/tk-534
- library/perl-5/tk-536
- library/perl-5/tk-538
- runtime/java/openjdk11
- runtime/java/openjdk17
- runtime/java/openjdk18
- runtime/java/openjdk19
- runtime/java/openjdk20
- runtime/java/openjdk21
- runtime/java/openjdk22
- runtime/swipl

All of them should be upgraded to libjpeg-turbo before Checking this PR. Unfortunately one or two we will have to update with the PR as they cannot look for libjpeg outside of /usr/lib as stated above. If none of the ones listed can be upgraded, We will need to do a Mass merge but We can Organize that with the people that have merge rights.

Toasterson:
I have checked all packages and added those that cannot be compiled against libjpeg-turbo as simple rebuilds to integrate immediately after this PR. I'll coordinate an upgrade pause for people while we merge this batch.

components/library/libjpeg-turbo/libjpeg-turbo.p5m
```diff
@@ -26,9 +27,6 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 # Require package with default JPEG API
 depend fmri=image/library/libjpeg8-turbo type=require
```
mtelka: You do not touch the `Makefile` for this component.  Will it be built?

mtelka:
> To address the concerns posted also in the previous PR and here.

I'm very sorry, but this does not address my concerns.  I still see this PR as too intrusive and I believe it could (and should) be done in smaller steps that would be well justified and understood.

components/library/libjpeg/pkg5
```diff
@@ -1,11 +1,11 @@
 {
     "dependencies": [
-        "SUNWcs",
         "system/library"
     ],
     "fmris": [
+        "image/library/libjpeg",
```
mtelka: Is my understanding correct that `image/library/libjpeg` will contain version `9e` after this change?

The `image/library/libjpeg` package is currently not needed by any other package/component in the oi-userland.  Why you cannot create new `image/library/libjpeg-9` (or similar) package without touching rest of libjpeg packages?

mtelka (changes requested):

components/library/libjpeg/pkg5
Toasterson: The change happened due to libjpeg-turbo being faster at the time. However we ended up bringing more maintenance burden into the system because of the small speed up. There is no need to name the package differently just because of a version change since there is no plan to keep around multiple versions for longer than we need.  Libjpeg is mostly considered finished so I expect little need for updates other than security upgrades. Neither do most linux distros.

As to the main situation, we must have a main libjepg in the default location As mentioned in the openssl case which falls into the same category some build scripts cannot be changed to build against libraries not under /usr/lib in the case of our 32bit to 64bit transition we added ISA subdirectories (i386/amd64) and there we also needed to invent hacks like runpath editing etc. to be able to package some software at all.

If it is unnecessary to put libraries into non-default paths we should not do so. This library is also required to be in the default path by all the packages in #16235 hence the name change to simply libjepg to reflect that.

components/library/libjpeg-turbo/libjpeg-turbo.p5m
Toasterson: Good catch, this got lost during the rebase.

Toasterson:
This is as small as it gets. How would you make this smaller?

components/library/libjpeg/pkg5
alarcher: Just some background information. Back then some binary-only software relied on libjpeg6 being in /usr/lib, while IJG's libjpeg9 was discouraged by most Linux distribution as being ABI incompatible and bringing inquestionable extensions. Libjpeg-turbo was the default version on nearly all the distributions when the change was made, so it was the natural option to adopt the same default implementation. It is still the case in most distributions and ports.

alarcher: To be clear, basically nobody uses this library as default libjpeg library for different reasons, technical and non-technical.

mtelka:
> This is as small as it gets. How would you make this smaller?

Very good question :-).

To propose anything meaningful (except some obvious minor improvements) I'd need to fully understand the current situation and also the desired final (optimal) state.  After both starting and final points are known, we can start to find the proper path from _here_ to _there_.

I'll try to draw my understanding of both the current and the optimal final state.  Please correct me if I miss, misunderstood or misinterpret something.  And please answer my questions below.  Thanks.

### Current state

We currently have these libjpeg packages:
```
image/library/libjpeg                             6.0.2-2020.0.1.0           i--
image/library/libjpeg-turbo                       3.0.2-2024.0.0.0           i--
image/library/libjpeg6                            6.0.2-2020.0.1.1           i--
image/library/libjpeg6-ijg                        6.0.2-2020.0.1.1           i--
image/library/libjpeg6-turbo                      3.0.2-2024.0.0.0           i--
image/library/libjpeg7-turbo                      3.0.2-2024.0.0.0           i--
image/library/libjpeg8-turbo                      3.0.2-2024.0.0.0           i--
```
In addition there is also `image/library/openjpeg`, but I believe it is out of scope.

1. `image/library/libjpeg`

This is built from [components/meta-packages/libjpeg](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/meta-packages/libjpeg).  It is pure meta package and it depends on `image/library/libjpeg6` only.  Nothing in OpenIndiana depends on `image/library/libjpeg`.

2. `image/library/libjpeg6-ijg`

This is built from [components/library/libjpeg6-ijg](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libjpeg6-ijg).  It contains the ancient version `6b` of jpeg.  The package contains `libjpeg.so.62`, four header files, and five binaries.  All of them are at alternative locations (with `libjpeg6-ijg` in their path names).   About 10 other components in oi-userland requires this package.

3. `image/library/libjpeg6`

This is built from [components/library/libjpeg6-ijg](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libjpeg6-ijg) together with `image/library/libjpeg6-ijg`.  It contains symlinks for libraries and header files from standard locations (`/usr/include`, `/usr/lib`) to their non-standard `libjpeg6-ijg` locations as provided by the `image/library/libjpeg6-ijg` package.  The package is needed by ca 10 other components in oi-userland (the same as for the `image/library/libjpeg6-ijg` package).  This package depends on `image/library/libjpeg6-ijg` (this is obvious and understandable) and also on `image/library/libjpeg-turbo` (this is surprising a bit).

4. `image/library/libjpeg6-turbo`

Built from [components/library/libjpeg-turbo](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libjpeg-turbo).  This looks (mostly) like a turbo drop-in replacement for `image/library/libjpeg6-ijg` with some additional development files and binaries.  Everything is located in `libjpeg6-turbo` subdirectories, so it is invisible by default.  It depends on `image/library/libjpeg-turbo`.  Nothing in OpenIndiana depends on `image/library/libjpeg6-turbo`.

5. `image/library/libjpeg7-turbo`

This is similar to `image/library/libjpeg6-turbo` but provides `libjpeg.so.7` instead.  Everything is in `libjpeg7-turbo` subdirs.  It depends on `image/library/libjpeg-turbo` and nothing in OpenIndiana depends on it.

6. `image/library/libjpeg8-turbo`

Similar to both `image/library/libjpeg6-turbo` and `image/library/libjpeg7-turbo`.  It provides `libjpeg.so.8`.  Everything is in `libjpeg8-turbo` subdirs.  It depends on `image/library/libjpeg-turbo` and *many* packages in OpenIndiana depends on it.

7. `image/library/libjpeg-turbo`

It is built from [components/library/libjpeg-turbo](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libjpeg-turbo).  This contains documentation for `libjpeg-turbo`, symlinks for binaries from `/usr/bin` to `libjpeg8-turbo` variants of bins, and symlinks for pkg-config files from their standard locations to `libjpeg8-turbo` variants of these files.  The package depends on `image/library/libjpeg6` (optionally) and `image/library/libjpeg8-turbo` (this is obvious).

### Final state

I think it is obvious that we want to get rid of the old ancient [components/library/libjpeg6-ijg](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libjpeg6-ijg) because it is dead horse for more years than anybody could accept.  It is also obvious that we need _at least_ one of `libjpeg-turbo` and/or `libjpeg 9` projects.

### Questions

- Do we want/need both `libjpeg-turbo` and `libjpeg 9` as the final state?  If not, which one?  Why that one?
- If we want to keep both, which one would be the preferred one?  For example, if we (re)build a package, will we prefer to build it with `libjpeg-turbo` (v8 of the API) or `libjpeg` (v9 of the API)?
- Do we need to keep v6 and v7 of the API once we reach the final state?
- How compatible the API is?  For example, if a software was developed for API v6 only, will it compile and run properly with API v7, v8, v9?
- Is `libjpeg 9` able to provide libraries with older API than v9?
- The `libjpeg-turbo` project claims their v6 libraries are API/ABI compatible with IJG v6 libraries.  Do we have some observations or knowledge in this area?  What will happen if we simply do something like `cp /usr/lib/libjpeg6-turbo/lib/amd64/lib* /usr/lib/amd64/`?  What will stop to work?
- What package names _must_ be preserved in the final state?

### Final thoughts and notes

- The above analysis shows that the state of `libjpeg` in OpenIndiana is not so bad.  I expected far worse state :-).
- I think the reason why some components still builds with old IJG v6 and are refusing to find `libjpeg-turbo` automatically is that they do not use pkg-config.
- I believe we do not need mass rebuild (like #16235) to reach the final state.  The transition could be done incrementally without any shocks or other surprises.

Thank you.

components/library/libjpeg/pkg5
alarcher: @mtelka Hello, not sure if you noticed the comments above. I made the switch to libjpeg-turbo back then. I can answer a few questions that you may have about the state of the userland w.r.t libjpeg.

mtelka: > @mtelka Hello, not sure if you noticed the comments above. I made the switch to libjpeg-turbo back then.

Sure, I noticed.  BTW, the change from IJG 6 to turbo was very good change.

Yes, you are saying that _others_ do not use `libjpeg9` and almost exclusively use `libjpeg-turbo`.  That means that we _probably_ want to go the same way.  I'm just asking to be sure what exactly we want to achieve because it is not rare that SunOS does many things _differently_ than others :-).

BTW, if we want to go the `libjpeg-turbo`-only way (i.e. with no IJG 6 and no IJG 9) then I see this PR as a step in wrong direction (hence my questions, see below).

> I can answer a few questions that you may have about the state of the userland w.r.t libjpeg.

I think the current state of `libjpeg` in oi-userland is clear to me, so I currently have no questions about the _current state_.  Maybe the only exception is my question re the API v6 compatibility between IJG and turbo, but the question is not oi-userland specific.  Please review what I wrote [here](https://github.com/OpenIndiana/oi-userland/pull/16035#issuecomment-1967078529) and if you find any issue, please let me know.

I do have questions about the _future_ and about stuff outside the oi-userland.  I'd appreciate at least some answers to my questions.  Either from you @alarcher, or from @Toasterson, or from @AndWac, or from ... anybody else who have anything to add to the matter :-).  The most important question is the first one:  _Do we want/need both `libjpeg-turbo` and `libjpeg 9` as the final state? If not, which one? Why that one?_

Thank you.

Toasterson: Hi @alarcher thanks for this info. I would need to look at which package exactly started this change, but we started to plan the move because there were some components starting to complain about wanting libjpeg9 and refusing to compile. From the two linux distros I have at hand (Ubuntu and archlinux) both do indeed use libjpeg-turbo as their base. At least Ubuntu manages to have both libjpeg9 and libjpeg-turbo under their equivalent of `/usr/lib/amd64` which sounds to me like the most sane approach for all edge cases. With that and a properly setup pkgconfig we should be ok for whatever comes in the future. libjpeg-turbo stated that they do not want to adopt the jpeg9 API as it was too small a change to warrant the effort and Color management was deemed too exotic apparently. Such statements tend to cause software devs to abandon the library. I can set up a test system to see how complex a PR I would need to place a copy of libjpeg-turbo under `/usr/lib/amd64` but we need to get rid of libjpeg6. We lost Solaris ABI compatibility some time ago so there is no longer a justification to keep libjpeg6 around.

mtelka: > Hi @alarcher thanks for this info. I would need to look at which package exactly started this change, but we started to plan the move because there were some components starting to complain about wanting libjpeg9 and refusing to compile. From the two linux distros I have at hand (Ubuntu and archlinux) both do indeed use libjpeg-turbo as their base. At least Ubuntu manages to have both libjpeg9 and libjpeg-turbo under their equivalent of `/usr/lib/amd64` which sounds to me like the most sane approach for all edge cases. With that and a properly setup pkgconfig we should be ok for whatever comes in the future. libjpeg-turbo stated that they do not want to adopt the jpeg9 API as it was too small a change to warrant the effort and Color management was deemed too exotic apparently. Such statements tend to cause software devs to abandon the library. I can set up a test system to see how complex a PR I would need to place a copy of libjpeg-turbo under `/usr/lib/amd64` but we need to get rid of libjpeg6. We lost Solaris ABI compatibility some time ago so there is no longer a justification to keep libjpeg6 around.

I decipher this as *both* for my question: _Do we want/need both `libjpeg-turbo` and `libjpeg 9` as the final state? If not, which one? Why that one?_

Am I right?

BTW, what components needs `libjpeg9`?  I'm just curious...

mtelka:
> > This is as small as it gets. How would you make this smaller?
> ### Final state
> 
> I think it is obvious that we want to get rid of the old ancient [components/library/libjpeg6-ijg](https://github.com/OpenIndiana/oi-userland/tree/oi/hipster/components/library/libjpeg6-ijg) because it is dead horse for more years than anybody could accept. It is also obvious that we need _at least_ one of `libjpeg-turbo` and/or `libjpeg 9` projects.

While thinking more about this I realized that we can do some reasonable steps to obsolete `libjpeg6-ijg` even without knowing the exact final state. Here is the first step: #16313

components/library/libjpeg/pkg5
alarcher: You may want to add libjpeg9 for component strictly requiring it but the vast majority want the usual 6-8 ABI backward compatible library. So the answer is: having both is good but the default should remain libjpeg-turbo version 8. Another argument I had read was about DRM-related changes in IJG's version. Back then I added all the flavours because I was not sure whether some libaries required strictly libjpeg-turbo 6, 7, or 8. Then later on I moved most components (nearly all of them) to the latest version after checking that the component worked. I had left libjpeg6-ijg in place as some binary distributions from OpenSolaris times expected it in /usr/lib and I preferred to avoid breaking things unnecessarily for some people. Since I was doing this in my spare time I made decisions based on the time available mostly and have been more conservative than required. Years have passed so compatibility is not relevant anymore. Thanks for improving the situation.

mtelka: Thanks for these notes, @alarcher.

For now I'd prefer to stay with `libjpeg-turbo` only and add `libjpeg9` only after there is some software really requiring it.  @Toasterson: Do we already have something in the queue where missing `libjpeg9` would be a showstopper?

Toasterson: It's been sometime that andreas and I looked at it. At the moment I can't find any in my notes. So we can leave libjpeg9 out of the picture for this PR.

Toasterson:
Alright, I changed the approaches as discussed and simply fixed up libjpeg-turbo to be in the default paths. I had to remove the links to the default headers from libjpeg-ijg so both components must be in the same PR, otherwise installation will fail on enduser devices.

mtelka:
> Alright, I changed the approaches as discussed and simply fixed up libjpeg-turbo to be in the default paths. I had to remove the links to the default headers from libjpeg-ijg so both components must be in the same PR, otherwise installation will fail on enduser devices.

Unless you accept the slow(er) but 100% safe way...

mtelka:
BTW, I looked at the latest version of this change and it is far, far, far better than previously when libjpeg9 was still in the picture.  Thank you for that!

But, OTOH, it still contains at least one bug that will cause the newly built packages won't be installable.  This is because you still force the 'do as much as possible in one go' approach.

mtelka:
> > Alright, I changed the approaches as discussed and simply fixed up libjpeg-turbo to be in the default paths. I had to remove the links to the default headers from libjpeg-ijg so both components must be in the same PR, otherwise installation will fail on enduser devices.
> 
> Unless you accept the slow(er) but 100% safe way...

... as I proposed in #16313.

Toasterson:
It only changes development headers. So no all packages will be installable. And more importantly it does not break existing links because existing .so version files still exist. And only new links are added.

mtelka:
If all packages were installable, then why you had to create #16368?

PR #16053, opened 2024-02-11
make-rules/meson.mk: switch to new libexec dir

AndWac:


AndWac:
I don't intend to rebuild everything with this like I did for mate.

PR #16095, opened 2024-02-12
python/python39: Turn on CTF for Python 3.9 and fix the pstack helper

Bill-Sommerfeld:
Two commits that should probably remain separate (I can restructure these as two pull requests if this would help):

 1) Define USE_CTF to enable CTF support in make macros and userland-mangler.

This is a prerequisite for pstack working as it needs to see function argument counts (which it gets from CTF information in the executable or shared library).  

 2) Fix debug helper module in libpython39_db.c to work with Python 3.9

There are a bunch of changes here.    Most notably, the function it was looking for was renamed, and the argument holding the Python frame information was moved from argv[0] to argv[1].   Additionally, the compressed line number information decoder was wrong, and the link to the file describing the compressed encoding was stale.

The illumos-gate side of this change is here: https://code.illumos.org/c/illumos-gate/+/3290

I'm fixing this in 3.9 first so I have a working baseline to compare against when I dive into fixing this for 3.12.

### DIFF
--- a/components/python/python39/Makefile
+++ b/components/python/python39/Makefile
@@ -26,11 +26,13 @@
 #
 
 USE_OPENSSL11= yes
+USE_PARALLEL_BUILD=	yes
+USE_CTF=		yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=	        Python
 COMPONENT_VERSION=      3.9.18
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 PYTHON_VERSION =	$(subst $(space),.,$(wordlist 1,2,$(subst ., ,$(COMPONENT_VERSION))))
 COMPONENT_SUMMARY=      The Python interpreter, libraries and utilities
 COMPONENT_PROJECT_URL=  https://python.org/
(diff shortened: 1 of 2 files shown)


mtelka:
> I can restructure these as two pull requests if this would help

Yes please.

Bill-Sommerfeld:
The USE_CTF infrastructure is spun off to #16098

mtelka:
Please change title of this PR to match the commit.  I'm not sure which one will be used once it is merged so better to align them.

I'll review the actual change later.

Bill-Sommerfeld:
> Please change title of this PR to match the commit. I'm not sure which one will be used once it is merged so better to align them.

Done

mtelka (approved):
LGTM.

mtelka:
@Bill-Sommerfeld in the meantime you can add the fix for https://www.illumos.org/issues/16272 as a patch for `illumos-gate` component here so we can start to benefit from your work immediately.

Thank you.

Bill-Sommerfeld:
Sorry I didn't respond earlier but I was already working on integrating it into illumos-gate.   That has now happened -- the necessary patch is now in illumos-gate as illumos/illumos-gate@616d60f414d125f2dfc25908ca90a7c4451076c6

PR #16098, opened 2024-02-13
shared-macros.mk: define USE_CTF to enable CTF

Bill-Sommerfeld:
CTF is the "Compact Type Format", a compressed debugging metadata format used by dtrace and mdb.  Most of illumos-gate is built with CTF; it would be helpful for observability if a few other packages from userland were also built with CTF.

### DIFF
--- a/make-rules/ips.mk
+++ b/make-rules/ips.mk
@@ -484,6 +484,9 @@ $(BUILD_DIR) $(MANGLED_DIR):
 	$(MKDIR) $@
 
 PKGMANGLE_OPTIONS = -D $(MANGLED_DIR) $(PKG_PROTO_DIRS:%=-d %)
+ifeq ($(strip $(USE_CTF)),yes)
+PKGMANGLE_OPTIONS += -c $(CTFCONVERT)
+endif
 $(MANIFEST_BASE)-%.mangled:	$(MANIFEST_BASE)-%.mogrified $(MANGLED_DIR)
 	$(PKGMANGLE) $(PKGMANGLE_OPTIONS) -m $< >$@
 
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -1097,6 +1097,7 @@ JQ =		/usr/bin/jq
 DOS2UNIX =	/usr/bin/dos2unix
 TAC =		/usr/bin/tac
 QUILT =		/usr/bin/quilt
+CTFCONVERT =	/opt/onbld/bin/$(MACH)/ctfconvert
 
 INS.dir=        $(INSTALL) -d $@
 INS.file=       $(INSTALL) -m 444 $< $(@D)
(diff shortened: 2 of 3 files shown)


make-rules/ips.mk
```diff
@@ -484,6 +484,10 @@ $(BUILD_DIR) $(MANGLED_DIR):
 	$(MKDIR) $@
 
 PKGMANGLE_OPTIONS = -D $(MANGLED_DIR) $(PKG_PROTO_DIRS:%=-d %)
+ifeq ($(strip $(USE_CTF)),yes)
+CTFCONVERT=/opt/onbld/bin/$(MACH)/ctfconvert
```
mtelka: I think this could be usable even without `USE_CTF` so I suggest to move this to `shared-macros.mk` and define it there unconditionally.

make-rules/shared-macros.mk
```diff
@@ -1244,6 +1244,24 @@ CFLAGS +=	$(CC_BITS)
...
+# Enable CTF if desired
+ifeq ($(strip $(USE_CTF)),yes)
+# -gdwarf-4 is the newest DWARF revision understood by illumos libctf
+#
+# -gstrict-dwarf prevents the compiler from including newer DWARF features
+#
+# -msave-args makes it possible for debugging tools (specifically pstack and
+# mdb) to retrieve function arguments, and is required for the python pstack
+# hooks.
+CFLAGS += -g -gdwarf-4 -gstrict-dwarf -msave-args
```
mtelka: Is `-msave-args` supported by `clang`?

make-rules/shared-macros.mk
```diff
@@ -1244,6 +1244,24 @@ CFLAGS +=	$(CC_BITS)
...
+# By default, GCC may put parts of functions in different named sub-sections
+# of .text based on their presumed or measured calling frequency.  This is not
+# currently handled by libctf and other illumos tools which assume that a
+# function's text lies in a single contiguous range of virtual addresses.
+# Disable this optimization if building with CTF
+CFLAGS += -fno-reorder-functions -fno-reorder-blocks-and-partition
```
mtelka: Are these supported by `clang`?

Bill-Sommerfeld: Not listed in their documentation (looked at https://clang.llvm.org/docs/ClangCommandLineReference.html)

clang appears to have a roughly equivalent -f{no-,}split-machine-functions, documented as 
"Enable late function splitting using profile information (x86 ELF)".

make-rules/shared-macros.mk
Bill-Sommerfeld: -msave-args is  an illumos special for GCC.   Not sure if it's available for clang (I found an email thread about it at https://lists.llvm.org/pipermail/llvm-dev/2017-April/112431.html  - unclear if that has gone anywhere.

what's the right way to condition makefiles on gcc vs clang?

mtelka: The compiler specific `CFLAGS` should go to `CFLAGS.$(COMPILER)`, so `CFLAGS.gcc` in this case.  Currently we do not have native support for other compilers (like `clang`) using this infrastructure, so `CFLAGS` and `CFLAGS.gcc` should always match.  Such facility was primarily used for `studio` (as alternative of `gcc`) but support for that is gone.

I assume that `USE_CTF` won't work with non-gcc compiler, so maybe you should also make sure that `USE_CTF` cannot be set for different `COMPILER` than `gcc`.

make-rules/ips.mk
Bill-Sommerfeld: Done.

make-rules/shared-macros.mk
Bill-Sommerfeld: > I assume that USE_CTF won't work with non-gcc compiler, so maybe you should also make sure that USE_CTF cannot be set for different COMPILER than gcc.

It's not that it's impossible to make it work - it's just that you likely need a different set of options for best results.
I've added a check to ensure that you can't enable CTF with a different compiler without thinking about it..

make-rules/shared-macros.mk
```diff
@@ -1244,6 +1245,31 @@ CFLAGS +=	$(CC_BITS)
...
+CTF_CFLAGS.gcc = -g -gdwarf-4 -gstrict-dwarf
+# -msave-args makes it possible for debugging tools (specifically pstack and
+# mdb) to retrieve function arguments, and is required for the python pstack
+# hooks.
+CTF_CFLAGS.gcc += -msave-args
+# By default, GCC may put parts of functions in different named sub-sections
+# of .text based on their presumed or measured calling frequency.  This is not
+# currently handled by libctf and other illumos tools which assume that a
+# function's text lies in a single contiguous range of virtual addresses.
+# Disable this optimization if building with CTF
```
mtelka: Period missing :-).

mtelka (approved):

make-rules/shared-macros.mk
Bill-Sommerfeld: Fixed.

mtelka (approved):

PR #16113, opened 2024-02-13
perl/Tk: fix libX11 loading; add test results

mtelka:


components/perl/Tk/patches/01-load-64-bit-X11.patch
```diff
@@ -0,0 +1,16 @@
...
+
+--- Tk-804.036/Tk.pm.orig
++++ Tk-804.036/Tk.pm
+@@ -90,8 +90,8 @@
+ my $boot_time = timeofday();
+ 
+ # This is a workround for Solaris X11 locale handling
+-Preload(DynaLoader::dl_findfile('-L/usr/openwin/lib','-lX11'))
+-  if (NeedPreload() && -d '/usr/openwin/lib');
++Preload(DynaLoader::dl_findfile('-L/usr/openwin/lib/64','-lX11'))
```
Bill-Sommerfeld, line 12: What's questionable here is that `/usr/openwin` is a symlink to `/usr/X11`, and `/usr/X11` is mostly symlinks pointing at files in the usual subdirectories of `/usr`, all left in the filesystem for reasons of backwards compatibility across two migrations of names.

If the code needs to load `-lX11`, I'd think it should get it from the default library path..

```
$ realpath /usr/openwin/lib/64/libX11.so
/usr/lib/amd64/libX11.so.4.0.0
```

PR #16126, opened 2024-02-14
libreoffice: update to 24.2.1.1

iigs:


components/desktop/libreoffice/Makefile
```diff
@@ -48,7 +48,7 @@ COMPONENT_FULL_VERSION=$(COMPONENT_FULL_VERSION)-buildfix$(BUILD_FIX)
 endif
 # repology wants to use HUMAN_VERSION
 HUMAN_VERSION=          $(COMPONENT_FULL_VERSION)
-COMPONENT_REVISION=		2
+COMPONENT_REVISION=     3
 COMPONENT_SUMMARY=      LibreOffice is a powerful office suite
```
AndWac, line 51: You don't need to increment COMPONENT_REVISION if you change COMPONENT_VERSION. Instead of this you should set COMPONENT_REVISION to 0 or delete it.

iigs:
Thank you for catching that the COMPONENT_REVISION should be removed

PR #16136, opened 2024-02-15
sbcl: fix build dependencies

mtelka:
Please note that `COMPONENT_REVISION` is not incremented here because build of this component is disabled and we do not need its rebuild.


mtelka:
Please also note that this adds circular build dependency on `runtime/sbcl`, but it is best we can do now, I think.

PR #16187, opened 2024-02-18
runtime/erlang: update to 26.2.2

RAC-oi:
voila


AndWac:
When I build this here, I'll get the following difference in the sample-manifest:
╰─➤  git diff manifests/sample-manifest.p5m
diff --git a/components/runtime/erlang/manifests/sample-manifest.p5m b/components/runtime/erlang/manifests/sample-manifest.p5m
index 6c67a423c1..8fe602d648 100644
--- a/components/runtime/erlang/manifests/sample-manifest.p5m
+++ b/components/runtime/erlang/manifests/sample-manifest.p5m
@@ -3215,6 +3215,11 @@ file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/examples/sudoku/sudoku_gui.erl
 file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/include/gl.hrl
 file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/include/glu.hrl
 file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/include/wx.hrl
+file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/priv/erl_gl.so
+file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/priv/erlang-logo128.png
+file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/priv/erlang-logo32.png
+file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/priv/erlang-logo64.png
+file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/priv/wxe_driver.so
 file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/src/gen/gl.erl
 file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/src/gen/glu.erl
 file path=usr/lib/$(MACH64)/erlang/lib/wx-2.4/src/gen/wxAcceleratorEntry.erl

When I add the 4 lines to the manifest and run gmake REQUIRED_PACKAGES I'll get the following differences to the Makefile:╰─➤  git diff Makefile
diff --git a/components/runtime/erlang/Makefile b/components/runtime/erlang/Makefile
index 687465882b..d76cfc7791 100644
--- a/components/runtime/erlang/Makefile
+++ b/components/runtime/erlang/Makefile
@@ -125,9 +125,12 @@ test:     $(TEST_64)
 REQUIRED_PACKAGES += SUNWcs
+REQUIRED_PACKAGES += library/graphics/wxwidgets-3
 REQUIRED_PACKAGES += library/security/openssl-31
+REQUIRED_PACKAGES += x11/library/glu

RAC-oi:
yes, then you have wxwidgets installed. unfortunately recent erlang versions don't have a configure option to disable that anymore.
i avoided that on purpose because i think it's hard to justify that a scripting language requires these things. at least not by default. and less dependencies also keep the repo processing leaner.

however, if you want this to be waterproof, i can create an additional patch to remove the wxwidgets related parts.

AndWac:
Maybe it would be best to just have the additional lines in the sample-manifest and comment them out in the manifest with an explanation.

RAC-oi:
that would be risky because if it's built with wxwidgets but wxwidgets itself and the required dependencies are not installed, it might crash if somebody tries to use the wxwidgets funtionality.
if the "official" buildsystem does not have the option to exclude certain packages and/or files, i'll make a patch. that way we're on the safe side.

components/runtime/erlang/Makefile
```diff
@@ -23,71 +23,57 @@
...
-COMPONENT_LICENSE=        Apache-2.0
-COMPONENT_LICENSE_FILE=   LICENSE.txt
+COMPONENT_LICENSE= Apache-2.0
+COMPONENT_LICENSE_FILE= LICENSE.txt
 COMPONENT_SUMMARY= Erlang/OTP
 COMPONENT_DESCRIPTION= Erlang programming language and OTP libraries
 
 # manpages
+# no 26.2 available as of 16.02.2024. hence using 26.1.
 COMPONENT_NAME_1= otp_doc_man
```
AndWac, line 51: https://github.com/erlang/otp/releases/tag/OTP-26.2.2 seem to offer it.

RAC-oi:
indeed. strange that they didn't have it on their own server. anyway i'll use that ...

AndWac:
You seem to have tried to drop the documentation package.
In such situations it's not enough to simply delete the old manifest file but you also need to add an entry for it in either a local history file (if possible) or in the global history file.
Thus, in this case add a file named history in the package folder with a single line:
runtime/erlang/documentation@24.2.2,5.11-2023.0.0.1

(That is the latest FMRI of it with an increased COMPONENT_REVISION). If you don't do so, the old package will still be available, eg.╰─➤  pkg list -a|grep erlang
SUNWerlang                                        12.1.1-0.133               --r
SUNWerlang-doc                                    12.1.1-0.133               --r
runtime/erlang                                    26.2.2-2024.0.0.0          i--
runtime/erlang/documentation                      24.2.2-2023.0.0.0          ---

RAC-oi:
oh i see, added.

AndWac:
Please squash the commits

components/runtime/erlang/Makefile
```diff
@@ -23,71 +23,55 @@
...
 COMPONENT_ARCHIVE_HASH= \
-	sha256:b6adfc0bf14d94348146ae26cc38d09dca545f8e14ebab7ddcf9482a6e8d1162
-COMPONENT_ARCHIVE_URL=  https://github.com/erlang/otp/archive/${COMPONENT_ARCHIVE}
-COMPONENT_POST_UNPACK_ACTION= ( $(MV) otp-OTP-$(COMPONENT_VERSION) $(COMPONENT_SRC))
-COMPONENT_PREP_ACTION = (cd $(@D); ./otp_build update_configure --no-commit)
-COMPONENT_VERSION_SHORT= $(shell echo $(COMPONENT_VERSION) | cut -d'.' -f1-2)
+	sha256:d537ff4ac5d8c1cb507aedaf7198fc1f155ea8aa65a8d83edb35c2802763cc28
+COMPONENT_ARCHIVE_URL= https://github.com/erlang/otp/releases/download/OTP-${COMPONENT_VERSION}/${COMPONENT_ARCHIVE}
+COMPONENT_POST_UNPACK_ACTION= ( $(RM) -r $(COMPONENT_SRC)/lib/wx )
+COMPONENT_VERSION_SHORT= 26.2
```
mtelka: What was wrong with the previous calculation of `COMPONENT_VERSION_SHORT` (except it is slower than it needs to be, but that's different story)?

It is dangerous to hardcode version numbers at more than single place because people tend to forget to update other occurrences when the `COMPONENT_VERSION` gets updated.  If the `COMPONENT_VERSION_SHORT` value is strictly bound to `COMPONENT_VERSION` (and it looks so) then please use some sort of automatic calculation of its value.

Thank you.

components/runtime/erlang/Makefile
```diff
@@ -23,71 +23,55 @@
...
+COMPONENT_LICENSE= Apache-2.0
+COMPONENT_LICENSE_FILE= LICENSE.txt
 COMPONENT_SUMMARY= Erlang/OTP
 COMPONENT_DESCRIPTION= Erlang programming language and OTP libraries
 
 # manpages
 COMPONENT_NAME_1= otp_doc_man
-COMPONENT_SRC_1=  $(COMPONENT_NAME_1)_$(COMPONENT_VERSION_SHORT)
-COMPONENT_ARCHIVE_1=  $(COMPONENT_SRC_1).tar.gz
+COMPONENT_SRC_1= $(COMPONENT_NAME_1)_26.2.2
```
mtelka: Same here.  Please use `COMPONENT_VERSION` instead of `26.2.2`.  Or something else appropriate.

components/runtime/erlang/Makefile
```diff
@@ -23,71 +23,55 @@
...
-COMPONENT_NAME_2= otp_doc_html
-COMPONENT_SRC_2=  $(COMPONENT_NAME_2)_$(COMPONENT_VERSION_SHORT)
-COMPONENT_ARCHIVE_2=  $(COMPONENT_SRC_2).tar.gz
-COMPONENT_ARCHIVE_HASH_2= \
-    sha256:f479cbc8a28532fd6a0a55fc26684b4e79312da4f86ee0735d0757f936672bbc
-COMPONENT_ARCHIVE_URL_2=  $(COMPONENT_PROJECT_URL)download/$(COMPONENT_ARCHIVE_2)
-UNPACK_ARGS_2=  -r $(COMPONENT_SRC_2)
-PKG_PROTO_DIRS += $(COMPONENT_SRC_2)
+    sha256:16c1c6a949d619bf63b5c2fe9872dc043322e1130416595f22f0bc4c70459f47
+COMPONENT_ARCHIVE_URL_1= https://github.com/erlang/otp/releases/download/OTP-26.2.2/$(COMPONENT_ARCHIVE_1)
```
mtelka: And here too:  `26.2.2` -> `COMPONENT_VERSION`.

components/runtime/erlang/Makefile
```diff
@@ -148,17 +113,14 @@ COMPONENT_POST_INSTALL_ACTION += \
 	$(PERL) -pi -e "s|$(PROTO_DIR)||g" \
 		$(PROTOUSRLIBDIR64)/erlang/bin/erl \
 		$(PROTOUSRLIBDIR64)/erlang/releases/RELEASES
+COMPONENT_POST_INSTALL_ACTION += ;( $(MKDIR) -p $(PROTO_DIR)/usr/share )
```
mtelka: We use the opposite format, i.e. `command ;` instead of `; command`.  Please change this to avoid confusion.  Thank you.

components/runtime/erlang/Makefile
```diff
@@ -148,17 +113,14 @@ COMPONENT_POST_INSTALL_ACTION += \
...
+COMPONENT_POST_INSTALL_ACTION += ;( $(MV) $(COMPONENT_SRC_1)/man $(PROTO_DIR)/usr/share/ )
 
 # common targets
 build:    $(BUILD_64)
```
mtelka, line 122: These common targets should be already provided by the `common.mk` so they should be not needed and thus removed.

components/runtime/erlang/Makefile
```diff
@@ -23,71 +23,55 @@
...
 include $(WS_MAKE_RULES)/common.mk
 
 SYSTEM_TEST_TARGET= configure $(SYSTEM_TEST_64)
 
-CFLAGS += -DOPENSSL_NO_MD4 -DOPENSSL_NO_RC2 -DOPENSSL_NO_RC4
-
 # Fix 64-bit linking.
 LDFLAGS += $(CC_BITS)
+CFLAGS += -finline-functions -funroll-loops
+CXXFLAGS += -finline-functions -funroll-loops -fno-exceptions -fno-rtti
```
mtelka, line 68: After digging a bit into the `no-rtti` stuff it looks like that its use could be dangerous and cause surprises.  [Here](https://stackoverflow.com/questions/4486609/when-can-compiling-c-without-rtti-cause-problems) is a nice example.

My preliminary conclusion is that for general purpose environments like OpenIndiana where we prefer stability over the memory footprint (how much we would save with `-fno-rtti`?) the usage of `-fno-rtti` should be both well justified and limited to parts of the system where it really cannot cause any problems.

I'm sorry, but -1 for `-fno-rtti` from me here.

For `-funroll-loops` the `gcc` man page says:
```
This option makes code larger, and may or may not make it run faster.
```
Do you have any numbers justifying this option here?

Re `-finline-functions`: since we build with `-O2` by default (and I believe it is same for erlang too) this option should not be needed to be set explicitly.

Re `-fno-exceptions`: this is to disable some compile warnings.  Why do we need that?

mtelka (changes requested):

components/runtime/erlang/Makefile
RAC-oi: no, they're not. that's why i added this.

components/runtime/erlang/Makefile
RAC-oi: i'm sorry but i don't have the time to discuss common compiler options in depth. these have proven themselves for almost 2 decades and among other things are the default for games and things like the mozilla browsers for good reason.

components/runtime/erlang/Makefile
mtelka: > no, they're not.

How so?
```
$ cd components/runtime/erlang/
$ gmake print-value-COMMON_TARGETS 2>/dev/null
yes
$ gmake print-value-BUILD_TARGET 2>/dev/null
...REDACTED.../components/runtime/erlang/build/amd64/.built
$
```

> that's why i added this.

Could you please be more specific about _this_?  If I look properly at your change in this area of the `Makefile` I just see you removed empty lines between `build:`, `install:` and `test:` lines and I see nothing added.  What _exactly_ you added and why?

mtelka:
I tried to run `env - gmake install` here and it failed with:
```
( cd /data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts/tmp && \
erlc -W   -I/data/builds/oi-userland/components/runtime/erlang/build/amd64/lib/kernel/ebin -I/data/builds/oi-userland/components/runtime/erlang/build/amd64/lib/stdlib/ebin -I/data/builds/oi-userland/components/runtime/erlang/build/amd64/lib/sasl/ebin +no_warn_sasl -o /data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts /data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts/start_clean.rel )
Crash: error:undef
[{systools,compile_rel,
           ["/data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts/start_clean",
            "/data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts/start_clean",
            {options,["/data/builds/oi-userland/components/runtime/erlang/build/amd64/lib/kernel/ebin",
                      "/data/builds/oi-userland/components/runtime/erlang/build/amd64/lib/stdlib/ebin",
                      "/data/builds/oi-userland/components/runtime/erlang/build/amd64/lib/sasl/ebin"],
                     "/data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts",
                     undefined,[],1,false,999,
                     [no_warn_sasl],
                     [],
                     "/data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts/tmp"}],
           []},
 {erl_compile,compile_file,4,[{file,"erl_compile.erl"},{line,359}]},
 {erl_compile,compile3,3,[{file,"erl_compile.erl"},{line,317}]},
 {erl_compile,compile,2,[{file,"erl_compile.erl"},{line,61}]},
 {erl_compile,'-compile_cmdline1/1-fun-0-',2,
              [{file,"erl_compile.erl"},{line,74}]}]
gmake[2]: *** [Makefile:84: /data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts/start_clean.script] Error 2
gmake[2]: Leaving directory '/data/builds/oi-userland/components/runtime/erlang/build/amd64/erts/start_scripts'
gmake[1]: *** [Makefile:73: local_setup] Error 2
gmake[1]: Leaving directory '/data/builds/oi-userland/components/runtime/erlang/build/amd64/erts'
make: *** [Makefile:849: local_setup] Error 2
gmake: *** [/data/builds/oi-userland/make-rules/configure.mk:181: /data/builds/oi-userland/components/runtime/erlang/build/amd64/.built] Error 2
$
```

components/runtime/erlang/Makefile
Toasterson: There is some background to this. illumos uses a forked version of GCC and we are still partially not upstreamed. It's a lot better but some things need testing when it comes to commandline options as they have in the past not exactly behaved as expected. However we thus far have found all problems and fixed them in the compiler. So if we are not using these flags and most other distributions (including linx and bsd) do we should start using them. Although we need to gather some packaging experience with them.

components/runtime/erlang/Makefile
Toasterson: Pleasse don't try to enforce codyng styles as we have nor formal rulebook. If you absolutely must include some cleanups into the pr please use Code suggestions to allow them to be included with one click and keep the amount of work for contributors low. https://docs.github.com/de/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

components/runtime/erlang/Makefile
Toasterson: This seems like a bug then if they are still needed... But since that's not your change we can also look at it independently. this is non blocking.

components/runtime/erlang/Makefile
Toasterson:
```suggestion
COMPONENT_VERSION_SHORT= $(shell echo $(COMPONENT_VERSION) | cut -d'.' -f1-2)
```

components/runtime/erlang/Makefile
Toasterson:
```suggestion
COMPONENT_SRC_1= $(COMPONENT_NAME_1)_$(COMPONENT_VERSION)
```

components/runtime/erlang/Makefile
Toasterson:
```suggestion
COMPONENT_ARCHIVE_URL_1= https://github.com/erlang/otp/releases/download/OTP-$(COMPONENT_VERSION)/$(COMPONENT_ARCHIVE_1)
```

components/runtime/erlang/Makefile
mtelka: > So if we are not using these flags and most other distributions (including linx and bsd) do we should start using them.

Okay, let's start to gather some information:
* Fedora Linux does *not* build `erlang` with `-fno-rtti`: https://src.fedoraproject.org/rpms/erlang/blob/rawhide/f/erlang.spec#_759
* nixpkgs uses `-fno-rtti` in four packages (out of 90k+): https://github.com/search?q=repo%3ANixOS%2Fnixpkgs%20fno-rtti&type=code

I'd really appreciate some reference to a repository with the massive usage of the `-fno-rtti` option.

mtelka: * Solaris - zero occurrences of `-fno-rtti` in their `solaris-userland`: https://github.com/search?q=repo%3Aoracle%2Fsolaris-userland%20fno-rtti&type=code

mtelka: > ... are the default for games and things like the mozilla browsers for good reason.

If some particular project (like a game or web browser) is confident they are safe to be built using `-fno-rtti` then they can do so internally and that's perfectly okay because the particular project knows its internals best.  They can also take some measurements to make sure their internal usage of `-fno-rtti` is still correct.  If they found it is no longer okay to be build with `-fno-rtti` they can just drop it.

I believe we build some packages with `-fno-rtti` and we do not know that just because this is hidden somewhere in their `Makefile` forrest (or whatever) and we simply do not care about project's internals too much.

But I consider too risky to force `-fno-rtti` from an outside of the project and ... hope?  or believe?  ... it will work properly.  And it will do so forever.  There are missing test results for `erlang` so we do not know the actual situation and we cannot evaluate the impact of added `-fno-rtti`.  And even with 100% passed test results the justification of added `-fno-rtti` is still weak.

RAC-oi: > Re `-finline-functions`: since we build with `-O2` by default (and I believe it is same for erlang too) this option should not be needed to be set explicitly.

true. in practive however it turned out that, depending on os/platform/gcc version, it's not as reliable as having that option explicitly. and since it doesn't hurt having it in addition ...

> Re `-fno-exceptions`: this is to disable some compile warnings.

no.

RAC-oi: > There is some background to this. illumos uses a forked version of GCC and we are still partially not upstreamed. It's a lot better but some things need testing when it comes to commandline options as they have in the past not exactly behaved as expected.

okay, that changes things of course. i'll pick this up again in my general post here a little later ...

RAC-oi:
so, to summarize the things here and from ejabberd ...


the important one first: wonky compiler.
that's of course a major factor and in such a case it does of course make sense to keep things as conservative and careful as possible. however, the risk is less than one might think at first, because the buildsystems of most programs add a number of compiler options on their own. everyone who has ever built some foss with a compiler other than gcc or clang knows how many of them are added sometimes (because they fail in such cases).
now one could of course say that in this case the developers have tested and verified these to work with/for their software as rightfully pointed out by @mtelka.
at second glance however this turns out to be a naive fallacy and here is why:

95% of all foss is made on and for linux. and what works fine there, doesn't necessarily have to work fine somewhere else. so the chances of getting into trouble because of one or more options, which have been added by the program's buildsystem, are much higher compared to 2 or 3 manually added options. simply because many programs add much more than just 2 or 3. and in fact some programs add the ones i added on their own anyway. mariadb is just one such example of many.
further, the chances of the program itself having a bug, which causes much more trouble than any compiler option could, are not only higher but also more severe and more likely. a simple look at the userland packages and their required updates makes that clear.


as to keeping my comments here short, that was by no means meant to be rude or something like that but honestly caused by a lack of time. if @mtelka or anyone else is not familiar with c++ or has any other questions about what i did and why, i'm happy to have a chat with them. maybe not right here in form of comments but at a slightly more cozy and not time critical place. because these things are neither short nor trivial, which is also why they don't fit in a comment in a source file.

i hope that clears things up and i'm looking forward to working with all of you in the future :-)

components/runtime/erlang/Makefile
mtelka: > > Re `-finline-functions`: since we build with `-O2` by default (and I believe it is same for erlang too) this option should not be needed to be set explicitly.
> 
> true. in practive however it turned out that, depending on os/platform/gcc version, it's not as reliable as having that option explicitly. and since it doesn't hurt having it in addition ...

I disagree with this approach[*], but I can live with it because added `-finline-functions` is effectively no-op.  If `-finline-functions` is a must for you, then please add a comment saying something similar as you wrote above or anything else explaining why you think it is needed.

[*] if we would accept this is the way how to do things then somebody else could come and add `-fvar-tracking` with the exactly same explanation, then somebody else will add `-fshrink-wrap`, etc.  Where it will end?  Makefiles will be full of unnecessary stuff making maintenance of components harder.

> > Re `-fno-exceptions`: this is to disable some compile warnings.
> 
> no.

Yes, you are right.  I messed this with `-Wno-exceptions`.  I'm sorry.

The `-fno-exceptions` option falls under the same category as `-fno-rtti`, i.e. I do not think it is safe to force it from outside of a project.

BTW, if both `-fno-rtti` and `-fno-exceptions` are so good (and safe) for `erlang` then why not to propose their addition directly [into the `erlang` project](https://github.com/erlang/otp)?  This would make `erlang` better not only for OpenIndiana, but also for everybody else.

mtelka:
> 95% of all foss is made on and for linux.

Could you please let us know where this information comes from?  To be clear, I'm not saying it is not true.  I'm just curious about the source of this information.

> and what works fine there, doesn't necessarily have to work fine somewhere else. so the chances of getting into trouble because of one or more options, which have been added by the program's buildsystem, are much higher compared to 2 or 3 manually added options.

Here you are doing wrong generalization.  It is not that all added options have the exactly same (or very similar) impact.  Some are more-or-less harmless, while others could be dangerous.  So you cannot say that any random 2 or 3 options are similar to other 2 or 3 options and their impact is same.  You also cannot say that _many added by a program_ (where _many_ is more than 3) are for sure worse than 2 or 3 added by you.  You also cannot say that _one added by a program_ is for sure far better than 2 or 3 added by you.

> simply because many programs add much more than just 2 or 3. and in fact some programs add the ones i added on their own anyway.

Here we are back to what I said few times: developers of the program are in far better position to add dangerous options than any average packager.

> mariadb is just one such example of many. further, the chances of the program itself having a bug, which causes much more trouble than any compiler option could, are not only higher but also more severe and more likely.

Any backing for your _higher_, _more severe_, and _more likely_?

> a simple look at the userland packages and their required updates makes that clear.

No it does not.  What you means by _required updates_?  Patches?  Are you trying to say that the number of patches makes clear that a chance of a program to contain a bug is _higher, more severe, and more likely_ than an issue caused by a random added compiler option?  Especially when the option is known to be dangerous.

> as to keeping my comments here short, that was by no means meant to be rude or something like that but honestly caused by a lack of time. if @mtelka or anyone else is not familiar with c++ or has any other questions about what i did and why, i'm happy to have a chat with them. maybe not right here in form of comments but at a slightly more cozy and not time critical place. because these things are neither short nor trivial, which is also why they don't fit in a comment in a source file.

Any chat is of little value because what really matters is the actual code that goes to the component's sources (Makefile, patches, etc.).  If there is no comment explaining the reason of some non-obvious things there then there is a big chance that the actual meaning of that code is misinterpreted and so the code could get mangled or removed in future as not necessary/superfluous (remember `-finline-functions`) or counterproductive/dangerous/out-of-scope (`-fno-rtti` and `-fno-exceptions`).

> i hope that clears things up

I'm sorry, but it does not much.

> and i'm looking forward to working with all of you in the future :-)

AndWac:
Please squash the commits.

components/runtime/erlang/Makefile
Toasterson: Ok, I have now looked through the compiler options in detail and had some discussions off the list with @RAC-oi these options give benefits and since compilation works, we save on C++ features which erlang does not use. All small improvements are good for small projects. Endless discussions just add frustration and are not helpful. As stated before the userbase for erlang on OpenIndina is small. We don't even have the most popular consumer namely rabbitmq packaged. So even if it where risky it's a good testbed. Marking as resolved and moving on.

PR #16202, opened 2024-02-21
web/haproxy: update to 2.8.6

RAC-oi:
updated and switched to LTS


RAC-oi:
seems to be a jenkins problem. if i can do something on my end let me know ...

components/web/haproxy/Makefile
```diff
@@ -12,26 +12,26 @@
...
 # Copyright 2016 Adam Stevko. All rights reserved.
 # Copyright 2022 Niklas Poslovski
+# Copyright 2024 Goetz Fischer, R-A-C. All rights reserved.
 #
 
-BUILD_STYLE=		justmake
-USE_OPENSSL11=		yes
+BUILD_STYLE= justmake
+USE_OPENSSL31= yes
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 20: USE_OPENSSL31 doesn't exist is shared-macros.mk. USE_OPENSSL10 and USE_OPENSSL11 is a relic from a time we only had these two versions. We have switched to OPENSSL_VERSION when we added 3.1. And this is the default at the moment (thus, can be omitted). As a result you can see that 3.1 is detected and used here.

RAC-oi:
okay but that's not the reason for the jenkins error.

PR #16212, opened 2024-02-22
web/server/ejabberd: update to 23.10

RAC-oi:
has to go together with erlang 26.2.2


AndWac:
When I run gmake REQUIRED_PACKAGES here I'll get (I have manually added the next three lines):
# Manually added build dependencies
REQUIRED_PACKAGES += runtime/erlang

# Auto-generated dependencies
REQUIRED_PACKAGES += SUNWcs
REQUIRED_PACKAGES += database/sqlite-3
REQUIRED_PACKAGES += library/expat
REQUIRED_PACKAGES += library/libyaml
REQUIRED_PACKAGES += library/security/openssl

Thus, the Makefile needs more work if you really want to use openssl-3.1. Its configure doesn't detect it automatically.

RAC-oi:
it detects the default which is 3.1 for userland according to you. hence:

```
# ldd build/prototype/i386/usr/lib/amd64/fast_tls-1.1.16/priv/lib/fast_tls.so 
	libssl.so.3 =>	 /lib/64/libssl.so.3
	libcrypto.so.3 =>	 /lib/64/libcrypto.so.3
	libc.so.1 =>	 /lib/64/libc.so.1
	libz.so.1 =>	 /usr/lib/64/libz.so.1
	libsocket.so.1 =>	 /lib/64/libsocket.so.1
	libnsl.so.1 =>	 /lib/64/libnsl.so.1
	libmp.so.2 =>	 /lib/64/libmp.so.2
	libmd.so.1 =>	 /lib/64/libmd.so.1
	libm.so.2 =>	 /lib/64/libm.so.2

```

AndWac:
It doesn't detect the default openssl on my system which makes me doubt that it will on the official build server.
This is a common problem if the package's configuration is not following standards strictly (eg. uses pkgconf).
Are you sure that on your system openssl's mediator hasn't been changed?

RAC-oi:
the mediator is set to 3.1 since that's the userland standard. is there a reason why it should be set to something else?

AndWac:
Mediator settings and oi-userland defaults are two separate things. oi-userland standard is being used when we build things which can be overwritten by local Makefiles. Mediator settings are being used by installed binaries that don't have explicit run paths. The mediator for openssl should be
openssl             vendor    1.0     vendor     openssl

Otherwise you'll likely break installed software and prohibit the detection of situations like with ejabberd.
You'll need to find a way (eg. by setting an appropriate combination of CONFIGURE_OPTIONS, LDFLAGS, CPPFLAGS, ...) to make it use the openssl version you want it to use. We still have some packages where we weren't able to do so.
Alas at least one package is still bound to 1.0 so this is the least common denominator and dictates the mediator settings.

RAC-oi:
not sure that's a problem. at least the libs have different names. let's try:
```
# pkg set-mediator -V 1.0 openssl
...
# openssl version
OpenSSL 1.0.2u  20 Dec 2019
# ldd build/prototype/i386/usr/lib/amd64/fast_tls-1.1.16/priv/lib/fast_tls.so
	libssl.so.3 =>	 /lib/64/libssl.so.3
	libcrypto.so.3 =>	 /lib/64/libcrypto.so.3
```
still getting the right ones. let's try the other way around:
```
# pkg set-mediator -V 3.1 openssl
...
# openssl version
OpenSSL 3.1.5 30 Jan 2024 (Library: OpenSSL 3.1.5 30 Jan 2024)
# nginx -V
nginx version: nginx/1.25.3
built with OpenSSL 1.0.2u  20 Dec 2019
# ldd /usr/sbin/nginx 
	libsocket.so.1 =>	 /lib/64/libsocket.so.1
	libnsl.so.1 =>	 /lib/64/libnsl.so.1
	libsendfile.so.1 =>	 /lib/64/libsendfile.so.1
	libpthread.so.1 =>	 /lib/64/libpthread.so.1
	libpcre2-8.so.0 =>	 /usr/lib/64/libpcre2-8.so.0
	libssl.so.1.0.0 =>	 /lib/64/libssl.so.1.0.0
	libcrypto.so.1.0.0 =>	 /lib/64/libcrypto.so.1.0.0
```
no problem that way either. i.e. once they're built with the right ssl they're fine unless of course some program uses the "openssl" binary directly. but that in turn would crash all programs which use a more recent ssl. so that's bad either way.
another problem could be rebuilding one of the programs which uses an older ssl. but does that happen? and if so would it make sense to build the old one with the old ssl again? nginx from my example for example, why was that built with ssl 1.0 in the first place? nginx 1.25.3 was released in october last year and by then at least ssl 3.0 should have been used.

AndWac:
It's best to use
pfexec pkg unset-mediator openssl
here. Make sure you have a clean build environment before you test it with the new mediator settings.
Thus, run gmake clean before and build it again.

mtelka:
> It's best to use pfexec pkg unset-mediator openssl here.

Exactly.  It is strongly recommended that on `oi-userland` dev machines **all** mediators are unset (IOW, set to defaults).

RAC-oi:
> Otherwise you'll likely break installed software

not the case as demonstrated.

> and prohibit the detection of situations like with ejabberd. You'll need to find a way (eg. by setting an appropriate combination of CONFIGURE_OPTIONS, LDFLAGS, CPPFLAGS, ...) to make it use the openssl version you want it to use. We still have some packages where we weren't able to do so.

well then the next logical step would be to tackle those. hardcoding all future packages to a current ssl, just because a handful of old ones are stuck, is very inefficient.
openssl 1.0.2 reached its eol in 2019. packages still using that should either be updated or thrown out.

RAC-oi:
latest commit: compromise solution. only using $(OPENSSL_INCDIR) and $(OPENSSL_LIBDIR) and not dependent on mediator settings.

components/web/ejabberd/Makefile
```diff
@@ -65,41 +65,30 @@ COMPONENT_PRE_CONFIGURE_ACTION+= \
...
 
-COMPONENT_TEST_ARGS = "CT_BACKENDS=mnesia"
-COMPONENT_TEST_TARGETS = test
-COMPONENT_TEST_TRANSFORMS= \
-    '-n ' \
-    '-e "/TEST COMPLETE/p"'
-
+REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
+REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
 REQUIRED_PACKAGES += SUNWcs
```
AndWac, line 91: Did you run gmake REQUIRED_PACKAGES or did you manually edit the Makefile?
I miss
# Auto-generated dependencies
in front of the required packages list which is automatically generated.

components/web/ejabberd/Makefile
```diff
@@ -23,36 +23,36 @@
...
+TEST_TARGET= $(NO_TESTS)
 include $(WS_MAKE_RULES)/common.mk
 
 CXX      += $(CC_BITS)
-CFLAGS   += $(JPEG_CPPFLAGS) $(JPEG_CFLAGS) -I$(USRINCDIR)/gd2
-CXXFLAGS += $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS) -I$(USRINCDIR)/gd2
-CPPFLAGS += $(JPEG_CPPFLAGS) -I$(USRINCDIR)/gd2
-LDFLAGS  += $(JPEG_LDFLAGS) -lnsl -lsocket
+CFLAGS   += $(JPEG_CPPFLAGS) $(JPEG_CFLAGS) -I$(USRINCDIR)/gd2 -finline-functions -funroll-loops
+CXXFLAGS += $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS) -I$(USRINCDIR)/gd2 -finline-functions -funroll-loops -fno-exceptions -fno-rtti
```
mtelka: Could you please add a note why these two new options are needed?  Also, please add them as a separate `CFLAGS += ` lines to keep these lines short.

Thank you.

components/web/ejabberd/Makefile
RAC-oi: i did both because ironically the automatic check missed runtime/erlang

components/web/ejabberd/Makefile
RAC-oi: erm, well, i can separate them but i can't explain what these options do or why it's usually good to have them in a comment in 2 or 3 sentences.
somebody who doesn't know these at all has to read a little more about them than a brief comment. i assumed that everyone working on a repo like this is familiar with compilers. it's not like inlining or unrolling loops is something exotic.

components/web/ejabberd/Makefile
```diff
@@ -23,36 +23,36 @@
...
 
 COMPONENT_NAME= ejabberd
 COMPONENT_FMRI= web/server/$(COMPONENT_NAME)
-COMPONENT_VERSION= 22.0.5
-HUMAN_VERSION= 22.05
-COMPONENT_PROJECT_URL=	http://www.ejabberd.im/
-COMPONENT_SRC= $(COMPONENT_NAME)-$(HUMAN_VERSION)
+COMPONENT_VERSION= 23.10
+HUMAN_VERSION= 23.10
+COMPONENT_PROJECT_URL= http://www.ejabberd.im/
```
mtelka, line 36: Please switch to https here.

components/web/ejabberd/Makefile
mtelka: > erm, well, i can separate them but i can't explain what these options do or why it's usually good to have them in a comment in 2 or 3 sentences. somebody who doesn't know these at all has to read a little more about them than a brief comment. i assumed that everyone working on a repo like this is familiar with compilers. it's not like inlining or unrolling loops is something exotic.

I'm very sorry to see you cannot explain why you want/need these options.  OTOH, I believe these options are something that deserves explanation so here is my view:

After digging a bit into the `no-rtti` stuff it looks like that its use could be dangerous and cause surprises.  [Here](https://stackoverflow.com/questions/4486609/when-can-compiling-c-without-rtti-cause-problems) is a nice example.

My preliminary conclusion is that for general purpose environments like OpenIndiana where we prefer stability over the memory footprint (how much we would save with `-fno-rtti`?) the usage of `-fno-rtti` should be both well justified and limited to parts of the system where it really cannot cause any problems.

I'm sorry, but -1 for `-fno-rtti` from me here.

For `-funroll-loops` the `gcc` man page says:
```
This option makes code larger, and may or may not make it run faster.
```
Do you have any numbers justifying this option here?

Re `-finline-functions`: since we build with `-O2` by default (and I believe it is same for ejabberd too) this option should not be needed to be set explicitly.

Re `-fno-exceptions`: ~this is to disable some compile warnings.  Why do we need that?~ Edit: This option falls under the same category as `-fno-rtti`, i.e. I do not think it is safe to force it from outside of a project.

RAC-oi: > I'm very sorry to see you cannot explain why you want/need these

that's not what i said. it seems you missed the last part of my sentence:

"i can't explain what these options do or why it's usually good to have them **in a comment in 2 or 3 sentences**"

components/web/ejabberd/Makefile
RAC-oi: seriously?

mtelka: http://www.ejabberd.im/ is just a redirect to https://www.ejabberd.im/ so yes, seriously.  What is wrong with https://www.ejabberd.im/ ?

Toasterson:
```suggestion
COMPONENT_PROJECT_URL= https://www.ejabberd.im/
```

components/web/ejabberd/Makefile
Toasterson: Same comment as in #16187 this software also has a small enough userbase that we can test it. I am unaware of any commercial user using it.

PR #16235, opened 2024-02-23
Rebuild all packages depending on libjpeg6 that cannot find their library in a non default location merge after libjpeg9 PR

Toasterson:
Depends on merging #16035 first


Toasterson:
Needs to be redone, and I can now split it.

PR #16255, opened 2024-02-24
Gnome mousetweaks4

isoriano1968:
Hi,

mousetweaks is one of the many packages to bring GNOME 3 to OI.

Cheers

Iggi

### DIFF
--- a/components/gnome/mousetweaks/Makefile
+++ b/components/gnome/mousetweaks/Makefile
@@ -1,0 +1,65 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2015, 2021, Oracle and/or its affiliates.
+#
+
+BUILD_BITS =            64
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		mousetweaks
+COMPONENT_VERSION=	3.32.0
+COMPONENT_PROJECT_URL=  https://wiki.gnome.org/Projects/Mousetweaks
+COMPONENT_ARCHIVE_HASH= \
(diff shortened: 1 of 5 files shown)


mtelka:
You accidentally deleted files in `tools` directory.  Please fix.

Toasterson:
@isoriano1968 asked me via chat if I can cleanup the commit I think I finally found out how :)

make-rules/gnome-component.mk
```diff
@@ -0,0 +1,60 @@
...
+
+#
+# Copyright (c) 2016, 2020, Oracle and/or its affiliates. All rights reserved.
+#
+
+#
+# This file sets up the standard, default options and base requirements for
+# GNOME userland components.
+#
+COMPONENT_PROJECT_URL ?=	https://www.gnome.org/
```
mtelka, line 30: We already have `gnome.mk`.  No need to introduce something similar.

components/gnome/mousetweaks/Makefile
```diff
@@ -0,0 +1,65 @@
...
+REQUIRED_PACKAGES += library/desktop/cairo
+REQUIRED_PACKAGES += library/desktop/gtk3
+REQUIRED_PACKAGES += library/desktop/pango
+REQUIRED_PACKAGES += library/glib2
+REQUIRED_PACKAGES += x11/library/libx11
+REQUIRED_PACKAGES += x11/library/libxcursor
+REQUIRED_PACKAGES += x11/library/libxfixes
+REQUIRED_PACKAGES += x11/library/libxtst
+# Auto-generated dependencies
+REQUIRED_PACKAGES += library/desktop/cairo
```
mtelka, line 57: I do not think it is good idea to bring this from `solaris-userland`.  We already diverged too much and many things makes no sense here when copied verbatim or are even counterproductive.

This is fairly simple component so package it from scratch should be straightforward.  And you would also get much smaller `Makefile`.  For example here are duplicate dependencies.  Above is `COMPONENT_BAID`, etc.

components/gnome/mousetweaks/Makefile
```diff
@@ -0,0 +1,65 @@
...
+#
+
+#
+# Copyright (c) 2015, 2021, Oracle and/or its affiliates.
+#
+
+BUILD_BITS =            64
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		mousetweaks
```
mtelka, line 29: ... and finally, this project is [deprecated](https://gitlab.gnome.org/Archive/mousetweaks), so it cannot be needed for GNOME 3.  The homepage (referenced from `COMPONENT_PROJECT_URL`) says so too.

Toasterson: It is Archived, that does not mean it's deprecated. It is a simple piece of software. It looks more finished, which is ok.

mtelka: > It is Archived, that does not mean it's deprecated. It is a simple piece of software. It looks more finished, which is ok.

Yes, Archived could also mean it is not active (see https://gitlab.gnome.org/Archive), but in this case it is deprecated.

What GNOME package depends on mousetweaks?

PR #16291, opened 2024-02-26
GLEW headers should be include from GL base

janbar:
see https://glew.sourceforge.net/basic.html


AndWac:
If you want to change anything you'll also need to touch the Makefile. If the Makefile isn't changed, jenkins will not create a new revision. Furthermore, with every change you'll either need to change COMPONENT_VERSION or COMPONENT_REVISION. If COMPONENT_REVISION is not yet set then introduce it and set it to 1. If you change COMPONENT_VERSION the either delete COMPONENT_REVISION or set it to 0.

janbar:
Thanks ! jenkins built it now.

PR #16313, opened 2024-02-29
libjpeg6-ijg: make development files mediated

mtelka:
This is first small and easily understandable step to obsolete the ancient `libjpeg6-ijg` component.  It effectively changes nothing on-disk (yet).  It just adds `jpeg` mediator for `libjpeg6-ijg` development files to make them replaceable.  This is preparation for future work.

### DIFF
--- a/components/library/libjpeg6-ijg/Makefile
+++ b/components/library/libjpeg6-ijg/Makefile
@@ -18,7 +18,7 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		libjpeg6-ijg
 COMPONENT_VERSION=	6.0.2
 LIBJPEG_API_VERSION= 6b
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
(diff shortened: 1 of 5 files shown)


alarcher:
Great!

Toasterson:
This won't be necessary and just adds an extra step which I'll have to cleanup in the next PR. We simply remove the links to the header from `libjpeg6` package and add appropriate links to `libjpeg-turbo`. No links to libjpeg-turbo conflict. as they have different filenames. So we can simply leave the binary files in place and thus not break anything.

mtelka:
> This won't be necessary and just adds an extra step which I'll have to cleanup in the next PR.

To cleanup everything you would need several steps anyway, because #16035 in the state as it is now is not the final state.  So one more step (is it really one more step? maybe, maybe not) is not a big deal.

mtelka:
> Great!

Not everybody share your view.  Unfortunately.

PR #16363, opened 2024-03-02
add postgresql-16.2

AndWac:


make-rules/shared-macros.mk
```diff
@@ -906,6 +906,7 @@ PKG_MACROS +=   PERL_VERSION=$(PERL_VERSION)
 # |         12         |   2024-11-14   |
 # |         14         |   2026-11-12   |
 # |         15         |   2027-11-11   |
+# |         16         |   2028-11-09   |
 # +--------------------+----------------+
 #
 # See https://www.postgresql.org/support/versioning/
```
mtelka, line 912: Please update `PG_VERSION` below too.

AndWac: I want to wait until some users had the chance to test the new version.

mtelka:
* Test results are same as for version 15 (`timestamptz` failed, everything else passed) so 16 should not be worse than 15.
* The default version will be set to 16 anyway once this is installed (see mediator).
* `PG_VERSION` is for other components only (when they are rebuild), so it will take effect slowly.

In short, no need to wait.

mtelka:
Great. Thank you!

PR #16368, opened 2024-03-03
Remove main symlink from libjpeg-turbo

Toasterson:


mtelka:
This change is wrong and moves us farther from the completion of https://www.illumos.org/issues/7391.

PR #16416, opened 2024-03-06
Update libjpeg meta-package to deliver libjpeg-turbo

Toasterson:


components/meta-packages/libjpeg/Makefile
```diff
@@ -15,7 +15,7 @@
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libjpeg
-COMPONENT_VERSION=	6.0.2
+COMPONENT_VERSION=	8.3.2
 COMPONENT_SUMMARY=	Joint Photographic Experts Group library
```
AndWac, line 20: Shouldn't we enhance the summary?

Toasterson: Sure, I commited what I could come up with. Open to suggestions though.

mtelka:
Since nothing in OpenIndiana depends on this package I suggest to obsolete it.  One less useless meta-package is always good.

Toasterson:
Agreed. Moving into new PR as I killed my git history in the process :)

PR #16432, opened 2024-03-08
web/server/nginx: switch to openssl-3.1

RAC-oi:
openssl 3.1 and current version

### DIFF
--- a/components/web/nginx/Makefile
+++ b/components/web/nginx/Makefile
@@ -13,63 +13,63 @@
 #
 
 BUILD_BITS= 64
-# Alas nginx ignores the openssl settings
-OPENSSL_VERSION= 3.1
-USE_PARALLEL_BUILD= yes
 include ../../../make-rules/shared-macros.mk
 
-COMPONENT_NAME=		nginx
-COMPONENT_VERSION=	1.25.4
-COMPONENT_SUMMARY=	Nginx Webserver
-COMPONENT_PROJECT_URL=	https://nginx.net/
-COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
-COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
-COMPONENT_ARCHIVE_HASH=	sha256:760729901acbaa517996e681ee6ea259032985e37c2768beef80df3a877deed9
-COMPONENT_ARCHIVE_URL=	https://nginx.org/download/$(COMPONENT_ARCHIVE)
-COMPONENT_SIG_URL=	$(COMPONENT_ARCHIVE_URL).asc
+COMPONENT_NAME= nginx
+COMPONENT_VERSION= 1.25.4
+COMPONENT_SUMMARY= Nginx Webserver
+COMPONENT_PROJECT_URL= https://nginx.org/
+COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= sha256:760729901acbaa517996e681ee6ea259032985e37c2768beef80df3a877deed9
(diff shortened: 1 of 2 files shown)


components/web/nginx/Makefile
```diff
@@ -13,63 +13,62 @@
...
-COMPONENT_SUMMARY=	Nginx Webserver
-COMPONENT_PROJECT_URL=	https://nginx.net/
-COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
-COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tar.gz
-COMPONENT_ARCHIVE_HASH=	sha256:760729901acbaa517996e681ee6ea259032985e37c2768beef80df3a877deed9
-COMPONENT_ARCHIVE_URL=	https://nginx.org/download/$(COMPONENT_ARCHIVE)
-COMPONENT_SIG_URL=	$(COMPONENT_ARCHIVE_URL).asc
+COMPONENT_NAME= nginx
+COMPONENT_VERSION= 1.25.4
+COMPONENT_SUMMARY= Nginx Webserver
```
AndWac, line 20: Please add COMPONENT_REVISION= 1 as we are already on 1.25.4

RAC-oi: well release versions are arbitrary. nginx has a third number while others just increase the second one instead or follow an even completely different naming. so using COMPONENT_REVISION this way would tell pretty much nothing.
wouldn't COMPONENT_REVISION make more sense if it was used for changes made by us to the same version? e.g. we update a package of the same program and same version with an additional patch or other configure options and similar.

AndWac: Your PR doesn't change the version but you only found a way to switch to openssl-3.1.
So we agree on the use of COMPONENT_REVISION here as we only change configure options.

AndWac: So, your commit message and the PR's title should also be changed.

RAC-oi: oh that was a misunderstanding, seems i missed the 1.25.4 update. when i prepared it, the repo still had 1.25.3.

AndWac: That happens because we don't use anything to announce work on items. And that'e because we are so few that it would only add bureaucracy and is not worth the efforts.

AndWac:
So please change the commit message and update the PR so I can merge.

components/web/nginx/Makefile
```diff
@@ -13,63 +13,62 @@
 #
 
 BUILD_BITS= 64
-# Alas nginx ignores the openssl settings
-OPENSSL_VERSION= 3.1
-USE_PARALLEL_BUILD= yes
```
mtelka, line 18: Why this got removed?  Does it fail to build when the parallel build is enabled?

RAC-oi:
> please change the commit message and update the PR

of course, coming up tonight

components/web/nginx/Makefile
```diff
@@ -13,63 +13,63 @@
...
+COMPONENT_ARCHIVE_HASH= sha256:760729901acbaa517996e681ee6ea259032985e37c2768beef80df3a877deed9
+COMPONENT_ARCHIVE_URL= https://nginx.org/download/$(COMPONENT_ARCHIVE)
+COMPONENT_SIG_URL= $(COMPONENT_ARCHIVE_URL).asc
 COMPONENT_FMRI=	web/server/nginx
-COMPONENT_CLASSIFICATION=	Web Services/Application and Web Servers
-COMPONENT_LICENSE=	BSD
+COMPONENT_CLASSIFICATION= Web Services/Application and Web Servers
+COMPONENT_LICENSE= BSD
 COMPONENT_LICENSE_FILE=	LICENSE
+COMPONENT_REVISION= 1
```
mtelka, line 31: Please move this right below `COMPONENT_VERSION` so it is not forgotten on the next version update.  Thank you.

PR #16532, opened 2024-03-15
python/scour: add 0.38.2

iigs:
This is dependency for one or more widgets in inkscape.  This is not built with python-integrate-project.  inkscape widgets successfully run with this python module.

### DIFF
--- a/components/python/scour/Makefile
+++ b/components/python/scour/Makefile
@@ -1,0 +1,52 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2024 Geoff Weiss
+#
+
+BUILD_STYLE=setup.py
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=         scour
+COMPONENT_VERSION=      0.38.2
+COMPONENT_FMRI=         library/python/scour
+COMPONENT_CLASSIFICATION=Development/Python
+COMPONENT_SUMMARY=      An SVG optimizer/cleaner written in Python that reduces the size of scalable vector graphics by optimizing structure and removing unnecessary data.
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= \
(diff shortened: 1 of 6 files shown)


AndWac:
gmake test fails here:
...
py39: remove tox env folder /export/home/andreas/oi-userland/components/python/scour/build/amd64-3.9/.tox/py39
py39: commands[0]> scour --version
0.38.2
py39: commands[1]> python -m coverage run --parallel-mode --source=scour test_scour.py
/export/home/andreas/oi-userland/components/python/scour/build/amd64-3.9/.tox/py39/bin/python: No module named coverage
py39: exit 1 (0.03 seconds) /export/home/andreas/oi-userland/components/python/scour/build/amd64-3.9> python -m coverage run --parallel-mode --source=scour test_scour.py pid=4198
  py39: FAIL code 1 (0.11=setup[0.03]+cmd[0.05,0.03] seconds)
  evaluation failed :( (0.23 seconds)
gmake: *** [/export/home/andreas/oi-userland/make-rules/setup.py.mk:568: /export/home/andreas/oi-userland/components/python/scour/build/amd64-3.9/.tested] Error 1

iigs:
This behavior was documented in the Makefile with these two lines:
```
# gmake test fails.  pypi does not have the test_scour.py file
TEST_STYLE = none
```
Please advise if I need to do something else in the Makefile.

It should be noted that I started working on scour in January 2023.  I used several existing (at the time) python module recipes as a basis to make this.  The trend around that time frame was to use source from pypi.  I realize that the OI trend has evolved, but then this package hasn't been updated since 2020.  Not that I am implying that it is ok for older released packages should use older build styles--I don't know what length of time will be needed to change the COMPONENT_ARCHIVE_URL to github and redo the package based on the contents of that archive.

components/python/scour/Makefile
```diff
@@ -0,0 +1,54 @@
...
+COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH= \
+    sha256:6881ec26660c130c5ecd996ac6f6b03939dd574198f50773f2508b81a68e0daf
+COMPONENT_ARCHIVE_URL=  $(call pypi_url)
+COMPONENT_PROJECT_URL=  https://pypi.org/project/scour/
+COMPONENT_LICENSE=      APACHE
+COMPONENT_LICENSE_FILE= scour.license
+
+include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 36: If tests don't work for us then disable them by adding
TEST_TARGET= $(NO_TESTS)
here with a comment why tests have been disabled.

iigs: Thank you.  I used the terminator python module to determine where these lines should be placed.  I ran gmake sample-manifest, gmake test, and gmake publish with this change.  I am now getting the expected behavior.

PR #16583, opened 2024-03-19
firefox: update to 124.0

iigs:
Removed 03-browser-extension-css.patch since Mozilla unified the css for the extension page.

Bug 1875369 adds new Wayland support to Firefox.  It introduced two regressions which caused WebRender to stop working on OpenIndiana.  One of the regressions got fixed by adding back the removed code.  I am not happy with the 10-regression-in-Bug1875369.patch , but WebRender works with if using nVidia 470 or newer.  I don't like that I disabled an ASSERT, but it looks like there are checks in place to avoid a crash.  I used this build as my primary browser for a week and it ran stable and I saw no regressions from 123.x.


AndWac:
With this version videos cannot be played while it works with 123.x. Do you know if anything has to be rebuild or changed?

iigs:
Audio/video fixed.  Revert default preference change in 124 to use new Utility in 03-force-audio-non-utility.patch
Fixed permissions for vaapitest file.
Update to 124.0.1 will be provided in a separate PR.

PR #16676, opened 2024-03-27
libgit2: split runtime lib to own pkg; add libgit2-1.7

mtelka:
This is needed to fix cargo (yes, rust rebuild would fix that too, but ...) and to make the future libgit2 updates easier (so no package rebuilds are needed).

### DIFF
--- a/components/library/libgit2-1.7/Makefile
+++ b/components/library/libgit2-1.7/Makefile
@@ -19,20 +19,17 @@ BUILD_STYLE= cmake
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		libgit2
-COMPONENT_VERSION=	1.8.0
-COMPONENT_REVISION=	1
+COMPONENT_VERSION=	1.7.2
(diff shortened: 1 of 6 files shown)


mtelka:
BTW, the problem is this:
```
$ cargo
ld.so.1: cargo: fatal: libgit2.so.1.7: open failed: No such file or directory
Killed
$
```

AndWac:
I am not happy with the idea to have a separate libgit package for cargo. When cargo needs a newer version we need to obsolete the package and create a new one.

Toasterson:
I am wondering if we are using IPS wrong here. It should be possible to have two versions of libgit installed and remove the nonversioned symlink from the old package.

AndWac:
I added a compatibility link into libgit2. With that we can rebuild rustc.

mtelka:
> I am not happy with the idea to have a separate libgit package for cargo. When cargo needs a newer version we need to obsolete the package and create a new one.

This was not needed for cargo only, but also for geany plugins.  This is also good design to avoid rebuilds in future.

PR #16722, opened 2024-03-29
xz Remove Backdoor activation code.

Toasterson:
This change removes the backdoored fourth file in the release tarball and regenerates a clean version with autoreconf. Unfortunately, we cannot use `autogen.sh` as we are missing the p4oa utility which is needed to build translated manfiles. Checking with diff shows that the backdoor activation was removed.

We are not affected by the backdoor as the code was already limiting itself to linux only as far as people can tell. However, it doesn't hurt to be cautious.

Our OpenSSH daemon does link to liblzma so on the off chance that it was activated it now should not be included in the build.

### DIFF
--- a/components/archiver/xz/Makefile
+++ b/components/archiver/xz/Makefile
@@ -27,6 +27,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		xz
 COMPONENT_VERSION=	5.6.1
+COMPONENT_REVISION= 1
 COMPONENT_SUMMARY=	XZ Utils - loss-less file compression application and library
 COMPONENT_PROJECT_URL=	https://tukaani.org/xz/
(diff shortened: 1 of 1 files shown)


Toasterson:
Upstream bug report here https://github.com/tukaani-project/xz/issues/92

onurmercury:
I believe you should use git + tag/commit like [Arch Linux packaging](https://gitlab.archlinux.org/archlinux/packaging/packages/xz/-/commit/881385757abdc39d3cfea1c3e34ec09f637424ad) (idk if it's possible here) or GitHub archives as the source instead. We're unsure if only that testfile is included in these tarballs.

Toasterson:
I first have to package all build dependencies. I am currently missing p4oa and dependencies. For now no other backdoors have been found and we do not have ifunc in our libc. So until somebody has a proper release tarball ready we are as safe as we can get. Also, the git repo got blocked fully so I can't change it now. I need to wait for a fork.

PR #16891, opened 2024-04-14
component(mate-session-manager): Enable delivery of gl acceleration checkers

Toasterson:


components/desktop/mate/mate-session-manager/Makefile
```diff
@@ -34,6 +34,10 @@ include $(WS_MAKE_RULES)/common.mk
 
 # Build requirements
 REQUIRED_PACKAGES += x11/library/xtrans
+REQUIRED_PACKAGES += developer/build/autoconf-archive
+
+## A note on nvidia dependencies. In order to things to properly depend on libGL or libEGL we have a line in the manifest.
+## Whoever rebuilds this will need to remove the REQUIRED_PACKAGES lines generated manually as we have no filtering mechanism
```
mtelka, line 41: This sounds really strange.  Could you please elaborate what is this all about?  For example, what you mean by _filtering mechanism_?

components/desktop/mate/mate-session-manager/mate-session-manager.p5m
```diff
@@ -26,6 +26,7 @@ license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
 depend type=require fmri=gnome/gnome-keyring
 depend type=require fmri=desktop/mate/marco
+depend fmri=pkg:/driver/graphics/nvidia fmri=pkg:/driver/graphics/nvidia-340 fmri=pkg:/driver/graphics/nvidia-470 fmri=pkg:/x11/library/mesa fmri=pkg:/driver/graphics/nvidia-390 type=require-any
```
mtelka: I believe it is a bad idea for MATE to depend directly and explicitly on a display driver.

Toasterson: This is not just the display driver but also the whole opengl stack which is nvidia driver dependant. Check out ogl-select for how this has been dealt with since OpenSolaris

components/desktop/mate/mate-session-manager/Makefile
Toasterson: We need to depend on any nvidia so people can install any graphics driver. ogl-select selects it at runtime and we have no way to properly handle require-any dependencies in the package that cannot be installed together.

components/desktop/mate/mate-session-manager/mate-session-manager.p5m
mtelka: ~What will happen if MATE is run without NVIDIA hardware?  What added value is dependency on `nvidia` then?~  EDIT: There is `mesa` in the list which I missed previously.

Toasterson: nvidia is in those cases installed aswell due to dependencies. However in the case no nvidia package is installed mesa is picked which delivers the opengl stack for OSS graphics drivers. Checkout ogl-select what we switch between at the moment.

mtelka: BTW, the `x11/server/xorg/driver/xorg-video` already depends on `nvidia` packages.  Yes, you added `mesa` to the list here, but still, shouldn't instead be `x11/server/xorg/driver/xorg-video`  modified accordingly (i.e. to add `mesa` to the list there together with `nvidia` packages)?

Toasterson: xorg-video driver are drivers. we depend on mesa because of libGL and libEGL. They have nothing to do with xorg-video drivers. Nvidia delivers it's own patched version of libGL and EGL and we thus have to depend on all possible versions of lib[E]GL. I do not see why we should introduce artificial dependencies on Xorg packages we do not need to depend on here.

mtelka: > xorg-video driver are drivers. we depend on mesa because of libGL and libEGL. They have nothing to do with xorg-video drivers. Nvidia delivers it's own patched version of libGL and EGL and we thus have to depend on all possible versions of lib[E]GL. I do not see why we should introduce artificial dependencies on Xorg packages we do not need to depend on here.

This explanation is what I missed.  Thank you!

Maybe something similar (stripped down, reworded) re libGL and libEGL should be added as a comment directly to the manifest so it is there for future reference.

For the actual code, please remove `pkg:/` from all fmris you added to the manifest.  It will cause that these packages will not appear in the `REQUIRED_PACKAGES` list.  I believe this is what you need here.  Also it would be great if the list of `require-any` packages is sorted somehow (maybe with one package per line?).

Thank you.

Toasterson: I can sort it but the pkg thing is not what is wrong. These packages cannot be installed simultaneously thus we have to remove them from the REQUIRED_PACKAGES so we don't kill people's dev setups.

components/desktop/mate/mate-session-manager/Makefile
```diff
@@ -52,3 +56,4 @@ REQUIRED_PACKAGES += x11/library/libx11
 REQUIRED_PACKAGES += x11/library/libxext
 REQUIRED_PACKAGES += x11/library/libxrender
 REQUIRED_PACKAGES += x11/library/libxtst
+REQUIRED_PACKAGES += x11/library/mesa
```
AndWac, line 60: When I run gmake REQUIRED_PACKAGES x11/library/mesa will be removed here.
So maybe it's a good idea to add a comment or move the dependency out of the generated area into a separate one.

components/desktop/mate/mate-session-manager/mate-session-manager.p5m
mtelka: The `pkg:/` thing is exactly what had to be removed (as you did) because when there are fmris with `pkg:/` in manifests then such packages are added automatically to `REQUIRED_PACKAGES`.  If fmris in manifests are without `pkg:/` then such packages are NOT added automatically to `REQUIRED_PACKAGES`.

components/desktop/mate/mate-session-manager/Makefile
mtelka: It got removed because of [this](https://github.com/OpenIndiana/oi-userland/pull/16891#discussion_r1564934797).

components/desktop/mate/mate-session-manager/Makefile
```diff
@@ -34,6 +34,10 @@ include $(WS_MAKE_RULES)/common.mk
 
 # Build requirements
 REQUIRED_PACKAGES += x11/library/xtrans
+REQUIRED_PACKAGES += developer/build/autoconf-archive
+
+## A note on nvidia dependencies. In order to things to properly depend on libGL or libEGL we have a line in the manifest.
```
mtelka, line 40: Please move this comment directly to the manifest.  Thanks.

components/desktop/mate/mate-session-manager/Makefile
mtelka, line 41: I believe this comment is no longer needed so please remove it.

components/desktop/mate/mate-session-manager/Makefile
```diff
@@ -34,6 +34,10 @@ include $(WS_MAKE_RULES)/common.mk
 
 # Build requirements
```
mtelka, line 36: You need to increment `COMPONENT_REVISION`.

components/desktop/mate/mate-session-manager/mate-session-manager.p5m
```diff
@@ -35,8 +41,8 @@ file path=usr/bin/mate-session-properties
 file path=usr/bin/mate-session-save
 file path=usr/bin/mate-wm
 file path=usr/libexec/mate/mate-session-check-accelerated
-#file path=usr/libexec/mate/mate-session-check-accelerated-gl-helper
-#file path=usr/libexec/mate/mate-session-check-accelerated-gles-helper
```
mtelka, line 39: While thinking more about this topic I start to suspect it should be enough to uncomment these two lines (without adding `pkg.depend.bypass-generate`) and run `gmake REQUIRED_PACKAGES`.  No manual `depend` line with `nvidia`/`mesa` should be needed.

PR #16894, opened 2024-04-15
library/objfw: update to 1.1.1

Midar:


Midar:
CI fails because the current oi-userland seems to depend on clang-18, but build-essential pulls in clang-17.

PR #16943, opened 2024-04-20
python/jaraco.packaging: update to 10.1.0

mtelka:


AndWac:
Is it really necessary to have daily (if not hourly) updates of python packages? These create a lot of work and make the repository grow fast without substantial value.

mtelka:
> Is it really necessary to have daily (if not hourly) updates of python packages?

No, it is not, obviously.  What would you prefer then?  Weekly updates?  Monthly?  Yearly?  Never to update Python packages and obsolete them?  I do not mind either way :-).   It is very easy to do nothing :-).

BTW, I understand that some Python projects are updated very often.  Or, to be precise, their updates are clustered.  For example `jaraco.packaging` had [5 version releases in last 3 days](https://pypi.org/project/jaraco.packaging/#history), but in 2023 there were only 3 releases in the whole year.  This is something I'm not happy with, but I cannot change it and it is up to the upstream maintainers how often they release their projects.

> These create a lot of work and make the repository grow fast without substantial value.

Yes, they create a lot of work, but mainly for me.  For you it is just one click per update I assume.  For the repository grow I disagree, because currently one Python package update creates two (usually small) new packages only (one unversioned and one `-39`), while for example illumos-gate nightly creates ca 500 packages every night.  And since most of them are binary (so every file is changed with the new build), then the repo's on-disk space is increasing significantly too.  Every night.

It is hard to discuss the "substantial value" because somebody could argue that OpenIndiana as a project have no substantial value either.

Here are some numbers for consideration for the "repository grow" argument:

```
# pkg list -a | egrep -v '(o|r)$' | grep 'library/python' | wc -l
    1407
# pkg list -a | egrep -v '(o|r)$' | wc -l
    5152
# pkg list -afv | egrep -v '(o|r)$' | grep 'library/python' | wc -l
    1878
# pkg list -afv | egrep -v '(o|r)$' | wc -l
   22444
#
```

Those numbers means there are currently 1407 Python packages out of 5152 packages, so IOW 27.3% of all packages are Python packages (when counted the latest versions of packages only).  But when looked at the ips repo history (since the last cleanup), we do have 1878 Python packages only there out of 22444 packages.  So only 8.3% of all packages in the current ips repo are Python packages.  The repo grow is not caused by Python packages, but by illumos-gate.

PR #16953, opened 2024-04-20
ffmpeg: update to 7.0

AndWac:


AndWac:
vlc is not compatible with the new version because libavcodec changed its interface with ffmpeg 6 and with 7 deprecated functions have been removed.

PR #16983, opened 2024-04-22
ffmpeg: bring back 6.1.1 with fake 7.0 version number as most depende…

AndWac:
…nts cannot cope with the new libavcodec


mtelka:
Wouldn't it be possible to add another package with 6.1.1 libraries only?  So we have both: old libraries for compatibility and new libraries so we can (slowly) rebuild other packages to use it.

AndWac:
We need to find out if and how both versions could co-exist. But that will need some time and efforts. So the fastest solution for the moment is this.

PR #16987, opened 2024-04-23
ffmpeg: move runtime libraries to own package

mtelka:
This is step towards the seamless upgrade to ffmpeg 7.0.

### DIFF
--- a/components/encumbered/ffmpeg/ffmpeg.p5m
+++ b/components/encumbered/ffmpeg/ffmpeg.p5m
@@ -169,29 +169,13 @@ file path=usr/include/libswscale/swscale.h
 file path=usr/include/libswscale/version.h
 file path=usr/include/libswscale/version_major.h
 link path=usr/lib/$(MACH64)/libavcodec.so target=libavcodec.so.60.31.102
-link path=usr/lib/$(MACH64)/libavcodec.so.60 target=libavcodec.so.60.31.102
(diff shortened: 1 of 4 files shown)


components/encumbered/ffmpeg/ffmpeg.p5m
```diff
@@ -169,29 +169,13 @@ file path=usr/include/libswscale/swscale.h
 file path=usr/include/libswscale/version.h
```
AndWac, line 169: This should depend on library/ffmpeg-6

mtelka: It does:
```
$ pkg contents -g $WS_TOOLS/../i386/encumbered-repo/ -mr video/ffmpeg@7.0-2024.0.0.2:20240422T215758Z | grep depend
depend fmri=pkg:/library/ffmpeg-6@6.1.1-2024.0.0.2 type=require
depend fmri=pkg:/library/sdl2@2.30.2-2024.0.0.0 type=require
depend fmri=pkg:/system/library/math@0.5.11-2024.0.0.22148 type=require
depend fmri=pkg:/system/library@0.5.11-2024.0.0.22148 type=require
$
```
This is because `/usr/bin/ffmpeg` needs those libraries and so `pkgdepend` automatically finds and adds the dependency.

PR #17031, opened 2024-04-26
thunderbird: update to 125.0

iigs:
Due to recent rust and/or clang updates on OpenIndiana, thunderbird 115 branch no longer compiles.  Bump to thunderbird to 125.

Integrate necessary firefox 125 patches. Remove license file since it is generated by the Makefile.


AndWac:
The license file doesn't get automatically created during publishing:

pkgsend: Action payload 'thunderbird.license' was not found in any of the provided locations:

iigs:
Added previous license file.  I'll look at why the code in the Makefile fails for another thunderbird release.

AndWac:
If you generate the license file with gmake license the resulting file is a lot bigger than the provided one.

PR #17048, opened 2024-04-28
libreoffice: update to 24.2.3.1

iigs:


mtelka:
Libreoffice fails to build and the cause seems to be this: https://bugs.documentfoundation.org/show_bug.cgi?id=160570

iigs:
Please advise.  I updated the initial PR with a fix.  I switched ICU to libreoffice's internal 73.2 instead of using the system ICU. I see that this PR now says closed.  Do I need to introduce a new PR for libreoffice 24.2.3.1?

AndWac:
Yes, you'll need a new PR because this one already had been merged. I noticed the failure then on the official build system.
You don't need to set COMPONENT_REVISION in this case because we haven't  successfully published this version yet.

iigs:
Thank you.  I'll make a new PR.

PR #17072, opened 2024-04-30
gtk+4: add gir lib files

iigs:
Add missing usr/lib/$(MACH64)/girepository-1.0 files (needed to compile gtk4 apps). Marcel gave bad advice about removing CONFIGURE_OPTIONS += -Dintrospection=enabled during initial gtk4 PR.


mtelka:
I assume the _initial gtk4 PR_ is #15757.  While looking there I'm unable to find any my advice related to `-Dintrospection` removal or inclusion.  First, you had this:
```
CONFIGURE_OPTIONS += -Dintrospection=disabled
```
and I asked: _Why is this needed?_
Then you changed it to:
```
CONFIGURE_OPTIONS += -Dintrospection=enabled
```
so I asked: _Do we need an explicit introspection enable? If so, please add a comment to the Makefile saying so (and possibly also explaining why)._
Then you removed the ` -Dintrospection` line.

@iigs, do I miss something?

PR #17112, opened 2024-05-06
gst-libav1: update to 1.22.12

mtelka:


mtelka:
According the [GStreamer 1.22 release notes](https://gstreamer.freedesktop.org/releases/1.22) there are no changes in `gst-libav` when compared to previous versions (1.22.11 and 1.22.10).

PR #17178, opened 2024-05-13
orc: update to 0.4.38; rename package to library/orc

mtelka:


components/library/orc/history
```diff
@@ -0,0 +1 @@
+system/library/orc@0.5.11,5.11-2022.0.0.2 library/orc
```
AndWac, line 1: Is this sufficient to let IPS see the dependency on system/library/orc as being satisfied?

mtelka: Yes.

PR #17227, opened 2024-05-17
setup.py.mk: fix numpy build

iigs:
numpy package is missing five files caused by the --skip-build argument.  See https://bugs.gentoo.org/717594

numpy will need to be rebuilt after this PR is integrated.


mtelka:
I'm very sorry, but this change is not suitable for general use and it will likely cause some issues in other packages.  If you really need to remove the `--skip-build` option from `COMPONENT_INSTALL_ARGS` then please do so directly in the `components/python/numpy/Makefile` file.

For example using something like:
```
COMPONENT_INSTALL_ARGS =
COMPONENT_INSTALL_ARGS +=       --root $(PROTO_DIR)
COMPONENT_INSTALL_ARGS +=       --install-lib=$(PYTHON_LIB)
COMPONENT_INSTALL_ARGS +=       --install-data=$(PYTHON_DATA)
COMPONENT_INSTALL_ARGS +=       --force
```

Thank you.

PR #17396, opened 2024-06-04
library/libssh2: update to 1.11.0

RAC-oi:
updated to 1.11.0 but still with openssl 1.0.2 because: https://openindiana.org/pipermail/oi-dev/2024-May/033625.html
also cleaned up the Makefile which had mixed openssl 3.1 and 1.1 settings ... which in the end didn't work anyway.


AndWac:
Did you run gmake test? If I do it here it fails for both 32 bit and 64 bit.
ndefined			first referenced
 symbol  			    in file
userauth_keyboard_interactive_decode_info_request test_auth_keyboard_info_request.o
ld: fatal: symbol referencing errors. No output written to test_auth_keyboard_info_request

components/library/libssh2/Makefile
```diff
@@ -47,13 +45,14 @@ include $(WS_MAKE_RULES)/common.mk
 
 CONFIGURE_OPTIONS += --enable-shared
 CONFIGURE_OPTIONS += --disable-static
-CONFIGURE_OPTIONS += --with-libssl-prefix=/usr/openssl/1.1
+CONFIGURE_OPTIONS += --disable-silent-rules
+CONFIGURE_OPTIONS += --disable-werror
+CONFIGURE_OPTIONS += --disable-debug
 
 # Test suite needs this
 unexport SHELLOPTS
```
AndWac, line 54: This shouldn't be needed anymore.

RAC-oi:
i don't know what the tests do but ldd shows no missing libs and a quick test with curl and sftp worked as well. i did in fact push this commit using the new version.

AndWac:
gmake test usually runs test that are provided by the project. The quality of these tests differ, though. But we try to use them because they can show possible problems with our package. You should run the tests before you start to work on a package and compare the results with what you get when you are working on an update or enhancement. If less tests fail then it's a good sign, if more tests fail you should at least investigate why. If building the tests fails it is an indication that an update won't be working as expected. One can assume that the creators of a library or application would know how to use it...

RAC-oi:
i meant the tests specific to this one right here. not the purpose of such tests in general of course.

anyway the problem with the tests seems to be known:
https://github.com/libssh2/libssh2/issues/1056
https://lists.openembedded.org/g/openembedded-core/topic/99295530

i ran it without --disable-static and then the test in question and 2 others worked. all others (more than 40 in total) however failed. the previous version btw only ran 2 tests and then called it good. that doesn't seem right either.

AndWac:
I am not sure how to proceed. At the moment the test results are outdated. You can build with static libs, update the sample-manifest and leave out the static libs in the manifest (or drop them with a transform). Then you can run the tests and update the test results file, too. Are you sure that during your tests with curl the library has been used?

AndWac:
I switched to libgcrypt instead of openssl and that made two more tests pass. The remaining tests fail because docker is being needed to run them.
I prefer the latest libgrypt over an outdated openssl version.

AndWac:
You can see the details after gmake test when you look in build/amd64/tests.

RAC-oi:
great, then just ditch my PR here and use your libgcrypt version. that's of course much better than using openssl 1.0.2.

PR #17583, opened 2024-06-18
16613 emacs-gtk vers 29.3 crashes on startup

gwr:
This fixes the emacs-gtk startup abort for me.

### DIFF
--- a/components/editor/emacs/Makefile
+++ b/components/editor/emacs/Makefile
@@ -29,6 +29,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         emacs
 COMPONENT_VERSION=      29.3
+COMPONENT_REVISION=	1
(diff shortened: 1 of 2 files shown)


gwr:
See https://www.illumos.org/issues/16613

AndWac:
Please insert COMPONENT_REVISION=1. With every change either the COMPONENT_VERSION or COMPONENT_REVISION has to be changed (incremented).
Then please run gmake REQUIRED_PACKAGES. This should update the Makefile. Then run gmake publish. This should also change the contents of pkg5. Then add pkg5 to your PR.

AndWac:
BTW I doubt that this is really the cause for your problems. It enables the use of webkitgtk2 which is broken on OI and causes core dumps but I haven't experienced crashes of emacs but only of webkitgtk2 instances.

PR #17827, opened 2024-06-24
Make userland-incorporator skip packages that don't want to be in the userland incorporation (mainly osnet)

Toasterson:


Toasterson:
reported by @Bill-Sommerfeld

Bill-Sommerfeld:
Thanks for the fix!  ONU is once again working as expected with the latest userland-incorporation package installed.

PR #17854, opened 2024-06-24
new package XCA - X Certificate and Key Management v2.6.0

fritzkink:


components/sysutils/xca/Makefile
```diff
@@ -0,0 +1,62 @@
...
+COMPONENT_FMRI= 	sysutils/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	Applications/System Utilities
+COMPONENT_LICENSE=	Christian Hohnstaedt
+COMPONENT_LICENSE_FILE=	COPYRIGHT
+
+include $(WS_MAKE_RULES)/common.mk
+
+# partly tests need a graphical desktop
+TEST_TARGET= $(NO_TESTS)
+
```
AndWac, line 44: We have a solution for this:
# Tests are run using an X virtual framebuffer.
COMPONENT_TEST_ENV= DISPLAY=:1

COMPONENT_PRE_TEST_ACTION = \
    $(SHELL) $(COMPONENT_DIR)/start_Xvfb $(COMPONENT_DIR)
COMPONENT_POST_TEST_ACTION = \
    $(SHELL) $(COMPONENT_DIR)/stop_Xvfb $(COMPONENT_DIR)

# Test dependencies
TEST_REQUIRED_PACKAGES += x11/server/xvfb

with the needed scripts (you can find them eg. in library/gtk+4/

fritzkink: Great I implemented, but unfortunately the first test expects interactive input, so fails anyway.
There is still a test failing which I don't understand. To compare time outputs it looks for the information about timezones in /usr/share/zoneinfo whereas Solaris/Illumos has this information in /usr/share/lib/zoneinfo. Does anyone know whether this is a configurable path?

fritzkink: When creating a link from /usr/share/lib/zoneinfo /usr/share/zoneinfo the test passes.

fritzkink: i found the issue:
strings /usr/lib/qt/5.15/lib/amd64/libQt5Core.so.5.15.13|grep zoneinfo
/usr/share/zoneinfo/
/usr/lib/zoneinfo/

Qt5 is the guilty package.

fritzkink: Now that qt5 is updated and the zoneinfo patch is included this package can be released as everything is working (incl. all tests).

PR #17860, opened 2024-06-24
lynx: update to 2.9.2; switch to openssl 3.1

iigs:


components/web/lynx/Makefile
```diff
@@ -25,20 +25,18 @@
 #
 
 BUILD_BITS=			64
-USE_OPENSSL11=		yes
+USE_OPENSSL31=		yes
```
mtelka: Our build framework does not use `USE_OPENSSL31`.  Please remove this line.

iigs:
Fixed CDDL header.  Fixed openssl selection.

PR #17882, opened 2024-06-25
w3m: update to 0.5.3; switch to openssl 3.1

iigs:


components/web/w3m/Makefile
```diff
@@ -14,18 +14,17 @@
...
 
-USE_OPENSSL11 = yes
+OPENSSL_VERSION= 3.1
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= w3m
-COMPONENT_VERSION=  0.5.3.20230121
-HUMAN_VERSION= 0.5.3+git20230121
-#COMPONENT_REVISION= 6
+COMPONENT_VERSION=  0.5.3.20210102
```
mtelka, line 21: Isn't this a downgrade?

mtelka: Obviously, it is.  This won't work as expected since it is not possible to update to older IPS package version automatically.  IOW, the `0.5.3.20230121` version will be listed as the latest one so nobody will update to `0.5.3.20210102` version introduced here.

iigs:
closing PR since, among reasons, the commit title is wrong -- this will not be a version update.

PR #17955, opened 2024-06-30
rsync: switch to openssl 3.1

iigs:


components/network/rsync/Makefile
```diff
@@ -1,22 +1,12 @@
 #
-# CDDL HEADER START
+# This file and its contents are supplied under the terms of the
```
mtelka: Is it okay to change the copyright notice of already existing file with the Oracle copyright in it?

AndWac: The original text references http://www.opensolaris.org/os/licensing which is not available anymore. The new text is the best replacement as it seems to be in the same vein and continues to use the CDDL.

iigs:
I found at least one example of an open file from Oracle with an original CDDL header, last modified post merge by Oracle.  The file contains an Oracle copyright, which an Illumos dev did not change the header when added for Illumos use.  Even  though this is slightly different use case, I will follow that example for header use here.

mtelka:
> I found at least one example of an open file from Oracle with an original CDDL header, last modified post merge by Oracle.

URL?  Thank you.

PR #17957, opened 2024-06-30
socat: update to 1.8.0.0; switch to openssl 3.1

iigs:
integrated omnios patch

### DIFF
--- a/components/network/socat/Makefile
+++ b/components/network/socat/Makefile
@@ -13,21 +13,23 @@
 # Copyright (c) 2018, Michal Nowak
 #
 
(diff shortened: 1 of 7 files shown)


components/network/socat/Makefile
```diff
@@ -14,20 +14,23 @@
 #
 
 PREFERRED_BITS=		64
+OPENSSL_VERSION= 3.1
```
AndWac, line 16: You can remove the PREFERRED_BITS settings as we only use 64 bit.

iigs: Removed PREFERRED_BITS line

PR #18016, opened 2024-07-03
squid - update to version 6.10

fritzkink:


AndWac:
Although configure searches and finds expat and libxml2 it is not being used?
checking for LIBEXPAT... yes
checking for expat.h... yes
checking for LIBXML2... yes
checking for libxml/parser.h... yes
checking for libxml/HTMLparser.h... yes
checking for libxml/HTMLtree.h... yes
configure: Enabling ESI processor:  -lexpat  -lxml2

I have some doubts. But if we keep it like this you also need to update pkg5.

PR #18050, opened 2024-07-07
rsyslog: update to 8.2406.0; switch to openssl 3.1

iigs:


AndWac:
I get the same error like the PR checker had:
/export/home/andreas/oi-userland/components/sysutils/rsyslog/build/manifest-i386-rsyslog.depend has unresolved dependency '
    depend type=require fmri=__TBD pkg.debug.depend.file=libmariadb.so.3 \
        pkg.debug.depend.reason=usr/lib/amd64/rsyslog/ommysql.so \
        pkg.debug.depend.type=elf \
        pkg.debug.depend.path=usr/gcc/13/lib/amd64 \
        pkg.debug.depend.path=usr/postgres/16/lib/amd64'.
make[1]: *** [/export/home/andreas/oi-userland/make-rules/ips.mk:523: /export/home/andreas/oi-userland/components/sysutils/rsyslog/build/.resolved-i386] Error 1
make[1]: Leaving directory '/export/home/andreas/oi-userland/components/sysutils/rsyslog'
gmake: *** [/export/home/andreas/oi-userland/make-rules/ips.mk:544: REQUIRED_PACKAGES] Error 2

iigs:
Please test the new Makefile that was changed in the recent force-push.  I think the reason why I did not run into this issue is that I have only one version of mariadb installed on the box where I made this package.  The change also includes references to mariadb and postgres no longer have version numbers hardcoded.

AndWac:
pkg5 needs an update

iigs:
pkg5 updated

PR #18052, opened 2024-07-08
nagios-plugins: bump COMPONENT_REVISION

mtelka:
This was forgotten in #18051 and it is needed to make sure the package is really updated on `pkg update` and also properly incorporated into `userland-incorporation`.


AndWac:
This has been superceded by #18053

PR #18054, opened 2024-07-08
squid -fix version 6.10

fritzkink:
in fact libxml and expat is not in use any longer


AndWac:
You have deletions (talloc) unrelated to squid in your commit!

PR #18055, opened 2024-07-08
squid -fix version 6.10

fritzkink:
libxml and expat still not needed


components/web/squid/Makefile
```diff
@@ -30,14 +30,15 @@ OPENSSL_VERSION=	3.1
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         squid
-COMPONENT_VERSION=      6.9
+COMPONENT_VERSION=      6.10
+COMPONENT_REVISION=     1
 COMPONENT_SUMMARY=      Squid Web Proxy Cache
```
AndWac, line 34: You don't need to set COMPONENT_REVISION as we haven't published 6.10 yet.

fritzkink:
libxml and expat are only needed when esi is enabled. I enabled esi now.

PR #18178, opened 2024-07-16
preserve new uid/gid for new system user freerad

fritzkink:


AndWac:
You should add this to your PR for freeradius integration.

PR #18255, opened 2024-07-24
python/types-setuptools: update to 71.1.0.20240724

mtelka:


AndWac:
May I ask of what value are updates of this kind? This seems to be a daily update of the same version number. In my eyes most of the python packages are not worth to keep in OI.

mtelka:
The `types-setuptools` package is needed to build and test `mypy` and to test `pytest-mypy-plugins`, `versioneer`.

The `mypy` package is test dependency for 29 other Python packages (complete list available on request).  The `pytest-mypy-plugins` package is needed to test 4 other Python packages (`attrs`, `consolekit`, `lxml-stubs`, `pyhamcrest`).  The `versioneer` package is needed to build the `constantly` package.

And this is just a beginning of the dependency forest.  I can continue and we will find at the end that something like `pkg` or so transitionally depends on `types-setuptools`.

If a Python package is not needed (and especially if it is deprecated) then I'd be first who would obsolete it from OI.  Unfortunately the Python ecosystem is very complex.  Far more complex than the Perl ecosystem.  It is also live and evolving ecosystem, while Perl is almost static and new package releases are rare there.

The value of the update itself is that it will make sure that I do not see unexpected failures when I test other Python packages (because they usually expects up to date dependants).  It is far easier to create a PR with the update than to try to analyze if the update is of any real value or not (especially for these `types-*` packages).

PR #18283, opened 2024-07-26
python/incremental: update to 24.7.1; obsolete package for Python 3.7

mtelka:


mtelka:
We need to postpone this update since `24.7.0` seems to cause serious regression: https://github.com/twisted/incremental/issues/100

PR #18323, opened 2024-07-30
ipmitool: fix oem enterprise numbers download url

cgrzemba:
the download url is wrong so the expected file contains html error text.


hoewweken:
Thank you! :-)

PR #18340, opened 2024-07-31
git: update to 2.46.0

AndWac:


mtelka:
We do have the `missing SANITY` problem here again.  Similarly as it was there [last time](https://github.com/OpenIndiana/oi-userland/pull/17378#issuecomment-2144017021).  @AndWac, could you please try to find why the following code (copied from `git-2.46.0/t/test-lib.sh`) thinks you have no `SANITY`?

Thank you.

```
# SANITY is about "can you correctly predict what the filesystem would
# do by only looking at the permission bits of the files and
# directories?"  A typical example of !SANITY is running the test
# suite as root, where a test may expect "chmod -r file && cat file"
# to fail because file is supposed to be unreadable after a successful
# chmod.  In an environment (i.e. combination of what filesystem is
# being used and who is running the tests) that lacks SANITY, you may
# be able to delete or create a file when the containing directory
# doesn't have write permissions, or access a file even if the
# containing directory doesn't have read or execute permissions.

test_lazy_prereq SANITY '
        mkdir SANETESTD.1 SANETESTD.2 &&

        chmod +w SANETESTD.1 SANETESTD.2 &&
        >SANETESTD.1/x 2>SANETESTD.2/x &&
        chmod -w SANETESTD.1 &&
        chmod -r SANETESTD.1/x &&
        chmod -rx SANETESTD.2 ||
        BUG "cannot prepare SANETESTD"

        ! test -r SANETESTD.1/x &&
        ! rm SANETESTD.1/x && ! test -f SANETESTD.2/x
        status=$?

        chmod +rwx SANETESTD.1 SANETESTD.2 &&
        rm -rf SANETESTD.1 SANETESTD.2 ||
        BUG "cannot clean SANETESTD"
        return $status
'
```

PR #18344, opened 2024-08-01
fix git test results

AndWac:


mtelka:
Thank you for this.

It fixes the problem mentioned [here](https://github.com/OpenIndiana/oi-userland/pull/18340#issuecomment-2261545972).

PR #18346, opened 2024-08-01
timescaledb: add postgresql extension timescaledb 2.16.0

cgrzemba:
This extension is a prerequisite for Zabbix server 7

### DIFF
--- a/components/database/postgresql-16-timescaledb/Makefile
+++ b/components/database/postgresql-16-timescaledb/Makefile
@@ -1,0 +1,41 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2024, Carsten Grzemba
+#
+BUILD_STYLE= justmake
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=	timescaledb
+COMPONENT_VERSION=	2.16.0
+COMPONENT_SUMMARY=	Timescaledb extension for PostgreSQL
+COMPONENT_SRC = $(COMPONENT_NAME)-$(COMPONENT_VERSION)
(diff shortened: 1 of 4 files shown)


components/database/postgresql-16-timescaledb/Makefile
```diff
@@ -0,0 +1,39 @@
...
+COMPONENT_ARCHIVE=	$(COMPONENT_VERSION).tar.gz
+COMPONENT_ARCHIVE_HASH= sha256:c68c12c3d62f2e3c46d277d2558c20e31d3826b84e15a8594d1084874a1ea9a4
+COMPONENT_ARCHIVE_URL=	https://github.com/timescale/timescaledb/archive/refs/tags/$(COMPONENT_ARCHIVE)
+COMPONENT_PROJECT_URL=	www.timescale.com
+COMPONENT_FMRI=	database/postgres-$(PG_VERNUM)/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	System/Databases
+COMPONENT_LICENSE=	Apache-2.0 TSL
+COMPONENT_LICENSE_FILE=	LICENSE
+
+include $(WS_MAKE_RULES)/common.mk
```
AndWac, line 32: TEST_TARGET= $(NO_TESTS)
is missing as gmake test fails with an error message.

components/database/postgresql-16-timescaledb/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+COMPONENT_BUILD_ACTION= cd $(@D); $(ENV) $(COMPONENT_BUILD_ENV) ./bootstrap -DUSE_OPENSSL=0 && cd build && gmake
+
```
AndWac, line 35: OpenSSL is explicitly disabled here. Is there a reason for this?

cgrzemba: yes, bootstrapping fails with

CMake Error at CMakeLists.txt:593 (message):
  PostgreSQL was built without OpenSSL support, which TimescaleDB needs for
  full compatibility.  Please rebuild PostgreSQL using `--with-openssl` or if
  you want to continue without OpenSSL, re-run bootstrap with
  `-DUSE_OPENSSL=0`
but it is not true and the extension works for me.

AndWac: Ok, you should probably add a comment with that information then.

PR #18347, opened 2024-08-02
timescaledb: workaround the bootsrap openssl check

cgrzemba:
small fix for Postgress Openssl check


components/database/postgresql-16-timescaledb/Makefile
```diff
@@ -31,7 +32,9 @@ COMPONENT_LICENSE_FILE=	LICENSE
 TEST_TARGET= $(NO_TESTS)
 include $(WS_MAKE_RULES)/common.mk
 
-COMPONENT_BUILD_ACTION= cd $(@D); $(ENV) $(COMPONENT_BUILD_ENV) ./bootstrap -DUSE_OPENSSL=0 && cd build && gmake
+COMPONENT_BUILD_ACTION= cd $(@D); \
+	$(ENV) $(COMPONENT_BUILD_ENV) ./bootstrap -DCMAKE_C_FLAGS=-I/usr/postgres/16/include && \
```
mtelka: Please use `PG_INCDIR` here instead of `/usr/postgres/16/include`.  BTW, why is `CMAKE_C_FLAGS` needed, but it was not needed earlier?

You also need to set `PG_VERSION` to `16` in this `Makefile`.

mtelka (changes requested):

components/database/postgresql-16-timescaledb/Makefile
cgrzemba: The CMAKE_C_FLAGS setting is only necessary for get the ugly OpenSSL check to work.

mtelka: So you basically traded `USE_OPENSSL=0` by `CMAKE_C_FLAGS=-I$(PG_INCDIR)`.  From my point of view it looks similar, but I've no objection for the change either.

The addition of `PG_VERSION` is crucial here (it is needed even without the trade).

Thank you.

mtelka: Or, does it mean that with the `CMAKE_C_FLAGS=-I$(PG_INCDIR)` we will find (and use) OpenSSL?

PR #18416, opened 2024-08-11
pari-elldata: J.E. Cremona Elliptic Curve Data for PARI/GP

cstes:
To test

 pkg install -v pari pari-elldata

Then run:

   gp

In gp run:

   ellinit("11a1") 

this will fail if the package pari-elldata is not installed, but it will work if the new package pari-elldata is installed at the IPS level for working with elliptic curve data.
`


AndWac:
If pari is needed in order to use this package you should add pari as a runtime dependency on it.

AndWac:
You have added a build dependency (which also creates a runtime dependency on it).
If you want to add an explicit runtime dependency you'll do that in the manifest, with a line like this:
depend type=require fmri=library/math/pari

You can find lots of examples in other manifests by searching for them:
find . -name "*.p5m" -exec grep -- depend  {} \; -print

cstes:
ok, I'll remove the build dependency on pari from the component Makefile, and add a runtime dependency on pari in the manifest ...

PR #18430, opened 2024-08-12
graphviz: update to 12.1.0

AndWac:


components/image/graphviz/graphviz.p5m
```diff
@@ -44,7 +44,9 @@ file path=usr/bin/diffimg
 file path=usr/bin/dijkstra
 file path=usr/bin/dot
 link path=usr/bin/dot2gxl target=gxl2gv
-file path=usr/bin/dot_builtins
+# Prevent missing dependency
+#   depend type=require fmri=__TBD pkg.debug.depend.file=libgvplugin_vt.so.6
```
mtelka: I assume the `libgvplugin_vt.so.6` file is created during the `graphviz` build and so it should be added to a package.

components/image/graphviz/graphviz.p5m
```diff
@@ -44,7 +44,9 @@ file path=usr/bin/diffimg
...
+file path=usr/bin/dot_builtins pkg.depend.bypass-generate=.*
 file path=usr/bin/edgepaint
 link path=usr/bin/fdp target=dot
 file path=usr/bin/gc
```
mtelka: None of `runtime/go-8`, `runtime/guile-8`, and `runtime/php-8` packages exist so the conditional dependencies at the end of the file have unusable predicate.  Please fix this issue too while working on `graphviz`.

Thank you.

PS: The issue was found using the [oi-pkg-checker](https://github.com/aueam/oi-pkg-checker).

components/image/graphviz/graphviz.p5m
AndWac: Yes, the file is part of graphviz but gmake sample-manifest doesn't include it.

AndWac: Hm, it looks like it is already there. I need to investigate further...

mtelka: It is in the sample manifest, but not in a real manifest used to publish the component.

mtelka:
```
$ find components/image/graphviz/ -name '*.p5m' | xargs grep _vt
components/image/graphviz/manifests/sample-manifest.p5m:link path=usr/lib/$(MACH64)/graphviz/libgvplugin_vt.so \
components/image/graphviz/manifests/sample-manifest.p5m:    target=libgvplugin_vt.so.6.0.0
components/image/graphviz/manifests/sample-manifest.p5m:link path=usr/lib/$(MACH64)/graphviz/libgvplugin_vt.so.6 \
components/image/graphviz/manifests/sample-manifest.p5m:    target=libgvplugin_vt.so.6.0.0
components/image/graphviz/manifests/sample-manifest.p5m:file path=usr/lib/$(MACH64)/graphviz/libgvplugin_vt.so.6.0.0
$
```

AndWac: Yes, I noticed it. There is another plugin library missing which I have added now.

components/image/graphviz/history
```diff
@@ -5,3 +5,4 @@ image/graphviz/graphviz-python-27@2.40.1-2020.0.1.6
 image/graphviz/graphviz-sharp@2.40.1-2020.0.1.6
 image/graphviz/graphviz-ocaml@7.1.0-2023.0.0.2
 image/graphviz/graphviz-perl-536@7.1.0-2023.0.0.2
+image/graphviz/graphviz-perl-538@12.0.0-2024.0.0.1
```
mtelka: Please add `,5.11` here.

AndWac: Why? It hasn't been done for the other obsoletions, even by you.

mtelka: Huh, really?  That's a mistake.  Where I forgot to add it?

mtelka: Ah, you mean in this file.  Yes, right.  Mistake done more than 2.5 years ago.  We should avoid similar mistake where possible.

AndWac: We had at least one discussion whether it is needed or not. As far as I remember the oucome was that it is not needed. Even "pkg info" doesn't show it.

mtelka: Yes, it is not needed now, but it could be needed in future (although the probability is very low).  Also, it is used in almost every `history` file so for consistency it is better to add it.

... and thank you for fixing this minor issue :-).

PR #18631, opened 2024-08-30
sbcl: version 2.4.7 introduced a bug; revert to 2.4.6 with fake 2.4.8…

AndWac:
… version number


AndWac:
Ok, sbcl needs a newer clx version and 
cd ~/quicklisp/local-projects/
git clone https://github.com/sharplispers/clx.git

did the trick

PR #18789, opened 2024-09-12
wireshark: update to 4.4.0

iigs:
Update to qt6.  
Reformat all p5m files to better match standard layout.


AndWac:
The pr-checker is right. If nghttp3 is installed (which it is on the build server) it is detected and being used by wireshark.

iigs:
added nghttp3 dependency.

iigs:
Added minizip-ng dependency.  I looked over pr-checker and reviewed the libraries it couldn't find.  It does not look like that OI has any of those.  Please do not hesitate to reach out to me if any more dependencies got missed.

PR #18824, opened 2024-09-16
libxml2: update to 2.13.3

mtelka:


mtelka:
For reference: the support for libxml2 2.13 has been [integrated to illumos-gate about three months ago](https://www.illumos.org/issues/16612#note-3).

PR #18958, opened 2024-09-26
libxml2: move 32 bit support to own package

mtelka:


mtelka:
This change caused the `illumos-gate` component build failure.  It should be solved via https://www.illumos.org/issues/16820 soon.

mtelka:
Discussion about the 32-bit packaging split with some details related to `libxml2` is here: https://github.com/OpenIndiana/oi-userland/pull/18982#issuecomment-2383359309

PR #18982, opened 2024-09-30
libjpeg-turbo: build and packaging cleanups; move 32 bit support to own packages

mtelka:
This is designed to be fully compatible with the previous `libjpeg-turbo` packages.  IOW, all `libjpeg` consumers (both existing packages and building of components in oi-userland) should produce same or possibly slightly better results than before.

This requires cmake component-based install from #18981.


Toasterson:
Can you elaborate on the rationale of having dedicated 32bit packages? This can break pkgdepend resolve (has already with libxml2) and increases the dependency tree unnecessarily. To slim down the filesystem we should use facets rather than seperate packages.

mtelka:
> Can you elaborate on the rationale of having dedicated 32bit packages?

The main idea behind `*/32` packages is to find what 32-bit support is really needed.  Once we found there is an unused `/32` package then we can obsolete it without breaking existing packages.

> This can break pkgdepend resolve (has already with libxml2)

Could you please elaborate about this?  I'm not aware about any breakage.

> increases the dependency tree unnecessarily.

I do not understand this.  What exactly is your concern here?

We currently do have these `*/32` packages:

1. `library/icu-74/32`
   - not needed by any other package
1. `library/icu-75/32`
   - needed by `library/libxml2/32` (and `developer/icu`, obviously)
1. `library/libxml2/32`
   - needed by `library/libxml2` (obviously)
1. `library/readline-8/32`
   - needed by `library/libxml2/32` (and `library/readline`, obviously)

In what sense is this a big dependency tree?  Especially when compared to the rest of the dependency tree we do already have.

> To slim down the filesystem we should use facets rather than seperate packages.

This would be the right thing to do when we would opt to fully (as much as possible) support 32-bit.  Instead, we are going different (opposite) way: torch as much as possible of 32-bit.  (BTW, I do agree with this goal.)

We are currently doing so using a way that could cause breakages of existing packages:  We remove 32-bit support when we either _believe_ (or we are more or less _sure_) we do not need it, or we cannot build it because some dependencies already dropped 32-bit support.  This can somehow work for non-essential packages (like desktop applications, for example), but it is very dangerous for critical pieces of the OS.  I consider all `icu`, `libxml2`, `readline` (and `libjpeg` too) as critical components of the system and I think it is better to provide 32-bit variants of them (even they are maybe not needed) than to drop them and then be surprised that something critical does not work.  OTOH, we would still like to remove 32-bit variants of them if they are not needed.  I believe the easiest way how to find if something needs them without breaking anything is to create `*/32` packages and wait until their dependants are rebuilt.

Thank you.

AndWac:
Publishing illumos-gate is broken at the moment.

mtelka:
> Publishing illumos-gate is broken at the moment.

Is this related to `libxml2`?

mtelka:
> > Publishing illumos-gate is broken at the moment.
> 
> Is this related to `libxml2`?

Hm, possibly yes.  I just started the `illumos-gate` build here to see and analyze.

BTW, when there is no feedback that I possibly broke something then there is a chance that I miss the breakage and so I won't fix it (unless I notice myself, which didn't happen in this case).

Anyway, I'm very sorry for the breakage and if it is caused by my recent `libxml2` change (I'll find in few hours) then I'll fix it ASAP.

AndWac:
I am not sure but the illumos-gate build is broken for some days and 32 bit libxml2 is a dependency of it.

mtelka:
> I am not sure but the illumos-gate build is broken for some days and 32 bit libxml2 is a dependency of it.

I filled https://www.illumos.org/issues/16820 to solve the `illumos-gate` build issue.

mtelka:
> > To slim down the filesystem we should use facets rather than seperate packages.
> 
> This would be the right thing to do when we would opt to fully (as much as possible) support 32-bit. Instead, we are going different (opposite) way: torch as much as possible of 32-bit. (BTW, I do agree with this goal.)

Facets could possibly work for slim down the filesystem (which is not a goal with the 32-bit split out) only in a case when every package would have its own facet to disable/uninstall the 32-bit (but it would be a lot of work anyway to achieve that).

Currently it won't be possible to use one common facet for all packages (for example, `system.32bit` facet or so, similarly to current `doc.man`).  The reason is that some packages currently contains 32-bit only bits, for example `SUNWcs`, and so we cannot turn off 32-bits for them.

So let's imagine there is `system.32bit` facet and `libxml2` is modified to use it for 32-bit, and `SUNWcs` is kept unmodified.  If we would set `system.32bit` to `false` to uninstall 32-bit `libxml2` then binaries from `SUNWcs` would stop to work because of missing 32-bit `libxml2`.

If we would modify `SUNWcs` to disable their 32-bit bits on `system.32bit` facet, then with `system.32bit=false` we would miss a lot of critical binaries making the OS unusable.  To make such a facet to work properly we would need to convert `SUNWcs` first to produce 64-bit variants for all its bits (a lot of work).

I'm very sorry I see no reasonable way how to achieve the main goal using facets with the current OpenIndiana and illumos-gate state.

Toasterson:
I was unaware of the main goal of the project and nobody else could answer that. As far as I am aware we where already down to a small enough portion of 32bit packages that only illumos-gate and SunRay are still left. Although a Accounting of the situation would be helpful.

Other parts of my comment are not related to the goal of tracking 32bit dependencies so please Ignore.

There was some talk in illumos-gate about the some of it but I am open to temporarily carry a patch for exception_lists for this project. 

Please note that until we have something to properly replace SunRay with we would need to find a way to provide the needed 32bit dependencies for it. SunRay has become rather popular with OpenIndiana Contributors for Family home setups and small retro fan setups. Even though alternative Remote desktop setups exist and it has become rather tedious and complex to setup, once setup it is still the smoothest remote desktop experience out there (according to reports). Contributors described it as wanted feature.

Efforts for 64bit only illumos-gate are ongoing inside the ARM branch as that Architecture has no 32bit support (with our kernel) and updating the binaries to be 64bit is less work.

tsoome:
> I was unaware of the main goal of the project and nobody else could answer that. As far as I am aware we where already down to a small enough portion of 32bit packages that only illumos-gate and SunRay are still left. Although a Accounting of the situation would be helpful.
> 
> Other parts of my comment are not related to the goal of tracking 32bit dependencies so please Ignore.
> 
> There was some talk in illumos-gate about the some of it but I am open to temporarily carry a patch for exception_lists for this project.
> 
> Please note that until we have something to properly replace SunRay with we would need to find a way to provide the needed 32bit dependencies for it. SunRay has become rather popular with OpenIndiana Contributors for Family home setups and small retro fan setups. Even though alternative Remote desktop setups exist and it has become rather tedious and complex to setup, once setup it is still the smoothest remote desktop experience out there (according to reports). Contributors described it as wanted feature.
> 
> Efforts for 64bit only illumos-gate are ongoing inside the ARM branch as that Architecture has no 32bit support (with our kernel) and updating the binaries to be 64bit is less work.

It is not only about arm64, the actual issue is wrap over of 32-bit time_t.

mtelka:
> I was unaware of the main goal of the project and nobody else could answer that. As far as I am aware we where already down to a small enough portion of 32bit packages that only illumos-gate and SunRay are still left. Although a Accounting of the situation would be helpful.

I do not know if the 32-bit removal is somewhere stated as the main goal of the project.  I doubt, especially the word _main_ sounds suspicious there. :-)  Anyway, the current state is that 32-bit is dropped everywhere where possible.

> Other parts of my comment are not related to the goal of tracking 32bit dependencies so please Ignore.
> 
> There was some talk in illumos-gate about the some of it but I am open to temporarily carry a patch for exception_lists for this project.

A temporary patch for `exception_lists` would make the `illumos-gate` component to become buildable again, but we still need the upstream solution for folks that would want to build the vanilla `illumos-gate` on OpenIndiana.

> 
> Please note that until we have something to properly replace SunRay with we would need to find a way to provide the needed 32bit dependencies for it. SunRay has become rather popular with OpenIndiana Contributors for Family home setups and small retro fan setups. Even though alternative Remote desktop setups exist and it has become rather tedious and complex to setup, once setup it is still the smoothest remote desktop experience out there (according to reports). Contributors described it as wanted feature.

The 32-bit package split should not affect SunRay.  At least I'm not aware about any issue related.  Actually, it will probably make the SunRay support more explicit, visible and clear.  Especially when somebody create a [metapackage for it](https://github.com/OpenIndiana/oi-userland/pull/15150#issuecomment-1880019898).

> Efforts for 64bit only illumos-gate are ongoing inside the ARM branch as that Architecture has no 32bit support (with our kernel) and updating the binaries to be 64bit is less work.

Yes, sure.  But I believe this is orthogonal to the 32-bit packaging split.

DrLou:
Any chance all of this is related to (recent) non-delivery of some 64-bit objects in postgres-16? Eg, pkg does not install plpython, others...

mtelka:
> Any chance all of this is related to (recent) non-delivery of some 64-bit objects in postgres-16? Eg, pkg does not install plpython, others...

No, as far as I know.

PR #18984, opened 2024-09-30
shared-macros.mk: set CLANG_VERSION conditionally

mtelka:
This will allow to set `CLANG_VERSION` even before the `shared-macros.mk` include.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -696,7 +696,7 @@ LD =		/usr/bin/ld
 
 # Clang definitions (we only have 64 bit clang)
 CLANG_DEFAULT =		18
-CLANG_VERSION =		$(CLANG_DEFAULT)
+CLANG_VERSION ?=	$(CLANG_DEFAULT)
 CLANG_FULL_VERSION =	$(CLANG_VERSION).1
 CLANG_PREFIX             = /usr/clang/$(CLANG_FULL_VERSION)
 CLANG_BINDIR =		$(CLANG_PREFIX)/bin


make-rules/shared-macros.mk
```diff
@@ -696,7 +696,7 @@ LD =		/usr/bin/ld
 
 # Clang definitions (we only have 64 bit clang)
 CLANG_DEFAULT =		18
-CLANG_VERSION =		$(CLANG_DEFAULT)
+CLANG_VERSION ?=	$(CLANG_DEFAULT)
 CLANG_FULL_VERSION =	$(CLANG_VERSION).1
 CLANG_PREFIX             = /usr/clang/$(CLANG_FULL_VERSION)
```
AndWac, line 701: CLANG_FULL_VERSION also needs ?=

mtelka: Why?

AndWac: It unconditionally appends .1
Either this or CLANG_PREFIX might need extra contol.

mtelka: > It unconditionally appends .1 Either this or CLANG_PREFIX might need extra contol.

I'm not sure I fully understand your concern, so let's try to explain:

Make variables works differently than you expect here.  I assume the problem you are referring to would be there only in a case the `CLANG_FULL_VERSION` is [Simply Expanded Variable](https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html).  In this case the `CLANG_FULL_VERSION` is [Recursively Expanded Variable](https://www.gnu.org/software/make/manual/html_node/Recursive-Assignment.html) and so I see no reason to use `?=` there.  The same is true for `CLANG_PREFIX`.

If you still thinks there is a problem then could you please elaborate in detail what exactly would be the scenario that would not work properly or as expected?

Thank you.

AndWac: How can you achieve that CLANG_PREFIX will be eg. /usr/clang/18.2?

mtelka: I see.

First, we should look at what `clang` packages we do have:
```
$ pkg list -a | grep clang | grep -v 'o$'
developer/clang-13                                13.0.1-2023.0.0.1          i--
developer/clang-15                                15.0.7-2023.0.0.1          i--
developer/clang-17                                17.0.6-2023.0.0.0          i--
developer/clang-18                                18.1.8-2024.0.0.0          i--
runtime/clang-13                                  13.0.1-2023.0.0.1          i--
runtime/clang-15                                  15.0.7-2023.0.0.1          i--
runtime/clang-17                                  17.0.6-2023.0.0.0          i--
runtime/clang-18                                  18.1.8-2024.0.0.0          i--
$
```
and where they are installed:
```
$ ls /usr/clang/
13.0  15.0  17.0  18.1
$
```
This brings a question: Why we do see a difference between the version number in the package name (major version only) and the actual disk location (major + minor)?

It looks like [their current release schedule](https://www.llvm.org/docs/HowToReleaseLLVM.html) expects only one minor version for every major version (currently `1`, in past it was `0`).  Last time when they [released](https://releases.llvm.org/) two minor versions for one major version was `11.0.1` released in January 2021 and `11.1.0` released in February 2021.  So it is reasonable to expect that the minor number have no significance and we could simply ignore it.

In terms of what we do have in `shared-macros.mk`, we could (as a quick fix) do some mapping, e.g. if `CLANG_VERSION` is `18`, then `CLANG_FULL_VERSION` should be `$(CLANG_VERSION).1`, if it is `17` or less, then `CLANG_FULL_VERSION` should be `$(CLANG_VERSION).0`.  With the current `CLANG_FULL_VERSION` the `CLANG_VERSION` set in component's Makefile won't work properly.  For example, I expect that the building of `qtcreator` will fail (it sets `CLANG_VERSION` to `13`).  Please note that the issue with `CLANG_FULL_VERSION` is not related to this PR.

However, what I propose above (and what you were apparently [talking about](https://github.com/OpenIndiana/oi-userland/pull/18984#discussion_r1781443025) too) is very likely not the right way to solve the issue.  I believe the better way would be to simply rename `/usr/clang/MAJOR.X` to `/usr/clang/MAJOR`.  Then we could simply remove `CLANG_FULL_VERSION`.  I started to work on this change for Clang 18.

mtelka: Further analysis shows that they apparently started to consider `18.1` as the (let say) major version, while for versions <= 17 it was just the single number, for example `17`:

```
$ ls -l /usr/clang/18.1/lib/ | grep '\.18'
lrwxrwxrwx   1 root     root          20 Mar 21  2024 libclang-cpp.so -> libclang-cpp.so.18.1
-r-xr-xr-x   1 root     bin      108165120 Jun 23 21:42 libclang-cpp.so.18.1
lrwxrwxrwx   1 root     root          16 Mar 21  2024 libclang.so -> libclang.so.18.1
lrwxrwxrwx   1 root     root          18 Jun 23 21:42 libclang.so.18.1 -> libclang.so.18.1.8
-r-xr-xr-x   1 root     bin      69402800 Jun 23 21:41 libclang.so.18.1.8
lrwxrwxrwx   1 root     root          15 Mar 21  2024 libLLVM-18.so -> libLLVM.so.18.1
lrwxrwxrwx   1 root     root          15 Mar 21  2024 libLLVM.so -> libLLVM.so.18.1
-r-xr-xr-x   1 root     bin      180313328 Jun 23 21:42 libLLVM.so.18.1
lrwxrwxrwx   1 root     root          14 Mar 21  2024 libLTO.so -> libLTO.so.18.1
-r-xr-xr-x   1 root     bin       114192 Jun 23 21:42 libLTO.so.18.1
lrwxrwxrwx   1 root     root          18 Mar 21  2024 libRemarks.so -> libRemarks.so.18.1
-r-xr-xr-x   1 root     bin         8360 Jun 23 21:42 libRemarks.so.18.1
$
```

Here is version `17`:
```
$ ls -l /usr/clang/17.0/lib/ | grep '\.17'
lrwxrwxrwx   1 root     root          18 Feb 20  2024 libclang-cpp.so -> libclang-cpp.so.17
-r-xr-xr-x   1 root     bin      103293768 Feb 20  2024 libclang-cpp.so.17
lrwxrwxrwx   1 root     root          14 Feb 20  2024 libclang.so -> libclang.so.17
-r-xr-xr-x   1 root     bin      66902984 Feb 20  2024 libclang.so.17
lrwxrwxrwx   1 root     root          12 Feb 20  2024 libLTO.so -> libLTO.so.17
-r-xr-xr-x   1 root     bin       114192 Feb 20  2024 libLTO.so.17
lrwxrwxrwx   1 root     root          16 Feb 20  2024 libRemarks.so -> libRemarks.so.17
-r-xr-xr-x   1 root     bin         8360 Feb 20  2024 libRemarks.so.17
$
```
And here version `13`:
```
$ ls -l /usr/clang/13.0/lib/ | grep '\.13'
lrwxrwxrwx   1 root     root          18 Feb 20  2024 libclang-cpp.so -> libclang-cpp.so.13
-r-xr-xr-x   1 root     bin      85387872 Feb 20  2024 libclang-cpp.so.13
lrwxrwxrwx   1 root     root          14 Feb 20  2024 libclang.so -> libclang.so.13
-r-xr-xr-x   1 root     bin      57808152 Feb 20  2024 libclang.so.13
lrwxrwxrwx   1 root     root          12 Feb 20  2024 libLTO.so -> libLTO.so.13
-r-xr-xr-x   1 root     bin       101808 Feb 20  2024 libLTO.so.13
lrwxrwxrwx   1 root     root          16 Feb 20  2024 libRemarks.so -> libRemarks.so.13
-r-xr-xr-x   1 root     bin        13800 Feb 20  2024 libRemarks.so.13
$
```

mtelka: > For example, I expect that the building of `qtcreator` will fail (it sets `CLANG_VERSION` to `13`). Please note that the issue with `CLANG_FULL_VERSION` is not related to this PR.

Interestingly, the `qtcreator` build does not fail.  Since its clang configuration is wrong (it needs clang 13, but `CLANG_PREFIX` points to `/usr/clang/13.1`, which does not exist), the build fails to detect clang at all and uses gcc to build `qtcreator` instead. :-)

But again, this is `CLANG_PREFIX` misconfiguration issue that is not related to this PR.

I still believe components should not set `CLANG_FULL_VERSION` and/or `CLANG_PREFIX` and so they do not need `?=`.  The `CLANG_FULL_VERSION`/`CLANG_PREFIX` misconfiguration for `CLANG_VERSION` < 18 is a separate issue (not related to this PR) and should be handled separately.  I'm looking at it.

@AndWac, please merge this.  Thank you.

PR #18997, opened 2024-10-01
firefox: update to 131.0

iigs:
Dropping firefox-bookmarks package since the way it distributes firefox bookmarks hasn't worked for new profiles in years. Please review the history file for accuracy.


components/web/firefox/history
```diff
@@ -0,0 +1 @@
+web/data/firefox-bookmarks@130.0.1,5.11-2024.0.0.1 web/data/firefox-bookmarks
```
AndWac: Please remove the trailing web/data/firefox-bookmarks
The 2nd fmri can optionally be used and defines it as a replacement for the first one. That is, it will automatically use the 2nd fmri in a pkg update to replace the first one.

iigs: Removed trailing package name. Thank you for checking this.

PR #19010, opened 2024-10-03
clang-18: move to /usr/clang/18

mtelka:
This solves the issue with `CLANG_FULL_VERSION` discussed in #18984 for Clang 18.  Please note that building with forced Clang 13 will be still broken, but I plan to work on the fix for it once this PR is merged.  I do not plan to fix this for Clang 15 (because of #18996) and for Clang 17 (because it should be obsoleted too for the same reason as Clang 15).

To make the future updates (and integrations of new Clang major versions) easier I replaced the Clang major version in manifests by `$(CLANGVER)` and did few cleanups in the `Makefile` too.  I also enabled parallel run of tests to speed the test run.


AndWac:
Note that #18996 has a merge conflict!

mtelka:
> Note that #18996 has a merge conflict!

Thanks for the notice.  I missed this.

Anyway, wouldn't it make more sense to add such a comment directly to #18996? :-)

PR #19011, opened 2024-10-03
openssh: update to 9.9p1; use libexecdir and move sshd to usr/sbin;

AndWac:
synchonize with OmniOS patches


AndWac:
Alas sshd doesn't listen on port 22 for IPv4 but only for IPv6.

Toasterson:
It's the way ListenAddress is configured in sshd_config Either both need to be commented out or not. But if only one is commented out ssh will only listen on that IP stack.

PR #19066, opened 2024-10-06
librsvg: revoke 2.59.1 update and publish 2.58.2 with a fake version …

AndWac:
…number

Alas the tests cannot be build with rust > 1.80


mtelka:
What exactly is the problem with `librsvg`?  I noticed few strange PRs recently related to it: #19042, #19056 and this one...

AndWac:
If you build gimp now, you won't get .svg files anymore because "Vector icons:              no (librsvg GdkPixbuf loader missing)"
If you run gtk4 tests now, you will get additional 70 failing tests.

Reverting librsvg to 2.58.2 will fix the problem with gimp but not the problem with gtk4.
I am actually testing to go back to rust 1.80.0 to find out whether the gtk4 problems are related to the new rust version.

iigs:
I was asked to comment on this PR based on a rust issue I have been looking at over the past few days.  The history is found in https://bugzilla.mozilla.org/show_bug.cgi?id=1795899 .  The three comments in the mozilla ticket of note are comment 30,  31, and 33.
Commnet 30: there are multiple cstddefs on the system (one from clang, one from gcc), and which one goes first makes it work or not.
Comment 31: So the difference is that in the "bad" case, the size_t in size_t len is Unexposed, rather than Typedef... So bindgen tries to go across the unexposed by looking at the canonical type...
Comment 33: patches for clang and rust's bindgen crate.

Even though this problem with Firefox manifests during compilation, are there other issues with clang and gcc header file differences that can manifest other types of problems?

mtelka:
> If you build gimp now, you won't get .svg files anymore because "Vector icons: no (librsvg GdkPixbuf loader missing)" If you run gtk4 tests now, you will get additional 70 failing tests.
> 
> Reverting librsvg to 2.58.2 will fix the problem with gimp but not the problem with gtk4. I am actually testing to go back to rust 1.80.0 to find out whether the gtk4 problems are related to the new rust version.

Rust was updated from 1.80.0 to 1.81.0 (#18714) before the gtk4 update from 4.10.5 to 4.16.1 (#18838) so gtk4 tests worked okay with rust 1.81.0 already.  I feel the possible cause for the gtk4 tests regression could be the recent  gdk-pixbuf update (#19034).

AndWac:
Yes, with rust 1.80.0 I get the same errors.

mtelka:
I downgraded few packages so now I'm at this:
```
# pkg list -u
image/library/librsvg                             2.58.2-2024.0.0.0          i--
library/desktop/gdk-pixbuf                        2.36.12-2023.0.0.6         i--
library/desktop/gtk2                              2.24.33-2023.0.0.1         i--
library/desktop/gtk2/gtk-backend-cups             2.24.33-2023.0.0.1         i--
library/desktop/libgsf                            1.14.52-2024.0.0.0         i--
#
```
and then I ran `gtk4` tests:
```
Ok:                 4585
Expected Fail:      0
Fail:               248
Unexpected Pass:    0
Skipped:            5
Timeout:            2
```
This (more or less) matches the test results in the git.

mtelka:
... then I updated the `library/desktop/gdk-pixbuf` package to `2.42.12-2024.0.0.0` and now the `gtk4` test results are:
```
Ok:                 4512
Expected Fail:      0
Fail:               322
Unexpected Pass:    0
Skipped:            5
Timeout:            1
```
So yes, the updated `gdk-pixbuf` is causing problems.  I'm looking at it.

mtelka:
There are three sets of newly failing `gtk4` tests: `texture-scale-stripes`, `texture-url`, `scale-textures-negative-ngl` (72 tests total).  All of them are failing with `ERROR   killed by signal 5 SIGTRAP`.

mtelka:
And the cause is that the `png` file format is not recognized.

I noticed some issue with `png` when I worked on the `gdk-pixbuf` update: https://github.com/OpenIndiana/oi-userland/blob/d9b03893c7111b1eec4677ac207e138016f7bb5b/components/library/gdk-pixbuf/Makefile#L43-L45
But I thought that making the `png` support builtin solves the problem.  It apparently didn't :-(.

Back to the drawing board to find what's the problem with `png` in `gdk-pixbuf`...

mtelka:
#19067 should add back the `png` support for `gdk-pixbuf` and so it should fix the problem with `gtk4` tests.

AndWac:
Gimp is still not able to create its .svg images and falls back to .png. I will merge this PR and hope we'll find a solution.

PR #19077, opened 2024-10-07
gtk4: update to 4.16.3

AndWac:


mtelka:
This is not `gtk+4`, but `gtk4`.

mtelka:
... see https://download.gnome.org/sources/gtk+/ vs. https://download.gnome.org/sources/gtk/

PR #19137, opened 2024-10-10
glib2: add a charset-alias package for 32 bit

AndWac:


mtelka:
Hold on please.

mtelka:
This is mistake in the `glib-32.p5m` file.  The solution is _not_ to add the `library/glib2/32/charset-alias` package.  I'm working on proper fix.

mtelka:
#19138 would be a bit better (and simpler) than this.

PR #19126, opened 2024-10-10
glib: add documentation; move 32 bit support to separate package

mtelka:

PR #19138, opened 2024-10-10
glib: fix charset-alias dependency for 32-bit package

mtelka:
This is just a first aid fix for [the issue I introduced recently](https://github.com/OpenIndiana/oi-userland/pull/19126#issuecomment-2405979921).  The proper (and more complex) solution will follow shortly.


mtelka:
The proper way how to solve this issue is via #19139 and #19139.

PR #19189, opened 2024-10-14
geeqie: update to 2.5

iigs:


iigs:
I see the check failed.  I saw this on one system--not the first system which I used to build this.  Is there a problem with g++ mediation?

PR #19202, opened 2024-10-15
shared-macros.mk: install 32/64-bit in desired order

mtelka:
I think it is sometimes silently expected that 64-bit is installed _after_ 32-bit when `BUILD_BITS` is `64_and_32` (and vice versa for `32_and_64`).  This is to make sure it is really like that.


mtelka:
It is apparently not so _silently_ because for example `vte-291` does this explicitly here: https://github.com/OpenIndiana/oi-userland/blob/459554e9d008a14f35fa4c6bf7e4873eca280a1a/components/library/vte-291/Makefile#L73

PR #19218, opened 2024-10-15
libexpat: move 32 bit support to separate package

mtelka:


mtelka:
rebase only

mtelka:
Added patch for `illumos-gate/exception_lists/packaging.deps` to make sure the build of the `illumos-gate` component succeeds.  The patch will be upstreamed later together with other similar changes in `packaging.deps`.

PR #19219, opened 2024-10-15
yaml-cpp: update to 0.8.0

AndWac:


components/library/yaml-cpp/yaml-cpp.p5m
```diff
@@ -107,24 +60,12 @@ file path=usr/include/yaml-cpp/parser.h
...
-file path=usr/lib/$(MACH64)/libgtest_main.so
-link path=usr/lib/$(MACH64)/libyaml-cpp.so target=libyaml-cpp.so.0.7
-file path=usr/lib/$(MACH64)/libyaml-cpp.so.$(COMPONENT_VERSION)
-link path=usr/lib/$(MACH64)/libyaml-cpp.so.0.7 \
-    target=libyaml-cpp.so.$(COMPONENT_VERSION)
-file path=usr/lib/$(MACH64)/pkgconfig/gmock.pc
-file path=usr/lib/$(MACH64)/pkgconfig/gmock_main.pc
-file path=usr/lib/$(MACH64)/pkgconfig/gtest.pc
-file path=usr/lib/$(MACH64)/pkgconfig/gtest_main.pc
+link path=usr/lib/$(MACH64)/libyaml-cpp.so target=libyaml-cpp.so.0.8
```
mtelka, line 67: Since the library version changed it will probably cause that its consumers will stop to work (`qtcreator` and `opencolorio`).

mtelka: I think the best way how to solve this is to do something similar here as I did with gspell in #19260.

PR #19222, opened 2024-10-15
mozilla-nspr: obsolete header-nspr; prefer 64 bit; move 32 bit support to separate package

mtelka:
The `library/nspr/header-nspr` package currently contains an (incomplete) set of developer files (includes and the `nspr-config` script).  I see no reason to have such files living separately so I merged all files from `library/nspr/header-nspr` to the main `library/nspr` package.

Previously, the `/usr/bin/nspr-config` script was for 32 bit.  Now the 64-bit is preferred so the `/usr/bin/nspr-config` script is for 64 bit while the 32-bit one is in `/usr/bin/i86/nspr-config`.

Finally, the 32-bit related stuff was split out to new `library/nspr/32` package.


mtelka:
This is blocked by #19221.

AndWac:
#19221 is already merged

mtelka:
> #19221 is already merged

Thank you.  I need to refresh, update (add illumos-gate patch) and re-test this one.  Should be done today.

mtelka:
rebase only

mtelka:
Added patch for `illumos-gate/exception_lists/packaging.deps` to make sure the build of the `illumos-gate` component succeeds.  The patch will be upstreamed later together with other similar changes in `packaging.deps`.

PR #19242, opened 2024-10-17
pari: enable tls (thread local stack)

cstes:
Remove patch for "out of source tree" build because, although the out of source tree build seems not to work, the build succeeds after setting up links, even without patch.

Enable TLS which in this context stands for "thread local stack" (not for transport layer security).  This is enable by setting the --mt=pthread option


components/scientific/pari/pari.p5m
```diff
@@ -1,19 +1,3 @@
...
-# A full copy of the text of the CDDL should have accompanied this
-# source.  A copy of the CDDL is also available via the Internet at
-# http://www.illumos.org/license/CDDL.
-#
-
-#
-# Copyright (c) 2021,2022 David Stes
-# Copyright (c) 2023 Andreas Wacknitz
-# Copyright (c) 2024 David Stes
-#
```
AndWac, line 1: The license information here is for exactly this manifest file and has nothing to do with the license of pari itself.
As its contents is generated and no additional contents it is ok for me but I suspect that the PARI team is misled here.

cstes: I am OK with the request and I hope that the PARI team continues to support OpenIndiana.   They can also reasonably ask for some options to be set in the Makefile or the manifest.

PR #19256, opened 2024-10-18
clang-18: fix bug D140075

iigs:
The problem is that there are multiple cstddefs on the system (one from clang, one from gcc). The wrong one gets used by a rust program, which effect firefox builds using gcc.  Firefox will not build with clang without a bunch of work. The area of code which is affected is used for debugging.  The code has to be compiled, even if it doesn't get used under normal firefox builds.   I am using an ugly workaround to bypass this problem https://github.com/OpenIndiana/oi-userland/blob/oi/hipster/components/web/firefox/patches/07-Regression-in-Bug1869835-profiler-rust-typecast-error.patch -- this firefox patch can be dropped with this clang patch

Building firefox with debug support on OpenIndiana has been broken since late December 2023.  The problem is found in Firefox bug report https://bugzilla.mozilla.org/show_bug.cgi?id=1795899 (reported by Oracle in October 2022).  Mozilla developed a solution which requires a patch to clang and to bindgen (at least the copy used by Firefox). Based on this bug report, Oracle expressed interest in having the clang patch integrated.  I can confirm this fixes debug build in Firefox 124.  Since that time, Mozilla moved code around and debug builds needs some fixes to work again. It will help to have this clang patch integrated in OI.  Troubleshooting firefox problems without a working debug is very slow.

### DIFF
--- a/components/developer/clang-18/Makefile
+++ b/components/developer/clang-18/Makefile
@@ -24,7 +24,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		clang
 COMPONENT_VERSION=	18.1.8
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 CLANG_VERSION=		$(firstword $(subst ., ,$(COMPONENT_VERSION)))
 COMPONENT_SUMMARY=	LLVM tools and Clang compiler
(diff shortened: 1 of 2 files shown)


mtelka:
~@iigs, can we get a link to bug D140075 please?~ EDIT: it is in the patch.

PR #19263, opened 2024-10-19
pari: update test results with elldata

cstes:
Ran "gmake test" with optional package "elldata" installed.

Updated results file for the test results when elldata is installed.

Added my patch for genfunclist again, as initially contributed for PARI 2.13.   This is because I think that under some conditions the file src/funclist is regenerated and without my patch, the file src/funclist will be populated with an empty list of functions.

The patch was used in the past for PARI 2.13, 2.15 and 2.17 initial version.

### DIFF
--- a/components/scientific/pari/Makefile
+++ b/components/scientific/pari/Makefile
@@ -22,7 +22,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		pari
 COMPONENT_VERSION=	2.17.0
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_SUMMARY=	The PARI Computer Algebra System
 COMPONENT_PROJECT_URL=	https://pari.math.u-bordeaux.fr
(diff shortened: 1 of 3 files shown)


components/scientific/pari/Makefile
```diff
@@ -22,7 +22,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		pari
 COMPONENT_VERSION=	2.17.0
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_SUMMARY=	The PARI Computer Algebra System
 COMPONENT_PROJECT_URL=	https://pari.math.u-bordeaux.fr
 COMPONENT_SRC=		pari-$(COMPONENT_VERSION)
```
AndWac, line 28: If pari-elldata is being used for the test you need to add

\# Manually added dependencies
TEST_REQUIRED_PACKAGES += library/math/pari-elldata

With TEST_REQUIRED_PACKAGES our build system makes sure to have it installed when gmake test is run.

PR #19307, opened 2024-10-24
gcc-10: add patch for kernel printf for %j and %z size specifiers

AndWac:
tagged as gcc-10.4.0-il-2 on illumos/gcc


components/developer/gcc-10/gcc-10.p5m
```diff
@@ -2034,16 +2034,16 @@ file path=usr/gcc/10/share/gcc-$(HUMAN_VERSION)/python/libstdcxx/__init__.py
 file path=usr/gcc/10/share/gcc-$(HUMAN_VERSION)/python/libstdcxx/v6/__init__.py
 file path=usr/gcc/10/share/gcc-$(HUMAN_VERSION)/python/libstdcxx/v6/printers.py
 file path=usr/gcc/10/share/gcc-$(HUMAN_VERSION)/python/libstdcxx/v6/xmethods.py
-file path=usr/gcc/10/share/info/cpp.info
-file path=usr/gcc/10/share/info/cppinternals.info
+#file path=usr/gcc/10/share/info/cpp.info
```
mtelka: Why are these files commented out?

AndWac: That was a relic.

mtelka: But these files are in the sample manifest (at different location) so we should probably deliver them (once relocated back into `usr/gcc/10/share/info`).

PR #19308, opened 2024-10-24
desktop-cache: drop obsoleted restart_fmri and find_newer scripts

mtelka:
This fixes https://www.illumos.org/issues/16855 together with all other services that used to use the `find_newer` script.


DrLou:
Thanks, Marcel, for your work on this. And to ptribble, ik5pvx: ...

PR #19314, opened 2024-10-25
botan: update to 3.6.0; keep documentation

AndWac:


components/library/botan/botan.p5m
```diff
@@ -204,233 +206,81 @@ file path=usr/include/botan-3/botan/zfec.h
 file path=usr/include/botan-3/botan/zlib.h
 file path=usr/lib/$(MACH64)/cmake/Botan-$(HUMAN_VERSION)/botan-config-version.cmake
 file path=usr/lib/$(MACH64)/cmake/Botan-$(HUMAN_VERSION)/botan-config.cmake
-link path=usr/lib/$(MACH64)/libbotan-3.so target=libbotan-3.so.5.5.0
-link path=usr/lib/$(MACH64)/libbotan-3.so.5 target=libbotan-3.so.5.5.0
```
mtelka, line 208: This will very likely cause `keepassxc` to stop working:
```
$ ldd /usr/bin/keepassxc | grep botan
        libbotan-3.so.5 =>       /usr/lib/64/libbotan-3.so.5
$
```

mtelka: There is #19315 which should fix the issue :-).

PR #19357, opened 2024-10-29
drawing: fix p5m manifest

iigs:
Fix drawing.p5m so python runtime is detected by gmake REQUIRED_PACKAGES


components/image/drawing/drawing.p5m
```diff
@@ -22,7 +22,13 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
 license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
 
+# python apps are unusable without python runtime binary
+depend type=require fmri=__TBD pkg.debug.depend.file=python$(PYVER) \
```
mtelka, line 26: Wouldn't it be simpler to adjust the `pkg.depend.bypass-generate` transform below to do not complain about missing `drawing/__init__.py` and leave `pkgdepend` to do its job instead?

iigs: I would love for gmake REQUIRED_PACKAGES auto detect python modules and add those REQUIRED_PACKAGES to the Makefile.  I have zero insight how to make it work for this package.

If I replace the line `<transform file path=usr/share/drawing/drawing/.*\.py$ -> add pkg.depend.bypass-generate .* >` with `<transform file path=usr/share/drawing/drawing/.*\.py$ -> add pkg.depend.bypass-
generate .* >` then gmake REQUIRED_PACKAGES completes without an error and the results of the Makefile does not change.  Maybe I do not understand what you are trying to say?  Pointers to an existing .p5m file that does what you suggest will help to solve this problem.

mtelka: > If I replace the line `<transform file path=usr/share/drawing/drawing/.*\.py$ -> add pkg.depend.bypass-generate .* >` with `<transform file path=usr/share/drawing/drawing/.*\.py$ -> add pkg.depend.bypass- generate .* >`

Could you please explain what do you want to achieve by the added space to the transform?  Maybe I miss something important but the only difference in your _"If I replace the line X with Y"_ sentence above between the _X_ and _Y_ seems to be literally the single added space only.

Anyway, my point is this:

I did this:
```
$ git diff drawing.p5m 
diff --git a/components/image/drawing/drawing.p5m b/components/image/drawing/drawing.p5m
index bfbe90b589..58db09d7ae 100644
--- a/components/image/drawing/drawing.p5m
+++ b/components/image/drawing/drawing.p5m
@@ -22,7 +22,7 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
-<transform file path=usr/bin/drawing -> add pkg.depend.bypass-generate .* >
+<transform file path=usr/bin/drawing -> add pkg.depend.bypass-generate .*drawing.* >
 
 # mate menu icon needs to be in /usr/share/pixmaps
 file usr/share/icons/hicolor/128x128/apps/com.github.maoschanz.drawing.png path=usr/share/pixmaps/com.github.maoschanz.drawing.png
$
```
and everything seems to work as expected.

So my question is: why to add several lines here to somehow achieve what can be done by simple online change?

iigs:
The whole goal of this PR is to add as many REQUIRED_PACKAGES as possible that can be detected by gmake REQUIRED_PACKAGES.  I learned yesterday that python modules won't be detected.  So the best result is to at least add runtime/python.  Changes that Marcel suggested remove runtime/python.

If there are any optimizations that can be added and not lose the REQUIRED_PACKAGES, then I can not figure it out.

I am happy with what is delivered here.

mtelka:
How so?  The change I suggested does this after `env - gmake REQUIRED_PACKAGES`:
```
$ git diff
diff --git a/components/image/drawing/Makefile b/components/image/drawing/Makefile
index fc2b50ed88..d1ed86a7b2 100644
--- a/components/image/drawing/Makefile
+++ b/components/image/drawing/Makefile
@@ -40,3 +40,5 @@ PYTHON_REQUIRED_PACKAGES += library/python/pygobject
 REQUIRED_PACKAGES += library/desktop/gtk3
 
 # Auto-generated dependencies
+PYTHON_REQUIRED_PACKAGES += library/python/pygobject
+PYTHON_REQUIRED_PACKAGES += runtime/python
diff --git a/components/image/drawing/drawing.p5m b/components/image/drawing/drawing.p5m
index bfbe90b589..58db09d7ae 100644
--- a/components/image/drawing/drawing.p5m
+++ b/components/image/drawing/drawing.p5m
@@ -22,7 +22,7 @@ set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
 
-<transform file path=usr/bin/drawing -> add pkg.depend.bypass-generate .* >
+<transform file path=usr/bin/drawing -> add pkg.depend.bypass-generate .*drawing.* >
 
 # mate menu icon needs to be in /usr/share/pixmaps
 file usr/share/icons/hicolor/128x128/apps/com.github.maoschanz.drawing.png path=usr/share/pixmaps/com.github.maoschanz.drawing.png
$
```
Where do you see the removed `runtime/python`?

mtelka:
> How so? The change I suggested does this after `env - gmake REQUIRED_PACKAGES`:

Please not that this is on top of the current `oi/hipster` branch (i.e. without your change proposed in this PR).

iigs:
I have since deleted my working copy of files for this PR.    gmake REQUIRED_PACKAGES did not work for me on Monday with just Marcel's changes.
Andreas: Please let me know if I should just close this PR.

And just for completeness in case anybody reads over this PR, key bits of data in my second comment got mangled due to a bug in OI's copy and paste (extra characters get added) and the first transform line with DEBUG got lost because I edited the comment with a new copy/paste and I forgot that the file contents were already changed with the new text--the comment was edited was made because of other text had gotten mangled by OI's copy/paste. github's markdown does not like these extra characters.

AndWac:
Superceded by #19471

PR #19418, opened 2024-11-02
Add cargo-c

mtelka:


mtelka:
This package is not installable.  I'm working on a fix.

mtelka:
> This package is not installable. I'm working on a fix.

Fixed via #19424.

PR #19419, opened 2024-11-03
illumos-gate: alternate way to handle /32 packages

Bill-Sommerfeld:


mtelka:
This change is not needed now because the `illumos-gate` component builds okay.  I propose to postpone the merge until the https://www.illumos.org/issues/16879 gets fixed upstream.  After the https://www.illumos.org/issues/16879 fix upstream we will very likely need to do something here anyway, so better to do it once instead of twice, I think.

Toasterson (approved):

Bill-Sommerfeld:
> This change is not needed now because the `illumos-gate` component builds okay. I propose to postpone the merge until the https://www.illumos.org/issues/16879 gets fixed upstream. 

I was hoping that this would integrate here sooner so I could point to it as part of the testing for https://www.illumos.org/issues/16879 when I file the RTI.

> After the https://www.illumos.org/issues/16879 fix upstream we will very likely need to do something here anyway, so better to do it once instead of twice, I think.

Once 16879 lands upstream it should be sufficient to delete `0009-bug16879.patch`

mtelka:
> > This change is not needed now because the `illumos-gate` component builds okay. I propose to postpone the merge until the https://www.illumos.org/issues/16879 gets fixed upstream.
> 
> I was hoping that this would integrate here sooner so I could point to it as part of the testing for https://www.illumos.org/issues/16879 when I file the RTI.

Interesting idea.  But for materialize that you would need to wait for the next `illumos-gate` rebuild.  It is unclear when it will happen.  Anyway, your local nightly should be more than enough for the RTI.

> 
> > After the https://www.illumos.org/issues/16879 fix upstream we will very likely need to do something here anyway, so better to do it once instead of twice, I think.
> 
> Once 16879 lands upstream it should be sufficient to delete `0009-bug16879.patch`

Yes.  Provided it lands in the shape as it is currently proposed.

Bill-Sommerfeld:
No longer relevant given the integration of https://www.illumos.org/issues/16879 upstream.

PR #19466, opened 2024-11-06
abiword: update to 3.0.6

mtelka:


components/desktop/abiword/Makefile
```diff
@@ -12,37 +12,47 @@
...
+# - we patched configure.ac
+COMPONENT_PREP_ACTION= ( cd $(@D) && $(ENV) PATH="$(PATH)" NOCONFIGURE=1 ./autogen.sh ) ;
 
-TEST_TARGET = $(NO_TESTS)
-include $(WS_MAKE_RULES)/common.mk
+# Testing needs an X server
+TEST_REQUIRED_PACKAGES += x11/server/xvfb
+COMPONENT_PRE_TEST_ACTION += /usr/bin/Xvfb :0 >/dev/null 2>&1 &
+COMPONENT_TEST_ENV += DISPLAY=:0
+COMPONENT_POST_TEST_ACTION += /usr/bin/pkill -x Xvfb ;
```
AndWac, line 48: Can we switch to use DISPLAY \:1 for this purpose? Most of my systems are using X11 and \:0 clashes with it.

mtelka: There are several possible solutions (they could be combined):
- use dedicated user for running tests (the one without running X) - strongly recommended,
- slowly replace `DISPLAY=:0` by something like `DISPLAY=$(DISPLAY)` and add `DISPLAY = :0` to `shared-macros.mk`.  Then you would be able to run tests using `env - gmake DISPLAY=:1 test` if you wish/need.

Feel free to implement both. :-)

mtelka: Also `/usr/bin/Xvfb :0` would need to be replaced by `/usr/bin/Xvfb $(DISPLAY)`, obviously.

PR #19468, opened 2024-11-06
Fix configure option --localedir, because Lynx does not support it.

klausz65:
set CONFIGURE_DEFAULT_DIRS to no, because lynx does not support --localedir.

### DIFF
--- a/components/web/lynx/Makefile
+++ b/components/web/lynx/Makefile
@@ -14,12 +14,16 @@
 # Copyright (c) 2019, Michal Nowak
 #
 
+# disable CONFIGURE_DEFAULT_DIRS, because lynx does not support --localdir
+CONFIGURE_DEFAULT_DIRS=no
+
 BUILD_BITS=			64
 OPENSSL_VERSION= 3.1
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		lynx
 COMPONENT_VERSION=	2.9.2
+COMPONENT_REVISION=	1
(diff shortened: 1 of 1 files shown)


mtelka:
The `configure` script for `lynx` says this:
```
# Generated by Autoconf 2.52.20231210 for lynx 2.9.1b.
```
Please note that autoconf 2.52 was released more than 23 years ago (on 2001-07-18), however I'm not sure what the `20231210` version suffix above means.  The only known active distribution shipping autoconf 2.52 is OpenBSD (see https://repology.org/project/autoconf/versions) so presumably the lynx maintainers are using OpenBSD and intentionally use the oldest possible autoconf to generate the `configure` file.

The `--localedir` option was added to `autoconf` in version 2.59c released on 2006-04-12.  One could expect (see #18959) that more than 18 years is enough so that all projects already switched to it (or to a bit newer version).  Apparently this is not the case here with `lynx`.

While the change in this PR is good and makes `lynx` buildable again (thank you for it!), I'm still thinking about some different approach here:  To use our (latest) `autoconf` to regenerate the `configure` script so it supports everything that our build framework expect to see there.  I'll try to prepare the change...

mtelka:
It turned out that the `configure.in` for `lynx` is very old and is incompatible with `autoconf` 2.72.  This is probably the main reason why `lynx` maintainers still use the old `autoconf` 2.52.  It makes no more sense for me to pursue this way.

PR #19508, opened 2024-11-09
xapian - new package (search engine library) v1.4.26

fritzkink:
COMPONENT_CLASSIFICATION maybe is not the desired one. Please advise.


AndWac:
We already have it in library/xapian

fritzkink:
> We already have it in library/xapian

oh dear, should have checked this before. Thanks for closing.

PR #19551, opened 2024-11-12
python/tox: update to 4.23.2 with Dependency Groups support in setup.py.mk

mtelka:


mtelka:
Damn, this had to be merged separately without other Python projects that use tox.  I completely forgot this.  I'm sorry.

Reference: https://hipster.openindiana.org/jenkins/job/oi-userland/12372/

PR #19584, opened 2024-11-15
trousers: move 32 bit support to separate package

mtelka:


mtelka:
@AndWac this is ready for merge.

PR #19640, opened 2024-11-17
WIP: Mesa/drm/dri fixes

xmhd:
As per title. This pull request serves as the beginning of fixing the currently broken state of Mesa / DRM / DRI in OpenIndiana, which has been broken since 7defdf189741d867232df830a2cc385e4ed20f21

The problems which I have identified thus far:

- Mesa currently tries to enable gallium swrast , which can't work as gallium swrast has a dependency on llvm (16 specifically), which we neither have nor enable in the Makefile
- No classic (DRI) drivers explicitly enabled, left entirely to auto detect magic
- No working gallium drivers enabled
- I think Mesa 21.x defaults to dri2, explicitly enable dri3
- xorg-server does not explicitly enable libdrm

With the above in context, starting from a fresh install of OpenIndiana Hipster and updating to the latest available packages, only Vesa (not related to Mesa for the uninformed) is enabled. In this state neither glxinfo nor glxgears are available.

With the changes in this pull request I haven't quite gotten things working to the state that they were previously i.e. accelerated Intel graphics on a Thinkpad X230, but glxinfo and glxgears are now working. I've spent quite a few hours on this problem over the weekend and I have been discussing it with @Toasterson throughout. I will not have any more time to work on this until next weekend at the earliest. Please contribute and we can get this working again before continuing on to further improving the graphical desktop experience with things such as libva, Vulkan and more in Mesa.

PR #19641, opened 2024-11-17
WIP: Mesa/drm/dri fixes

xmhd:
Obsoletes #19640 

As per title. This pull request serves as the beginning of fixing the currently broken state of Mesa / DRM / DRI in OpenIndiana, which has been broken since 7defdf189741d867232df830a2cc385e4ed20f21

The problems which I have identified thus far:

- Mesa currently tries to enable gallium swrast , which can't work as gallium swrast has a dependency on llvm (16 specifically), which we neither have nor enable in the Makefile
- No classic (DRI) drivers explicitly enabled, left entirely to auto detect magic
- No working gallium drivers enabled
- I think Mesa 21.x defaults to dri2, explicitly enable dri3
- xorg-server does not explicitly enable libdrm

With the above in context, starting from a fresh install of OpenIndiana Hipster and updating to the latest available packages, only Vesa (not related to Mesa for the uninformed) is enabled. In this state neither glxinfo nor glxgears are available.

With the changes in this pull request I haven't quite gotten things working to the state that they were previously i.e. accelerated Intel graphics on a Thinkpad X230, but glxinfo and glxgears are now working. I've spent quite a few hours on this problem over the weekend and I have been discussing it with @Toasterson throughout. I will not have any more time to work on this until next weekend at the earliest. Please contribute and we can get this working again before continuing on to further improving the graphical desktop experience with things such as libva, Vulkan and more in Mesa.

Note:

- I haven't generated any new manifests for these changes as of yet
- xorg-server currently does not build due to a missing header, specifically <GL/internal/dri_interface.h>. Our previous Mesa v13.x correctly installed this header, whereas our new Mesa v21.3.9 installs this header to <mesa/internal/dri_interface.h>. I didn't bother to patch this just yet - I symlinked it on my test system.

### DIFF
--- a/components/x11/mesa/Makefile
+++ b/components/x11/mesa/Makefile
@@ -67,9 +67,18 @@ DRI_DRIVERS='i915'
 # Other Gallium drivers: virgl (virtip-gpu), svga (vmware)
 GALLIUM_DRIVERS='i915, crocus'
 
+# 'Vulkan' DRI drivers
+# We have some work to do before these can be enabled
+# Intel: intel, intel_hasvk
+# AMD: amd
+# nvidia: nouveau, nvk
+# swrast: swrast
+# virtio-gpu: virtio
+VULKAN_DRIVERS=''
+
 # Command line options to GNU autoconf configure script
 CONFIGURE_OPTIONS += -Dgallium-drivers=$(GALLIUM_DRIVERS)
-CONFIGURE_OPTIONS += -Dvulkan-drivers=''
+CONFIGURE_OPTIONS += -Dvulkan-drivers=$(VULKAN_DRIVERS)
 CONFIGURE_OPTIONS += -Dgbm=enabled
 CONFIGURE_OPTIONS += -Dglvnd=false
 # Wayland currently requires linux specific socket options and epoll for events.
--- a/components/x11/mesa/Makefile
+++ b/components/x11/mesa/Makefile
@@ -49,6 +49,14 @@ CONFIGURE_OPTIONS += --libdir='lib/mesa/amd64'
 PKG_HARDLINKS += usr/lib/xorg/modules/dri/kms_swrast_dri.so
 PKG_HARDLINKS += usr/lib/xorg/modules/dri/i915_dri.so
 
+# 'Classic' DRI drivers
+# Only add Intel i915 for now as these start getting removed in Mesa 22.x
+# Other Intel classic DRI drivers: i965, iris
+# Other AMD classic DRI drivers: r100, r200, r300, r600
+# Other NVIDIA classic DRI drivers: nv, nouveau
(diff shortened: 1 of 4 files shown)


iigs:
Please find package x11/library/glslang version 14.3.0 is now available.

Toasterson:
Hi @xmhd good to see someone coming into OpenIndiana packaging. And thanks for the work!

Usually what we do while reviewing, is we try to compile this on a set of machines to coax out some hidden things that auto-magic is detecting so we can make this reproducible. I found some things during that process.

Apparently, only one type of i915 driver can be enabled at the same time. I get 
```
meson.build:318:2: ERROR: Problem encountered: Only one i915 provider can be built
```

When commenting out either `DRI_DRIVERS='i915'` or `GALLIUM_DRIVERS='i915'`  i get further but then get stuck on
```
Found CMake: /usr/bin/cmake (3.31.1)
Run-time dependency libunwind found: NO (tried pkgconfig and cmake)

meson.build:1985:4: ERROR: Problem encountered: OSMesa gallium requires gallium softpipe or llvmpipe.
```

It could be that we also need libunwind https://github.com/libunwind/libunwind which seems to be straight forward or that some option is requiring llvm now we have already started to provide llvm16 again on the system to support this effort here.

lastly we will need to split each component into it's own PR so that we can make sure the build system builds them in the right order. We still have to do this manually. We can keep it in this PR for now to make sharing code easier during development phase.

Loving the work so far. Let's get this over the finish line.

xmhd:
@Toasterson Thanks for initial review - I realise that I missed submitting a few changes hence why you hit issues. And yeah no worries re: splitting this into multiple PRs etc afterwards, this can just serve as the staging ground for getting this working again.

I'll pick it up this weekend and see if I can get it over the finish line.

components/x11/mesa/Makefile
```diff
@@ -32,6 +33,16 @@ include $(WS_MAKE_RULES)/common.mk
 
 PATCH_LEVEL= 0
 
+# Mesa 21.x requires 2.4.100 - 2.4.109
+# Mesa 22.x requires 2.4.9 - 2.4.110
```
mtelka, line 37: Shouldn't this read 2.4.109 instead of 2.4.9 ?

xmhd: Yes indeed, good catch. I'll fix this tonight.

iigs:
Please find LLVM 16 in packages developer/clang-16 and runtime/clang-16 .

I looked at libunwind--it doesn't have enough Solaris/illumos support to be useful.  Adding this to the mesa Makefile will ensure libunwind does not get used:
CONFIGURE_OPTIONS += -Dlibunwind=disabled

xmhd:
Closing this - individual PRs incoming.

PR #19652, opened 2024-11-18
mercurial: update to 6.8.2

AndWac:


mtelka:
Could you please use `python-integrate-project` to update this?  Thank you.

BTW, I do have this in my queue too, but some tests fails here and I had no time yet to look at it deeper.

PR #19677, opened 2024-11-21
zeromq: build only 64bit, because dependent library: libsodium is alr…

klausz65:
…eady 64bit only.

### DIFF
--- a/components/library/zeromq/Makefile
+++ b/components/library/zeromq/Makefile
@@ -12,15 +12,15 @@
 #
 # Copyright 2014 Alexander Pyhalov.  All rights reserved.
 # Copyright 2016 Jim Klimov
+# Copyright 2024 Klaus Ziegler
 #
 
-BUILD_BITS= 64_and_32
 USE_DEFAULT_TEST_TRANSFORMS= yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= libzmq
 COMPONENT_VERSION= 4.3.5
-COMPONENT_REVISION= 1
+COMPONENT_REVISION= 2
 COMPONENT_SUMMARY= Open source message queue optimised for performance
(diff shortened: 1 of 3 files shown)


components/library/zeromq/Makefile
```diff
@@ -44,13 +44,10 @@ COMPONENT_PREP_ACTION = \
 
 COMPONENT_PRE_CONFIGURE_ACTION = ( $(CLONEY) $(SOURCE_DIR) $(@D) )
 
-# Build docs just once (for one ARCH), parameter varies from version to version
-CONFIGURE_OPTIONS.32 +=	--with-docs
-CONFIGURE_OPTIONS.64 +=	--without-docs
+CONFIGURE_OPTIONS +=	--with-docs
 CONFIGURE_OPTIONS +=	--enable-static=no
 CONFIGURE_OPTIONS +=	--with-libsodium
```
mtelka, line 49: I assume the other option would be to build 32-bit `zeremq` with `--without-libsodium` (or so).  But I do not know if we need 32-bit `zeremq` for anything.  If we don't then dropping 32-bit support here is okay.

klausz65: I checked all other libsodium packages and all of them are allready at 64bit:
network/megasync
network/chat/toxcore
library/c++/zeromq
web/php-81/php-common
mail/dovecot
library/python/pynacl-39
service/network/dns/dnscrypt-proxy
web/php-82/php-common
desktop/remote-desktop/remmina
editor/vim/vim-core
editor/gvim

and the only package which requires zeromq is: library/python/pyzmq-39 and this is of course also only 64bit.

mtelka: The other package that requires `zeromq` is `library/c++/czmq`.

klausz65: Okay, Marcel, I converted it to draft and will first do a 64 bit pull for library/c++/czmq - thanks for the heads up!

PR #19678, opened 2024-11-21
illumos-gate: SPARC systems need package: system/library/processor to…

klausz65:
… be able to at least boot.

I really learned this the hard way - after creating a new gate an trying to boot from it.

### DIFF
--- a/components/openindiana/illumos-gate/Makefile
+++ b/components/openindiana/illumos-gate/Makefile
@@ -196,9 +196,17 @@ endif
 # The list of FMRIs to ignore is also useful for other things, and is easier to
 # use when not phrased directly as a sed script.
 #
+# Package: system/library/processor has been obsoleted on x86/amd64 systems.
+# SPARC systems can't boot without it.
+#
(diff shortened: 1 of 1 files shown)


mtelka:
When you build the `illumos-gate` then you'll get the `system/library/processor` package obsoleted (see https://www.illumos.org/issues/14802) and it is so for more than two years now.  Since the `openindiana/illumos-gate` component is designed to build the `HEAD` of the `illumos-gate` then you cannot get unobsoleted `system/library/processor` package for very long time now.

I'm very sorry, but this change makes no sense with the current `components/openindiana/illumos-gate/Makefile` file.

klausz65:
I checked all other libsodium packages and all of them are allready at 64bit:
network/megasync
network/chat/toxcore
library/c++/zeromq
web/php-81/php-common
mail/dovecot
library/python/pynacl-39
service/network/dns/dnscrypt-proxy
web/php-82/php-common
desktop/remote-desktop/remmina
editor/vim/vim-core
editor/gvim

and the only package which requires zeromq is: library/python/pyzmq-39 and this is of course also only 64bit.

PR #19688, opened 2024-11-22
Replace the openssl-31 package by OpenSSL 3.4.0 packaged as openssl-3

mtelka:


AndWac:
Are you sure that this version can replace 3.1 without introducing compatibility problems?

mtelka:
> Are you sure that this version can replace 3.1 without introducing compatibility problems?

I'm running this on a machine with (almost) all OpenIndiana packages installed and so far I see no issue.

BTW, OpenSSL 3.x is by design fully API and ABI compatible - see https://openssl-library.org/policies/releasestrat/index.html for details.

PR #19698, opened 2024-11-22
clang-16: add 16.0.6

iigs:
WIP: Mesa/drm/dri fixes PR#19641 mentions that llvm 16 is needed for gallium swrast.  This package is to support that effort.


mtelka:
Is _exactly_ clang 16 needed, or _at least_ clang 16?  Please note that clang 16 is deprecated.

BTW, this integration won't work properly.  Please read #19010 for details.

AndWac:
As I understand #19641 mesa (gallium) depends exactly on llvm 16:
"* Mesa currently tries to enable gallium swrast , which can't work as gallium swrast has a dependency on llvm (16 specifically), which we neither have nor enable in the Makefile".

iigs:
I do not understand why this will not integrate correctly.  I looked over PR #19010 and only see that $(CLANGVER) got added to clang-18.  This package, as provided, publishes and installs.  During its development, I had to rip out all the instances of the generated $(CLANGVER) in the clang*.p5m files since they had a vale of 18 instead of 16, which broke gmake publish.

This Makefile was based on clang-17.  The patches got ported to 16 and all of the missing omnios patches got added.

iigs:
The dollarsigns got eaten by the markdown editor that should be included in front of the (CLANGVER).  Here is the mangled text with spaces:
got added to clang-18.  This package, as provided, publishes and installs.  During its development, I had to rip out all of the instances of the generated

mtelka:
> As I understand #19641 mesa (gallium) depends exactly on llvm 16: "* Mesa currently tries to enable gallium swrast , which can't work as gallium swrast has a dependency on llvm (16 specifically), which we neither have nor enable in the Makefile".

Sure, #19641 claims so, but I'd appreciate some link to some kind of authoritative source of this information.  Just to make sure we understand the matter properly.

mtelka:
> This Makefile was based on clang-17. The patches got ported to 16 and all of the missing omnios patches got added.

This is exactly the problem, because `clang-17` does not work well with our build framework.  Any new `clang` integrations should be based on `clang-18`.

Toasterson:
clang 16 is the newest one that still works with mesa oldest is clang 13 I suggest we reintroduce either of these while we work on mesa and the software rendering driver which should give us a large range of desktop hardware support.

mtelka:
> clang 16 is the newest one that still works with mesa oldest is clang 13 I suggest we reintroduce either of these while we work on mesa and the software rendering driver which should give us a large range of desktop hardware support.

We still do have `clang-13` packaged (and not obsoleted yet).

BTW, where the information about the clang version range (13 to 16 only) comes from?

xmhd:
> > clang 16 is the newest one that still works with mesa oldest is clang 13 I suggest we reintroduce either of these while we work on mesa and the software rendering driver which should give us a large range of desktop hardware support.
> 
> We still do have `clang-13` packaged (and not obsoleted yet).
> 
> BTW, where the information about the clang version range (13 to 16 only) comes from?

My experience packaging Mesa elsewhere. v21 indeed works with llvm13 through llvm16. If I need to dig through the Mesa source code to point you to that, I can do.

mtelka:
> > BTW, where the information about the clang version range (13 to 16 only) comes from?
> 
> My experience packaging Mesa elsewhere. v21 indeed works with llvm13 through llvm16. If I need to dig through the Mesa source code to point you to that, I can do.

Hmmm.  Looking elsewhere I see this (as an example): https://src.fedoraproject.org/rpms/mesa/blob/rawhide/f/mesa.spec#_149 and `clang-devel` is at version `19.1.3` in Fedora Rawhide (see https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/c/).

What am I missing here?

Toasterson:
> > > BTW, where the information about the clang version range (13 to 16 only) comes from?
> > 
> > 
> > My experience packaging Mesa elsewhere. v21 indeed works with llvm13 through llvm16. If I need to dig through the Mesa source code to point you to that, I can do.
> 
> Hmmm. Looking elsewhere I see this (as an example): https://src.fedoraproject.org/rpms/mesa/blob/rawhide/f/mesa.spec#_149 and `clang-devel` is at version `19.1.3` in Fedora Rawhide (see https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/c/).
> 
> What am I missing here?

Their mesa is 24 ours 21, we cannot go to 24 just yet.
And yes there is quite some coordination behind this in the channels since we still need patches for mesa 21x to get things working and then we need even more for 24.x as 24 has many more linux isms in it's code by what we where able to look at.

xmhd:
> > > BTW, where the information about the clang version range (13 to 16 only) comes from?
> > 
> > 
> > My experience packaging Mesa elsewhere. v21 indeed works with llvm13 through llvm16. If I need to dig through the Mesa source code to point you to that, I can do.
> 
> Hmmm. Looking elsewhere I see this (as an example): https://src.fedoraproject.org/rpms/mesa/blob/rawhide/f/mesa.spec#_149 and `clang-devel` is at version `19.1.3` in Fedora Rawhide (see https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Everything/x86_64/os/Packages/c/).
> 
> What am I missing here?

https://github.com/gentoo/gentoo/blob/0ab20faf2aeb36af54cb0de6c4fb36d0e85a6e80/media-libs/mesa/mesa-21.3.8.ebuild

As @Toasterson commented - Mesa v21 (What we are using) is llvm13 - llvm16. llvm19 is for Mesa v24 (what Fedora are using).

mtelka:
> Their mesa is 24 ours 21

I see.  Thanks for explanation.

Does this all means that our Mesa 21 got build using `clang-18` which is incompatible with Mesa 21 and so this is causing problems in Mesa?  Like the swrast issue mentioned in #19641?

If that's the case and if Mesa 21 works okay with clang 13 to 16, then why not to simply recompile our Mesa using our already readily available and fully working `clang-13`?

xmhd:
> > Their mesa is 24 ours 21
> 
> I see. Thanks for explanation.
> 
> Does this all means that our Mesa 21 got build using `clang-18` which is incompatible with Mesa 21 and so this is causing problems in Mesa? Like the swrast issue mentioned in #19641?
> 
> If that's the case and if Mesa 21 works okay with clang 13 to 16, then why not to simply recompile our Mesa using our already readily available and fully working `clang-13`?

It isn't a case of compiling it with clang-13 breaking Mesa, it's that the current Makefile doesn't actually enable any gallium drivers (which in turn depend on llvm), nor we do explicitly enable llvm. Using llvm 13 is potentially a workable solution, and I plan to explore it this weekend.

mtelka:
> > Does this all means that our Mesa 21 got build using `clang-18` which is incompatible with Mesa 21 and so this is causing problems in Mesa? Like the swrast issue mentioned in #19641?
> > If that's the case and if Mesa 21 works okay with clang 13 to 16, then why not to simply recompile our Mesa using our already readily available and fully working `clang-13`?
> 
> It isn't a case of compiling it with clang-13 breaking Mesa

Wait.  I'm confused.  How is it then?  Is `clang-13` okay for Mesa 21 or not?

mtelka:
> it's that the current Makefile doesn't actually enable any gallium drivers

What is this then? https://github.com/OpenIndiana/oi-userland/blob/215e0371baac7947426632b654ce2bd122fda258/components/x11/mesa/Makefile#L51

xmhd:
> > it's that the current Makefile doesn't actually enable any gallium drivers
> 
> What is this then?
> 
> https://github.com/OpenIndiana/oi-userland/blob/215e0371baac7947426632b654ce2bd122fda258/components/x11/mesa/Makefile#L51

What I should have said the current Makefile isn't enabling any gallium drivers that work. They all depend on llvm, which we are explicitly disabling.

See: https://github.com/OpenIndiana/oi-userland/blob/215e0371baac7947426632b654ce2bd122fda258/components/x11/mesa/Makefile#L58

iigs:
I found the diffs with clang-18 (Makefile and the two clang*p5m files) and integrated them here.

AndWac:
I cannot publish it here as it depends on itself.

╰─➤  gmake publish

  /developer/clang-16

	/developer/clang-16
pkg install returned 1
gmake: *** [/export/home/andreas/oi-userland/make-rules/environment.mk:51: component-environment-prep] Error 1

mtelka:
> I cannot publish it here as it depends on itself.
> 
> ╰─➤ gmake publish
> 
> pkg list: no packages matching the following patterns are installed: /developer/clang-16 Adding required packages to build environment...
> 
> pkg install: The following pattern(s) did not match any allowable packages. Try using a different matching pattern, or refreshing publisher information:
> 
> ```
> /developer/clang-16
> ```
> 
> pkg install returned 1 gmake: *** [/export/home/andreas/oi-userland/make-rules/environment.mk:51: component-environment-prep] Error 1

Oops.  I think that's a bug in the `Makefile` for `clang-18`.

mtelka:
> Oops. I think that's a bug in the `Makefile` for `clang-18`.

Or, to be more precise, the current implementation expects that when there is new major clang version integrated then the `CLANG_DEFAULT` (in `shared-macros.mk`) is set to the same version number together with the initial integration (in the same commit).  Since this assumption cannot be met when there is some older clang integrated (as in this case) then we need to change this.

I'm working on a solution...

mtelka:
@iigs what `clang` version you need/want to use to build `clang-16`?  Will it compile properly using `clang-18`?

mtelka:
> @iigs what `clang` version you need/want to use to build `clang-16`? Will it compile properly using `clang-18`?

That question probably makes no sense because it looks like `clang` is compiled using `gcc` :-).

mtelka:
> I'm working on a solution...

It should be fixed by #19704.  However, to make the fix working properly the `clang` component needs to be named `clang` (not `clang-16`), so @iigs please change the `COMPONENT_NAME` here.

Thank you.

AndWac:
If all clang versions will have COMPONENT_NAME=clang then clang 18 will obsolete all other versions. Or did I miss anything?

mtelka:
> If all clang versions will have COMPONENT_NAME=clang then clang 18 will obsolete all other versions. Or did I miss anything?

Yes :-).  The `COMPONENT_NAME` is not necessarily related to package fmri.  The `COMPONENT_NAME` for `clang-13` is `clang` too.  Another example is this:
```
$ grep COMPONENT_NAME= components/runtime/openjdk-*/Makefile
components/runtime/openjdk-11/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-17/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-18/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-21/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-22/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-23/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-24/Makefile:COMPONENT_NAME=          openjdk
components/runtime/openjdk-8/Makefile:COMPONENT_NAME=         openjdk
$
```
IOW, all openjdk versions does have the same `COMPONENT_NAME`.

AndWac:
Ok, I got it.
What I don't like is that all clang packages, be it developer or runtime, for all versions share the same COMPONENT_SUMMARY.

mtelka:
> What I don't like is that all clang packages, be it developer or runtime, for all versions share the same COMPONENT_SUMMARY.

This is so since the initial `clang-18` integration (see #16569).  I didn't went through the whole `clang-18` component history but I assume it was always so.

The fix is easy.  Just add `COMPONENT_SUMMARY.clang-runtime=something` to the `Makefile` and possibly update the main `COMPONENT_SUMMARY` (if needed) since it is used for the devel package.

mtelka:
> and possibly update the main `COMPONENT_SUMMARY` (if needed) since it is used for the devel package.

Maybe you want to add the major clang version there, for example.  Something like:
```
COMPONENT_SUMMARY=LLVM $(CLANG_VERSION) tools and Clang $(CLANG_VERSION) compiler
```
and then you can do this, for example:
```
COMPONENT_SUMMARY.clang-runtime=$(COMPONENT_SUMMARY) (runtime)
```
You can also try something like this (although I'm not 100% confident it will work, so please double check it does) if you want:
```
COMPONENT_SUMMARY.clang=$(COMPONENT_SUMMARY) (devel files)
```

PR #19725, opened 2024-11-24
freeradius - rebuild due to new openssl version

fritzkink:


mtelka:
Why is this needed?  I'm curious.  Please note that this change is okay, but I simply do not understand why do we need it.  There are ca 60 other components where you could do something similar like you did here and I doubt we should do that.

fritzkink:
root@secdienst:/root# /usr/sbin/radiusd -X
libssl version mismatch.  built: 30100060 linked: 30400000

mtelka:
> root@secdienst:/root# /usr/sbin/radiusd -X libssl version mismatch. built: 30100060 linked: 30400000

Thank you.  Really appreciated.

However, I do not understand why is `radiusd` so strict here.  This looks like a bug in `radiusd`.  Maybe should be reported upstream?  Did it work properly with OpenSSL 3.1.7?  It looks like it was built with OpenSSL 3.1.6 and so there had to be mismatch too.

mtelka:
Reported upstream as https://github.com/FreeRADIUS/freeradius-server/issues/5467

fritzkink:
thanks. I cheched on the source but I only found #if OPENSSL_VERSION_NUMBER >= 0x30000000L. There was never a different 3.x version referenced,

PR #19729, opened 2024-11-24
php83 - new package version php 8.3

fritzkink:


components/web/php/php-8_3/Makefile
```diff
@@ -0,0 +1,252 @@
...
+CFLAGS +=		-I/usr/include/openldap
+CPPFLAGS +=      	$(CPP_XPG6MODE) 
+LDFLAGS +=       	-lldap_r
+
+# build with the system default libjpeg
+CFLAGS+=                $(JPEG_CPPFLAGS) $(JPEG_CFLAGS)
+CXXFLAGS+=              $(JPEG_CPPFLAGS) $(JPEG_CXXFLAGS)
+CPPFLAGS+=              $(JPEG_CPPFLAGS)
+LDFLAGS+=               $(JPEG_LDFLAGS)
+
```
AndWac: Is this really still needed?

hoewweken:
Hello!

We have to rebuild PHP 8.3 and 8.2 because of the libtidy update from 1.0 to 5.8 #21125 .

Alas, php-8_3 doesn't build in my newly created build-zone:
```
...
checking for crypt in -lcrypt... yes
configure: error: Cannot find imap library (libc-client.a). Please check your c-client installation.
gmake: *** [/root/src/oi-userland/make-rules/common-rules.mk:43: /root/src/oi-userland/components/web/php/php-8_3/build/amd64/.configured] Error 1

```
why `libc-client.a`this is a  static lib ??

fritzkink:
This is coming from the umap package which is currently broken as the dynamic lib has some unresolved dependencies

AndWac:
What is the umap package? We don't have a package named umap. You probably meant uw-imap. But that package is not broek as far as I can tell. I have even rebuilt it recently.

PR #19735, opened 2024-11-25
revert set defaults for X11 libraries to 64 bit only

AndWac:


cgrzemba:
I identified the following X11 packages which needed for SunRay in 32bit:
libx11
libxext
libxau
libxfixes
libxft
libxcursor
libxrender
libxv
libxinerama
libxrandr
libxpm
libxcb
libxdmcp
libxt
libice
libsm

PR #19737, opened 2024-11-25
php83 - fix mediator settings

fritzkink:


AndWac:
You also need to set COMPONENT_REVISION=1 in the Makefile.

PR #19763, opened 2024-11-27
tiff: fix dependencies for jpeg library and add FILE_OFFSET_BITS patc…

klausz65:
…h for 32bit compile

### DIFF
--- a/components/library/tiff/patches/01-fix-FILE_OFFSET_BITS-in-32bit-mode.patch
+++ b/components/library/tiff/patches/01-fix-FILE_OFFSET_BITS-in-32bit-mode.patch
@@ -1,0 +1,18 @@
+
+This is to fix the following warnings, during the 32 bit build:
+
+/usr/include/sys/feature_tests.h:263: note: this is the location of the previous definition
+  263 | #define _FILE_OFFSET_BITS       32
+
+--- tiff-4.7.0/configure.orig	2024-09-11 09:42:32.000000000 +0200
++++ tiff-4.7.0/configure	2024-11-06 21:14:53.704371346 +0100
+@@ -18075,6 +18075,9 @@
+   break
+ done
+ fi
++case $CFLAGS in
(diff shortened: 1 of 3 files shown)


components/library/tiff/patches/01-fix-FILE_OFFSET_BITS-in-32bit-mode.patch
```diff
@@ -0,0 +1,12 @@
+--- tiff-4.7.0/configure.orig	2024-09-11 09:42:32.000000000 +0200
++++ tiff-4.7.0/configure	2024-11-06 21:14:53.704371346 +0100
+@@ -18075,6 +18075,9 @@
+   break
+ done
+ fi
++case $CFLAGS in
++  *-m32*) ac_cv_sys_file_offset_bits=32 ;;
```
mtelka, line 14: Could you please add comment explaining why is this needed?  Thank you.

mtelka:
Thanks for the added comment.  But, it is ... strange.  Does it mean that `_FILE_OFFSET_BITS` is set to `64` without this patch?

Yes, correct, during configure much of the autoconf software sets this often wrong, but I didn't catch all of them,
just one after another if I recognize this warning.

mtelka:
> Yes, correct, during configure much of the autoconf software sets this often wrong, but I didn't catch all of them, just one after another if I recognize this warning.

With `_FILE_OFFSET_BITS` forcefully set to `32` you are probably loosing the ability to work with large files.  I think this is not what we want here.  See https://illumos.org/man/7/lfcompile for more info.  IOW, I'd say that `_FILE_OFFSET_BITS` set to `64` is okay and if you want to solve this you should try to find why is the build complaining.  Maybe it is because the `_FILE_OFFSET_BITS` is set to `64` too late.  After the `feature_tests.h` include.

PS: @klausz65, please do not edit _my_ comments here.  Create your own instead.  Thank you.

PR #19764, opened 2024-11-27
mozilla-nss: fix dependencies for library/nspr/32

klausz65:


mtelka:
@klausz65 I'm curious if we really need all of these rebuilds.  Could you please try to build your SPARC platform incrementally?  I.e. in the same order as it was done here?  This would save you a lot of work and headache.

You cannot reasonably expect that the `oi-userland` is always buildable to the same state as it is published in the ips repo.  It is simply evolving all the time.  What you can reasonably expect is, that when a change is merged then it should build on the current system and produce new package(s) that should be installed on the build machine.  Then you can go to the next changeset.  If you do all of this incrementally, then you'll find that almost everything works.

OTOH, these rebuilds are not bad per se and the cleanup work would need to be done anyway.  I just find them unnecessary.

Thank you.

AndWac:
I appreciate this kind of work because sooner or later we will need to fix the builds.

PR #19765, opened 2024-11-27
php-8_3-ext-imagick - new package for php83

fritzkink:


hoewweken:
This package is not in pkg.openindiana.org, does it build ?
https://pkg.openindiana.org/hipster/en/search.shtml?token=php-imagick&action=Search

i have no success to build it locally, cause i currently have a linker reference error in tidy.so ..

```
$ php -v
PHP Warning:  PHP Startup: Unable to load dynamic library 'tidy.so' (tried: /usr/php/8.3/extensions/tidy.so (ld.so.1: php: fatal: relocation error: file /usr/php/8.3/extensions/tidy.so: symbol tidyBufFree: referenced symbol not found)
```
Maybe it doesn't build since March, 19 after the commit of  #21125  `tidy: upgrade from 1.0.0 to 5.8.0`, which at least partially broke php...
I'll check on an older BE when i get some minutes of time..

PR #19778, opened 2024-11-28
pulseaudio: fix dependencies for openssl-3

klausz65:


components/library/openssl-3/patches/04-amd64.patch
```diff
@@ -1,4 +1,4 @@
-We prefer amd64 over 64.  It fits better our packaging.
+We prefer amd64/sparcv9 over 64.  It fits better our packaging.
```
AndWac: This file doesn't belong to pulseaudio!

klausz65: Sorry, my mistake - how to get rid of it in this commit?

AndWac: Fix it locally in your branch and the use -f (force) in order to update your Github branch. This will automatically update the PR.

PR #19811, opened 2024-12-01
mpg123: new webside / download url

klausz65:
Sorry for the trouble


AndWac:
The original site is still available for both web and download. So this change is not necessary.

PR #19843, opened 2024-12-03
dejagnu: fix package dependencies

klausz65:


components/developer/dejagnu/Makefile
```diff
@@ -53,4 +54,5 @@ ASLR_MODE = $(ASLR_ENABLE)
 REQUIRED_PACKAGES += shell/expect
```
mtelka: Please remove this line.  It is duplicate.

klausz65: done

PR #19856, opened 2024-12-03
text-utilities: build 64-bit programs only / new shared-makros setup

klausz65:
I tried to build cfdisk, but it still doesn't build, but I left the man-page cfdisk.8 in the sample-manifest for completness


components/text/text-utilities/Makefile
```diff
@@ -90,7 +83,7 @@ COMPONENT_TEST_ENV +=	TS_CMD_HEXDUMP=$(PROTOUSRBINDIR)/hexdump
 COMPONENT_TEST_ENV +=   TS_CMD_RENAME=$(PROTOUSRBINDIR)/rename
 COMPONENT_TEST_ENV +=	PATH="$(PATH.gnu)"
 # Needed for the libsmartcols.so library
-COMPONENT_TEST_ENV +=	LD_LIBRARY_PATH=$(PROTOUSRLIBDIR64)
+#COMPONENT_TEST_ENV +=	LD_LIBRARY_PATH=$(PROTOUSRLIBDIR64)
```
mtelka: What is this?

The comment says: **Needed for the libsmartcols.so library**

So, is the line needed or not?  If it _is_ needed then why is it commented out?  If it _is not_ needed then why the comment says it is needed?

klausz65: Sorry, forgot to remove that line before pushing.

AndWac:
Did you really run the tests? The result file is still named results-64.master and thus cannot be found. It should be renamed to results-all.master.

AndWac:
My test results here differ (the first test run; the 2nd produces no diffs):
Lots of similar diffs like this:
-       rename: symlink check                  ... OK
+       rename: symlink check                  ... SKIPPED (rename not found)
Which make me think that the dropped COMPONENT_TEST_ENV might still be needed.

AndWac:
No, that doesn't change the results. But something seems to be broken during the 1st run.

klausz65:
Well, I receive no differences at all on oibuild...

mtelka:
Typo in the commit message: `makros` vs `macros`

@AndWac, @klausz65 just an idea: do you have the `text/text-utilities` package installed when you run tests?

AndWac:
On my system the text-utilities package was installed but after de-installation the test results are the same like before.

mtelka:
> On my system the text-utilities package was installed but after de-installation the test results are the same like before.

This very likely means that the testing actually tests the installed `text-utilities` not the currently built one.  This is obviously wrong.

AndWac:
Am 09.12.24 um 06:02 schrieb Marcel Telka:
>
>     On my system the text-utilities package was installed but after
>     de-installation the test results are the same like before.
>
> This very likely means that the testing actually tests the installed
> |text-utilities| not the currently built one. This is obviously wrong.
>
I don't think so. The test results don't depend on the installation of
the package. This is not an indicator that the tests are using the
installed files.
It is not even an indication for the opposite.

mtelka:
Right.  I looked at it a bit deeper and indeed the problem is something else.  I'll prepare a PR with the fix.

PR #19876, opened 2024-12-05
mlocate: build only 64 bit binaries new shared-macros setup

klausz65:


components/sysutils/mlocate/Makefile
```diff
@@ -23,11 +23,12 @@
 # Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
 #
 
+BUILD_BITS= 64
```
mtelka, line 26: Please remove this line.  It is default.

klausz65: done

PR #19878, opened 2024-12-05
openssl-3: add compatibility symlink for 64 bit libraries

mtelka:


klausz65:
Thank you for this!!! However I think your local branch missed pull #19809, because patch 04-lib-64.patch had allready the fix for sparcv9 in it, but was still named 04-amd64.patch. So I would also check both manifests for variant.arch=i386 regarding the different configuration.h file. They are not different on SPARC, so your diff loop didn't catch them. And BTW. the perl modules do build now with openssl-3 - thanks again, much appreciated.

mtelka:
@klausz65 could you please double check what exactly I did in this PR?  I obviously do not miss anything here (especially not #19809).  If you still think I miss anything, then please tell us how exactly you found that I miss what you think I miss.

Thank you.

klausz65:
I'm all fine. Please go ahead with merging.

PR #19879, opened 2024-12-05
postfix: recompile for library/security/openssl-3 and update to 3.9.1

klausz65:


mtelka:
To give some sense to this non-sense: Could you please update `postfix` to 3.9.1 while here?

Thank you.

klausz65:
updated to 3.9.1

PR #19916, opened 2024-12-09
appstream: update to 1.0.4

AndWac:


mtelka:
@AndWac are all tests passing for you?

AndWac:
No:
-as-test_pool             OK
+as-test_pool             FAIL   killed by signal 6 SIGABRT
But that also happens with the previous version.

mtelka:
Yes.  Same here.  Something apparently changed since the 1.0.3 integration and so the testing fails now.  It should be investigated (or at least test results should be updated).

AndWac:
The test fails because it's missing the samples folder. Alas CLONEY doesn't work because meson complains about two meson.build files.

AndWac:
With the samples folder in place all tests are ok.

AndWac:
Or not, the trick was working for 1.0.3 but failed now.

AndWac:
Running the test manually within the tests folder finishes ok, though.

PR #19927, opened 2024-12-10
text-utilities: install before testing

mtelka:


klausz65:
Thanks, for fixing this testing make target!

PR #19985, opened 2024-12-12
which: obsolete shell/which component, has been replaced by BSD imple…

klausz65:
…mentation in illumos-gate with: #15471


mtelka:
I'm confused here.  The `shell/which` is GNU which while the illumos which is currently BSD which.  Why we cannot have both?

components/meta-packages/history/history
```diff
@@ -1258,6 +1258,8 @@ service/network/load-balancer/pen@0.18.0,5.11-2013.0.0.1
 service/network/samba/samba36@3.6.25,5.11-2015.0.2.3
 service/postrun@1.0,5.11-2015.0.2.0
 service/storage/avs/cache-management@0.5.11,5.11-2022.0.1.0
+shell/whichh2.21-2020.0.1.2
+shell/whichh2.21-2023.0.0.2
 slim_install@0.1,5.11-2015.0.2.1
```
AndWac, line 1263: The actual version is 2.21-2020.0.1.1, so the 2nd line is superfluous.

klausz65:
The first one is x86 the second is SPARC, why shouldn't I use the same technics as x86 to obsolete the package.

AndWac:
There seem to be packages dependent on this:
mate_install
server_install
shell/parallel

So we cannot obsolete this package until the dependency is dropped.

AndWac:
The entry in the history file only needs to be higher than the highest provided one. In this case the first line is superfluous.

klausz65:
Okay, I never had package shell/paralell buildfor SPARC - I'll check that first - will convert to draft for now....

components/meta-packages/history/history
mtelka: Besides of that there is missing `@`.

mtelka:
> There seem to be packages dependent on this: mate_install server_install shell/parallel
> 
> So we cannot obsolete this package until the dependency is dropped.

Also, `encumbered/audacity` component needs `shell/which` for building.

mtelka:
BTW, I'm -1 for this PR.

klausz65:
closed

PR #19992, opened 2024-12-13
shared-macros.mk: fix GNU_CPU

mtelka:
This is broken since #19420.

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -124,14 +124,10 @@ PKG_OS_VERSION ?= 0.$(PKG_SOLARIS_VERSION)
 
 # GNU target triplet
 GNU_TRIPLET=$(GNU_CPU)-$(GNU_VENDOR)-$(GNU_OS)
-# The cpu part of the triplet is basically the same as $(MACH):
-# i386/sparc     - for GCC version older than 9
-# x86_64/sparcv9 - for GCC version 9 and newer
-ifeq ($(MACH), i386)
-GNU_CPU = $(if $(filter $(GCC_VERSION),3 4 7),$(MACH),$(MACH:i386=x86_64))
-else
-GNU_CPU = $(if $(filter $(GCC_VERSION),3 4 7),$(MACH),$(MACH:sparc=sparcv9))
-endif
+# The cpu part of the triplet is this:
+# i386/sparc     - for GCC version older than 9 - so same as $(MACH)
+# x86_64/sparcv9 - for GCC version 9 and newer - $(MACH64) is close to this
+GNU_CPU = $(if $(filter $(GCC_VERSION),3 4 7),$(MACH),$(MACH64:amd64=x86_64))
 # The vendor part of the triplet is:
 # pc - for x86
 # sun - for SPARC


klausz65:
Shall I fix #19420 with this:
ifeq ($MACH64), amd64)
GNU_CPU = $(if $filter $(GCC_VERSION),3 4 7),$(MACH),$(MACH64:amd64=x86_64))
else
GNU_CPU = $(if $filter $(GCC_VERSION),3 4 7),$(MACH),$(MACH64:sparcv9=sparcv9))
endif

would that be a sufficient fix for you?

mtelka:
> Shall I fix #19420 with this: ifeq ($MACH64), amd64) GNU_CPU = $(if f i l t e r (GCC_VERSION),3 4 7),$(MACH),$(MACH64:amd64=x86_64)) else GNU_CPU = $(if f i l t e r (GCC_VERSION),3 4 7),$(MACH),$(MACH64:sparcv9=sparcv9)) endif
> 
> would that be a sufficient fix for you?

No, this does not work either.  This PR works.  You do not need to do anything.

klausz65:
But then sparc will not have a GNU_CPU variable in shared-makros - correct?

mtelka:
> But then sparc will not have a GNU_CPU variable in shared-makros - correct?

No.

klausz65:
Okay, please let me double check this. The first component which uses GNU_CPU is ruby-32 - this will take some time on my SPARC build server.

mtelka:
> Okay, please let me double check this. The first component which uses GNU_CPU is ruby-32 - this will take some time on my SPARC build server.

You do not need to do any complex tests.  Simply run `env - gmake print-value-GNU_CPU` (in any component dir) and you'll see the result immediately.

klausz65:
Woouu, that's a real good one - thanks quite much, didn't knew about that one!

mtelka:
> ... didn't knew about that one!

I _know_ :-(.

klausz65:
Okay, now I get it that filter sets x86_64 only if MACH64 equals amd64 and on sparc it does nothing, but where the hell is sparcv9 now set for GNU_CPU - please enlight me in this!

mtelka:
> Okay, now I get it that filter sets x86_64 only if MACH64 equals amd64 and on sparc it does nothing, but where the hell is sparcv9 now set for GNU_CPU - please enlight me in this!

There is only one line in `oi-userland` where the `GNU_CPU` is set (ignoring the `ruby-32` component).  I'm touching it in this PR.

klausz65:
Okay, thanks - but after all the setting I've done in ruby-32/Makefile should then also be corrected by me right? because on x86_64 I should not have set GNU_CPU at all in the first place, if the shared-macros setting would not have been broken by me. However I need to set it from sparcv9 to sparc because the install-path's for sparc don't conform to the contents of GNU_CPU on sparc, they install as sparc and not as sparcv9.

mtelka:
> Okay, thanks - but after all the setting I've done in ruby-32/Makefile should then also be corrected by me right? because on x86_64 I should not have set GNU_CPU at all in the first place, if the shared-macros setting would not have been broken by me. However I need to set it from sparcv9 to sparc because the install-path's for sparc don't conform to the contents of GNU_CPU on sparc, they install as sparc and not as sparcv9.

In general, you should not set `GNU_CPU` anywhere (unless you really knows what you are doing) because what is done here in `shared-macros.mk` should be okay and sufficient.  The `GNU_CPU` set in `shared-macros.mk` primarily exists for `GNU_TRIPLET` (right above the `GNU_CPU` in `shared-macros.mk`).  You can use `GNU_CPU` for other purposes if you find it useful, but you need to accept the comment right above `GNU_CPU` here in `shared-macros.mk`.  If the `GNU_CPU` value is not what you want/need you are free to invent any other variable locally that would fit your needs better.

klausz65:
Thanks quite much - and sorry that I broke this var, it was really not my intention. Andreas, please go ahead and merge.

PR #19993, opened 2024-12-13
ntp: resection manual-pages according to IPD4, rebuild for library/se…

klausz65:
…curity/openssl-3


components/network/ntp/Makefile
```diff
@@ -44,6 +45,7 @@ COMPONENT_FMRI=	service/network/ntp
 COMPONENT_CLASSIFICATION=	System/Services
 COMPONENT_LICENSE=	ntp license
 COMPONENT_LICENSE_FILE=	COPYRIGHT
+COMPONENT_BUILD_ARGS=	-j20
```
mtelka: Please _never_ do this.  Use `USE_PARALLEL_BUILD` instead.

components/network/ntp/Makefile
```diff
@@ -53,6 +55,7 @@ include $(WS_MAKE_RULES)/common.mk
 
 # We modified configure.ac and m4 macros
 COMPONENT_PREP_ACTION = ( cd $(@D) && PATH="$(PATH)" ./bootstrap )
+COMPONENT_POST_CONFIGURE_ACTION += ( cd $(BUILD_DIR_64)/ntpd; gmake ntp_parser.h)
```
mtelka: Please add a comment explaining why is this needed.  Thank you.

components/network/ntp/patches/10-etcfix.patch
```diff
@@ -17,9 +17,13 @@ can go away.
...
-@@ -66,7 +66,7 @@
- static struct audio_device device; /* audio device ident */
- #endif /* HAVE_SYS_AUDIOIO_H */
+@@ -62,11 +62,8 @@
+ /*
+  * Global variables
+  */
+-#ifdef HAVE_SYS_AUDIOIO_H
+-static struct audio_device device; /* audio device ident */
+-#endif /* HAVE_SYS_AUDIOIO_H */
```
mtelka: This change is not related to paths customization (see top of this file).  If this is really needed please add such a change as a new patch with a comment explaining why it is needed.  Thank you.

components/network/ntp/ntp.p5m
```diff
@@ -19,7 +19,7 @@
 # CDDL HEADER END
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
-# Copyright (c) 2023, Friedrich Kink. All rights reserved.
```
mtelka: Please never remove copyrights of others without their consent.

components/network/ntp/patches/60-rem-vpath_after-ntpd.patch
```diff
@@ -0,0 +1,11 @@
+--- ntpd/Makefile.am.orig	2023-05-31 12:31:38.000000000 +0200
```
mtelka: Please add a comment explaining why we do need this.  Thank you.

components/network/ntp/patches/70-warrnigs.patch
```diff
@@ -0,0 +1,62 @@
+--- util/ntptime.c.orig	2020-03-04 00:41:29.000000000 +0100
```
mtelka: Please add a comment explaining why we do need this.  Thank you.

mtelka (changes requested):

components/network/ntp/patches/70-warrnigs.patch
mtelka: BTW, the patch filename contains typo (two `r`, missing `n`).

components/network/ntp/Makefile
klausz65: removed

components/network/ntp/Makefile
klausz65: removed

components/network/ntp/ntp.p5m
klausz65: changed back

components/network/ntp/patches/10-etcfix.patch
klausz65: removed

components/network/ntp/patches/70-warrnigs.patch
klausz65: removed

components/network/ntp/patches/60-rem-vpath_after-ntpd.patch
klausz65: removed

klausz65:
The new results-all.master test file needs pull #20032

mtelka:
> The new results-all.master test file needs pull #20032

It does not.  You can simply do something like `COMPONENT_TEST_ARGS += -k` here.

components/network/ntp/Makefile
```diff
@@ -26,12 +26,14 @@
 # Copyright (c) 2023, Niklas Poslovski
 #
 
+# for the moment the SHAKE128 digest fails
+USE_DEFAULT_TEST_TRANSFORMS= yes
```
mtelka: Is the `results-all.master` reproducible with this?

components/network/ntp/Makefile
```diff
@@ -26,12 +26,14 @@
 # Copyright (c) 2023, Niklas Poslovski
 #
 
+# for the moment the SHAKE128 digest fails
```
mtelka: I believe this comment does not belong here.  Why do we need it at all?  Isn't the test failure visible in `results-all.master`.  If not, why?

Isn't it possible to skip the test instead?

components/network/ntp/ntp.p5m
mtelka: Where is Friedrich Kink?  What exactly you added back?

components/network/ntp/test/results-all.master
```diff
@@ -1,3380 +1,107 @@
...
-1 XPASS 
-0 Ignored 
-
-FAIL
-PASS test-packetProcessing (exit status: 0)
-=========================================
-   sntp : tests/test-suite.log
-=========================================
-
+FAIL: test-crypto
```
mtelka: With the new `results-all.master` I miss the list of tests.  It is always good to have such a list so we are able to detect when the number of failing tests didn't change, but instead of `test1` failure we now see the `test2` failing.

mtelka (changes requested):

components/network/ntp/ntp.p5m
klausz65: have forgotten git add for ntp.p5m

components/network/ntp/Makefile
klausz65: yes

components/network/ntp/Makefile
klausz65: removed

components/network/ntp/test/results-all.master
klausz65: I don't get you here

mtelka: > I don't get you here

Since you reverted this change then it makes little sense to discuss it.

components/network/ntp/Makefile
```diff
@@ -54,6 +54,8 @@ include $(WS_MAKE_RULES)/common.mk
 # We modified configure.ac and m4 macros
 COMPONENT_PREP_ACTION = ( cd $(@D) && PATH="$(PATH)" ./bootstrap )
 
+COMPONENT_TEST_ARGS += -k
```
mtelka, line 59: Please add a comment explaining why is this needed.  Thank you.

components/network/ntp/test/results-all.master
```diff
@@ -1,22 +1,25 @@
-crypto.c:14:test_MakeSHAKE128Mac:PASS
+make_mac: MAC SHAKE128 Digest Final failed.
```
mtelka, line 1: This is apparently a regression.  I've no idea what would be the impact of this but it could possibly cause that something won't work as expected.  Since `ntpd` is core part of system (and I would like to avoid reports that people are no longer able to sync their time properly) then please try to find why is this failing now.

Thank you.

mtelka (changes requested):

components/network/ntp/test/results-all.master
klausz65: I allready  tried, but sorry I'm too dump to fix this, if you want to have that IPD4 part, you'll have to take this PR as it is otherwise tell me and I'll cancel it.

mtelka: > I allready tried, but sorry I'm too dump to fix this, if you want to have that IPD4 part, you'll have to take this PR as it is otherwise tell me and I'll cancel it.

Quick search revealed that the problem is very likely caused by a change in OpenSSL 3.4.

Related links:

- https://github.com/openssl/technical-policies/issues/93
- https://github.com/erlang/otp/issues/8997
- https://github.com/erlang/otp/pull/9002

klausz65: I can't fix this - sorry

Bill-Sommerfeld: Pull request #21930 should fix the SHAKE-128 regressions in ntp.

PR #20032, opened 2024-12-14
shared-macros.mk: add gmake -k flag to test target, to be able to tes…

klausz65:
…t as much as possible

### DIFF
--- a/make-rules/shared-macros.mk
+++ b/make-rules/shared-macros.mk
@@ -561,8 +561,9 @@ COMPONENT_TEST_ENV_CMD =	$(ENV)
 # set the default command to use for test of the component
 COMPONENT_TEST_CMD =	$(GMAKE)
 
-# set the default target for test of the component
-COMPONENT_TEST_TARGETS =	check
+# set the default target for test of the component,
+# add -k flag to gmake, to test as much as possible.
+COMPONENT_TEST_TARGETS =	check -k
 
 # set the default directory for test of the component
 COMPONENT_TEST_DIR =	$(@D)


make-rules/shared-macros.mk
```diff
@@ -561,8 +561,9 @@ COMPONENT_TEST_ENV_CMD =	$(ENV)
 # set the default command to use for test of the component
 COMPONENT_TEST_CMD =	$(GMAKE)
 
-# set the default target for test of the component
-COMPONENT_TEST_TARGETS =	check
+# set the default target for test of the component,
+# add -k flag to gmake, to test as much as possible.
+COMPONENT_TEST_TARGETS =	check -k
```
mtelka, line 566: If we decide to do this then I think `-k` should go into `COMPONENT_TEST_ARGS` instead.  However, I'm not sure it is good idea.  For Python projects I do exactly the opposite:
https://github.com/OpenIndiana/oi-userland/blob/10217e3733a099c6cf258b50191236260d2e4545/make-rules/setup.py.mk#L490-L493

I do that because my main goal is to make sure that all tests pass (not only Python or Perl ones, but all).  If some tests do not pass and I do not know how to make it working, then I report such failure(s) upstream and then I make the test to skip.  When I debug the test failures and so I need to see all of them then I simply run `env - gmake PYTEST_FASTFAIL= test`.

So I think the fast fail for tests is good.  I also think that adding an option to do not do the fast fail is good idea.  Please try to do something like that instead.

Thank you.

mtelka (changes requested):

make-rules/shared-macros.mk
klausz65: will put that change into the  ntp component Makefile - thanks.

PR #20033, opened 2024-12-14
sysding: follow IPD4 man-page change

klausz65:


components/openindiana/sysding/Makefile
```diff
@@ -18,7 +18,6 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		sysding
 COMPONENT_SRC=		$(COMPONENT_NAME)
 COMPONENT_VERSION=	0.5.11
-COMPONENT_REVISION=	1
 COMPONENT_LICENSE=	CDDL
```
AndWac, line 21: Why have you deleted COMPONENT_REVISION?

klausz65: Because in the previous change of that component I thought the revision has to be upgraded as usual, but the revision comes out of the merges in the underlying repo of sysding - and that one has already been updated by me. So there is no need to raise REVISIONS in that Makefile at all.

PR #20085, opened 2024-12-17
Fix include not packaged by library/objfw

Midar:


mtelka:
I think the `sample-manifest.p5m` update is missing.

mtelka:
Also, if you publish on the latest OI with up-to-date oi-userland repo you should get a bit different `pkg5` file.

Midar:
Updated

PR #20086, opened 2024-12-17
bhyve-fw: drop CSM

mtelka:
The CSM support got [removed from the FreeBSD ports](https://cgit.freebsd.org/ports/commit/?id=2f2c84c6fc9514b1965aafb8a58a7228136a60cd) in April 2024.  It got also [removed from the Tianocore upstream](https://github.com/tianocore/edk2/pull/5122) in December 2023.  The best option here is simply to follow and to accept the fact that legacy BIOS boot support in bhyve is dead.


Toasterson:
@citrus-it are you also dropping csm from.your bhyve or is there a effort around to keep it going we are not aware of?

Toasterson:
I could confrm that indeed this has happened upstream. And FreeBSD people are also a bit sad about it. It looks like we are not the only ones missing legacy BIOS boot in bHyve now. For now we should take this as is and if a upstream SeaBIOS port is available we have a seperate package anyway. We should announce this as this is a breaking change for people that configured their VM's with Legacy Boot. After this they cannot boot any longer unless we leave the blob in the OS filesystem.

mtelka:
I can send something like this to openindiana-discuss@openindiana.org once this PR is merged:
```
Subject: Flag Day: EOL of legacy bhyve BIOS (CSM) boot

With the https://github.com/OpenIndiana/oi-userland/pull/20086 merged
the legacy bhyve BIOS boot support (CSM) has been removed.  This is a flag day
for those who configured their bhyve VMs to use such ancient boot method.

To solve or workaround this you can:

1) reconfigure your VMs to use UEFI boot, or
2) save the CSM firmware (located in /usr/share/bhyve/firmware/) before
   the system/bhyve/firmware package update and then copy it back after
   the package update.

We removed the CSM firmware because it got dropped upstream and we
have no better option than to follow this here too.

Thank you.
```

citrus-it:
> @citrus-it are you also dropping csm from.your bhyve or is there a effort around to keep it going we are not aware of?

Sorry, missed this earlier. No, we are continuing to ship the CSM ROMs in OmniOS, although they are based on the old 2014 version.

https://github.com/omniosorg/omnios-build/blob/master/build/bhyve-fw/build.sh#L113-L116

PR #20105, opened 2024-12-19
gettext: fix build with gcc-14; enhance test environment

AndWac:


components/text/gettext/Makefile
```diff
@@ -42,6 +43,11 @@ COMPONENT_LICENSE_FILE=	COPYING
 
 include $(WS_MAKE_RULES)/common.mk
 
+PATH= $(PATH.gnu)
```
mtelka, line 47: Could you please add a comment explaining why is this needed?  Thank you.

AndWac: I thought that it was obvious that this change made two mode tests pass instead of skip.

mtelka: No, it is not.  Thanks for the comment.   However ... (to be continued)

components/text/gettext/Makefile
```diff
@@ -42,6 +43,12 @@ COMPONENT_LICENSE_FILE=	COPYING
 
 include $(WS_MAKE_RULES)/common.mk
 
+# Our sed misbehaves during the tests and let msgfilter-2 and msgfilter-4 tests being skipped.
```
mtelka, line 46: ... if our sed misbehaves, then it is a bug in our sed.  I'd expect some reference to an illumos bug then.

Isn't it instead that GNU gettext testing simply uses some sed GNUisms?  If that's the case then maybe testing should be fixed.

PR #20115, opened 2024-12-21
flamegraph: Fix download hash

klausz65:


mtelka:
If you need to update a Python project, then please use `python-integrate-project` to do so properly.  Thank you.

klausz65:
Hm:
./python-integrate-project -d /ws/klausz/oi-userland/components/python/flamegraph -f Makefile -l 2 -o library/python/flamegraph-37 -u flamegraph
ERROR: Too many arguments

Usage: python-integrate-project [-d DIR] [-f] [-l VERSION] [-o OBSOLETE].. [-u] PROJECT

Which options should I omit?

mtelka:
> Hm: ./python-integrate-project -d /ws/klausz/oi-userland/components/python/flamegraph -f Makefile -l 2 -o library/python/flamegraph-37 -u flamegraph ERROR: Too many arguments
> 
> Usage: python-integrate-project [-d DIR] [-f] [-l VERSION] [-o OBSOLETE].. [-u] PROJECT
> 
> Which options should I omit?

All.  Use the exactly same command as shown in the `Makefile`.  Why do you expect you have to use different command line options than shown in the `Makefile`?

BTW, In your particular case there is extra `Makefile` after `-f`.  The `-f` option does not have a parameter.  But, as I said above, you do not need to (and should not) use `-f` at all.  Especially, when you do not know what the option does.

And, to make the `python-integrate-project` script working properly you should start from scratch, i.e. you should drop all your changes you did for `flamegraph`.  The `python-integrate-project` script will do everything for you and it will do so properly.

klausz65:
Okay, I'll re-try - will drop this PR

klausz65:
As written above

mtelka:
BTW, it is strange that the hash changed.  I compared the old tarball with the new one and their content differs.  So it looks like the `flamegraph` maintainer(s) simply uploaded new tarball with different content without bumping the version number.  This is bad.

Also, it looks like this project is dead upstream and since nothing in OpenIndiana depends on it then maybe we should simply obsolete it, instead of riding on a dead horse.

PR #20116, opened 2024-12-21
component(developer/golang-123): Add Component

LophophanesCristatus:


Toasterson:
I wouldn't worry too much about the testing infrastructure I know the tests pass and we have a golang testing vm dedicated to the upstream project so these tests are more a formality. LGTM and thankyou very much for the contribution. Hope to see more and let me or anybody on the oi-dev mailinglist know if you get stuck so we can help.

-Toasty

PR #20128, opened 2024-12-22
xorriso: Fix download hash and package dependencies

klausz65:


mtelka:
For the record: the hash is changing because version 1.5.7 is devel snapshot.

PR #20132, opened 2024-12-22
minimal_amd64: add driver/network/igc

iigs:
Add igc to install media and installed systems after next update.


AndWac:
You also need to touch the corresponding Makefile. In this case you'll need to increment the COMPONENT_VERSION otherwise no jenkins job is being run on this.

iigs:
Please advise if any other Makefile needs to be changed.

PR #20147, opened 2024-12-25
libpoppler: update to 24.12.0

AndWac:


AndWac:
inkscape and libreoffice need patches to build against this.

PR #20193, opened 2024-12-29
add openjdk-25-3

AndWac:


mtelka:
@AndWac I'm curious why do we need development versions of OpenJDK?  Especially in so early development phase.  The same question stands for openjdk-24 integrated via #18530 too.  Thank you.

PR #20231, opened 2025-01-01
sysutils/stunnel: update to 5.74

RAC-oi:
some tests are skipped but all remaining ones pass. "gmake test" quits with an error anyway for whatever reason.

### DIFF
--- a/components/sysutils/stunnel/Makefile
+++ b/components/sysutils/stunnel/Makefile
@@ -12,27 +12,27 @@
 #
 # Copyright (c) 2013, Oracle and/or its affiliates. All rights reserved.
 # copyright (c) 2021, Tim Mooney.  All rights reserved.
+# Copyright (c) 2025 Goetz Fischer, R-A-C. All rights reserved.
 #
 
-OPENSSL_VERSION= 3.1
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         stunnel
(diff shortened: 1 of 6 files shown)


AndWac:
You should try to make the test results more complete. The project changed its output so we should adapt the filter rules.
Eg. I tried these with success:
COMPONENT_TEST_TRANSFORMS += \
	'-n ' \
	'-e "/^[0-9]/p" ' \
	'-e "/succeeded/p" ' \
	'-e "/skipped/p" ' \
	'-e "/failed/p" ' \
	'-e "/^ /p" ' \
	'-e "/^Summary:/p" '

Then copy the new test results from build/amd64/results-64.snapshot to test/results-all.master.
After that gmake test should run fine.
Furthermore, unexport SHELLOPTS is not necessary anymore and should be removed from the Makefile.

RAC-oi:
didn't work. after running "gmake test" with the new COMPONENT_TEST_TRANSFORMS there's no build/amd64/results-64.snapshot.

AndWac:
Sorry, I have the wrong path for the snapshot file. The correct one is build/test/amd64/.

RAC-oi:
it was my fault as well. the path is actually displayed above the results. the file in question is: build/test/amd64/results-64.snapshot
now "gmake test" passes and thanks for the COMPONENT_TEST_TRANSFORMS!

PR #20241, opened 2025-01-03
boost: Fix dependencies to new ICU library

klausz65:


mtelka:
While here, could you please try to update boost?

klausz65:
Sorry, I'm really much too busy to keep SPARC alive, just wanted to get it going again.

PR #20253, opened 2025-01-04
windowmaker successfully compiled and ported to openindiana

lumentrail:
This is a successful compilation and port of the windowmaker window manager to openindiana.   After the package is installed, it is necessary to edit ~./xprofile and add the following lines in order for lightDM to launch windowmaker instead of the default.

xrandr --output default --mode (insert resolution here)
exec /usr/local/bin/wmaker


components/desktop/wmaker/Makefile
```diff
@@ -0,0 +1,69 @@
...
+# Copyright 2021 Gary Mills
+# Copyright (c) 2017 Dariusz Sendkowski
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=         WindowMaker
+COMPONENT_VERSION=      0.96.0
+COMPONENT_REVISION=     1
+COMPONENT_SUMMARY=      Window Maker is an X11 window manager originally designed to provide integration support for the GNUstep Desktop Environment, although it can run stand alone. In every way possible, it reproduces the elegant look and feel of the NeXTSTEP user interface.
```
AndWac, line 22: Please delete COMPONENT_REVISION. Incrementing it is only needed when you need to make changes to the package without changing the COMPONENT_VERSION.

components/desktop/wmaker/Makefile
```diff
@@ -0,0 +1,69 @@
...
+COMPONENT_INSTALL_ENV=	DEST_DIR=$(PROTO_DIR)
+
+#CONFIGURE_OPTIONS += --sysconfdir=$(ETCDIR)
+#CONFIGURE_OPTIONS += --disable-static
+#CONFIGURE_OPTIONS += --disable-imlib2
+
+#CONFIGURE_ENV+= PYTHON=$(PYTHON)
+
+#LDFLAGS += -lsocket
+
```
AndWac, line 46: Why are there commented out lines? Please remove them.

AndWac:
Please also add manifests/sample-manifest.p5m and pkg5 files to your PR.

AndWac:
I cannot build it here:
| echo '/export/home/andreas/oi-userland/components/desktop/wmaker/WindowMaker-0.96.0/util/'`wmiv.c
/export/home/andreas/oi-userland/components/desktop/wmaker/WindowMaker-0.96.0/util/wmiv.c:27:10: fatal error: WINGs/WINGsP.h: No such file or directory
   27 | #include <WINGs/WINGsP.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [Makefile:794: wmiv-wmiv.o] Error 1

components/desktop/wmaker/Makefile
```diff
@@ -0,0 +1,69 @@
...
+COMPONENT_ARCHIVE_HASH= sha256:4fe130ba23cf4aa21c156ec8f01f748df537d0604ec06c6bbcec896df1926f6d
+COMPONENT_ARCHIVE_URL=  https://www.windowmaker.org/pub/source/release/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI=         desktop/window-manager/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=System/X11
+COMPONENT_LICENSE=      GPL	
+COMPONENT_LICENSE_FILE= COPYING
+
+TEST_TARGET= $(NO_TESTS)
+include $(WS_MAKE_RULES)/common.mk
+
```
AndWac, line 35: Looking at the output of config.log makes me want to add the following lines to the Makefile:
\# GNU gettext is needed                                                                                                                                                              
PATH= $(PATH.gnu)

lumentrail: Hello AndWac. I've cleaned up the Makefile (removed extra lines).  Also working on a few other things but need help.  

Here's the issue:

I am able to build the package outside of oi-userland. However a problem arises when building within the oi-userland. When running the gmake sample-manifest process, it pulls down the archive from the web and creates a build/amd64 staging location where it places the files during the build. That's all fine and well. However, within the build/amd64 directory is a subdirectory called _WINGs_.  Within _WINGs_ directory is another _WINGs_ and it contains the _WINGsP.h_ file that the compiler is looking for, however the include is looking for it to be within _WINGs/WINGsP.h_ but the file is not there -- it exists another level deeper.  This is what's causing the error.  It's almost as though I need some sort of custom build rule to manually copy the compiled _WINGsP.h_  from _build/amd64/WINGs/WINGs_ into the parent directory _build/amd64/WINGs_ and then continue the compile.  

Here is a workaround:

1.  From within the wmaker directory run:  cp ~/oi-userland/components/software/wmaker/build/amd64/WINGs/WINGs/WINGsP.h ~/oi-userland/components/software/wmaker/build/amd64/WINGs

2.  Run gmake sample-manifest

You will then find that it then successfully compiles.  I'm not entirely sure how or why it fails building within the oi-userland but not outside of it.  I'm trying to understand what's going on in the oi-userland environment specifically that's causing this issue.

What I'm thinking of is to implement this workaround in the Makefile such that I create a post build script which performs the above task.  Or perhaps some sort of custom script that hooks into the build process and moves that WINGsP.h file one level up when possible?  I just don't see why I have to go through all this trouble if the application builds successfully outside of oi-userland.  Is it an easy fix that I'm just not seeing right now?

lumentrail: Or I was thinking a component pre build action to edit the wmiv.c on the source and point the #include it to the WINGs/WINGs/WINGsP.h location.  Sort of like this:

```

COMPONENT_PRE_BUILD_ACTION =  \
     ( sed -i 's|#include <WINGs/WINGsP.h>|#include <WINGs/WINGs/WINGsP.h>|' $(SOURCE_DIR)/util/wmiv.c ; )

```
          
Not sure if that's the correct syntax.

mtelka:
@influxparadise please mark this PR as a Draft because it is clearly not ready yet for the merge.  Thank you.

components/desktop/wmaker/Makefile
lumentrail: I determined the fix. I needed to create a pre build script to the Makefile in order to ./configure at source && gmake under WINGs/WINGs which produces the WINGsP.h file.  The compiler then finds the needed included file as it is pre-compiled before the rest of the build transpires. oi-userland takes care of the rest of the build. This is one way to do the workaround I think it's the cleanest. If there's a better way, let me know.

PR #20254, opened 2025-01-04
nmap: update to 7.95, migrate pcre to pcre2 and openssl-1 to openssl-…

klausz65:
…3, enable ndiff.

### DIFF
--- a/components/sysutils/nmap/Makefile
+++ b/components/sysutils/nmap/Makefile
@@ -23,14 +23,13 @@
 # Copyright (c) 2011, 2012, Oracle and/or its affiliates. All rights reserved.
 # Copyright 2016 Jim Klimov
 # Copyright 2022 Nona Hansel
+# Copyright 2025 Klaus Ziegler
 #
 
-BUILD_BITS=64
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= 		nmap
-COMPONENT_VERSION=		7.80
-COMPONENT_REVISION=		2
+COMPONENT_VERSION=		7.95
 COMPONENT_PROJECT_URL=	http://nmap.org/
 COMPONENT_FMRI=			diagnostic/$(COMPONENT_NAME)
 COMPONENT_CLASSIFICATION=	System/Administration and Configuration
@@ -52,19 +51,20 @@ PKG_OPTIONS +=	-D COMPONENT_DESCRIPTION="$(COMPONENT_DESCRIPTION)"
 
 COMPONENT_PREP_ACTION = (cd $(@D)/libdnet-stripped && autoconf)
 
-COMPONENT_PRE_CONFIGURE_ACTION = $(CP) -rp $(SOURCE_DIR)/. $(@D)
+CLONEY_MODE = copy
 
-CONFIGURE_OPTIONS  +=   CPPFLAGS="$(CPPFLAGS)"
+CONFIGURE_OPTIONS  +=   CPPFLAGS="-I$(USRINCDIR)/pcre"
+# otherwise it picks up the 32 bit library instead
+CONFIGURE_OPTIONS  +=   LDFLAGS="-L$(OPENSSL_LIBDIR)"
 CONFIGURE_OPTIONS  +=   LIBS="-lsocket -lnsl"
 CONFIGURE_OPTIONS  +=   --srcdir=$(BUILD_DIR_$(BITS))
 CONFIGURE_OPTIONS  +=   --localstatedir=/var
 CONFIGURE_OPTIONS  +=   --without-localdirs
 CONFIGURE_OPTIONS  +=   --without-ndiff
 CONFIGURE_OPTIONS  +=   --without-zenmap
-CONFIGURE_OPTIONS  +=   --with-openssl
+CONFIGURE_OPTIONS  +=   --with-openssl=$(OPENSSL_PREFIX)
 CONFIGURE_OPTIONS  +=   --with-libpcap
-CONFIGURE_OPTIONS  +=   --with-libpcre
-CONFIGURE_OPTIONS  +=   --with-liblua
+CONFIGURE_OPTIONS  +=   --with-libpcre=$(USRDIR)
 CONFIGURE_OPTIONS  +=   --with-ncat
 
 TEST_TARGET= $(NO_TESTS)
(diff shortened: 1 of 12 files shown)


components/sysutils/nmap/Makefile
```diff
@@ -54,29 +52,33 @@ COMPONENT_PREP_ACTION = (cd $(@D)/libdnet-stripped && autoconf)
 
 COMPONENT_PRE_CONFIGURE_ACTION = $(CP) -rp $(SOURCE_DIR)/. $(@D)
 
-CONFIGURE_OPTIONS  +=   CPPFLAGS="$(CPPFLAGS)"
+COMPONENT_POST_INSTALL_ACTION = $(GSED) \
+    -i -e 's:\#\!/usr/bin/python3:\#\!/usr/bin/python$(PYTHON_VERSION):' $(PROTOUSRBINDIR)/ndiff ; \
```
mtelka: Please try to use `PYTHON_SCRIPTS` instead of this.

components/sysutils/nmap/Makefile
```diff
@@ -54,29 +52,33 @@ COMPONENT_PREP_ACTION = (cd $(@D)/libdnet-stripped && autoconf)
...
+    $(MV) $(PROTOUSRLIBDIR)/python$(PYTHON_VERSION)/site-packages $(PROTO_DIR)/$(PYTHON_VENDOR_PACKAGES)
+
+CONFIGURE_OPTIONS  +=   CPPFLAGS="-I$(USRINCDIR)/pcre"
+CONFIGURE_OPTIONS  +=   LDFLAGS="-L$(OPENSSL_LIBDIR.64)"
 CONFIGURE_OPTIONS  +=   LIBS="-lsocket -lnsl"
 CONFIGURE_OPTIONS  +=   --srcdir=$(BUILD_DIR_$(BITS))
 CONFIGURE_OPTIONS  +=   --localstatedir=/var
```
mtelka, line 61: Is this line really needed?

components/sysutils/nmap/patches/01-climits.patch
```diff
@@ -0,0 +1,21 @@
+--- nmap-7.95/charpool.cc.orig	2024-02-28 20:46:45.000000000 +0200
```
mtelka, line 1: Hint: you can use `env - gmake refresh-patches` to normalize patches.

components/sysutils/nmap/Makefile
```diff
@@ -54,29 +52,33 @@ COMPONENT_PREP_ACTION = (cd $(@D)/libdnet-stripped && autoconf)
 
 COMPONENT_PRE_CONFIGURE_ACTION = $(CP) -rp $(SOURCE_DIR)/. $(@D)
 
-CONFIGURE_OPTIONS  +=   CPPFLAGS="$(CPPFLAGS)"
+COMPONENT_POST_INSTALL_ACTION = $(GSED) \
```
mtelka: Is there any reason that `+=` is not used here?

components/sysutils/nmap/Makefile
```diff
@@ -54,29 +52,33 @@ COMPONENT_PREP_ACTION = (cd $(@D)/libdnet-stripped && autoconf)
 
 COMPONENT_PRE_CONFIGURE_ACTION = $(CP) -rp $(SOURCE_DIR)/. $(@D)
```
mtelka: Instead of this you could use `CLONEY_MODE = copy` to achieve the same, I think.

components/sysutils/nmap/Makefile
klausz65: removed

components/sysutils/nmap/Makefile
klausz65: removed

components/sysutils/nmap/Makefile
klausz65: removed

components/sysutils/nmap/patches/01-climits.patch
klausz65: thanks

components/sysutils/nmap/Makefile
klausz65: it just works

mtelka: Yes, it works, but `CLONEY_MODE = copy` is far far better. :-)

components/sysutils/nmap/Makefile
```diff
@@ -54,29 +52,29 @@ COMPONENT_PREP_ACTION = (cd $(@D)/libdnet-stripped && autoconf)
...
 
-CONFIGURE_OPTIONS  +=   CPPFLAGS="$(CPPFLAGS)"
+CONFIGURE_OPTIONS  +=   CPPFLAGS="-I$(USRINCDIR)/pcre"
+CONFIGURE_OPTIONS  +=   LDFLAGS="-L$(OPENSSL_LIBDIR.64)"
```
mtelka: Is this really needed with `--with-openssl=$(OPENSSL_PREFIX)` below?

klausz65: Yes, because otherwise it picks up the 32 bit library instead.

mtelka: I see.  _(rant on)_ If they would use pkgconfig then we would not need to do this at all _(rant off)_.  Anyway, I assume you can safely drop `.64` from this line.  Also, please add short comment above this line explaining why we do need it (something like `# otherwise it picks up the 32 bit library instead` would be okay).  Thank you.

components/sysutils/nmap/Makefile
klausz65: done

components/sysutils/nmap/Makefile
klausz65: done

components/sysutils/nmap/Makefile
```diff
@@ -39,7 +38,6 @@ COMPONENT_DESCRIPTION=	Nmap is useful for inventorying the network, managing \
 service upgrade schedules, and monitoring host or service uptime
 COMPONENT_SRC=			$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=		$(COMPONENT_SRC).tar.bz2
-COMPONENT_ARCHIVE_HASH=	sha256:fcfa5a0e42099e12e4bf7a68ebe6fde05553383a682e816a7ec9256ab4773faa
```
mtelka: I just noticed this.  I think we should keep the hash here to make sure the tarball does not change in future.

klausz65: done

PR #20259, opened 2025-01-05
transforms/defaults: drop vendor mediator priority for java

mtelka:
I do not see any reason why we do want/need to prefer Java 17.  If there is any please let me know.

### DIFF
--- a/transforms/defaults
+++ b/transforms/defaults
@@ -151,11 +151,6 @@ set name=variant.arch value=$(MACH)
 <transform link mediator=ruby mediator-version=2.3 -> default mediator-priority vendor>
 
 #
-# Set the default Java for mediated links
-#
-<transform link mediator=java mediator-version=17 -> default mediator-priority vendor>
-
-#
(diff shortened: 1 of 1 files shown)


AndWac:
I don't have an answer why we should prefer Java 17 over other versions but looking at the other settings here makes me wonder why we have them at all, especially
#                                                                                                                                    
# Set the default Apache for mediated links                                                                                          
#                                                                                                                                    
<transform link mediator=apache mediator-version=2.2 -> default mediator-priority vendor>                                            

The only apache version we have atm is 2.4 so 2.2 doesn't make sense at all.
                                                                                                                                     
#                                                                                                                                    
# Set the default Ruby for mediated links                                                                                            
#                                                                                                                                    
<transform link mediator=ruby mediator-version=2.3 -> default mediator-priority vendor>                                              

Ruby 2.3 is fairly outdated and should be removed if possible.

mtelka:
Yes, I agree.  I'll do the cleanup later.

PR #20265, opened 2025-01-05
Enable database/opensearch

mtelka:
I suspect the opensearch build [failed previously](https://hipster.openindiana.org/jenkins/job/oi-userland/12648/) because it was built concurrently with the heavyweight openjdk-23 that eat most of the build server resources.  Let's try to build the opensearch again and alone this time.

### DIFF
--- a/components/components.ignore
+++ b/components/components.ignore
@@ -20,7 +20,6 @@
 # /^package$/d
 #
 # Temporarily disable failing builds:
-/^database\/opensearch/d
 /^web\/php\/php-8_2-ext-redis$/d
 # Our build server fails to build the following packages (the builds succeed on other build systems):
 /^desktop\/libayatana-appindicator$/d


AndWac:
I fixed the problem on the build server by explicitly changing the java mediator to 21. With java 17 the build failed repeatedly.

mtelka:
> I fixed the problem on the build server by explicitly changing the java mediator to 21. With java 17 the build failed repeatedly.

Interesting and strange.  The opensearch build should not depend on java mediator.

mtelka:
Ha, the `gradlew` uses the `~/.gradle` directory (likely as a cache) so maybe there was something at the build server in  `~/.gradle` that was incompatible with the `opensearch` build.  Maybe this is also the reason why the `opensearch` build failed for me when I tried with Java 21:

https://github.com/OpenIndiana/oi-userland/blob/eba576e298c45a84161733955a9fe0e1173c510c/components/database/opensearch/Makefile#L35-L36

We should try to force gradle to use different location - some directory under `components/database/opensearch/build`.

... I'm going to do some experiments.

mtelka:
Okay, testing showed that it is easy to relocate the `.gradle` directory using `GRADLE_USER_HOME` (a PR will follow soon).  I also confirmed that building with OpenJDK 21 fails and so we need to force version 17.

PR #20270, opened 2025-01-06
sqlite: update to 3.47.2

iigs:
tcl component did not get compiled with -fPIC.  A new tcl script was added for the 3.47 branch--force -fPIC usage calling the scripts and within the script.

Changed ENV to get more information from test results.

Removed unsupported CONFIGURE_OPTIONS.

### DIFF
--- a/components/database/sqlite/Makefile
+++ b/components/database/sqlite/Makefile
@@ -23,16 +23,17 @@
 #
 
 BUILD_BITS= 64_and_32
+ENV=/usr/bin/env
 USE_PARALLEL_BUILD= yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         sqlite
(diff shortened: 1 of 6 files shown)


components/database/sqlite/Makefile
```diff
@@ -23,16 +23,17 @@
 #
 
 BUILD_BITS= 64_and_32
+ENV=/usr/bin/env
```
mtelka, line 26: Please do not do this, unless _really_ needed.  If it is really needed, then please add a comment explaining why it is needed.  Thank you.

mtelka: In addition, it is a no-op because `ENV` is set unconditionally in `shared-macros.mk` anyway.  So please do not bother with a comment and simply remove this line.  Thank you.

PR #20303, opened 2025-01-06
windowmaker: fix fmri to conform name conventions

lumentrail:
To improve package searchability and ensure consistency, the FMRI of desktop/window-manager/WindowMaker is being obsoleted and transitioned to "desktop/window-manager/wmaker." This change aims to align with oi naming conventions and simplify identification of the package.


components/desktop/wmaker/history
```diff
@@ -0,0 +1 @@
+desktop/window-manager/WindowMaker@0.96.0,5.11-2025.0.0.0 /desktop/window-manager/wmaker
```
AndWac: Two changes are necessary here to make it
desktop/window-manager/WindowMaker@0.96.0,5.11-2025.0.0.1 desktop/window-manager/wmaker

(COMPONENT_REVISION needs to be increased from 0 to 1 to make this version the latest and remove the leading / from the forwarded fmri)

lumentrail: Syntax in the history file has been corrected.

PR #20351, opened 2025-01-09
gnome-commander: update to 1.18.1

mtelka:


AndWac:
Have you tested the new version? I had an upgrade but that one didn't work.

mtelka:
All tests passed so I assume it works.

AndWac:
No, it doesn't. I have installed it on my machine. No window opens when starting it.

mtelka:
Right, confirmed here too.  It does not work.

mtelka:
Similar problem: https://gitlab.gnome.org/GNOME/gnome-commander/-/issues/161

mtelka:
With the [proposed fix](https://gitlab.gnome.org/GNOME/gnome-commander/-/merge_requests/213) for [the startup issue](https://gitlab.gnome.org/GNOME/gnome-commander/-/issues/161) the `gnome-commander` 1.18.1 started to work properly so this PR is ready for merge.

PR #20352, opened 2025-01-09
shared-macros.mk: add more definitions for ruby

AndWac:


make-rules/shared-macros.mk
```diff
@@ -753,9 +753,12 @@ RUBY_VERSION = 3.2
 RUBY_LIB_VERSION.2.3 = 2.3.0
 RUBY_LIB_VERSION.3.2 = 3.2.0
 
-RUBY =		/usr/ruby/$(RUBY_VERSION)/bin/ruby
+RUBY_PREFIX = /usr/ruby/$(RUBY_VERSION)
+RUBY = $(RUBY_PREFIX)/bin/ruby
+RUBY_LIBDIR = $(RUBY_PREFIX)/lib/$(MACH64)
 RUBY_LIB_VERSION = $(RUBY_LIB_VERSION.$(RUBY_VERSION))
-PATH.prepend += /usr/ruby/$(RUBY_VERSION)/bin
+RUBY_PKG_CONFIG_PATH = $(RUBY_LIB)/pkgconfig
```
mtelka, line 760: Where is `RUBY_LIB` defined?

While looking deeper I believe all files from `/usr/ruby/*/lib/amd64/pkgconfig` should be moved to  `/usr/lib/amd64/pkgconfig` and so `RUBY_PKG_CONFIG_PATH` should not be needed.

AndWac: Sorry, I missed the update of RUBY_LIB to RUBY_LIBDIR to make it conform to similar definitions.

PR #20355, opened 2025-01-09
shared-macros.mk: fix RUBY_PKG_CONFIG_PATH

AndWac:


mtelka:
Hmmm, we do have ruby pkgconfig files in `/usr/lib`:
```
$ ls -l /usr/lib/amd64/pkgconfig/ | grep ruby
lrwxrwxrwx   1 root     root          49 Feb 20  2024 ruby-3.2.pc -> ../../../ruby/3.2/lib/amd64/pkgconfig/ruby-3.2.pc
lrwxrwxrwx   1 root     root          49 Jan  3 18:49 ruby-3.4.pc -> ../../../ruby/3.4/lib/amd64/pkgconfig/ruby-3.4.pc
$ ls -l /usr/lib/pkgconfig/ | grep ruby
lrwxrwxrwx   1 root     root          40 Feb 20  2024 ruby-2.3.pc -> ../../ruby/2.3/lib/pkgconfig/ruby-2.3.pc
lrwxrwxrwx   1 root     root          30 Feb 20  2024 ruby-3.2.pc -> ../amd64/pkgconfig/ruby-3.2.pc
lrwxrwxrwx   1 root     root          30 Jan  3 18:49 ruby-3.4.pc -> ../amd64/pkgconfig/ruby-3.4.pc
$
```

Why do we need `RUBY_PKG_CONFIG_PATH` then?

AndWac:
Oh, I didn't see it (didn't expect it to be there).

PR #20371, opened 2025-01-10
fuse: drop hidden and unused dependency on GCC 4

mtelka:


mtelka:
The [build failed](https://hipster.openindiana.org/jenkins/job/oi-userland/12692/).  @AndWac please run `env - gmake clobber` in `components/openindiana/fuse` and then restart the build.  Thank you.

mtelka:
It looks like the `fuse` build failure is common.  It failed last time too when we [tried to build it](https://hipster.openindiana.org/jenkins/job/oi-userland/12566/).

AndWac:
This happens only when the build is being run under jenkins. I typically
run gmake publish in a shell and the next successful jenkins run will
update the pkg repository.

Am 10.01.25 um 14:01 schrieb Marcel Telka:

PR #20378, opened 2025-01-10
byacc: initial version 20241231

cstes:
byacc is a modified Berkeley Yacc

there is some argument perhaps to package the old Robert Corbett byacc 1.9.1 as available for example on the FTP site of the University of Utah

https://ftp.math.utah.edu/pub/byacc/index.html

However this pull request packages the byacc as packaged in Debian Linux

Debian Linux has already a long time byacc as a Debian package

Note that my pull request is not a criticism of the Stephen Johnson yacc (illumos yacc) or the bison packages

### DIFF
--- a/components/developer/byacc/Makefile
+++ b/components/developer/byacc/Makefile
@@ -1,0 +1,62 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2025 David Stes
+# IPS Makefile for Berkeley Yacc
+# Berkeley Yacc was written by Robert Corbett according to the README 
+# Also see ACKNOWLEDGEMENTS and AUTHORS in the sources
+# See sources yysccsid[] = \"@(#)yaccpar	1.9 (Berkeley) 02/21/93\" */"
+# The invisible-island.net byacc version is a version of byacc maintained by Thomas E. Dickey
+# The component version is taken from VERSION in that version of byacc
+# Other versions of byacc exist, see for example https://ftp.math.utah.edu/pub/byacc
+# or see for example the FreeBSD/NetBSD/OpenBSD source code or FreeBSD ports collection
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		byacc
+COMPONENT_VERSION=	20241231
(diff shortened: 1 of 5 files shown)


AndWac:
Why do you need it?

cstes:
Because Debian Linux has it.    Because FreeBSD has it.

For FreeBSD see for example

https://cgit.freebsd.org/ports/commit/?id=830d39c127125876bb128417b9f7ffad3be901fc

The changes for byacc are at

https://invisible-island.net/byacc/CHANGES.html

This has contributions from FreeBSD , NetBSD, Debian Linux and Solaris (see the change log).

An IPS package would be convenient to avoid downloading the sources and compiling installing it from source ... which can be done because it compiles easily.

I'm avoiding to discuss here which particular target software can be built using byacc, since the argument is rather that it's in FreeBSD and Debian as a FreeBSD port and a Debian package.

components/developer/byacc/Makefile
```diff
@@ -0,0 +1,51 @@
...
+COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE=	$(COMPONENT_SRC).tgz
+COMPONENT_ARCHIVE_HASH=	\
+    sha256:192c2fae048d4e7f514ba451627f9c4e612765099f819c19191f9fde3e609673
+COMPONENT_ARCHIVE_URL=	$(COMPONENT_PROJECT_URL)/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI=		developer/parser/byacc
+COMPONENT_CLASSIFICATION=	Development/Other Languages
+# Berkeley Yacc is reportedly not BSD software but PublicDomain (see LICENSE file in the package)
+COMPONENT_LICENSE=	PublicDomain
+
```
AndWac, line 50: Please add
COMPONENT_LICENSE_FILE= LICENSE
as the sources already have it and remove the explicit license file you added to the PR.

components/developer/byacc/Makefile
```diff
@@ -0,0 +1,51 @@
...
+#
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2025 David Stes
+#
+
+BUILD_BITS= 64
+include ../../../make-rules/shared-macros.mk
```
AndWac, line 34: BUILD_BITS= 64 is default nowadays and not necessary here.

AndWac:
Please rename test/results-64.master to test/results-all.master
results-all.master is the default for quite some time now. Is you build environment up-to-date?

components/developer/byacc/test/results-64.master
```diff
@@ -0,0 +1,436 @@
+** making check
+** Thu Jan  9 18:55:26 CET 2025
+** testing flags help -z
```
AndWac: You should add a COMPONENT_TEST_TRANSFORMS line to the Makefile in order to get rid of the date and time here. Otherwise every gmake test will fail because of the changed time and date.

components/developer/byacc/Makefile
```diff
@@ -0,0 +1,63 @@
...
+# Other versions of byacc exist, see for example https://ftp.math.utah.edu/pub/byacc
+# or see for example the FreeBSD/NetBSD/OpenBSD source code or FreeBSD ports collection
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=		byacc
+COMPONENT_VERSION=	20241231
+COMPONENT_SUMMARY=	byacc - Berkeley Yacc
+COMPONENT_DESCRIPTION=  Berkeley Yacc is an LALR(1) parser generator
```
mtelka: This line seems to be unused.

AndWac: It would probably best to remove COMPONENT_DESCRIPTION and move its text over to COMPONENT_SUMMARY .

PR #20404, opened 2025-01-12
components/x11/mesa: link additional (currently missing) GL headers

xmhd:
These GL headers are currently missing from /usr/include/GL, yet are expected by e.g. xorg-server during build.

Currently the headers stored /usr/include/GL are symlinks created at runtime via ogl-select. This has been the case since ye olde days.

I don't particularly want to add further headers to ogl-select, especially when the current supported method of doing this is via libglvnd which actually supports multiple libgl implementations on a single system.

Once mesa is in a better working state, I will proceed with enabling libglvnd and we can retire ogl-select, and remove the links created in this package.

### DIFF
--- a/components/x11/mesa/Makefile
+++ b/components/x11/mesa/Makefile
@@ -20,6 +20,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME= mesa
 COMPONENT_VERSION= 21.3.9
+COMPONENT_REVISION= 1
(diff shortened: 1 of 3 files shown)


xmhd:
This PR will result in glxinfo and glxgears working again. Confirmed tested on multiple machines.

xmhd:
I've split the xorg-server changes into it's own branch and PR.

Toasterson:
Thanks for the work. Looking forward to the other mesa work.

PR #20405, opened 2025-01-12
components/x11/xorg-server: explicitly enable libdrm

xmhd:
Depends on #20404

With the updates to Mesa and libdrm in 2024, glxinfo + glxgears were in a broken state. Explicitly enabling libdrm within the xorg-server package fixes this.

With the recent change in default GCC_VERSION to gcc-14, at least one patch (maybe more) will require tweaking due to gcc-14 now defaulting to -Werror=sizeof-pointer-memaccess. I have chosen to set GCC_VERSION to gcc-13 for xorg-server as I will be working on updating xorg-server to the current upstream release in the near future.

I am not too happy about needing to pass --disable-dependency-tracking to pass configure, but for the same reason as above, I don't want to spend any further time on it right now.

### DIFF
--- a/components/x11/xorg-server/Makefile
+++ b/components/x11/xorg-server/Makefile
@@ -17,9 +17,11 @@
 include ../../../make-rules/shared-macros.mk
 include $(WS_MAKE_RULES)/x11.mk
 
+GCC_VERSION=13
+
 COMPONENT_NAME=         xorg-server
 COMPONENT_VERSION=      1.19.7
-COMPONENT_REVISION=     11
+COMPONENT_REVISION=     12
 COMPONENT_SUMMARY=      Xorg - X11R7 X server
 COMPONENT_SRC=          $(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=      $(COMPONENT_SRC).tar.bz2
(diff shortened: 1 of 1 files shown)


xmhd:
IPS is not yet my strong point - but my understanding is that gcc-13 will need to be declared as a build time dependency , ditto for runtime?

Toasterson:
Yes, Both should be added to REQUIRED_PACKAGES for the Runtime we have GCC_RUNTIME_PKG which automatically points to GCC_VERSION Runtime Pkg.

A quick grep in `make-rules` gives you a definition list which Runtimes we define 

```
make-rules/shared-macros.mk:GCC_RUNTIME_PKG =   system/library/gcc-$(GCC_VERSION)-runtime
make-rules/shared-macros.mk:GXX_RUNTIME_PKG =   system/library/g++-$(GCC_VERSION)-runtime
make-rules/shared-macros.mk:GCCGO_RUNTIME_PKG = system/library/gccgo-$(GCC_VERSION)-runtime
make-rules/shared-macros.mk:GFORTRAN_RUNTIME_PKG =      system/library/gfortran-$(GCC_VERSION)-runtime
make-rules/shared-macros.mk:GOBJC_RUNTIME_PKG = system/library/gobjc-$(GCC_VERSION)-runtime
```

Toasterson:
For example Audacity defines

```
```

xmhd:
> Yes, Both should be added to REQUIRED_PACKAGES for the Runtime we have GCC_RUNTIME_PKG which automatically points to GCC_VERSION Runtime Pkg.
> 
> A quick grep in `make-rules` gives you a definition list which Runtimes we define
> 
> ```
> make-rules/shared-macros.mk:GCC_RUNTIME_PKG =   system/library/gcc-$(GCC_VERSION)-runtime
> make-rules/shared-macros.mk:GXX_RUNTIME_PKG =   system/library/g++-$(GCC_VERSION)-runtime
> make-rules/shared-macros.mk:GCCGO_RUNTIME_PKG = system/library/gccgo-$(GCC_VERSION)-runtime
> make-rules/shared-macros.mk:GFORTRAN_RUNTIME_PKG =      system/library/gfortran-$(GCC_VERSION)-runtime
> make-rules/shared-macros.mk:GOBJC_RUNTIME_PKG = system/library/gobjc-$(GCC_VERSION)-runtime
> ```

Thanks! Will update and comment when this is ready for merging.

xmhd:
Closing this for now.

PR #20412, opened 2025-01-13
configure.mk: conditionally remove --localedir from CONFIGURE_OPTIONS

mtelka:
The support for the `--localedir` option in `configure` was added in version 2.60 of autoconf.  Several components that use the `configure` script generated using an older autoconf are unable to run the script properly when the `CONFIGURE_DEFAULT_DIRS` is set to `yes` (this is the default).  Examples of such components: `lynx` or `byacc`.

This PR adds support for the new `CONFIGURE_DEFAULT_DIRS` value: `pre-2.60`.  When the `CONFIGURE_DEFAULT_DIRS` variable is set to `pre-2.60` in a component's `Makefile` then it is ensured that the `--localedir` option is not added to the `CONFIGURE_OPTIONS`.

This fixes the problem described at https://www.illumos.org/issues/17056 but differently than the solution proposed there.  I believe this one is a bit more generic and will allow some other values (for example `pre-2.50` or `pre-3.0`) when such a need arrives.


klausz65:
Very cool - this absolutely very usefull - thanks

PR #20482, opened 2025-01-18
illumos-gate: add patch to set cubic instead of sunreno congestion-co…

AndWac:
…ntrol algorithm

### DIFF
--- a/components/openindiana/illumos-gate/patches/0010-set-modern-congestion-control.patch
+++ b/components/openindiana/illumos-gate/patches/0010-set-modern-congestion-control.patch
@@ -1,0 +1,16 @@
+For an overview of several congestion control algorithms see
+https://www.geeksforgeeks.org/tcp-congestion-control-algorithms-reno-new-reno-bic-cubic/
+
+diff --git a/usr/src/uts/common/inet/cc.h b/usr/src/uts/common/inet/cc.h
+index d92535cd48..aa72c46162 100644
+--- a/usr/src/uts/common/inet/cc.h
++++ b/usr/src/uts/common/inet/cc.h
+@@ -64,7 +64,7 @@ extern "C" {
+ 
+ #define	CC_ALGO_NAME_MAX	16	/* max congestion control name length */
+ 
+-#define	CC_DEFAULT_ALGO_NAME	"sunreno"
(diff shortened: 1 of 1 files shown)


mtelka:
Why do we need this patch?  Isn't it possible to set this via tunables where needed?

AndWac:
Because others (eg. FreeBSD 14.2, Debian Linux 12) have it as their default algorithm, too.

mtelka:
Then why not to propose this change upstream?  I believe we should not diverge in non-critical areas from the vanilla illumos-gate.  Especially when it is extremely easy to run `ipadm set-prop ...` to change this.

AndWac:
illumos-gate is ver conservative and I don't want to argue there. Most OI users don't know these details but suffer from ancient settings in illumos-gate / OI.

mtelka:
You do not need to argue much.  Just file a bug explaining how exactly users suffer and that should be enough as a first step.

AndWac:
I don't have the time and energy to do that. I'll ask and wait others before I merge this. In the end being as near as possible to illumos-gate is not something that valuable. Illumos-gate is the least common denominator of the involved parties and not a product. Parts of it are ancient.

mtelka:
The problem with these kind of changes is that somebody somewhere claims something but such a claim is often not backed by any real data and/or proper root cause analysis.  Such random claims lead to changes that does not fix or improve anything.  They only do pollute the code.  And here the "users suffer" looks exactly like such a claim with no backing data.

Looking at the ticket where the pluggable congestion algorithms were introduced I see [nice benchmark there](https://www.illumos.org/issues/11553#note-1).  It is clear that an average user won't see any measurable improvement with the `sunreno` -> `cubic` change.  For a special case (machine in data center with big pipe to internet) there was 10% improvement measured.  If somebody is using an OI machine in such an environment where he is able to measure the difference then I'm sure he is smart enough to run the simple `ipadm set-prop` command to implement the improvement even without this PR.  In any case, 10% performance difference under special conditions is not something I'd label as "users suffer".

AndWac:
The ticket is old and so are the results. At least FreeBSD and Debian think that cubic should be the default for them.
The reactions so far are such that you are the only one against it, while most don't care.

PR #20485, opened 2025-01-18
x11/components/mesa: fix erroneous dri-drivers-path

xmhd:
Xorg expects /usr/lib/xorg/modules/dri/amd64 , and previous versions of Mesa provided this path. With the version update to v21.3.9 this changed to /usr/lib/xorg/modules/dri , which causes problems, i.e. as previously described with Xorg expecting /usr/lib/xorg/modules/dri/amd64 , which is defined in x11.mk along with drivers, extensions, input and multimedia paths for Xorg.


xmhd:
This commit results in working glxinfo and glxgears which has been broken since the update from Mesa v13 -> v21.3.9.

PR #20496, opened 2025-01-19
opensmalltalk cog-spur: set GCC_VERSION=13

cstes:
This version of OpenSmalltalk Cog Spur runs Cuis Smalltalk,
and it also runs Squeak 6.1alpha 64bit

However when running the SUnit tests it fails on 
MCMczInstallerTest test suite.

I'll raise an OpenSmalltalk issue for that SUnit failure for the MczInstaller (which is meant to load the chunk-format code inside an mcz archive without using the normal Monticello machinery).

### DIFF
--- a/components/runtime/smalltalk/cog-spur/Makefile
+++ b/components/runtime/smalltalk/cog-spur/Makefile
@@ -36,6 +36,14 @@ USE_OPENSSL10=yes
 # the default OpenIndiana JPEG_IMPLEM at time of writing is libjpeg8-turbo
 JPEG_IMPLEM=libjpeg8-turbo
 
+# problems OpenSmalltalk and gcc
+# the default gcc_OPT flags with -O2 do not work (see below)
+# this has been the case for many years
+# extra problems with gcc 14.2 
+# gcc_OPT+=		-Wno-error=implicit-function-declaration
+# problem with iconv and restrict keyword
+GCC_VERSION=	13
+
 include ../../../../make-rules/shared-macros.mk
 
 # there is no official triplet version for opensmalltalk-vm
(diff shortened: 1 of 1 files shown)


cstes:
Actually before raising an OpenSmalltalk issue for this, I'll do some more debugging myself.   I think the Squeak community considers the OpenIndiana version of OpenSmalltalk as something unsupported by Squeak anyway ...

cstes:
When I use the command "inisqueak" (which is a script that will download and create a Smalltalk Squeak VM in the current directory) it downloads the older 5.3 64bit image (update 19486) and the SUnit tests for that version (5.3) works fine with this version of cog-spur (compiled with gcc 13.3).   It runs 4735 tests and for me 4688 tests pass.   There were a few BitmapStreamTest failures for me but I think these could be normal.   Note that the  older Squeak 5.3 image is a 68021 Smalltalk image and not a 68533 SistaV1 image like the Squeak 6.1alpha image where I noticed a SUnit failure on MCMczInstallerTest .   Anyway, this version of cog-spur is as good as any of the older versions of cog-spur packaged for OpenIndiana.    Regarding the C compiler gcc 14.2,  I think noone so far has compiled OpenSmalltalk with gcc 14.2, but in the following weeks I'll give it a try when I update the cog-spur again, and move the package to gcc 14.2.

AndWac:
You can silence gcc-14 with these lines:
# gcc 14 complains about this
CFLAGS += -Wno-error=implicit-function-declaration
CFLAGS += -Wno-error=incompatible-pointer-types
CFLAGS += -Wno-error=return-mismatch

mtelka:
@cstes please report this upstream and add link to the bug report into the `Makefile` as  a comment.  Thank you.

PR #20521, opened 2025-01-22
sound-juicer: fix build dependencies

mtelka:


AndWac:
I would have expected an updated pkg5, too.

mtelka:
Hmm, you are right.  I forgot to run the metadata update.  OTOH, since there is no consumer for `dependencies` in `pkg5` files then nobody cares anyway.

PR #20532, opened 2025-01-23
gnu-make: update to 4.4.1; drop 32 bit

mtelka:


AndWac:
I have installed this on a second system with the same results:
╰─➤  gmake setup
/bin/sh: git: not found [No such file or directory]
/bin/bash: line 1: uname: command not found
make-rules/shared-macros.mk:140: /make-rules/ips-buildinfo.mk: No such file or directory
gmake: *** No rule to make target '/make-rules/ips-buildinfo.mk'.  Stop.

The new make version seem to have problems with its environment. I don't know what makes my installations special as I try to be as close as possible to the build server. So I expect similar problems when I merge this PR.

mtelka:
@AndWac please paste output from following commands:
```
pwd
pkg list -afv gnu-make
gmake --version
git rev-parse HEAD
git status
env - gmake setup
```

mtelka:
Okay, I reproduced the problem.  When gmake is run without `env -` then it fails here too.  Investigating...

mtelka:
It looks like the following change introduced in [GNU make 4.4](https://lists.gnu.org/archive/html/info-gnu/2022-10/msg00008.html) is causing the problem:
```
* WARNING: Backward-incompatibility!
  Previously makefile variables marked as export were not exported to commands
  started by the $(shell ...) function.  Now, all exported variables are
  exported to $(shell ...).  If this leads to recursion during expansion, then
  for backward-compatibility the value from the original environment is used.
  To detect this change search for 'shell-export' in the .FEATURES variable.
```
Namely, the problem is with the `PATH` variable.

To solve the problem we need some changes and cleanups in our Makefiles.  Until they are implemented I'll convert this PR to draft.

mtelka:
This is now ready for merge.

PR #20575, opened 2025-01-26
librewolf: update to 134.0.2-1

AndWac:


mtelka:
Is `gnome/config/gconf` really needed to build this?  GConf is deprecated for more than 10 years now...

AndWac:
You can ask the same question for firefox. This package is just adapted from it.

iigs:
The file config/system-headers.mozbuild header list includes gconf/gconf-client.h . A quick grep does not find gconf-client.h being included in any file in the firefox tree (nor any other gconf header or gconf library).  I can not confirm that gconf is not being used at this time--it appears that it is not being used. Could the gconf line in system-headers.mozbuild have been used for Flash support through Firefox 84 and not yet removed?

I can make a determination if this manual dependency can be dropped for Firefox 136.

AndWac:
We can further investigate. For the moment it doesn't make sense to wait with this PR.

mtelka:
> I can make a determination if this manual dependency can be dropped for Firefox 136.

That would be great.  Thank you.

PR #20587, opened 2025-01-27
sendmail: fix lookup of ipv4-mapped addresses

hrosenfeld:
I noticed that our sendmail has trouble resolving IPv4-mapped addresses when listening on an AF_INET6 socket, complaining that the client's address may be forged when connecting.

When a client connects, sendmail first gets the remote address of the socket. For a IPv4 client connecting to a sendmail listening on an AF_INET6 socket, this will be an IPv4-mapped IPv6 address (in the form ::ffff:a.b.c.d). Sendmail then performs a reverse-lookup to get the hostname of the client. When that succeeds, it does a lookup of the hostname and tries to find the address of the client in the list of addresses returned. This latter lookup fails on illumos (and probably on Solaris, too).

The reason for this is that sm_gethostbyname() in sendmail/conf.c calls getipnodebyname() with the default setting of SM_IPNODEBYNAME_FLAGS, which is AI_ADDRCONFIG. As the comment in the code says (and our getipnodebyname(3socket) confirms this), to also return IPv4-mapped addresses we need to use AI_DEFAULT | AI_ALL.

Testing: I've run this for a few days now on both sendmail servers that I maintain, the smart host in my home network and a public-facing SMTP server that I maintain for a few domains with a handful of users.


AndWac:
Please increment COMPONENT_REVISION.

PR #20715, opened 2025-02-08
libreoffice: update to 25.2.0.3

iigs:
* Removed gcc version 13 comment in Makefile.  This builds with gcc 14.
* Switched to jdk 21 to build to be in better synch with userland.  On my system, system mediator for jdk is still 17 (jdk 21 is installed).  Only tested with jdk 17 runtime.
* Newer skia doesn't have Solaris/Illumos support to access self executable. Added 10-skia-proc-a.out.patch.
* New PDF Tiling code fails to compile.  Disable using 27-pdf-tiling-not-work.patch.
* If using MATE screensaver, support now added for LibreOffice to inhibit screensaver (matches support for all other platforms).


components/desktop/libreoffice/Makefile
```diff
@@ -311,7 +311,7 @@ COMPONENT_ARCHIVE_HASH_55 = sha256:4d1dac6ef66528b7a2bc910ff45c1015a1b83cb928515
 COMPONENT_ARCHIVE_URL_55 = https://dev-www.libreoffice.org/src/$(COMPONENT_ARCHIVE_55)
 
 # ignore archive 21 (boost)
-NUM_EXTRA_ARCHIVES = $(shell seq 1 20) $(shell seq 22 54)
+NUM_EXTRA_ARCHIVES = $(shell seq 1 20) $(shell 22 54)
```
mtelka: What is `$(shell 22 54)` supposed to do?

components/desktop/libreoffice/Makefile
```diff
@@ -487,12 +485,12 @@ LIBREOFFICE_SHARE_DIR = /usr/lib/amd64/libreoffice/share
 
 NSS_LIB_DIR=/usr/lib/mps/amd64
 
-JAVA17_LIB_DIR=/usr/jdk/openjdk17/lib
+JAVA21_LIB_DIR=/usr/jdk/openjdk21/lib
```
mtelka, line 488: I suggest to define something like `JAVA_LIB_DIR = $(JAVA_HOME)/lib` here instead of `JAVA21_LIB_DIR`.  This will make sure `libreoffice` always builds with the latest LTS java version without further changes needed here.  If you would need (or want) different (a non-default) java version then simple addition of `JAVA_VERSION = 11` or similar would do the trick.

components/desktop/libreoffice/Makefile
```diff
@@ -467,7 +465,7 @@ CONFIGURE_OPTIONS += --with-system-openldap
 CONFIGURE_OPTIONS += --enable-python=system
 CONFIGURE_OPTIONS += --disable-cve-tests
 CONFIGURE_OPTIONS += --with-java
-CONFIGURE_OPTIONS += --with-jdk-home="/usr/jdk/openjdk17/"
+CONFIGURE_OPTIONS += --with-jdk-home="/usr/jdk/openjdk21/"
```
mtelka, line 468: Also here, please use `$(JAVA_HOME)` instead of hardcoded `/usr/jdk/openjdk21/`.

iigs:
fixed shell seq.  25.5 Makefile was based on my developement contents which didn't include the fix in the 24.8 release.  Development contents now has the second seq.

Not changing JAVA.  JAVA_HOME was not maintained well from 2020 to some more current time period--the first implementation of JAVA_HOME for either 11 or 17 (or maybe both??) did not work with LibreOffice.  Sorry, I don't recall how things were and I it is not a good use of my time to look through the past userland commits.
By maintaining the JAVA related info in this Makefile, it ensures LO can compile no matter what the userland make files are using.

PR #20811, opened 2025-02-16
pulseaudio: correct dependencies for gstreamer

klausz65:


components/desktop/pulseaudio/Makefile
```diff
@@ -35,6 +35,9 @@ include $(WS_MAKE_RULES)/common.mk
 
 PATH= $(PATH.gnu)
 
+# build fails with gcc14.
+GCC_VERSION= 13
+
```
AndWac, line 39: I prefer silencing gcc-14 over using gcc-13. Try this:
\# gcc 14 complains about incompatible pointer types
CFLAGS += -Wno-error=incompatible-pointer-types

mtelka: Even better would be to report the issue upstream (if not fixed there yet).  If it is already fixed there then we should backport the fix.

PR #20822, opened 2025-02-17
rtmpdump: update to 2.6; switch to openssl 3.4

iigs:
Archive download no longer available.  Used openindiana/sysding/Makefile as the template to use GIT for download.

This is part 1.  Part 2 is gst-plugins-bad rebuild due to rtmpdump .so version number bump.


components/encumbered/rtmpdump/Makefile
```diff
@@ -14,30 +14,37 @@
...
-	http://rtmpdump.mplayerhq.hu/download/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL=  http://rtmpdump.mplayerhq.hu/
 COMPONENT_LICENSE=      GPLv2,LGPLv2
 COMPONENT_LICENSE_FILE= rtmpdump.license
 COMPONENT_CLASSIFICATION=Applications/Sound and Video
 COMPONENT_FMRI=         video/rtmpdump
 
+GIT=git
+GIT_REPO=https://git.ffmpeg.org/rtmpdump
+GIT_BRANCH=master
```
mtelka, line 30: If you track master then it won't be version 2.6 soon.  We should never integrate such moving targets for 3rd party software.  Please pin this to some changeset.  It looks like 6f6bb1353fc84f4cc37138baa99f586750028a01 is version 2.6.

Thank you.

mtelka:
> This is part 1. Part 2 is gst-plugins-bad rebuild due to rtmpdump .so version number bump.

Both `library/ffmpeg-6` and `library/ffmpeg-7` depends on `librtmp.so.0` too.

components/encumbered/rtmpdump/Makefile
iigs: I don't see an example of commit number in any userland Makefile that pulls content with a git clone. Do I just change GIT_BRANCH to 6f6bb1353fc84f4cc37138baa99f586750028a0? That does not seem right.

iigs:
My due diligence was  that I checked for packages that depended on rtmpdump using the command pkg search -rp depend:require:video/rtmpdump .  ffmpeg-6 and ffmpeg-7 was not in the result.

mtelka:
> My due diligence was that I checked for packages that depended on rtmpdump using the command pkg search -rp depend:require:video/rtmpdump . ffmpeg-6 and ffmpeg-7 was not in the result.

Then either the command you used is not sufficient to find all dependants, or it does not work as expected.

In any case, both `ffmpeg` packages definitely depends on `rtmpdump`:
```
$ pkg contents -mr library/ffmpeg-6 | grep ^depend.*rtmp
depend fmri=pkg:/video/rtmpdump@2.4-2022.0.0.2 type=require
$ pkg contents -mr library/ffmpeg-7 | grep ^depend.*rtmp
depend fmri=pkg:/video/rtmpdump@2.4-2022.0.0.2 type=require
$
```

components/encumbered/rtmpdump/Makefile
mtelka: It looks like `GIT_COMMIT_ID` is designed to do the job.

components/encumbered/rtmpdump/Makefile
```diff
@@ -14,30 +14,37 @@
...
-	sha256:0be975011d116bbdf1befad2a984ac83d09b1061c86550326a7ec496d2f188fc
-COMPONENT_ARCHIVE_URL=  \
-	http://rtmpdump.mplayerhq.hu/download/$(COMPONENT_ARCHIVE)
 COMPONENT_PROJECT_URL=  http://rtmpdump.mplayerhq.hu/
 COMPONENT_LICENSE=      GPLv2,LGPLv2
 COMPONENT_LICENSE_FILE= rtmpdump.license
 COMPONENT_CLASSIFICATION=Applications/Sound and Video
 COMPONENT_FMRI=         video/rtmpdump
 
+GIT=git
```
mtelka: You do not need this.  `GIT` is already set to `/usr/bin/git` in `shared-macros.mk`.

PR #20830, opened 2025-02-18
network/openssh: update to 9.9p2

RAC-oi:
network/openssh: update to 9.9p2

### DIFF
--- a/components/network/openssh/Makefile
+++ b/components/network/openssh/Makefile
@@ -29,14 +29,13 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		openssh
 #   OpenSSH <x>.<y>p<n>     => IPS <x>.<y>.0.<n>
 #   OpenSSH <x>.<y>.<z>p<n> => IPS <x>.<y>.<z>.<n>
-COMPONENT_VERSION=	9.9.0.1
-HUMAN_VERSION=		9.9p1
-COMPONENT_REVISION=	1
+COMPONENT_VERSION=	9.9.0.2
+HUMAN_VERSION=		9.9p2
 COMPONENT_SUMMARY=	OpenSSH client and associated utilities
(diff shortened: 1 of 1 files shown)


components/network/openssh/Makefile
```diff
@@ -29,14 +29,14 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		openssh
 #   OpenSSH <x>.<y>p<n>     => IPS <x>.<y>.0.<n>
 #   OpenSSH <x>.<y>.<z>p<n> => IPS <x>.<y>.<z>.<n>
-COMPONENT_VERSION=	9.9.0.1
-HUMAN_VERSION=		9.9p1
+COMPONENT_VERSION=	9.9.0.2
+HUMAN_VERSION=		9.9p2
 COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	OpenSSH client and associated utilities
```
AndWac, line 34: Please reset/delete COMPONENT_REVISION if COMPONENT_VERSION changes.

RAC-oi: ah missed that. changed.

AndWac:
How did you test the new version? Did you run gmake REQUIRED_PACKAGES, sample-manifest, publish?
Did you install it on a system to make sure it isn't broken? This is an essential package. If we break it almost everybody will have problems.

RAC-oi:
of course ssh is a central part of an os these days, hence the swift pr. and also the reason why i had mixed feelings after spotting the number of patches. but since this is just a small step from p1 to p2 it's a low-risk update.

i ran gmake install, test and publish. the dependencies don't change from a p1 to a p2. the tests had one problem with percent expansions but that's because i ran them as root and they expected root's home to be / which it is not in my case. the rest seemed fine.

but of course you can test it in whichever way you like.

PR #20904, opened 2025-02-23
illumos-gate: add guards in fab_xpath_query() to prevent from crashes

AndWac:


mtelka:
Is this issue reported upstream?  If not, could you please report it?

Thank you.

AndWac:
https://www.illumos.org/issues/17213

Bill-Sommerfeld:
I picked up the upstream issue and have an RTI pending; once it lands be prepared to revert this patch in favor of the upstream fix.

PR #20933, opened 2025-03-01
rustc - update to version 1.85.0

fritzkink:


components/developer/rust/Makefile
```diff
@@ -146,7 +146,6 @@ REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
 REQUIRED_PACKAGES += $(GXX_RUNTIME_PKG)
```
AndWac, line 142: Is developer/gcc-13 still needed?

mtelka: There are far more things in the `Makefile` that are not needed.  For example the touch of `META.depend-runtime.res` or `CLANG_VERSION` setting.

PR #20974, opened 2025-03-04
facter: migrate from ruby-23 to ruby-32

klausz65:


components/ruby/facter/Makefile
```diff
@@ -43,21 +43,16 @@ COMPONENT_LICENSE_FILE=	facter.license
 COMPONENT_POST_UNPACK_ACTION= \
     $(GEM) spec $(USERLAND_ARCHIVES)/$(COMPONENT_ARCHIVE) --ruby > $(SOURCE_DIR)/$(GEMSPEC)
 
-include $(WS_TOP)/make-rules/prep.mk
-include $(WS_TOP)/make-rules/ips.mk
+include $(WS_TOP)/make-rules/common.mk
 include $(WS_TOP)/make-rules/gem.mk
```
mtelka: I think you should use something like `BUILD_STYLE = gem` instead.

components/ruby/facter/Makefile
```diff
@@ -43,21 +43,16 @@ COMPONENT_LICENSE_FILE=	facter.license
 COMPONENT_POST_UNPACK_ACTION= \
```
mtelka: This should be moved after the `common.mk` include (if still needed).

components/ruby/facter/Makefile
```diff
@@ -43,21 +43,16 @@ COMPONENT_LICENSE_FILE=	facter.license
...
 
-# common targets
-build:		$(BUILD_32)
-
-install:	$(INSTALL_32)
-
 # facter requires the installation of the following rubygems to run
 # the unittests:  rspec rspec-expectations mocha rake rack hiera
 # Disable the tests until these rubygems are available.
 test:		$(NO_TESTS)
```
mtelka: Please use `TEST_TARGET = $(NO_TESTS)` instead (see `templates/ruby.mk`).

components/ruby/facter/Makefile
klausz65: done

klausz65:
All changes done as requested.

components/ruby/facter/Makefile
```diff
@@ -22,6 +22,7 @@
 #
 include ../../../make-rules/shared-macros.mk
 
+BUILD_STYLE= gem
```
mtelka, line 24: This should go before the `shared-macros.mk` include.  I think technically it does not matter, we just do it that way.  See also `templates/ruby.mk` for an example.

components/ruby/facter/Makefile
```diff
@@ -39,25 +40,20 @@ COMPONENT_CLASSIFICATION=	Development/Ruby
 COMPONENT_LICENSE=	Apache v2.0
 COMPONENT_LICENSE_FILE=	facter.license
 
+include $(WS_TOP)/make-rules/common.mk
+include $(WS_TOP)/make-rules/gem.mk
```
mtelka: With `BUILD_STYLE= gem` you should not need the `gem.mk` include.  Or, is it really needed?

components/ruby/facter/Makefile
```diff
@@ -39,25 +40,20 @@ COMPONENT_CLASSIFICATION=	Development/Ruby
...
+include $(WS_TOP)/make-rules/ruby.mk
```
mtelka, line 50: Is this really needed?

components/ruby/facter/Makefile
```diff
@@ -39,25 +40,20 @@ COMPONENT_CLASSIFICATION=	Development/Ruby
...
-# common targets
-build:		$(BUILD_32)
-
-install:	$(INSTALL_32)
-
 # facter requires the installation of the following rubygems to run
 # the unittests:  rspec rspec-expectations mocha rake rack hiera
 # Disable the tests until these rubygems are available.
-test:		$(NO_TESTS)
+TEST_TARGET=	$(NO_TESTS)
```
mtelka: Please move this before the `common.mk` include (see `templates/ruby.mk` for reference).

components/ruby/facter/Makefile
klausz65: done

klausz65: If I remove common.mk gmake clobber fails and if I remove ruby.mk gmake REQUIRED_PACKAGES fails

PR #21005, opened 2025-03-05
clutter: correct dependencies

klausz65:


iigs:
Please update COMPONENT_REVISION

klausz65:
Please describe, why I shall update COMPONENT_REVISION, there is no technical reason to do so - as it will get a new Branch: 2025 and is therefore newer than the current one. Descision made in today's OI-Maintainer-Meeting.

iigs:
I wasn't at the meeting today to get the updated information.  You are right--REVISION does not need to be increased.

PR #21008, opened 2025-03-05
thunderbird: update to 128.8.0

iigs:
Drop dependency for gnome/config/gconf since thunderbird doesn't need it any more.


iigs:
Also, fixed rustup since the gcc-compat package isn't being used any more.  The precompiled rust bins/libs that come from rustup does not have the rpath where openindiana keeps libgcc_s.so.1. Set LD_LIBRARY_PATH so rust can find libgcc_s.so.1 .

PR #21040, opened 2025-03-11
berkeley-db: fix configure silliness to allow build with gcc14

hrosenfeld:
There are a number of little test programs embedded in the configure script that don't specify the return type of main(), leading to compilation failures and thus mis-detected features during configure.


AndWac:
Please increment the COMPONENT_REVISION in the Makefile. The Makefiles needs to be touched in order to make jenkins build a new package.

PR #21053, opened 2025-03-13
ftgl: make build work again using gcc-11/13 and gcc-14.

klausz65:


components/x11/ftgl/Makefile
```diff
@@ -12,14 +12,15 @@
 #
 # Copyright 2018 Harry Liebel <hliebel@gmail.com>
 # Copyright 2021 Andreas Wacknitz
+# Copyright 2025 Klaus Ziegler
 #
 
 BUILD_BITS=	64
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 18: BUILD_BITS= 64 can be dropped as it's the default.

components/x11/ftgl/manifests/sample-manifest.p5m
```diff
@@ -154,6 +154,7 @@ file path=usr/share/doc/ftgl/html/dir_48342f6a4ef5e4a9839cb593b06f225a.html
 file path=usr/share/doc/ftgl/html/dir_60d10ebdb15d931f338a2058dca6095d.html
 file path=usr/share/doc/ftgl/html/dir_bcba83cc3296faf84eada02145c0eef0.html
 file path=usr/share/doc/ftgl/html/doxygen.css
+file path=usr/share/doc/ftgl/html/doxygen_crawl.html
 file path=usr/share/doc/ftgl/html/faq_8dox.html
```
AndWac, line 158: Is there a reason why doxygen_crawl.html is not also added to real the manifest?

PR #21060, opened 2025-03-14
bundler: migrate to ruby-32 build with new environment.

klausz65:


AndWac:
Publishing failed on the build server:

The following packages deliver conflicting action types to usr/ruby/3.2/bin/bundler:

  link:
    pkg://openindiana.org/library/ruby/bundler@1.13.6,5.11-2025.0.0.1:20250314T110507Z
  file:
    pkg://openindiana.org/runtime/ruby-32@3.2.6,5.11-2024.0.0.1:20241227T130702Z

klausz65:
Yeep, also happend here on SPARC, so where to change what? put the file into library/ruby/bundler and link it in runtime/ruby-32 or the other way around? - what do you think?

AndWac:
I am not sure whether our bundler package is still needed. If ruby-3.2 installs its own bundler that might be sufficient.
Especially as our bundler version is quite outdated (1.13 vs. latest 2.6).

klausz65:
Okay, the bundled bundler in runtime/ruby-32 is at version 2.4.19 also much newer than 1.13 - shall we just get rid of it in the meta/history file?

klausz65:
looks like no other package has dependencies to package: library/ruby/bundler - at least pkg search does not tell any...

AndWac:
Yes, please obsolete the package.

klausz65:
Where do you want me to obsolete it:
1.) in components/ruby/ruby-32
2.) in the metapackages/history/history file
I assume as:
library/ruby/bundler@1.13.6-2020.0.1.1 runtime/ruby-32@3.2.6-2024.0.0.1

Correct ?

AndWac:
The obsoletion belongs to metapackages/history/history. But please leave out the forward part as it is most likely not helpful.

klausz65:
you mean that part: runtime/ruby-32@3.2.6-2024.0.0.1 - right?

AndWac:
exactly

PR #21068, opened 2025-03-14
mixlib-cli: build using runtime/ruby-32

klausz65:


AndWac:
The obsoletion is missing.

klausz65:
added

PR #21074, opened 2025-03-14
dvd+rw-tools: make the build work with new environment

klausz65:


components/sysutils/dvd+rw-tools/Makefile
```diff
@@ -29,9 +29,10 @@ COMPONENT_LICENSE_FILE=	LICENSE
 COMPONENT_FMRI=		media/dvd+rw-tools
 COMPONENT_CLASSIFICATION=	System/Media
 
-include $(WS_MAKE_RULES)/prep.mk
+include $(WS_MAKE_RULES)/common.mk
 include $(WS_MAKE_RULES)/justmake.mk
-include $(WS_MAKE_RULES)/ips.mk
+
```
AndWac: Why not BUILD_STYLE= justmake?

PR #21092, opened 2025-03-15
liboauth: make build work with new environment, get rid of 32bit libs.

klausz65:


components/library/liboauth/Makefile
```diff
@@ -39,9 +40,7 @@ COMPONENT_FMRI=	library/liboauth
 COMPONENT_LICENSE= MIT
```
AndWac, line 40: Please change this to https:// while at it.

PR #21101, opened 2025-03-15
minicom: make it compile using >= gcc-11/13 and 14.

klausz65:


components/sysutils/minicom/Makefile
```diff
@@ -19,6 +20,7 @@ include ../../../make-rules/shared-macros.mk
```
AndWac, line 18: Please remove BUILD_BITS= 64

PR #21110, opened 2025-03-17
serf: build using library/security/openssl-3.

klausz65:


components/library/serf/Makefile
```diff
@@ -24,11 +24,11 @@
 #
 BUILD_BITS= 64
 BUILD_STYLE= justmake
```
AndWac, line 25: Please remove BUILD_BITS= 64.

PR #21125, opened 2025-03-19
tidy: upgrade from 1.0.0 to 5.8.0 - now in 64bit only.

klausz65:
The API documentation is long gone from the main distribution -but can by viewed online on the  web.


components/text/tidy/Makefile
```diff
@@ -21,62 +21,40 @@
 
 #
 # Copyright (c) 2011, 2013, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2025 Klaus Ziegler
 #
 
+BUILD_STYLE= cmake
+USE_PARALLEL_BUILD= yes
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 29: Please add
DROP_STATIC_LIBRARIES = yes
here. This should change the sample-manifest.

hoewweken:
This affects the php tidy extension, too.
This php extension is in the php-common package, namely

```
web/php-83/php-common
web/php-82/php-common
```

PR #21128, opened 2025-03-19
libodfgen: fix dependency for library/libxml2/32, remove CXXFLAGS for…

klausz65:
… old boost-library.

### DIFF
--- a/components/library/libodfgen/manifests/sample-manifest.p5m
+++ b/components/library/libodfgen/manifests/sample-manifest.p5m
@@ -10,7 +10,7 @@
 #
 
 #
-# Copyright 2023 <contributor>
+# Copyright 2025 <contributor>
 #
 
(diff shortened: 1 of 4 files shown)


components/library/libodfgen/manifests/sample-manifest.p5m
```diff
@@ -42,6 +42,7 @@ file path=usr/share/doc/libodfgen/html/bc_s.png
 file path=usr/share/doc/libodfgen/html/bc_sd.png
 file path=usr/share/doc/libodfgen/html/closed.png
 file path=usr/share/doc/libodfgen/html/doxygen.css
+file path=usr/share/doc/libodfgen/html/doxygen_crawl.html
 file path=usr/share/doc/libodfgen/html/index.html
```
AndWac, line 46: This should also be in the the manifest

PR #21130, opened 2025-03-19
timidity++: fix dependency to codec/flac-15 and no --localedir support.

klausz65:


components/multimedia/timidity++/Makefile
```diff
@@ -12,12 +12,14 @@
 # Copyright 2018 Harry Liebel <hliebel@gmail.com>
 #
 
-BUILD_BITS= 64
+
+# disable CONFIGURE_DEFAULT_DIRS, because timidity++ does not support --localdir
+CONFIGURE_DEFAULT_DIRS=no
```
mtelka: I assume it should read `--localedir` instead of `--localdir`.  If so, then please set `CONFIGURE_DEFAULT_DIRS` to `pre-2.60` to solve the issue.

mtelka: ... for details please see #20412.

klausz65:
would that also fix the setting for the wrong mandir? which I needed to set to /usr/share/man

mtelka:
> would that also fix the setting for the wrong mandir? which I needed to set to /usr/share/man

The `mandir` is by default set to `/usr/share/man` so you do not need to do anything special (provided you do not mess with `CONFIGURE_DEFAULT_DIRS=no`).

components/multimedia/timidity++/Makefile
```diff
@@ -34,6 +36,9 @@ COMPONENT_LICENSE_FILE=	COPYING
 TEST_TARGET= $(NO_TESTS)
 include $(WS_MAKE_RULES)/common.mk
 
+CFLAGS += -DSTDC_HEADERS
```
mtelka, line 39: Please add a comment explaining why is this needed.  Thanks.

mtelka: Hmmm, this sounds like a bug either in illumos or in timidity++ and so it should be reported.  Please do so.

hoewweken:
Hi!
Something went wrong, timidity doesn't play MIDI files (?)

$ timidity Downloads/TEST.MID 
Byte order is miss configured.

klausz65:
Okay, but I really didn't change anything except the dependency and that STDC_HEADERS definiton. Fom where do you know that the byte order is miss configured?

hoewweken:
I don't know, its an error message from the program.

From TiMidity++-2.15.0/timidity/timidity.c


    /* Check the byte order */
    i = 1;
#ifdef LITTLE_ENDIAN
    if(*(char *)&i != 1)
#else
    if(*(char *)&i == 1)
#endif
    {
	fprintf(stderr, "Byte order is miss configured.\n");
	exit(1);
    }


I think that the first Char of the MIDI byte stream ist checked here.. (4D 54 or 54 4D ??)
Maybe the configure script does not get it right (?)


MIDI File Type 0:
4D 54 68 64 MThd ID
00 00 00 06 Length of the MThd chunk is always 6.
00 00

Type 1:
4D 54 68 64 MThd ID
00 00 00 06 Length of the MThd chunk is always 6.
00 01

Type 2:
4D 54 68 64 MThd ID
00 00 00 06 Length of the MThd chunk is always 6.
00 02

hoewweken:
$ grep -i byte Timidity.configure.txt 
checking whether byte ordering is bigendian... yes

And this is Intel not sparc..

klausz65:
Okay, I now have installed multimedia/timidity++ and I'm testing on my home system now. It looks like there are 2 bugs: 1.) the one mentioned by Marcel and that one you discovered - stay tuned I'm trying to fix this...

klausz65:
Okay I've prepared a new pull and it looks like it's working - at least I can here something comming out of my speakers:
klausz@gandalf % /tmp/timidity 100428.mid
Warning: Audio buffer is too small.
Playing 100428.mid
MIDI file: 100428.mid
Format: 1  Tracks: 13  Divisions: 240
Track name: Whistle (+24)
Track name: Guitar (nylon string)
Track name: Accordian (+12)
Track name: Johnny, I Hardly Knew Ye
Track name: Sequenced by Brian Hicks
Track name: Remixed by Rory McGowen
Track name: Early 19th century song
Track name: about a woman whose man
Track name: was sent off to war for
Track name: the English, and what he
Track name: looked like when he came
Track name: home to Ireland.
(By Brian Hicks)(Generated by NoteWorthy Composer)

Will push the Pull-Request now...

PR #21132, opened 2025-03-19
dconf: add dependency for library/glib2/32 to successfully build 32 b…

klausz65:
…it libs.!


components/desktop/gnome2/dconf/dconf.p5m
```diff
@@ -47,7 +47,8 @@ link path=usr/lib/$(MACH64)/libdconf.so target=libdconf.so.1
 link path=usr/lib/$(MACH64)/libdconf.so.1 target=libdconf.so.1.0.0
 file path=usr/lib/$(MACH64)/libdconf.so.1.0.0
 file path=usr/lib/$(MACH64)/pkgconfig/dconf.pc
-file path=usr/lib/dconf-service mode=0555
+file path=usr/libexec/dconf-service mode=0555
+file path=usr/libexec/$(MACH32)/dconf-service mode=0555
 file path=usr/lib/gio/modules/libdconfsettings.so
```
AndWac, line 52: You don't need mode=0555 for files in libexec because that's the default (set in transforms/defaults):
defaults:# files in libexec should have mode 0555
defaults:<transform file path=usr/libexec/.* -> default mode 0555>

PR #21206, opened 2025-03-21
babl: make sure package builds without w3m and add SPARC support.

klausz65:


AndWac:
What are the consequences if babl is build without w3m installed?

components/library/babl/babl.p5m
```diff
@@ -59,30 +59,30 @@ file path=usr/lib/$(MACH64)/babl-0.1/sse4-int8.so
...
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-float.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-gegl-fixups.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-gggl-lies.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-gggl.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-grey.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-simple.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-u16.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-u32.so
-file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v3-ycbcr.so
+file path=usr/lib/$(MACH64)/babl-0.1/x86-64-v2-CIE.so variant.arch=i386
```
mtelka: Please do this using a transform to keep changes between this and the sample manifest as minimal as possible.

Thank you.

klausz65:
Hmm, I simply don't know how. These binaries are only build for x64/86 and don't exist for SPARC as replacement by any other name.

klausz65:
There are no consequences  if babl will be build without w3m, must be in there as a manual dependency froom older versions ob babl. I also de-installed w3m before building this new version and there were no errors/warnigs regarding w3m. Anyhow w3m just can't be build atm. and I'm planning to obsolete it in the near future.

AndWac:
Did you run gmake test when w3m was not installed?

mtelka:
> Hmm, I simply don't know how. These binaries are only build for x64/86 and don't exist for SPARC as replacement by any other name.

By creating a `<transform ...>` line in the `babl.p5m` file instead of modifying tons of lines there.

klausz65:
Yeep, I just now ran gmake test  - No differences found.

components/library/babl/Makefile
```diff
@@ -70,7 +71,6 @@ LD_Z_IGNORE=
 REQUIRED_PACKAGES += developer/vala
```
mtelka, line 71: Are these other manually added deps really needed?

klausz65:
I really don't know! - and I didn't tested the removal of those, because I build on a shared dev-server together with other maintainers and certainly those packages are needed in general - just w3m can go I think.

PR #21069, opened 2025-03-14
mixlib-config: build using runtime/ruby-32

klausz65:

PR #21222, opened 2025-03-23
tomlrb: fix manifest - followup on Pull request #21069

klausz65:


components/ruby/tomlrb/pkg5
```diff
@@ -3,7 +3,8 @@
         "runtime/ruby-32"
     ],
     "fmris": [
-        "library/ruby/tomlrb"
+        "library/ruby/tomlrb",
+        "library/ruby/tomlrb-3.2.0"
     ],
     "name": "tomlrb"
```
AndWac, line 8: I expect that library/ruby/tomlrb would disappear and needs to be obsoleted?

klausz65:
Hi Andreas, Marcel expects it to be installable, for details see his comment in: #21069

mtelka:
Do we really need/want a package named `library/ruby/tomlrb-3.2.0`?  I doubt.

mtelka:
> Hi Andreas, Marcel expects it to be installable, for details see his comment in: #21069

If it is only me who expects a ruby package to be installable, then it would be easier to simply obsolete all of them, because I do not need any ruby packages at all.

AndWac:
We now have an obsoleted tomlrb-23 and a tomlrb package which should rather be tomlrb-32.
So you need to get the fmri right and need to obsolete the tomlrb package (the one we accidentely introduced).

PR #21227, opened 2025-03-23
ibus-hangul : update to 1.5.5

jwntree:


jwntree:
Because of dependencies, I think I'll have to split PR to libhangul and ibus-hangul...

AndWac:
Yes, that is needed. Furthermore, they need to be merged in a strict order.

mtelka:
This package conflicts with `system/input-method/ibus/hangul`:
```
           Packages to install:  1
       Create boot environment: No
          Rebuild boot archive: No

Changed packages:
openindiana.org
  system/input-method/library/ibus/hangul
    None -> 1.5.5-2025.0.0.0

pkg install: The following packages all deliver file actions to usr/share/ibus-hangul/data/symbol.txt:

  pkg://openindiana.org/system/input-method/ibus/hangul@1.3.9,5.11-2013.0.0.0:20151027T074846Z
  pkg://openindiana.org/system/input-method/library/ibus/hangul@1.5.5,5.11-2025.0.0.0:20250325T194217Z

...
```
I'm not sure why this was integrated with different package name (fmri) than version 1.3.9 (`system/input-method/ibus/hangul@1.3.9` vs. `system/input-method/library/ibus/hangul@1.5.5`).  If the new fmri was intentional then we need to rename `system/input-method/ibus/hangul` to `system/input-method/library/ibus/hangul`.  If it was a mistake, then we need to rename `system/input-method/library/ibus/hangul` to `system/input-method/ibus/hangul`.

Or, are `system/input-method/ibus/hangul@1.3.9` and `system/input-method/library/ibus/hangul@1.5.5` completely unrelated?

@jwntree could you please try to fix this issue?  Thank you.

PR #21229, opened 2025-03-23
libhangul : update to 0.1.0

jwntree:


components/inputmethod/libhangul/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+PATH=$(PATH.gnu)
+
+# Missing files in build dir without this.
+COMPONENT_PREP_ACTION =       ( cd $(SOURCE_DIR) && ./autogen.sh )
+COMPONENT_PRE_CONFIGURE_ACTION= ( $(CLONEY) $(SOURCE_DIR) $(@D) )
+
+
+CONFIGURE_OPTIONS += --sysconfdir=/etc
+CONFIGURE_OPTIONS += --disable-static
```
AndWac, line 38: You don't need to set sysconfdir here because its default is set in configure.mk

components/inputmethod/libhangul/Makefile
```diff
@@ -0,0 +1,46 @@
...
+
+#
+# Copyright 2025 Jiwon Na
+#
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=           libhangul
+COMPONENT_VERSION=        0.1.0
+COMPONENT_GIT_REV=	154a5e0
```
AndWac, line 21: The exisiting version of libhangul is 0.5.11 which is > 0.1.0.
So you need to add
IPS_COMPONENT_VERSION=	0.5.11
and should also add
COMPONENT_REVISION= 1
(which is probably not strictly needed)

components/inputmethod/libhangul/Makefile
```diff
@@ -0,0 +1,46 @@
...
+COMPONENT_SUMMARY=        libhangul - Hangul hanja input support
+COMPONENT_SRC=		libhangul-$(COMPONENT_GIT_REV)
+COMPONENT_ARCHIVE=        $(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_HASH=   sha256:d5d4ee0e8e5ab30e5d190b1bc989b943a044844b95bd83b5533de550fd7af3b9
+COMPONENT_ARCHIVE_URL=    https://codeload.github.com/libhangul/$(COMPONENT_NAME)/tar.gz/$(COMPONENT_GIT_REV)
+COMPONENT_FMRI=           system/input-method/library/libhangul
+COMPONENT_CLASSIFICATION= System/Localizations
+COMPONENT_LICENSE=        LGPL v2.1
+COMPONENT_LICENSE_FILE=   COPYING
```
mtelka, line 30: The `libhangul.license` file is unused (and so it should be removed) because you pointed `COMPONENT_LICENSE_FILE` to `COPYING`.

components/inputmethod/libhangul/Makefile
```diff
@@ -0,0 +1,46 @@
...
+COMPONENT_PROJECT_URL=    https://github.com/libhangul/libhangul
+
+TEST_TARGET= $(NO_TESTS)
+include $(WS_MAKE_RULES)/common.mk
+
+PATH=$(PATH.gnu)
+
+# Missing files in build dir without this.
+COMPONENT_PREP_ACTION =       ( cd $(SOURCE_DIR) && ./autogen.sh )
+COMPONENT_PRE_CONFIGURE_ACTION= ( $(CLONEY) $(SOURCE_DIR) $(@D) )
```
mtelka: Please replace this line by `CLONEY_MODE =` to achieve the same result using modern approach.

mtelka (changes requested):
Thank you for this work!

PR #21242, opened 2025-03-23
uwimap: provide 32bit library for php-8.3

klausz65:


components/library/uwimap/Makefile
```diff
@@ -38,10 +40,11 @@ COMPONENT_FMRI=	library/uw-imap
 COMPONENT_SUMMARY=	University of Washington IMAP library
 COMPONENT_LICENSE=	Apache v2.0
 COMPONENT_LICENSE_FILE=	LICENSE.txt
+COMPONENT_BUILD_ARGS=	-j20
```
AndWac, line 44: Please remove this line. You should never put it into a Makefile but use an environment variable instead.

mtelka:
This PR is non-sense and simply wrong as properly described by Peter Tribble [here](https://openindiana.org/pipermail/oi-dev/2025-March/033764.html).

PR #21032, opened 2025-03-11
cairo 1.18.4

RAC-oi:
fixes pdf printing:
https://bbs.archlinux.org/viewtopic.php?id=299255
https://bugs.gentoo.org/939607
https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/595

PR #21247, opened 2025-03-24
cairo: update to 1.18.4

iigs:


components/library/cairo/Makefile
```diff
@@ -20,11 +20,11 @@ DEFAULT_COMPONENT_TEST_TRANSFORMS= yes
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 20: I need to correct my own error. The correct setting is
USE_DEFAULT_TEST_TRANSFORMS= yes
Please fix this, too.

iigs:
Thank you for figuring out why the test result had extra noise in it.  Fixed.

mtelka:
This supersedes #21032 and so it should be closed.

PR #21249, opened 2025-03-24
p11-kit: provide 32bit p11-kit for 32bit gnutls

klausz65:


mtelka:
Why do we need 32-bit gnutls?

klausz65:
Because, it's needed by libsmbclient (which is still 32/64bit), which in turn is needed by gvfs and after that needed by caja - so all in all for the SunRay support - sorry forot to mention that in the commit title.

mtelka:
caja is 64-bit only, so it should not need 32-bit gvfs.

klausz65:
Your right, but libsmbclient still needs gnutls in 32 bit to be compiled inc. 32 bit support. I would rather leave libsmbclient as it is 32/64 bit, until we solved all Ray issues. I will take care of ripping out all not needed 32 bit libs when we have fixed the current issues. BTW. I have an update for libsmbclient 4.22.0 ready to go in 32/64 bit.

mtelka:
Why do we need 32-bit libsmbclient?  BTW, the SunRay community is failing to properly cooperate with OpenIndiana...

cgrzemba:
Why do you think the SunRay community is failing to properly cooperate?

mtelka:
> Why do you think the SunRay community is failing to properly cooperate?

Because I asked several times to create a metapackage for SunRay and nothing happened (yet).  See https://github.com/OpenIndiana/oi-userland/pull/15150#issuecomment-1880019898 for an example.

cgrzemba:
There is such package sunray-essential on https://pkg.toc.de/sunray. 
If you want this in OI, some packages that have been classified as obsolete must be revived.

mtelka:
> There is such package sunray-essential on https://pkg.toc.de/sunray. If you want this in OI, some packages that have been classified as obsolete must be revived.

Anything out of OpenIndiana (oi-userland) does have limited impact on oi-userland.  If you need/want the SunRay support working properly and reliably you needs to create such a package in oi-userland.  Please create such a package (a not-fully-working draft would be enough for now), otherwise the fiddle with obsoletions, removals and revivals will never end.

mtelka:
> > There is such package sunray-essential on https://pkg.toc.de/sunray. If you want this in OI, some packages that have been classified as obsolete must be revived.
> 
> Anything out of OpenIndiana (oi-userland) does have limited impact on oi-userland. If you need/want the SunRay support working properly and reliably you needs to create such a package in oi-userland. Please create such a package (a not-fully-working draft would be enough for now), otherwise the fiddle with obsoletions, removals and revivals will never end.

BTW, you promised to create the metapackage ["in the near future"](https://openindiana.org/pipermail/oi-dev/2024-October/033684.html), but after almost 6 months nothing materialized.

klausz65:
Okay, I try the SunRay desktop without 32bit libsmbclient support - let's see how it works out...

klausz65:
see above

PR #21252, opened 2025-03-24
python/coverage: update to 7.7.1

mtelka:


AndWac:
Running gmake publish here runs into the same problem like the PR-checker.

mtelka:
This is probably some incompatibility with the new setuptools (merged today morning)...

mtelka:
Yes, since setuptools 77.0.0 the location of license files is different.

PR #21284, opened 2025-03-25
emacs: Fix: configure: WARNING: unrecognized options: --with-treesitter.

klausz65:


AndWac:
Remove this option with the next emacs update. We can ignore the warning until then.

PR #21290, opened 2025-03-26
sunray: add meta-package for required packages

cgrzemba:
These are the legacy and 32 bit packages required for SunRay Server Software.

This includes empty packages for the installation, retired package for GDM and Gnome greeter Session as well as 32bit libraries that are required of SRSS tools.

For additional details see https://www.illumos.org/issues/17332


mtelka:
Perfect!  Thanks for starting this work!

Few (unsorted) notes and questions after the quick look:
- Could you please convert dependencies to list of files as much as possible?  For example, when `rpm2cpio` (binary) is needed then there should be something like `depend type=require fmri=__TBD pkg.debug.depend.file=rpm2cpio pkg.debug.depend.path=usr/bin` or `depend type=require fmri=__TBD pkg.debug.depend.file=usr/bin/rpm2cpio` or so, instead of current `depend fmri=package/rpm type=require`.  This will make sure that the right dependency is detected even we would rename `package/rpm` to something else.  This will also prevent problems when we would remove `rpm2cpio` from `package/rpm` (and move it possibly to some other package, or remove it at all).
- The manifest filename should be changed to match the `COMPONENT_NAME` (so `sunray-essential.p5m`).  Similarly I think it would be less confusing if the directory where this lives match the `COMPONENT_NAME`.
- Re fmri: the `consolidation/srss` part sounds a bit strange.
- Re `COMPONENT_CLASSIFICATION`: is it really `Virtualization`?  Similarly `COMPONENT_VERSION`, `COMPONENT_REVISION`, `COMPONENT_PROJECT_URL` needs some love.
- Please include `shared-macros.mk` the same way as other components do.
- Please elaborate more on the `service/network/dhcp` dependency.
- This component should be well and carefully commented so it is crystal clear why every single line needs to be in the manifest.
- For dependencies like `depend fmri=compatibility/packages/SUNWxwplt@0.5.11-2024.0.0.0 type=require`, do we really need to specify the version there?  Will SRSS really refuse to work with older package versions?
- Are packages like `SUNWut` or `SUNWutkio-libs` IPS packages or old SVR4 packages?
- Is `mate_install` really needed?  Is it really _essential_?

Thank you.

cgrzemba:
The component classification is taken over from SRSS packages. Virtualization stands for Desktop Virtualization
The SRSS packages like SUWut are available in both package formats.

For this meta package getting to work the OI repository needs the packages:
- service/network/dhcp (empty)
- consolidation/X/X-incorporation (empty)
- compatibility/packages/SUNWxwplt (empty)

and these in the versions before these was retired 
- system/display-manager/gdm@2.30.7
- gnome/gnome-session@2
- library/desktop/gnome-desktop@2
- gnome/gnome-settings-daemon

the closed bins
- cde/cde-runtime@0.5.11
- library/tooltalk@0.5.11
- library/motif@0.5.11

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+
+# SUNWutkio-libs package dependency, empty packages only
+# for satisfy installation requirements
+depend fmri=consolidation/X/X-incorporation type=require
+depend fmri=compatibility/packages/SUNWxwplt type=require
+
+# for Sun JRE
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libCrun.so.1
+
+# SUNWut configure requirements rpm2cpio, see docs.openindiana.org
```
mtelka: There is nothing about SUNWut nor rpm2cpio at docs.openindiana.org

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+# empty package only for satisfy installation requirements
+# SRSS on Solaris uses the old Sun DHCP server but on OI we use ISC-DHCP
+depend fmri=service/network/dhcp type=require
+
+# SUNWutkio-libs package dependency, empty packages only
+# for satisfy installation requirements
+depend fmri=consolidation/X/X-incorporation type=require
+depend fmri=compatibility/packages/SUNWxwplt type=require
+
+# for Sun JRE
```
mtelka: Why is this needed?

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+
+# SUNWut configure requirements rpm2cpio, see docs.openindiana.org
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/bin/rpm2cpio
+
+# in OI legacy/retired 64bit packages
+# GDM and Gnome greeter, only gdm-session-worker has to run in 32bit
+#
+# pkg.debug.depend.file does not working yet, there is no appropriate package in the OI repo
+# depend type=require fmri=__TBD pkg.debug.depend.file=usr/sbin/gdm-binary
+depend type=require fmri=system/display-manager/gdm@2.30.7
```
mtelka: Isn't this essentially the same as the dependency on `usr/sbin/gdm-binary` above?

There is no `system/display-manager/gdm` package (it was obsoleted) so how this line is supposed to work?  Similarly for few other packages below.

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+depend fmri=gnome/gnome-settings-daemon@2.32.1 type=require
+depend fmri=library/gnome/libgnomekbd@2.91.92 type=require
+
+# 32bit packages
+# only for loginGUI
+depend fmri=cde/cde-runtime@0.5.11 type=require
+depend fmri=library/tooltalk@0.5.11 type=require
+depend fmri=library/motif@0.5.11 type=require
+
+# gdm-session-worker 32bit libraries
```
mtelka: Could you please elaborate what exactly this means?

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libglib-2.0.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/mps/libnspr4.so
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/mps/libnss3.so
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libdbus-glib-1.so.2
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libdbus-1.so.3
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libpcre2-8.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libffi.so.6
+
+# utaudio 32bit libs
```
mtelka: Could you please elaborate what exactly this means?

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libmpg123.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libFLAC.so.14
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libsndfile.so.1
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libSDL2-2.0.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libvorbis.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libopus.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libogg.so.0
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libmp3lame.so
+
+# uttsc-bin (RDP client)
```
mtelka: Could you please elaborate what exactly this means?

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libXpm.so.4
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libz.so.1
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libpng16.so.16
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libfontconfig.so.1
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libfreetype.so.6
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libbrotlidec.so.1
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libbz2.so.1
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libexpat.so.1
+
+# utresexec
```
mtelka: Could you please elaborate what exactly this means?

Maybe these cryptic comments are clear enough for you, but these should be fully understandable for anybody with zero SRSS knowledge.

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,107 @@
...
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libXss.so.1
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/libsasl.so.1
+
+depend type=require fmri=__TBD pkg.debug.depend.file=usr/lib/liblzma.so.5
```
mtelka: Why is this needed?

mtelka (changes requested):

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: I will improve this, it is documented here: https://docs.openindiana.org/handbook/sunray/#configure-isc-dhcp

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: SunRay service auth.utauthd.utaut is a Java application. This uses also native libraries from SRSS.

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: Sure, I mentioned this in my last PR comment.

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: These are libraries which are used by gdm-seesion-worker.

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: These are libraries which are used by utaudio, the SunRay audio driver

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: These are libraries which are used by uttsc-bin, the SunRay RDP-client

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: These are libraries which are used by utresexec, an exec wrapper

components/meta-packages/sunray-essential/sunray-essential.p5m
cgrzemba: is not needed, will removed.

cgrzemba:
In https://docs.openindiana.org/handbook/sunray/#configure-isc-dhcp is a script for repackaging an SRSS IPS package for remove the pkg://service/network/dhcp dependency.
It would be possible to extend the script to modify other packages for remove the dependencies for 
pkg://consolidation/X/X-incorporation
pkg://compatibility/packages/SUNWxwplt.
But this extents the package installation effort.

components/meta-packages/sunray-essential/sunray-essential.p5m
```diff
@@ -0,0 +1,111 @@
...
+
+# in OI legacy/retired 64bit packages GDM and Gnome greeter,
+# only gdm-session-worker has to run in 32-bit, it needs SRSS libraries only available in 32-bit
+#
+# pkg.debug.depend.file does not working yet, there is no appropriate package in the OI repo
+# hence the kind of definition with named packages
+#
+depend fmri=pkg:/system/display-manager/gdm-2 type=require
+depend fmri=pkg:/gnome/gnome-session-2 type=require
+depend fmri=pkg:/library/desktop/gnome-desktop-2 type=require
```
mtelka: Why do you need `pkg:/` prefixes here?
Why are these packages needed at all (some good explanation is missing)?

components/meta-packages/sunray-essential/sunray-essential.p5m
mtelka: Please update the comment in the file to contain the answer to my question.  BTW, (almost) all questions asked here should be answered by improving comments.

components/meta-packages/X-incorporation/Makefile
```diff
@@ -0,0 +1,26 @@
+#
```
mtelka: Please do not pollute this PR by incorporating other components than sunray-essential.  Thank you.

PR #21292, opened 2025-03-26
ibus-hangul: fix ibus/hangul fmri name

jwntree:
I'm really sorry. I made a big mistake.
I just found out that the package name is wrong.
I’m not sure if there’s a mechanism in place to roll back an already released package.

I think I fell asleep for a moment before sending the PR....
I apologize again.

### DIFF
--- a/components/inputmethod/ibus-hangul/history
+++ b/components/inputmethod/ibus-hangul/history
@@ -1,0 +1,1 @@
+system/input-method/library/ibus/hangul@1.5.5,5.11-2025.0.0.1 system/input-method/ibus/hangul
--- a/components/inputmethod/ibus-hangul/Makefile
+++ b/components/inputmethod/ibus-hangul/Makefile
@@ -16,12 +16,12 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=           ibus-hangul
 COMPONENT_VERSION=        1.5.5
-COMPONENT_SUMMARY=        libhangul - The hangul engine for IBus
+COMPONENT_SUMMARY=        ibus-hangul - The hangul engine for IBus
 COMPONENT_SRC=		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=        $(COMPONENT_SRC).tar.xz
 COMPONENT_ARCHIVE_HASH=   sha256:a5aac88286cd18960229860e3e1a778978a7aeaa484ad9acfa48284b87fdc3bb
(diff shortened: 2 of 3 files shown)


mtelka:
> I'm really sorry. I made a big mistake. I just found out that the package name is wrong. I’m not sure if there’s a mechanism in place to roll back an already released package.

Not a big problem at all :-).  Really.  The IPS is well prepared to handle such issues easily.

This PR looks mostly okay with one improvement needed: we miss a line in the `history` file (preferably `components/inputmethod/ibus-hangul/history`) to create the renamed `system/input-method/library/ibus/hangul` package.

BTW, this PR fixes the issue mentioned here: https://github.com/OpenIndiana/oi-userland/pull/21227#issuecomment-2753791306

jwntree:
I'm not sure if it's correct, but I added the history file.
Thanks.

mtelka:
You are almost there.  The version number in the `history` file needs to be newer than the current package version in the ips repo.  So please change `1.5.5,5.11-2025.0.0.0` to `1.5.5,5.11-2025.0.0.1`.  And please add missing end-of-line to the file.

components/inputmethod/ibus-hangul/history
```diff
@@ -0,0 +1 @@
+system/input-method/library/ibus/hangul@1.5.5,5.11-2025.0.0.0 system/input-method/ibus/hangul
```
AndWac: Please replace this line with
system/input-method/library/ibus/hangul@1.5.5,5.11-2025.0.0.1

The fmri in a history file needs to be the latest. To achieve this the COMPONENT_REVISION has to be incremented. The optional 2nd column is only needed if you want to make sure that existing installations of the old fmri will be replaced with the new one in the 2nd column. This is most probably not needed here because of the file collision in the package.

mtelka: > Please replace this line with system/input-method/library/ibus/hangul@1.5.5,5.11-2025.0.0.1
> 
> The fmri in a history file needs to be the latest. To achieve this the COMPONENT_REVISION has to be incremented. The optional 2nd column is only needed if you want to make sure that existing installations of the old fmri will be replaced with the new one in the 2nd column. This is most probably not needed here because of the file collision in the package.

This is needed, of course, to properly handle a (rare, admitedly) case when somebody installed `system/input-method/library/ibus/hangul` without `system/input-method/ibus/hangul` installed.  That was possible to do.

And, you do not need to touch `COMPONENT_REVISION` at all here.  Please do what I said above (just replace `1.5.5,5.11-2025.0.0.0` by `1.5.5,5.11-2025.0.0.1`).

jwntree:
fixed

PR #21293, opened 2025-03-26
dovecot: build against library/security/openssl-3 and library/icu-77.

klausz65:


components/mail/dovecot/Makefile
```diff
@@ -57,9 +60,7 @@ CONFIGURE_OPTIONS += --with-mysql
 CONFIGURE_OPTIONS += --with-pgsql
 CONFIGURE_OPTIONS += --enable-static=no
 CONFIGURE_OPTIONS += --without-systemd
-CONFIGURE_OPTIONS += SSL_CFLAGS=-I$(OPENSSL_INCDIR)
-CONFIGURE_OPTIONS += SSL_LIBS="-L$(OPENSSL_LIBDIR) -lssl -lcrypto"
-CONFIGURE_OPTIONS += LDFLAGS="-lldap_r"
+CONFIGURE_OPTIONS += LDFLAGS="-lldap_r -licuuc"
```
mtelka, line 64: Please add a comment explaining why do we need `-licuuc` in `LDFLAGS`.

mtelka:
While touching this, could you please update `dovecot` to latest version (2.4.0) to bring some added value to this PR?

Thank you.

components/mail/dovecot/Makefile
klausz65: will do

klausz65:
No, I will not update to newer version now - because this leaves users of dovecot with broken configurations, for details please see: https://doc.dovecot.org/2.4.0/installation/upgrade/2.3-to-2.4.html

klausz65:
Discovered that also migration to lua-54 should be done for dovecot - but lua54's pkgconfig is broken in 64bit - will fix this first...

components/mail/dovecot/Makefile
```diff
@@ -57,9 +60,8 @@ CONFIGURE_OPTIONS += --with-mysql
 CONFIGURE_OPTIONS += --with-pgsql
 CONFIGURE_OPTIONS += --enable-static=no
 CONFIGURE_OPTIONS += --without-systemd
-CONFIGURE_OPTIONS += SSL_CFLAGS=-I$(OPENSSL_INCDIR)
-CONFIGURE_OPTIONS += SSL_LIBS="-L$(OPENSSL_LIBDIR) -lssl -lcrypto"
-CONFIGURE_OPTIONS += LDFLAGS="-lldap_r"
+# to compile against new icu77 library -licuuc is needed.
```
mtelka, line 63: This is strange.  Any idea why it worked with ICU 75, but does not with ICU 77?

klausz65: I think, just because symbols have moved within ICU libs...

mtelka: > I think, just because symbols have moved within ICU libs...

Any ... proof?  Some link to ICU documentation?  And, even if that had happened, then why ICU does not detect the change automatically?

mtelka: BTW, `libsmbclient` (the update merged yesterday) was the first component built against ICU 77 and it didn't need anything similar...

klausz65: BTW. it's not possible to compile this dovecot release against lua-5.4 - no way - therefore the update to 2.4.x has to be done. One OI-SPARC user I know of, runs dovecot productive on it. I chase him to provide at least a script for converting the needed dove configuration changes to 2.4.x - Makes more sense than to just assume what needs to changed where -I think/hope I can convince him to do that...

klausz65: I didn't look after any ICU docu, but here is the error if that lib isn't included with LDFLAGS:
ucasemap_utf8ToLower_77             .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
u_strFromUTF8Lenient_77             .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
u_errorName_77                      .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
u_cleanup_77                        .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
ucasemap_close_77                   .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
u_strToUTF8WithSub_77               .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
ucasemap_open_77                    .libs/fts-icu.o  (symbol belongs to implicit dependency /usr/lib/amd64/libicuuc.so.77)
ld: fatal: symbol referencing errors. No output written to test-fts-icu

is that enough explanition?

mtelka:
I reported the issue upstream: https://dovecot.org/mailman3/archives/list/dovecot@dovecot.org/thread/VT7D6XWB5ZRL7KIER445D3JR464CALPU/

PR #21303, opened 2025-03-26
netbeans: update to 25

iigs:


components/editor/netbeans/Makefile
```diff
@@ -52,8 +57,15 @@ $(BUILD_DIR_64)/.installed: $(BUILD_DIR_64)/.built
 	touch $(BUILD_DIR_64)/.installed;
 
 install: $(BUILD_DIR_64)/.installed
+	$(GSED) -i -e  "1s?.*?$(PYTHON_SHEBANG)?" $(PROTO_DIR)/usr/netbeans/$(COMPONENT_VERSION)/extide/ant/bin/runant.py;
```
mtelka, line 60: Please use `PYTHON_SCRIPTS` instead of this.

components/editor/netbeans/Makefile
```diff
@@ -52,8 +57,15 @@ $(BUILD_DIR_64)/.installed: $(BUILD_DIR_64)/.built
...
 
+# Runtime dependencies
```
mtelka, line 62: Runtime-only dependencies are not needed to be manually listed in `REQUIRED_PACKAGES`.

components/editor/netbeans/Makefile
```diff
@@ -52,8 +57,15 @@ $(BUILD_DIR_64)/.installed: $(BUILD_DIR_64)/.built
...
+REQUIRED_PACKAGES += runtime/java/openjdk21
+REQUIRED_PACKAGES += runtime/perl
```
mtelka: This is already listed below (automatically detected), so please remove the duplicate here.

iigs:
Updated Makefile.  Removed duplicate runtime/perl

pkg5 file needs to list all dependencies. Keeping runtime dependencies.
I am not going to test to see if PYTHON_SCRIPTS works on the runant.py file--a separate PR would need to be submitted after somebody tests it.  This is beyond the scope of this package update.

mtelka:
> pkg5 file needs to list all dependencies.

Why do you think so?

iigs:
It is my understanding that the pkg5 should be a list of all of the package dependencies.  Some magic will take the contents and make sure a published package has that list.  As a package developer, one can look at the pkg5 file to make sure the dependencies are listed to sanity check the right package dependencies are listed.  The Netbeans package should have python 3.9 and the version of openjdk that was used to build the package listed in the pkg5 file since they are needed when the package is installed on systems.  Having REQUIRED_PACKAGES as runtime dependencies in the Makefile should be the solution to do this.

mtelka:
Apparently, some mythbusting work is needed here.

First, `pkg5` files: there are three entries in `pkg5` files: `dependencies`, `fmris`, `name`.

The `dependencies` list is basically a copy of `REQUIRED_PACKAGES` list from the component's `Makefile`.  You can get the same list using `env - gmake print-value-REQUIRED_PACKAGES`.  The only difference between the `env - gmake print-value-REQUIRED_PACKAGES` list and the `dependencies` entry in the `pkg5` file is that the `pkg5` file is frozen at a time when the file was created, while the `env - gmake print-value-REQUIRED_PACKAGES` output could change over the time (even without the component's `Makefile` change).  Please note that there is no known consumer for `dependencies` entry in the `pkg5` file (except humans that could find somehow enjoying looking at `dependencies` in `pkg5`).  IOW, if we would completely drop `dependencies` from `pkg5` files then everything will still work as before.

The `fmris` list is very similar to `dependencies`, with these differences: it is same as the `env - gmake print-package-names` output and there is one known consumer for these data: `oi-pkg-checker`.

The `name` entry is similar to `dependencies`, with this difference: its value is same as the `env - gmake print-value-COMPONENT_NAME` output (IIRC, I'm lazy to check now to be 100% sure :-)).  This value rarely (I'm not aware about any such component) changes without component's `Makefile` change.  There is no known consumer for this info either.  All of this makes the added value of `name` very close to zero.

It is clear from what I wrote above that almost no non-human entity cares about `pkg5` files and their contents.

mtelka:
Second, runtime dependencies in packages:

There are two ways how a package (please note the difference between a _package_ and a _component_; if the difference is unclear for you feel free to ask) gets its runtime dependencies listed inside the package: the _explicit_ way and the _implicit_ way.  By the _runtime dependencies for a package listed inside the package_ I mean output of commands like `pkg contents -mr runtime/python-39 | grep ^depend`.

The _explicit_ way how such a dependency gets into the package is to simply add such dependency explicitly into the package manifest, for example:
```
$ grep ^depend components/python/python39/idle.p5m 
depend fmri=pkg:/library/python/tkinter-39 type=require
$
```
Please note that the full runtime dependencies for a package could differ significantly from the explicit list of deps.  Please compare output from the following two commands: `grep ^depend components/python/python39/python.p5m` and `pkg contents -mr runtime/python-39 | grep ^depend` to see the difference (please note that the `runtime/python-39` package is constructed from the `components/python/python39/python.p5m` manifest).  The difference is caused by _implicit_ dependencies.  Please note that both terms _implicit_ and _explicit_ are my simplification for this comment and you'll very likely not find them in any other documentation.

The _implicit_ dependencies is something that magically appears in packages without any clearly visible and easily understandable reason :-).  However, the magic behind this is a little tool named `pkgdepend`.  This tool is able to magically find all such implicit dependencies (see `man pkgdepend` for details how it works).  There is small drawback in the `pkgdepend` magic: it is able to work properly only in a case when the runtime dependency package is already installed on the build machine.

Example with the `runtime/python-39` package:
```
$ pkg contents -mr runtime/python-39 | grep ^depend.*bzip2
depend fmri=pkg:/compress/bzip2@1.0.8-2022.1.0.1 type=require
$
```
The `pkgdepend` command is able to find the `compress/bzip2` dependency only when the `compress/bzip2` package is installed when you run the `env - gmake publish` command.

And now, finally, the `REQUIRED_PACKAGES` list comes to the game: this is the list of packages that should be installed to get the package fully built (and published).  It should contain:
- _implicit_ runtime dependencies (so `pkgdepend` works properly during `env - gmake publish`) - these are detected automatically using the `env - gmake REQUIRED_PACKAGES` command,
- build dependencies - any packages that are not _implicit_ runtime dependencies, but that are needed to properly build the package - these are _(usually)_ not detected automatically and they should be listed manually.  Please note the _(usually)_ word.  I wrote it because there are few build-only dependencies that are already automatically detected using the `env - gmake REQUIRED_PACKAGES` command and so they looks like _implicit_ runtime dependencies for an untrained eye.

Anything else should _not_ be in the `REQUIRED_PACKAGES` list because there is simply no reason to have it there.

Back to your `# Runtime dependencies` comment: after reading all above it should be clear that such a comment is non-sense and it had to read `# Build dependencies` (or so) instead.

Disclaimer: The full logic behind this is a bit more complex than described above (for example, there are other lists involved here like `USERLAND_REQUIRED_PACKAGES` and/or `TEST_REQUIRED_PACKAGES`) but it does not affect the reason why the `# Runtime dependencies` comment is a non-sense :-).

Questions?

PR #21333, opened 2025-03-28
illumos-gate: drop fmd guard patch after upstream fix

AndWac:


Bill-Sommerfeld:
The upstream fix has landed in illumos-gate as commit 71487f875a8d459180cea558ee6dc176c579bfe8

PR #21367, opened 2025-03-30
bacula: SPARC needs additional dependency to system/library/math.

klausz65:


mtelka:
Any idea why some components needs this and some other does not?  If there is some rule then we can do this automatically and system-wide.

klausz65:
I guess it depends on some math instrinsics which are there on x86/64 but not on SPARC - but I can't name them exactly.

mtelka:
Isn't this a bug in illumos then?

mtelka:
... or in gcc?

klausz65:
Not in the one nor in the other - just because of the different types of processors.

mtelka:
Maybe it would be better to do this in `shared-macros.mk`, or some other similar file (and so system-wide) instead of touching many components.

klausz65:
Yes, would be great, but then you would probably introduce dependencies where they are not needed if you do it in general for SPARC. There weren't so many components where I needed to add the math library for SPARC since the last few years.

mtelka:
Since `system/library/math` comes from `illumos-gate` then it should not be a big problem to have all components to depend on it.  We had something similar in past (for all platforms unconditionally) and it didn't cause any big issue.  For example, we had `shell/ksh93` there between Apr 2021 and Mar 2023.

klausz65:
Well, it sounds great to me, to do this on a central basis - but I think we should discuss this next Wednesday in the maintainer-meeting togeter with the other guys - would you also attend the meeting?

mtelka:
> Well, it sounds great to me, to do this on a central basis - but I think we should discuss this next Wednesday in the maintainer-meeting togeter with the other guys - would you also attend the meeting?

I see no reason to discuss this more than we did here.  I think you should just create a PR and if it gets merged then you'll already get the discussion completed :-).

I do not plan to attend the meeting.  I think I was there once (or so) few years back and it was mostly waste of (my) time, IIRC.

PR #21391, opened 2025-03-31
gtksourceview-4: drop 32 bit libs and correct dependency to gnome/accessibility/at-spi2-core.

klausz65:


AndWac:
The more relevant dropping of 32 bit is not mentioned.

PR #21394, opened 2025-03-31
lighttpd: update to 1.4.78 and fix dependency to library/security/ope…

klausz65:
…nssl-3.


AndWac:
Are you sure that the configuration patch isn't needed anymore? It changed the user/group from lighttpd to webservd which is being used for SMF.

AndWac:
The man pages patch should also be kept. The changes therein are forced by the other patch.

PR #21424, opened 2025-04-02
python/multidict: update to 6.3.1

mtelka:


mtelka:
This version was yanked at PyPI.

PR #21435, opened 2025-04-03
rebuild vlc after libtheora update

AndWac:


mtelka:
Wait.  This will clash with the protobuf update (#21434).  Please merge #21434 first and then rebuild this.

Thanks.

mtelka:
> ... and then rebuild this.

Once the updated protobuf is installed.

components/encumbered/vlc/Makefile
```diff
@@ -38,6 +39,8 @@ include $(WS_MAKE_RULES)/common.mk
 
 PATH= $(PATH.gnu)
 
+CONFIGURE_ENV += BUILDCC="/usr/bin/gcc"
```
mtelka, line 42: Is this the compiler used to build this component?  If so, then `/usr/bin/gcc` could point to any version.  Wouldn't be `$(CC)` or so better?

PR #21463, opened 2025-04-05
gst-devtools: enable gst-dots-viewer

iigs:
Patch single-instance rust crate (dependency) to add Illumos support.   Since the single-instance crate source  is not in gsts-devtools, it needs to be downloaded after gst-devtools is extracted.  The crate patch gets applied after configure stage, but before build stage.

The Linux inner code in the crate can't be used on Illumos since there is no support for abstract sockets.  Instead, copy/paste macos inner code.  Illumos doesn't have __error so replace with ___errno . Does any rust coders want to verify that this looks good?

I have no idea how to test gst-dots-viewer to determine if this works.  The code compiles.  The web server starts.  Connecting to server with a web client gets some text.

### DIFF
--- a/components/encumbered/gst-devtools/Makefile
+++ b/components/encumbered/gst-devtools/Makefile
@@ -34,10 +34,8 @@ COMPONENT_LICENSE_FILE=	validate/COPYING
 include $(WS_MAKE_RULES)/encumbered.mk
 include $(WS_MAKE_RULES)/common.mk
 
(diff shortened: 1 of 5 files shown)


mtelka:
Nice work!

Could you please try to upstream this change directly to single-instance too to solve https://github.com/WLBF/single-instance/issues/18?

Thank you!

PR #21499, opened 2025-04-07
brasero: fix dependencies to library/audio/gstreamer library/audio/gs…

klausz65:
…treamer/plugin/base, fix building with gcc-14.


AndWac:
Please change the title and the commit message to something simple like "rebuild brasero"

mtelka:
> Please change the title and the commit message to something simple like "rebuild brasero"

"rebuild brasero to pollute ips repo with unneeded rebuilds" would be even more accurate.

AndWac:
Superceded by #21643

PR #21500, opened 2025-04-07
rhythmbox: add grillo plugin

klausz65:
…rary/audio/gstreamer library/audio/gstreamer/plugin/base and add library/gnome/grilo, compile using gcc-14.


mtelka:
Do we really need so long commit messages when simple "update dependencies" would be more than enough?

mtelka:
I'd also mention the addition of the grilo plugin.  This is far more important than the dependencies.

BTW, you are not fixing anything here, just updating.  (And so I'd completely avoid the _dependencies_ in the commit message.)

klausz65:
done

PR #21600, opened 2025-04-11
sdl2: add 32-bit library

cgrzemba:
this PR adds the 32-bit library package nessessary for sunray-essential

### DIFF
--- a/components/library/SDL2/Makefile
+++ b/components/library/SDL2/Makefile
@@ -12,13 +12,15 @@
 # Copyright (c) 2016 Jim Klimov
 # Copyright (c) 2020 Andreas Wacknitz
 #
-
+# SunRay utaudio needs 32bit library
+BUILD_BITS= 64_and_32
 USE_PARALLEL_BUILD= yes
 DROP_STATIC_LIBRARIES= yes
 include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		SDL2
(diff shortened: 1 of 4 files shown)


components/library/SDL2/sdl2-32.p5m
```diff
@@ -0,0 +1,34 @@
...
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# We don't deliver static libraries
+<transform path=.+/lib/.+\.a -> drop>
+
```
AndWac: This transform shouldn't be necessary because DROP_STATIC_LIBRARIES is set and cares for it.

components/library/SDL2/Makefile
```diff
@@ -13,12 +13,14 @@
 # Copyright (c) 2020 Andreas Wacknitz
 #
 
+BUILD_BITS= 64_and_32
 USE_PARALLEL_BUILD= yes
```
AndWac, line 17: Please add a comment why 32 bit is needed. Otherwise it is prone to be removed again.

PR #21601, opened 2025-04-11
mpg123: add 32 bit library package

cgrzemba:
this PR adds the 32-bit library package necessary for sunray-essential


cgrzemba:
breaking cyclic dependeny pulseaudio - mpg123 - libsndfile - pulseaudio  similar to https://bugs.gentoo.org/915858

PR #21674, opened 2025-04-15
memcached: update to: 1.6.38

klausz65:


components/network/memcached/Makefile
```diff
@@ -81,8 +81,8 @@ COMPONENT_TEST_TRANSFORMS += \
 TEST_REQUIRED_PACKAGES += library/perl-5/io-socket-ssl-538
```
AndWac, line 82: Please replace that with
TEST_REQUIRED_PACKAGES.perl += library/perl-5/io-socket-ssl
This should create a dependency on the default perl version (5.40) of this package.

components/network/memcached/Makefile
```diff
@@ -50,7 +50,7 @@ COMPONENT_PRE_CONFIGURE_ACTION = \
 # otherwise the link with ssl library does not work
 # standard linker path does not make sense anyway
 COMPONENT_POST_CONFIGURE_ACTION = \
-	($(GSED) -i -e 's:-L/usr/lib/amd64::' $(@D)/config.status)
+	($(GSED) -i -e 's:-L/usr/lib/$(MACH64)::' $(@D)/config.status)
```
mtelka: What about `$(USRLIBDIR.$(BITS))`?

components/network/memcached/Makefile
klausz65: done

components/network/memcached/Makefile
klausz65: done

components/network/memcached/Makefile
```diff
@@ -78,11 +78,11 @@ COMPONENT_TEST_TRANSFORMS += \
 	'-e "/^Result/p" '
 
 # Manually added test dependency
-TEST_REQUIRED_PACKAGES += library/perl-5/io-socket-ssl-538
+TEST_REQUIRED_PACKAGES += library/perl-5/io-socket-ssl
```
AndWac, line 82: .perl is missing

components/network/memcached/Makefile
```diff
@@ -50,7 +50,7 @@ COMPONENT_PRE_CONFIGURE_ACTION = \
 # otherwise the link with ssl library does not work
 # standard linker path does not make sense anyway
 COMPONENT_POST_CONFIGURE_ACTION = \
-	($(GSED) -i -e 's:-L/usr/lib/amd64::' $(@D)/config.status)
```
mtelka, line 53: While looking at this again...

Is this `gsed` really needed?  What happens if you completely remove this `COMPONENT_POST_CONFIGURE_ACTION`?

components/network/memcached/Makefile
mtelka: @klausz65 FYI, without `.perl` it works too, but it is sub-optimal (usually brings in more packages than the needed minimum).

klausz65: I don't understand where to put ".perl" ?

components/network/memcached/Makefile
klausz65: It simply does not link!

components/network/memcached/Makefile
AndWac: Look at my original comment.

components/network/memcached/Makefile
mtelka: I see.  It is because `-L/usr/lib/amd64` goes before the `-L/usr/openssl/3/lib/amd64` on the command line and so linker tries to link with openssl 1.0 instead of 3.

BTW, this is a bug in memcached's `configure.ac` because they come with non-sense like:
```
S["LDFLAGS"]="-L/usr/lib/amd64 -m64 -Wl,-rpath,/usr/lib/amd64/lib"
```
in `config.status` (note, `/usr/lib/amd64/lib` does not exist).

I think it would be worth reporting this issue upstream.

(Note: I played with old version 1.6.23, not 1.6.38)

PR #21685, opened 2025-04-16
libgnomekbd: readd

cgrzemba:
readd libgnomekbd as a requirement for sunray-essential

### DIFF
--- a/components/desktop/gnome2/libgnomekbd/Makefile
+++ b/components/desktop/gnome2/libgnomekbd/Makefile
@@ -1,0 +1,51 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2015 Alexander Pyhalov
+#
+
+include ../../../../make-rules/shared-macros.mk
+
+BUILD_STYLE=configure
+COMPONENT_NAME= libgnomekbd
+
+COMPONENT_VERSION= 2.91.92
+COMPONENT_REVISION= 0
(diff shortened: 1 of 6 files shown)


components/desktop/gnome2/libgnomekbd/Makefile
```diff
@@ -0,0 +1,53 @@
...
+COMPONENT_REVISION= 0
+COMPONENT_SUMMARY= GNOME kbd library
+COMPONENT_SRC= $(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE= $(COMPONENT_SRC).tar.bz2
+COMPONENT_ARCHIVE_HASH= \
+  sha256:3c50c54ee3a74a1d7a17517d4d3e72237fe272770f05879d6053b4e8bc6fc8d8
+COMPONENT_ARCHIVE_URL= \
+  http://ftp.gnome.org/pub/GNOME/sources/$(COMPONENT_NAME)/$(basename $(COMPONENT_VERSION))/$(COMPONENT_ARCHIVE)
+COMPONENT_PROJECT_URL = http://www.gnome.org
+COMPONENT_CLASSIFICATION = System/Libraries
```
AndWac, line 31: Please use https where possible

components/desktop/gnome2/libgnomekbd/Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+include $(WS_MAKE_RULES)/common.mk
+
+PATH=$(PATH.gnu)
+
+CFLAGS+= -Wno-error=incompatible-pointer-types
+CFLAGS+= -Wno-error=implicit-function-declaration
+
+CONFIGURE_OPTIONS += --sysconfdir=/etc
+
```
AndWac: This shouldn't be needed as it is the default.

AndWac:
You should also add a comment at a prominent location with an explanation why this package is needed even if no other package from oi-userland depends on it.

components/desktop/gnome2/libgnomekbd/Makefile
cgrzemba: done

PR #21687, opened 2025-04-16
gnome-settings-daemon: add renamed package gnome-settings-daemon-2

cgrzemba:
(re)add Gnome2-settings-daemon as a requirement for sunray-essential


mtelka:
I think it would be better first to complete the sunray-essential package (#21290) and only after that proceed with these re-additions.

Thank you.

cgrzemba:
sunray-essential cannot work without these others.

mtelka:
How so?  You can easily comment out unsatisfied (yet) dependencies.

PR #21722, opened 2025-04-17
clutter-gst: fix dependencies

klausz65:


components/library/clutter-gst/Makefile
```diff
@@ -15,11 +15,11 @@
 #
 
 BUILD_STYLE=configure
-BUILD_BITS=64
 include ../../../make-rules/shared-macros.mk
```
AndWac, line 17: BUILD_STYLE=configure is also superfluous because it's our default.

klausz65: done -  thanks

PR #21749, opened 2025-04-19
Add backports_datetime_fromisoformat Python project

mtelka:
This is needed for marshmallow 4.0.0.


AndWac:
Why not switching to python 3.13 like OmniOS did instead of this?

mtelka:
> Why not switching to python 3.13 like OmniOS did instead of this?

Because we do not have Python 3.13 integrated yet?

PR #21755, opened 2025-04-19
libvncserver: update to 0.9.15

klausz65:


components/library/libvncserver/Makefile
```diff
@@ -51,12 +47,14 @@ CMAKE_OPTIONS += -DJPEG_INCLUDE_DIR=$(JPEG_INCDIR)
 CMAKE_OPTIONS += -DJPEG_LIBRARY=$(JPEG_LIBDIR)/libjpeg.so
```
AndWac: You should be able to remove the jpeg-related flags completely.
It should even be possible to remove the CMAKE_OPTION related to jpeg.

PR #21787, opened 2025-04-22
libX11: update to 1.8.12

AndWac:


AndWac:
There is a problem with this PR I haven't been able to solve yet:
The symbols in mapfile-vers-64 should be global but aren't anymore for an unknown reason.

AndWac:
Ok, I found the missing variable in the patch for the abi. This works now.

PR #21801, opened 2025-04-23
web/php-82/extension/php-xdebug: update to 3.4.2 & fix xdebug.ini

hoewweken:
Hello!
i found this tiny error:

$ cat /etc/php/8.2/conf.d/xdebug.ini
[xdebug]
zend_extension="/usr/php/8.1/extensions/xdebug.so"

and this PR should be the fix for it,
 along with an update to the most recent upstream version 3.4.2 wrt https://pecl.php.net/package/xdebug.


hoewweken:
Ah shit. Can i remove the commit for 8.3 here ?

PR #21853, opened 2025-04-26
ibus-pinyin : update to 1.5.1

jwntree:


AndWac:
This is not an update but a new package

jwntree:
yes, ibus/pinyin is already available on pkg.openindia.org, but there is no build recipe.
By the way, I force pushed the same commit again, but the build failed. I'll check if I missed something.

AndWac:
You are right, I didn't know that. Failing builds on the PR-checker are common. I have tested it locally.

PR #21886, opened 2025-04-30
Update to Valgrind 3.25

paulfloyd:
Hello from upstream.

I'm not certain that I got all the dependency changes right.

A summary of changes

- Removed all patches. There are all either merged or not needed.
- Removed --enable-inner from the build This is only needed for Valgrind developers wanting to run Valgrind under Valgrind. Definitely not for everyday use.
- Added new headers (mostly for riscv64 and arm64-freebsd)

Regtest is now almost clean. There are 3 kinds of outstanding error.

1. gdbserver errors where there is a ctrl-c. That is supposed to interrupt Valgrind and give control back to gdb/vgdb. That seems to work, but continuing afterwards in gdb results in Valgrind degfaulting.
2. pthread barrier issues in Helgrind. I think this is due to the pthread_barrier* implementation in illlumos libc causing nondeterministic behaviour in Helgrind.
3. One testcase with a 64Mbyte deep stackframe (64 times 1Mbyte) is producing false positive invalid read/write errors.


AndWac:
Please also update test/results-all.master after running gmake test:
cp build/test/amd64/results-64.snapshot test/results-all.master

AndWac:
You seem to have worked on the master branch. This is highly discouraged in any project. Please always work on branches instead. Your PR now has 2 commits because of that.

AndWac:
Running the tests on my machine gives
== 913 tests, 6 stderr failures, 0 stdout failures, 16 stderrB failures, 3 stdoutB failures, 1 post failure ==
Probably the tests depend on some packages that are not yet installed on your syste.

AndWac:
Please also run
gmake REQUIRED_PACKAGES
and then either gmake publish or gmake update-metadata

The former will update the build dependencies in the Makefile (the region marked with "Auto-generated dependencies".
The ladder will update the pkg5 file (which you also should add to your PR).

paulfloyd:
> Running the tests on my machine gives == 913 tests, 6 stderr failures, 0 stdout failures, 16 stderrB failures, 3 stdoutB failures, 1 post failure == Probably the tests depend on some packages that are not yet installed on your syste.

Don't just try to guess what is happening. I ran the tests on an old machine (I couldn't get OI installed on my new PC). At configure time Valgrind detects many OS and hardware dependent features. That also determines some of the CPU specific tests that get built. So on old hardware fewer tests get built and executed. On FreeBSD I get about 20 more testcases executed on my new PC compared to the old one.

paulfloyd:
> You seem to have worked on the master branch. This is highly discouraged in any project. Please always work on branches instead. Your PR now has 2 commits because of that.

Let me know if you'd like me to start over with a new PR.

paulfloyd:
> Please also run gmake REQUIRED_PACKAGES and then either gmake publish or gmake update-metadata
> 
> The former will update the build dependencies in the Makefile (the region marked with "Auto-generated dependencies". The ladder will update the pkg5 file (which you also should add to your PR).

The diff that I get for Makefile is

```
diff --git a/components/developer/valgrind/Makefile b/components/developer/valgrind/Makefile
index 58f8618fc4..de76214ec1 100644
--- a/components/developer/valgrind/Makefile
+++ b/components/developer/valgrind/Makefile
@@ -82,8 +82,6 @@ COMPONENT_TEST_TRANSFORMS += '-n -e "/^== /p"'
 
 # Build dependencies
 # Auto-generated dependencies
-REQUIRED_PACKAGES += $(GCC_RUNTIME_PKG)
 REQUIRED_PACKAGES += SUNWcs
-REQUIRED_PACKAGES += developer/gcc-13
 REQUIRED_PACKAGES += runtime/perl
```

I don't understand why the GCC runtime has been removed, and why there was a specific GCC version dependency that has also been removed.

Is this list also used for "gmake check"? If so then it is missing python39 and gdb.

AndWac:
When I run it here gcc-13 will be replaced by gcc-14 which is our default gcc and correct.

paulfloyd:
I'm abandoning this change and will start over with a clean change from a branch.

PR #21919, opened 2025-05-02
Packages for database-specific drivers for libdbi.  

Bill-Sommerfeld:
Note that this package includes tests but they don't build cleanly with gcc-14 and they require manual configuration (to point at a running database instance of each type).

### DIFF
--- a/components/library/libdbi-drivers/Makefile
+++ b/components/library/libdbi-drivers/Makefile
@@ -1,0 +1,72 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 Bill Sommerfeld
+#
+
+DROP_STATIC_LIBRARIES= yes
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME= libdbi-drivers
+COMPONENT_VERSION= 0.9.0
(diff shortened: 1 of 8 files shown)


components/library/libdbi-drivers/mysql.p5m
```diff
@@ -0,0 +1,27 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/lib/$(MACH64)/dbd/libdbdmysql.a
+file path=usr/lib/$(MACH64)/dbd/libdbdmysql.so
```
AndWac, line 26: Usually we don't deliver static libs if shared are available. Is there a good reason to break this rule here?

Bill-Sommerfeld: No reason to break the rule here.   I removed `*.a` from packaging.

components/library/libdbi-drivers/Makefile
```diff
@@ -0,0 +1,71 @@
...
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 Bill Sommerfeld
+#
+
```
AndWac, line 15: You can add
DROP_STATIC_LIBRARIES= yes
here and the static libs won't show up in the sample-manifest.

Bill-Sommerfeld: Done.

PR #21925, opened 2025-05-02
stack-spur: update to VMMaker.oscog.eem-3541

cstes:


cstes:
I tested the OpenSmalltalk stack-spur VM and it opens v6.1 images from www.squeak.org.

I'll open later a pull request for the OpenSmalltalk cog-spur (just in time compiler).

PR #21946, opened 2025-05-05
unrar: fix unrar -v

iigs:
My previous attempt to fix unrar -v was faulty since I had tested on an earlier BE.  The problem was not an alignment issue (removed that patch).  The problem is that the illumos fputws does not work correctly.  The workaround for unrar is to replace fputws with fprintf.


Bill-Sommerfeld:
> The problem is that the illumos fputws does not work correctly.

Can you file a bug against illumos describing how it misbehaves so we can get this fixed?

PR #21947, opened 2025-05-05
ibus-table : update to 1.17.12

jwntree:
Currently, ibus/table's tables are splitted into several packages.
I suggest shipping only ibus-table-chinese and ibus-table-others and sunset old packages.

In this case, the following input engine will not be delivered:
ibus-table-ziranma
ibus-table-zhengma
ibus-table-xinhua

They are still on ibus Google Code, so anyone who needs it can install it themselves.
It is not shipped with most Linux distros too, might patent issue. (As far as I checked, they already expired.)


AndWac:
Then please add a PR where these obsoletions happen. Probably a local history file.

PR #21961, opened 2025-05-05
gnucash-docs: Add docs for version 5.11

Bill-Sommerfeld:


Bill-Sommerfeld:
Question for reviewers: any idea if there's a conventional way to tag multi-lingual docs so you can pick & choose which language gets installed?

AndWac:
facet.locale.%1 should do the trick.
My recommendation in such cases is always: look for examples in oi-userland. Most problems have already be fixed there it's just a matter to find them. find . -name "*.p5m" -exec grep facet.locale \{\} \\; -print gives you examples; eg. in desktop/gnome2/GConf/gconf.p5m.

Bill-Sommerfeld:
> facet.locale.%1 should do the trick.

Wasn't sure what to grep for; that did the trick.  Thanks!

PR #21987, opened 2025-05-08
history: obsolete old apache-24 python modules.

klausz65:


mtelka:
Please do not move the history data to `meta-packages/history/history`.

Please note that the `pkg5` file still contains the `-37` package.  To properly obsolete the `-37` package you should add it to the `web/apache2-modules/mod_wsgi/history` file and make sure the component builds properly (and all metadata are updated accordingly).

Thank you.

PR #21988, opened 2025-05-08
virtualbox: add fix for illumos 17347

cgrzemba:


components/sysutils/virtualbox/patches/29-illumos17347.patch
```diff
@@ -0,0 +1,13 @@
...
++++ b/src/VBox/Runtime/r3/posix/RTTimeZoneGetCurrent-posix.cpp
+@@ -51,7 +51,7 @@
+ #include <iprt/symlink.h>
+ #include <iprt/stream.h>
+ 
+-#if defined(RT_OS_DARWIN) || defined(RT_OS_SOLARIS)
++#if defined(RT_OS_DARWIN)
+ # include <tzfile.h>
+ #else
+ # define TZDIR                   "/usr/share/zoneinfo"
```
Bill-Sommerfeld, line 16: Our `TZDIR` is `/usr/share/lib/zoneinfo`.

Bill-Sommerfeld (changes requested):
Maybe we should ship a compat symlink but we don't currently.

mtelka:
The [illumos#17347](https://www.illumos.org/issues/17347) is already closed with the change integrated upstream.  Why do we need a fix for the already solved case?

cgrzemba:
This change has generated the problem.

mtelka:
Then the commit message is confusing.  Something like `illumos no longer ships tzdata.h` would be far better.  Preferably with a bugreport to virtualbox so they can update their sources upstream.

Thank you.

cgrzemba:
We are sitting together in the boat with Oracle Solaris and they still ship tzdata.h so its unlikely that upstream would change something here.

mtelka:
> We are sitting together in the boat with Oracle Solaris and they still ship tzdata.h so its unlikely that upstream would change something here.

Not entirely.  We already do have `__illumos__` - see https://www.illumos.org/issues/13726.

components/sysutils/virtualbox/patches/29-illumos17347.patch
cgrzemba: But the file /usr/include/sys/tzfile.h does not define TZ_MAGIC ...

cgrzemba: I will update the patch

cgrzemba: It seems that all will not really help, because we have no /etc/localtime or /etc/timezone file.
/etc/default/init has the information, but have to parsed.
What would work is a symlink like
lrwxrwxrwx   1 root     root          39 Jul  4  2023 /etc/localtime -> ../usr/share/lib/zoneinfo/Europe/Berlin
which has Solaris, but not illumos.

src/VBox/Runtime/r3/posix/RTTimeZoneGetCurrent-posix.cpp:131

cgrzemba: or an environment variable TZ

mtelka:
> virtualbox: fix for illumos 17347, illumos no longer ships tzdata.h

This is not a fix for 17347.

PR #22018, opened 2025-05-11
xorg-cyrillic: update xorg-cyrillic fonts.

klausz65:


mtelka:
Please do the update by splitting out individual fonts to separate packages - see #22023 for an example and rationale.

Thank you.

klausz65:
Hi, I'm also trying to update xorg-core fonts - those will be 25 updated fonts - shall I also split therse into seperate packages?

mtelka:
I prefer to do so.  However, the `xorg-core` is a bit more complex than `xorg-cyrillic`. I'll try to do one font split-out as a proof of concept later.

klausz65:
This one has been obsoleted

PR #22027, opened 2025-05-12
font-screen-cyrillic: update to 1.0.5

klausz65:


mtelka (changes requested):
You need to regenerate sample manifest for `xorg-cyrillic` too.

klausz65:
What to do if the last one in cyrillic has been sperated? - decom xorg-cyrillic ? rename the last in it? - I can sent you a Makefile diff for xorg-core, so you do need to figure out the hashes in it - would save some time of course!

klausz65:
Or better, if the last one has been seperated - move xorg-cyrillic to metapackages with dependencies to the 4 seperated cyrillic ones, we could do the same for xorg-core - what do you think ?

components/fonts/xorg-cyrillic/Makefile
```diff
@@ -29,24 +29,19 @@ include ../../../make-rules/shared-macros.mk
...
-	sha256:824231e8dffe15299454e47259f29d98001c9cf8ad3d6b5171399e4d71705e79
-
 COMPONENT_NAME_3=	font-winitzki-cyrillic
 COMPONENT_VERSION_3= 	1.0.3
 COMPONENT_ARCHIVE_HASH_3=	\
 	sha256:abd13b63d02fcaec488686c23683e5cf640b43bd32f8ca22eeae6f84df0a36a0
 
 # Use font-screen-cyrillic version number as solaris-userland does
-IPS_COMPONENT_VERSION=$(COMPONENT_VERSION_2)
+IPS_COMPONENT_VERSION=$(COMPONENT_VERSION)
```
mtelka: This line does have two problems now:

- it no longer matches the comment above, and
- it will cause the package version decrease (1.0.4 -> 1.0.3).

I suggest to hardcode `IPS_COMPONENT_VERSION` to `1.0.4`.

mtelka (changes requested):
You should also remove the `font-screen-cyrillic` license from the `xorg-cyrillic.license` file.

klausz65:
okay I hardcoded version to 1.0.4 - but what is the part for font-screen-cyrillic in the xorg-cyrillic.license file?

mtelka:
> what is the part for font-screen-cyrillic in the xorg-cyrillic.license file?

What exactly you did to find the answer yourself?

klausz65:
Had a look in xorg-cyrillic.license file and found no font-screen relevant part in it.

mtelka:
What about to look into the `font-screen-cyrillic-1.0.x/COPYING` file too?

klausz65:
now okay

mtelka (approved):

PR #22050, opened 2025-05-14
mod_wsgi: update to 5.0.2 and obsolete python-3.7 module

klausz65:


mtelka (changes requested):
Did you try to build this?  I doubt because there is missing the `pkg5` file update.

Also, it would be great to get this component updated because version 4.9.4 is almost three years old.

klausz65:
done

mtelka (approved):
Looks far better now.  Thanks.

PR #22061, opened 2025-05-16
pystring: add SPARC support

klausz65:


components/library/pystring/Makefile
```diff
@@ -29,10 +29,12 @@ COMPONENT_FMRI=				library/pystring
 COMPONENT_CLASSIFICATION=	System/Libraries
 COMPONENT_LICENSE=		MIT
 COMPONENT_LICENSE_FILE=	LICENSE
+COMPONENT_BUILD_ARGS=	-j12
```
mtelka: Please _never_ do this.

components/library/pystring/patches/01-Makefile.patch
```diff
@@ -5,7 +5,7 @@
 -LIBDIR ?= /usr/lib
 +PREFIX ?= /usr
 +INCLUDEDIR = ${PREFIX}/include/pystring
-+LIBDIR ?= ${PREFIX}/lib/amd64
++LIBDIR ?= ${PREFIX}/lib/MACH64
```
mtelka, line 8: Why about to set `LIBDIR` during build instead of doing this?

mtelka (changes requested):

components/library/pystring/Makefile
klausz65: fixed.

components/library/pystring/patches/01-Makefile.patch
klausz65: tried to add: COMPONENT_BUILD_ENV += LIBDIR="/usr/lib/$(MACH64)" in Makefile, but the result is still amd64.

mtelka: What happens when you try to add this to `COMPONENT_BUILD_ARGS` instead?

klausz65: tried: COMPONENT_BUILD_ARGS - same result! - we did the sed approach quite often in the Makefile, so why do want to change it to anything else?

PR #21774, opened 2025-04-22
perl/Math-Pari: update to 2.030528

mtelka:

PR #22097, opened 2025-05-18
Math-Pari: update to 2.030528

klausz65:


mtelka:
Few notes:

- this is basically a duplicate of #21774,
- please never manually modify `Makefile`s created by `perl-integrate-module`,
- the `Math-Pari` update needs more work because of [this](https://github.com/OpenIndiana/oi-userland/pull/21774#issuecomment-2821296980).

klausz65:
Okay, as mtelka had the same I'll closed this one

PR #22158, opened 2025-05-23
python/setuptools: update to 80.8.0

mtelka:


AndWac:
I am not sure whether we should continue to update python packages until we can manage the situation about IPS.
Probably the best thing would be to decouple IPS from the development of the python packages to guarantee its stability.
Otherwise any package update needs thorough checks which is a thing I cannot provide.

mtelka:
> I am not sure whether we should continue to update python packages until we can manage the situation about IPS. Probably the best thing would be to decouple IPS from the development of the python packages to guarantee its stability. Otherwise any package update needs thorough checks which is a thing I cannot provide.

IPS works.  Or, is there some issue I'm not aware about? The only problem is that we cannot build it now.  We built it successfully ca 9 months ago and so the build breakage happened anytime between August 2024 and now.  We (or at least I) do not know when exactly.

I'm not sure I understand how the outdated `setuptools 80.4.0` (our current version) is better than `80.8.0` in regard to IPS when both equally prevents the [pkg5](https://github.com/OpenIndiana/pkg5) from being built.  BTW, there exists simple workaround for [pkg5](https://github.com/OpenIndiana/pkg5) to make its build working with latest `setuptools`, but unfortunately the build does not work even with that because of some other issue (probably not related to Python at all, but I'm not sure yet).

BTW, it is always that a package update could break some other packages (either runtime or build) and we obviously cannot guarantee it will never happen with the rolling release model and limited resources we do have.  OTOH, it should be possible to create a daily/weekly/monthly jenkins job to rebuild all oi-userland to see detect new build breakages ASAP.

PR #22171, opened 2025-05-24
tidy: add back 32bit lib for PHP build

klausz65:


mtelka:
Why do we need 32-bit tidy to build 64-bit only PHP?

mtelka:
I suspect this is similar non-sense as it was previously with uwimap - see #21242.

klausz65:
No need for 32bit tidy lib

klausz65:
No Need for 32bit tidy lib

PR #22173, opened 2025-05-24
gphoto2: fix dependencies

klausz65:


components/desktop/gphoto2/Makefile
```diff
@@ -46,10 +46,10 @@ unexport SHELLOPTS
 # test031 fails
```
AndWac, line 39: Please also remove the jpeg related flags.
unexport SHELLOPTS should also not be needed anymore.

PR #22195, opened 2025-05-25
python/aiohttp: update to 3.12.0

mtelka:


components/python/aiohttp/Makefile
```diff
@@ -99,6 +102,7 @@ TEST_REQUIRED_PACKAGES.python += library/python/frozenlist
 TEST_REQUIRED_PACKAGES.python += library/python/gunicorn
 TEST_REQUIRED_PACKAGES.python += library/python/multidict
 TEST_REQUIRED_PACKAGES.python += library/python/mypy
+TEST_REQUIRED_PACKAGES.python += library/python/pkgconfig
 TEST_REQUIRED_PACKAGES.python += library/python/propcache
```
AndWac, line 106: Does it make sense to have pkgconfig and propcache as TEST_REQUIRED_PACKAGES if they are needed to build the package anyways?

mtelka: Such duplicates are harmless and comes from the `aiohttps` project itself.  They are detected automatically by our build framework.

It is possible to remove such duplicates (either by patching `aiohttps` or using some generic approach in our `make-rules` directory), but it would be non-trivial work for almost zero profit.

PR #22268, opened 2025-05-29
qtcreator: update to 8.0.2

klausz65:
Okay 8.0.2 isn't the newset one, but releases 9.0.2 and 10.0.2 use a so called QT6 -> QT5 wrapper mode, which I wasn't able to initialze. Second this update uses it's internal copy of yaml-cpp, because our current yaml-cpp is to new and can't be used to build qtcreator anymore (also releases 9/10.x). The main difference to the old 7.0.2 build is that the libClangSupport.so and clangbackend are missing in this build. Other than, I installed the old version on my system and tried it out a little. After that I installed this update and tested again - from the look and feel they are both useable - however I'm not a real qtcreator user - so I can't tell if all plugins do work the same as before, but in general it should be useable.
The main propose of this update is to get rid of runtime|developer/clang-13 and it would be very nice/friendly if I could obsolete both of them - as this was quite some work - thanks.
Klaus


mtelka:
I think it is obvious that nobody is using `qtcreator` at OI, so we should probably simply obsolete it to make things easier.

Otherwise, I'm glad to see steps to enable obsoletion of `clang-13` :-).  Thank you.

PR #22334, opened 2025-06-06
desktop-cache: fix manifest, use /tmp cache files while installing

klausz65:


AndWac:
There is a typo in the commit message: mainfest -> manifest

klausz65:
Okay, now the SMF cache services are all Online during install, but there is still the cups/scheduler SMF service to be fixed, so if you think to just re-create installation images - wait! that's the next one I'm going to fix.

AndWac:
How did you test these changes?

klausz65:
By always booting with these new SMF methods from the net and see what svcs -xv tells me, but before I did a remove of all the cache files from an installed system. Then rebooted and looked if the cache files got re-created - okay so far. Of course all only on SPARC. However as you know I'don't have a X11 desktop on SPARC, so it would probably be a good idea to test this also using a X11 desktop to be 100% sure - sorry I only could test recreation / refresh of the services and of course the behavior during install.

PR #22372, opened 2025-06-09
asterisk - fix missing symbol (__gxx_personality_v0) by extra linking…

fritzkink:
… stdc++


AndWac:
You also need to re-run gmake REQUIRED_PACKAGES and update pkg5.

PR #22399, opened 2025-06-10
squeak: modifications for cmake 4.0.2 build

cstes:
attempt to build using cmake 4.0.2

the build system of Squeak is a mix of a configure script and cmake


components/runtime/smalltalk/squeak/Makefile
```diff
@@ -71,7 +71,7 @@ include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		squeak
 COMPONENT_VERSION=	4.20.10
-COMPONENT_REVISION=	3
+COMPONENT_REVISION=	4
 COMPONENT_SUMMARY=	The Squeak Virtual Machine
```
AndWac: You don't need to increment COMPONENT_REVISION because the last PR hasn't been published.

PR #22447, opened 2025-06-16
libevent: obsolete libevent-1.4.15

klausz65:
modified version of libevent2 will follow shortly.


mtelka:
This package is still needed by `web/browser/links` so we cannot obsolete it yet.

klausz65:
Hi, I know, as soon as #22448 a follow up version of web/browser/links will follow - is already ported to libevent2 :-)

mtelka:
Then you should do it the opposite way: first update `links` and then obsolete `libevent`.

klausz65:
Yes, your right, but that new links package does not build the way libevent2 is installed right now ...

mtelka:
Then you should adapt it.

PR #22550, opened 2025-06-25
libsndfile: add 32bit

cgrzemba:


components/library/libsndfile/Makefile
```diff
@@ -22,12 +22,13 @@
 # Copyright (c) 2021 Andreas Wacknitz
 #
 
+BUILD_BITS= 64_and_32
 USE_COMMON_TEST_MASTER= no
```
AndWac, line 27: Please do yourself a favour and add a comment here that 32 bit is needed to support SUNRay software.

AndWac:
As this needs /audio/mpg123/32 it cannot be build at the moment. Please make sure to provide an order for the PR's to be merged. Otherwise the build server will fail to publish it and the whole process will be more tedious than necessary.

PR #22589, opened 2025-06-28
gegl: new babl rebuild

iigs:
babl 0.1.114 needs to be published before this is built.  babl 0.1.114 has a new .so version number.


iigs:
I also want to mention that the current gegl 0.4.62 does not use lua in its build.  Since the time it was built, something changed with lua on OI so the detection of lua works, but gegl fails to link with luajit.  Disable lua to match the earlier build of 0.4.62.

PR #22658, opened 2025-07-04
uw-imap fix IPS_COMPONENT_VERSION variable

fritzkink:


components/library/uwimap/Makefile
```diff
@@ -28,8 +28,8 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		imap
 COMPONENT_VERSION=	2007f
-COMPONENT_REVISION=	4
-IPS_COMPONENT_VERSION=	0.2007.6
+COMPONENT_REVISION=	5
+IPS_COMPONENT_VERSION=	2007.6
```
AndWac, line 32: COMPONENT_REVISION doesn't need to be updated as we don't have 4 with that version yet.

PR #22660, opened 2025-07-04
Updates to Emacs package

nealsid:
- Update README to remove out of date information
- Update site-start.el to set variables to GNU variants of several utilities, which should help the out-of-the-box experience by reducing the number of errors to debug due to calling programs with GNU-variant-only switches
- Update .gitignore
- Update duplicate package name in emacs script
- Remove 01-find.patch because the variable is now set in site-start.el


AndWac:
You also need to touch the Makefile otherwise our jenkins wouldn't create a new package with your changes.
When a package needs re-publishing without changing its COMPONENT_VERSION we need to set/increment COMPONENT_REVISION in the Makefile.

nealsid:
> You also need to touch the Makefile otherwise our jenkins wouldn't create a new package with your changes. When a package needs re-publishing without changing its COMPONENT_VERSION we need to set/increment COMPONENT_REVISION in the Makefile.

Done, thanks!

PR #22670, opened 2025-07-06
install-types: add mate support for SPARC

klausz65:


klausz65:
library/desktop/gtk2 needs to be included in core, because otherwise SMF service:
application/desktop-cache/input-method-cache:default fails after installation of minimal installation profile.

PR #22954, opened 2025-07-20
opensmalltalk stack-spur: upgrade to 3566, disable OpenGL support

cstes:
compile with the switch :   --without-gl

upgrade to VMMaker.oscog-eem.3566

should investigate how to reenable OpenGL support in the future (add perhaps in a package or use a mediator)


cstes:
this PR removes the "mesa" dependency (at least for the moment)

also the manifest comments out a transform that I was not happy about:
# <transform file -> add pkg.depend.bypass-generate libGL\.so\.1>

I should find a way to enable OpenGL without that transform

AndWac:
This transform is needed by all packages that use OpenGL.

cstes:
For the moment because --without-gl is used, the IPS transform for the libGL bypass is not required. I tested with a Squeak and a Cuis image, and the Smalltalk images for Cuis and Squeak are launched fine. The only limitation could be in very specific usages for Squeak software that tries to use OpenGL, but most Smalltalk software does not depend on OpenGL. Some Smalltalk users like to use OpenGL and I can try to find a solution in the future for that, perhaps after better testing. I must add that I never tested with graphics cards, perhaps I should test say on a Lenovo LOQ laptop with NVIDIA RTX 3050 or NVIDIA RTX 5050 (gaming laptops) where OpenGL support in the graphics card is important, and then see whether Squeak OpenGL support is supporting those graphics cards.

AndWac:
I recommend to test this with an ordinary desktop PC with a supported NVIDIA graphics card first. Notebooks have problems on their own. Usually it's hard to get the graphics card working (need to fiddle with BIOS settings) especially when you have a combination of NVIDIA and Intel graphics which is quite common.

PR #22956, opened 2025-07-20
squeak: patch for cmake 4.0.3 and disable OpenGL support

cstes:
compile with the switch :   --without-gl

should investigate how to reenable OpenGL support in the future (add perhaps in a package or use a mediator)

also update /patch the CMakeLists.txt for cmake 4.0.3

should check upstream whether this change will be integrated in Squeak repository (a while ago, I submitted the cmake patches and I was told that the patch for cmake 4.0.3 was going to be integrated upstream)

### DIFF
--- a/components/runtime/smalltalk/squeak/Makefile
+++ b/components/runtime/smalltalk/squeak/Makefile
@@ -71,7 +71,7 @@ include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		squeak
 COMPONENT_VERSION=	4.20.10
-COMPONENT_REVISION=	3
+COMPONENT_REVISION=	4
 COMPONENT_SUMMARY=	The Squeak Virtual Machine
 COMPONENT_PROJECT_URL=	http://www.squeak.org
 COMPONENT_ARCHIVE_URL=  http://squeakvm.org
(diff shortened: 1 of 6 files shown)


cstes:
this PR removes the "mesa" dependency (at least for the moment)

also the manifest comments out a transform that I was not happy about:
# <transform file -> add pkg.depend.bypass-generate libGL\.so\.1>

I should find a way to enable OpenGL without that transform

AndWac:
You cannot find one without this transform. We have multiple implementations for OpenGL (two at the moment): mesa and NVIDIA. Which one is being used depends on whether NVIDIA is available or not. A fix dependency on either is wrong.

cstes:
For the moment because --without-gl is used,  the IPS transform for the libGL bypass is not required.  I tested with a Squeak and a Cuis image, and the Smalltalk images for Cuis and Squeak are launched fine.    The only limitation could be in very specific usages for Squeak software that tries to use OpenGL, but most Smalltalk software does not depend on OpenGL.    Some Smalltalk users like to use OpenGL and I can try to find a solution in the future for that, perhaps after better testing.    I must add that I never tested with graphics cards,  perhaps I should test say on a Lenovo LOQ laptop with NVIDIA RTX 3050 or NVIDIA RTX 5050 (gaming laptops) where OpenGL support in the graphics card is important, and then see whether Squeak OpenGL support is supporting those graphics cards.

AndWac:
You are right the transform is only needed if you compile your binaries with OpenGL support. If you don't provide OpenGL support then this transform is superfluous. You can check this easily, when you successfully publish without this transform then it isn't needed. But why disable OpenGL support? Wasn't it working in the past?

cstes:
I started from a build with default settings which enabled OpenGL by default.   "configure" enables OpenGL by default.   Then I created an IPS package and made fixes to the IPS package for example by using the bypass file transform.   However, lately I have been thinking that I could have simply disabled OpenGL support.   As I wrote before, I have not done a lot of testing.   I ran a Smalltalk game with three dimensional graphics in the past on an NVIDIA graphics adapter (GT 440) and I have now also a GT 730 card which also supports OpenGL.   I never tested a more high end NVIDIA graphics card (something that could be done).   Anyway, I can close this PR (or those 3 PR's) and keep it as it was in the past.      So with OpenGL support enabled (and with the file transform for the bypass libGL).

PR #22983, opened 2025-07-21
python312: add python 3.12.11

iigs:
This is a replacement PR for #15955.  

This does not use PR #16200.  GCC_GNU_SHORT_TRIPLET is used internally in Makefile.  If GCC_GNU_SHORT_TRIPLET is so important, PR #16200 would have been integrated already.

All of the tests pass.  This is because no effort was made to limit the amount of memory for each python module.

All of the patches came from #15955 --those that needed to be updated for this release were done so.  51-expat-2-6-test-fail.patch was not included--there is no evidence it is needed.

Python 3.12 is not getting bug fixes any more.  It will get security fixes until 2028.

transform autopyc has to be updated to support minor versions that contains more than one digit.

No effort was made to update ips.mk or shared-macros.mk to start supporting this python release from userland.  I will let someone else do that when it is ready to do so.


components/python/python312/Makefile
```diff
@@ -0,0 +1,211 @@
...
+#
+
+#
+# Copyright (c) 2011, 2020, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2021, Aurelien Larcher
+#
+
+# fewer tests fail
+ENV = /usr/bin/env
```
mtelka, line 29: Please do not do this.

components/python/python312/Makefile
```diff
@@ -0,0 +1,211 @@
...
+# CDDL HEADER END
+#
+
+#
+# Copyright (c) 2011, 2020, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2019, Michal Nowak
+# Copyright (c) 2021, Aurelien Larcher
+#
+
+# fewer tests fail
```
mtelka: In the PR description you claim that all tests pass.  Here that fewer tests fail.  So how is it?

mtelka (changes requested):
Just short and fast initial review.

mtelka (changes requested):
Please move this to `components/python/python-312`.

mtelka:
BTW, I like the idea of doing this without touching the `make-rules` directory.

iigs:
Changed comment in Makefile.  ENV is not changed.
It will be very time consuming to rename every file in this commit to change python312 to python-312.  It is not worth my time to spend an hour or more to manually rename *every* file in this PR to have a new directory structure for python-312.  By keeping the name python312, this matches the existing directory naming scheme for pythonc versions.

iigs:
Fixed p5m files, Makefile, and pkg5.  I had mixed up that I edited the p5m files for a newer python, not the 3.12 files.  I was very confused this past weekend trying to understand why `REQUIRED_PACKAGES += $(OPENSSL_PKG)` wasn't showing up in the Makefile (even though the _ssl library was clearly in the build directory and the tests for SSL passed). Thank you for checking this.

PR #22980, opened 2025-07-20
gettext: update to 0.26

AndWac:

PR #22984, opened 2025-07-21
gettext: disable nls to prevent libintl creation

AndWac:


mtelka:
This fixes https://github.com/OpenIndiana/oi-userland/pull/22980#issuecomment-3095317307

PR #23152, opened 2025-07-29
python-313: Add 3.13.5

iigs:
Several tests do not pass due to patch 34-selectmodule.patch (all tests pass if this is removed).

All of the patches from omnios were used except for cgiserver.patch and ustack.patch.  Additional patches came from Oracle until the source compiling completed successfully.

No effort was made to update ips.mk or shared-macros.mk.  I will let somebody else do that when it is ready.

This is not following prior python releases for naming of directories.  The directory where files are stored now has a hyphen (python-313).  This directory is the same as the package name.


iigs:
Removed unused lines from Makefile

PR #23174, opened 2025-07-30
sunray: add configuration utilities for OI

cgrzemba:
These scritps and patches simplify the configuration of SunRay Server software on Openindiana.

Guide will be updated on docs.openindinana.org

### DIFF
--- a/components/openindiana/sunray/Makefile
+++ b/components/openindiana/sunray/Makefile
@@ -1,0 +1,32 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"), version 1.0.
+# You may only use this file in accordance with the terms of version
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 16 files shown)


components/openindiana/sunray/Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+BUILD_STYLE = archive
+BUILD_BITS = NO_ARCH
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME= oi-utils
+COMPONENT_VERSION= 0.9
+COMPONENT_SUMMARY=	script to patch SunRay Server for use on Openindiana
+COMPONENT_FMRI=		sunray/$(COMPONENT_NAME)
+COMPONENT_LICENSE=	CDDL
```
AndWac, line 25: COMPONENT_CLASSIFICATION is referenced from the manifest but net set here.

AndWac:
Please add the pkg5 file.
I am a little bit unhappy with the FMRI. sunray/oi-utils doesn't sound right if only some patches for another package is delivered by it. The package name should be more revealing.

PR #23242, opened 2025-08-03
perl/XML-LibXML: rebuild for Perl 5.42 and obsolete package for Perl 5.36

mtelka:


mtelka:
The check failed for some strange reason.  It looks like `pkgdepend` does not work properly on that machine.

@AndWac could you please try to build and publish this on your machine to see if it fails similarly?  If not, please merge.  Thanks.

mtelka:
The problem is that `XML-LibXML` linked statically with `libxml2`.  It did so because there is outdated `library/perl-5/alien-libxml2` installed on the machine.

This PR should be safe to merge and safe to build/publish on a machine with up-to-date packages.

PR #23253, opened 2025-08-05
fish: do not build docs

iigs:
I didn't have sphinx installed when building fish 4.0.2.  Even after installing sphinx, I don't see the error that jenkins see--maybe because I don't have latex installed? Fix by not building docs.


mtelka:
FYI, the build failure was caused by the incompatibility between our (old) `sphinx` and (too new) `docutils`.  I'm working on a fix...

PR #23518, opened 2025-08-21
perl/ExtUtils-Install: obsolete package for Perl 5.36

mtelka:


AndWac:
Can we also obsolete Perl 5.38 and 5.40? I don't see a reason to keep three perl versions. This and frequent python package updates blow the IPS catalog. I had to shrink the repository again because it couldn't be handled anymore with our tools. Even Toasterson's Rust based tools had problems because of the size of the catalog.

mtelka:
> Can we also obsolete Perl 5.38 and 5.40? I don't see a reason to keep three perl versions. This and frequent python package updates blow the IPS catalog. I had to shrink the repository again because it couldn't be handled anymore with our tools. Even Toasterson's Rust based tools had problems because of the size of the catalog.

@AndWac I already asked you several times for a proof that Perl and Python packages are significant helpers to the problem.  I never got any data.  I suspect that the problem is primarily caused by a lot of `userland-incorporation` packages.  We definitely need some better way to thin the ips repository, for example by removing old packages automatically and selectively without manual interaction needed.  But, unfortunately, I cannot help you much in this area since I've no access to the build and publishing infrastructure and I do not know how exactly everything works there.

Re Perl 5.38 and 5.40 obsoletion:
- We cannot do that yet because some other packages depends on them (this one is easily solvable, but see below).
- We should not do that if we want to provide 1st class support for any Perl developers/users that relies on OI Perl.  To support Perl for three years we create enough timeframe for everybody to port their Perl stuff to new Perl version.
- There is really no reason to do that since Perl packages are updated rarely and so they do not create a lot of garbage in the ips repo.  There are only ca 250 - 350 Perl components updated every year (and I'm probably too optimistic anyway).  This creates 350 * 4 = 1 400 new packages every year.  One `illumos-gate` rebuild creates ca 500 new packages, so three `illumos-gate` rebuilds are worse than one year of Perl updates (with three supported versions).  IOW, one Perl version obsoletion would save yearly less packages than one `illumos-gate` build.  In addition, one `history` component rebuild creates more than 1900 new packages.  This is another problem you should focus on.  @AndWac these are my numbers, show me yours please.  No FUD, no feelings, just real hard numbers.
- @AndWac you are the guy who prefers rebuilds when technically not needed.  Focus on this area too, please.

@AndWac, you are grumbling all the time there is not enough people working on OpenIndiana.  I can easily start to do anything else in my free time than `oi-userland`.  Just say a word that I'm not welcome here and you'll never see any new PR from me here.  Yes, it is so easy.

AndWac:
You don't need to feel offended just because I asked whether obsoleting other outdated perl versions.
I haven't heard a perl developer needing several perl versions for OI yet.

mtelka:
> You don't need to feel offended just because I asked whether obsoleting other outdated perl versions.

How should I feel when you accuses me again and again without any data backing your claims and once I show you my data then the discussion always stops?

@AndWac if you are tired merging my PRs (you said something similar at least once) then either grant me permissions so I can merge my PRs myself, or let me know I should stop to contribute, or take your role and merge, or find somebody else who will do the job.  It is really easy.

@AndWac to solve the problem with full ips repository there are at least two options: either do it your way (and schizophrenically swear that there are both not enough contributors and at the same time there is the full ips repo overloaded by contributors), or you can ask me for help.  The only problem here is that without access to servers I can do literally nothing here.

PR #23566, opened 2025-08-24
librsvg - update to version 2.61.3

fritzkink:


fritzkink:
Please note - as of 2.61.2 [#1186 (closed)](https://github.com/GNOME/librsvg/-/issues/1186) - Disable compilation of the gdk-pixbuf loader module by
default.  It used to be that if meson detected that
gdk-pixbuf-query-loaders(1) is available, then it would build the
gdk-pixbuf loader module.  Now that GNOME has switched to glycin
loaders for images, which are sandboxed, gdk-pixbuf loaders are
deprecated.

PR #23641, opened 2025-08-29
rust: fix 1.89.0 build

iigs:
Add bootstrap patch.
Remove rust from components.ignore.
Increase revision.

### DIFF
--- a/components/components.ignore
+++ b/components/components.ignore
@@ -20,7 +20,7 @@
 # /^package$/d
 #
 # Our build server fails to build the following packages (the builds succeed on other build systems):
-/^developer\/rust$/d
+
 # squeak source will be dynamically downloaded which often has timeouts.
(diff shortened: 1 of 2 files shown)


components/developer/rust/Makefile
```diff
@@ -36,6 +36,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		rustc
 COMPONENT_VERSION=	1.89.0
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	Rust - Safe, concurrent, practical language
```
AndWac: There is no need to set COMPONENT_REVISION as the original PR has not yet successully been published.

iigs:
removed COMPONENT_REVISION line from Makefile

PR #23675, opened 2025-09-01
add scryer-prolog package

AndWac:


AndWac:
The resulting app seems to have problems with terminal access.

PR #23688, opened 2025-09-02
SDL3: static lib is really needed

AndWac:


mtelka:
Why do we need it?

AndWac:
-- Check floating point format - unknown
CMake Error at /usr/lib/amd64/cmake/SDL3/SDL3testTargets.cmake:95 (message):
  The imported target "SDL3::SDL3_test" references the file

     "/usr/lib/amd64/libSDL3_test.a"

  but this file does not exist.  Possible reasons include:

mtelka:
I see, thank you.

PR #23709, opened 2025-09-05
flatbuffers - new package version 25.2.10

fritzkink:


components/library/flatbuffers/flatbuffers/flatbuffers.p5m
```diff
@@ -0,0 +1,72 @@
...
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/FlatBuffersSharedTargets-noconfig.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/FlatBuffersSharedTargets.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/FlatBuffersTargets-noconfig.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/FlatBuffersTargets.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/FlatcTargets-noconfig.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/FlatcTargets.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/flatbuffers-config-version.cmake
+file path=usr/lib/$(MACH64)/cmake/flatbuffers/flatbuffers-config.cmake
+file path=usr/lib/$(MACH64)/libflatbuffers.a
+link path=usr/lib/$(MACH64)/libflatbuffers.so \
```
AndWac: Is there a good reason to deliver a static library?
If noit, then please add DROP_STATIC_LIBRARIES= yes to the Makefile and rerun gmake sample-manifest-

components/library/flatbuffers/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+COMPONENT_TEST_TARGETS= test
+COMPONENT_TEST_TRANSFORMS += \
+        "-e '/^Total Test time/d' " \
+        "-e 's/[[:digit:]]\+\.[[:digit:]][[:digit:]].*$$//' "
+
+include $(WS_MAKE_RULES)/common.mk
+
+DROP_STATIC_LIBRARIES= yes
+
```
AndWac: Please move this line up (eg. aftter BUILD_STYLE= cmake) otherwise it doesn't have an effect.
It does exactly what its name expresses - all static libs will be removed from the proto area and thus won't end in the sample-manifest.

PR #23725, opened 2025-09-06
Repology metadata and package.kdl

Toasterson:
This PR brings in the package KDL and repology metadata generation built around that. package.kdl is more streamlined metadata version of the Makefile that makes it easy to draw statistics from the components. One transformed information we build is the repology metadata so repology can in the future retrieve that one JSON file and import it rather than needing to clone this repo. Which is what they wanted. The import into the package.kdl format works both with a top level target and happens automatically during next publish. Unfortunately I needed to change a lot of Makefiles to include the targets. This may cause a lot of rebuilds where we then need to fix the build but hopefully it shouldnt


tools/find-missing-package-kdl.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
```
mtelka, line 1: This file is missing license header.

.gitignore
```diff
@@ -16,3 +16,4 @@ tools/pkglint
 tools/time-*o
 
 .vagrant
+.idea
```
mtelka: Why this change?
BTW, there is missing EOL.

userland-gate.kdl
```diff
@@ -0,0 +1,19 @@
...
+metadata-transform matcher="--libdir" replacement="--libdir=${LIBDIR}"
+metadata-transform matcher="--with-jobs" drop=true
+metadata-transform matcher="CC" drop=true
+metadata-transform matcher="CXX" drop=true
+metadata-transform matcher="F77" drop=true
+metadata-transform matcher="FC" drop=true
+metadata-transform matcher="FFLAGS" drop=true
+metadata-transform matcher="CFLAGS" drop=true
+metadata-transform matcher="LDFLAGS" drop=true
+metadata-transform matcher="PKG_CONFIG_PATH" drop=true
```
mtelka: Missing EOL.

mtelka:
What "KDL" stands for?

components/closed/as/Makefile
```diff
@@ -29,6 +29,7 @@ COMPONENT_ARCHIVE_HASH=	sha256:1758eb4721981d76c86dfe05b20d8dbccadc5938800554d2d
 
 include ../../../make-rules/prep.mk
 include ../../../make-rules/ips.mk
+include $(WS_MAKE_RULES)/forge.mk
```
mtelka, line 32: Changes in all these Makefiles will cause that all affected components will be rebuild and republished.  I suspect this is not a good idea because:

- it is likely that many such component builds will fail,
- there are missing `COMPONENT_REVISION` updates and so the built packages will not be consistent,
- there are also missing component metadata updates (`pkg5` files).

We should find other way to do this (prefereably by converting all components to use `common.mk`).

Makefile
```diff
@@ -37,13 +37,46 @@ clobber:	TARGET = clobber
...
+# Bootstrap pkgdev before running component setup
+setup: bootstrap-pkgdev $(SUBDIRS)
+
+bootstrap-pkgdev:
+	@echo "Bootstrapping pkgdev via cargo from toasterson/forge"
+	cargo install --git https://github.com/toasterson/forge pkgdev
+
+update-pkgdev:
+	@echo "Updating pkgdev via cargo from toasterson/forge"
+	cargo install --git https://github.com/toasterson/forge pkgdev --force
```
mtelka, line 63: Could you please publish forge at crates.io and then we should package it regularly.  This `cargo install` looks like a nasty hack we should avoid.

Thank you.

make-rules/shared-macros.mk
```diff
@@ -26,8 +26,8 @@
 # whenever PATH is to be defined there:
 #     PATH = $(PATH.illumos)
 #     PATH = $(PATH.gnu)
-PATH.illumos =	$(subst $(space),:,$(strip $(PATH.prepend))):$(USRBINDIR$(BITS)):$(USRBINDIR):$(GNUBIN):$(USRSBINDIR$(BITS)):$(USRSBINDIR)
-PATH.gnu =	$(subst $(space),:,$(strip $(PATH.prepend))):$(GNUBIN):$(USRBINDIR$(BITS)):$(USRBINDIR):$(USRSBINDIR$(BITS)):$(USRSBINDIR)
+PATH.illumos =	$(subst $(space),:,$(strip $(PATH.prepend))):$(or $(USRBINDIR$(BITS)),$(USRBINDIR64)):$(USRBINDIR):$(GNUBIN):$(or $(USRSBINDIR$(BITS)),$(USRSBINDIR64)):$(USRSBINDIR):$(HOME)/.cargo/bin
+PATH.gnu =	$(subst $(space),:,$(strip $(PATH.prepend))):$(GNUBIN):$(or $(USRBINDIR$(BITS)),$(USRBINDIR64)):$(USRBINDIR):$(or $(USRSBINDIR$(BITS)),$(USRSBINDIR64)):$(USRSBINDIR):$(HOME)/.cargo/bin
```
mtelka, line 30: Please do not add `$(HOME)` to `$(PATH)`.  This will cause several problems:

- it is not guaranteed that `HOME` is set (think of `env - gmake publish`)
- an user could have basically anything in his `$(HOME)/.cargo/bin` and this could have strange side effects to the build framework and cause hard to debug issues.

Thank you.

make-rules/shared-macros.mk
```diff
@@ -252,7 +252,7 @@ PROTO_DIR =	$(BUILD_DIR)/prototype/$(MACH)
 
 ARCHLIBSUBDIR32	=
 ARCHLIBSUBDIR64	= $(MACH64)
-ARCHLIBSUBDIR	= $(ARCHLIBSUBDIR$(BITS))
+ARCHLIBSUBDIR	= $(if $(filter 32,$(BITS)),$(ARCHLIBSUBDIR32),$(ARCHLIBSUBDIR64))
```
mtelka, line 255: Why this change?

make-rules/shared-macros.mk
```diff
@@ -688,7 +688,7 @@ GCC_GNU_TRIPLET  = $(GNU_TRIPLET)
 GCC_BINDIR =	$(GCC_ROOT)/bin
 GCC_LIBDIR.32 =	$(GCC_ROOT)/lib
 GCC_LIBDIR.64 =	$(GCC_ROOT)/lib/$(MACH64)
-GCC_LIBDIR =	$(GCC_LIBDIR.$(BITS))
+GCC_LIBDIR =	$(or $(GCC_LIBDIR.$(BITS)),$(GCC_LIBDIR.64))
```
mtelka, line 691: Why this change?

make-rules/shared-macros.mk
```diff
@@ -742,10 +742,10 @@ PYTHON.3.9.VENDOR_PACKAGES.64 = /usr/lib/python3.9/vendor-packages
...
 PYTHON.3.9.VENDOR_PACKAGES = $(PYTHON.3.9.VENDOR_PACKAGES.$(BITS))
 
-CC =		$(CC.$(COMPILER).$(BITS))
-CXX =		$(CXX.$(COMPILER).$(BITS))
-F77 =		$(F77.$(COMPILER).$(BITS))
-FC =		$(FC.$(COMPILER).$(BITS))
+CC =		$(or $(CC.$(COMPILER).$(BITS)),$(CC.$(COMPILER).64))
+CXX =		$(or $(CXX.$(COMPILER).$(BITS)),$(CXX.$(COMPILER).64))
+F77 =		$(or $(F77.$(COMPILER).$(BITS)),$(F77.$(COMPILER).64))
+FC =		$(or $(FC.$(COMPILER).$(BITS)),$(FC.$(COMPILER).64))
```
mtelka, line 748: Why these changes?

make-rules/shared-macros.mk
```diff
@@ -864,8 +864,8 @@ QT5_BINDIR.32 = $(QT5_BASEDIR)/bin
 QT5_LIBDIR.32 = $(QT5_BASEDIR)/lib
 QT5_BINDIR.64 = $(QT5_BASEDIR)/bin/$(MACH64)
 QT5_LIBDIR.64 = $(QT5_BASEDIR)/lib/$(MACH64)
-QT5_BINDIR = $(QT5_BINDIR.$(BITS))
-QT5_LIBDIR = $(QT5_LIBDIR.$(BITS))
+QT5_BINDIR = $(or $(QT5_BINDIR.$(BITS)),$(QT5_BINDIR.64))
+QT5_LIBDIR = $(or $(QT5_LIBDIR.$(BITS)),$(QT5_LIBDIR.64))
```
mtelka, line 868: Why do we need all of these?  I'm sure these changes are trying to solve some non-problem.  Or some problem that could and should be solved differently (and easier).

AndWac:
Many change will trigger package builds without changing the COMPONENT_REVISION.
Are you sure that every package that has been touched still builds?

tools/find-missing-package-kdl.sh
```diff
@@ -0,0 +1,45 @@
...
+# Exit code:
+#   - Always 0 (for easy piping/consumption)
+
+set -eu
+
+# Move to repo root if invoked from elsewhere
+ROOT_DIR=$(cd "$(dirname "$0")/.." && pwd)
+cd "$ROOT_DIR"
+
+# Collect component root directories (heuristic):
```
mtelka, line 40: For list of component directories you can use `components.mk`.

mtelka (changes requested):

Toasterson:
> What "KDL" stands for?

https://kdl.dev/ A popular Data language in Oxide Work and in general in the Rust ecosystem.

make-rules/shared-macros.mk
Toasterson: It turns out that Variables in Makefiles are somewhat target related. Thus when I use several variables in my own targets I get a self referential Variable. What it turns out is that BITS is not always set so sometimes when you are running outside the existing machinery BITS will cause issues. So in order to make this variables compatible with targets I added a fallback path that was needed. It's the same change for all variables with BITS.

components/closed/as/Makefile
Toasterson: I have yet to find a way to have them included in the metadata and me not needing to update the makefile. I could add forge.mk through shared-macros.mk but otherwise I cannot include forge.mk in all components that need metadata generation. For repology we need all component metadata relatively fast and not in a it will be eventually there. I am open for suggestions how to include it though

.gitignore
Toasterson: So my IDE does not add it's metadata to the repository.

tools/find-missing-package-kdl.sh
Toasterson: I wanted something that does not rely on the setup target to find components with issues.

Toasterson:
> Many change will trigger package builds without changing the COMPONENT_REVISION. Are you sure that every package that has been touched still builds?

No they most certainly don't build. I hoped it would not trigger the build. I am open to suggestions how to include forge.mk without Updating the makefiles. I am somewhat reluctant to add it to shared-macros due to the purpose of that makefile but If people are ok with adding it there or have a suggestion I am open.

tools/find-missing-package-kdl.sh
mtelka: Why?

make-rules/shared-macros.mk
mtelka: Could you show an example how can I reproduce this issue?

components/closed/as/Makefile
mtelka: I do not understand this.  Why you cannot include forge.mk in all components that needs the metadata?  How will you generate the metadata then?

BTW, could you show an example how the metadata will look like?

.gitignore
mtelka: I think this does not belong to the repo, but no strong objection against this.

mtelka:
> > Many change will trigger package builds without changing the COMPONENT_REVISION. Are you sure that every package that has been touched still builds?
> 
> No they most certainly don't build. I hoped it would not trigger the build. I am open to suggestions how to include forge.mk without Updating the makefiles. I am somewhat reluctant to add it to shared-macros due to the purpose of that makefile but If people are ok with adding it there or have a suggestion I am open.

I do not understand the plan yet and what exactly is the big picture and the goal you want to achieve.  Without this it is hard to propose any valuable solution to your problems.

Toasterson:
After coordinating about changes, we will break this up into multiple PR's

PR #23801, opened 2025-09-13
minio: add minio

jwntree:


components/network/minio/Makefile
```diff
@@ -0,0 +1,63 @@
...
+COMPONENT_ARCHIVE_HASH= sha256:1a63209927b9d1fb239b2a776b7b01d1902eff19379a108f4bc7005edcc0968c
+COMPONENT_FMRI= network/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION= Web Services/Application and Web Servers
+
+include $(WS_MAKE_RULES)/common.mk
+
+CLONEY_MODE=copy
+
+PATH=/usr/gnu/bin:/usr/bin
+
```
AndWac, line 36: The need for preferring GNU binaries is common so we have a variable for this:
PATH= $(PATH.gnu)

jwntree:
I agree, but, due to a path issue, the following occurs
usage: go build [-o output] [build flags] [packages]
btw, I updated the version.

PR #23803, opened 2025-09-13
seaweedfs: add seaweedfs

jwntree:
Another S3 compatible storage.
It seems that Tribblix has been providing packages for years.

### DIFF
--- a/components/network/seaweedfs/Makefile
+++ b/components/network/seaweedfs/Makefile
@@ -1,0 +1,58 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 Jiwon Na
+#
+
+BUILD_STYLE= justmake
+
+include ../../../make-rules/shared-macros.mk
+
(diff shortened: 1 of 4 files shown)


components/network/seaweedfs/Makefile
```diff
@@ -0,0 +1,56 @@
...
+COMPONENT_ARCHIVE_HASH= sha256:a2600d3af418d085ec6cb2c1afeea88e1cfcf3ce1ac0f9b423bcf2970d70a66c
+COMPONENT_FMRI= network/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION= Web Services/Application and Web Servers
+
+include $(WS_MAKE_RULES)/common.mk
+
+CLONEY_MODE=copy
+
+PATH=/usr/gnu/bin:/usr/bin
+
```
AndWac, line 36: Please replace this with
PATH= $(PATH.gnu)

jwntree:
When using $(PATH.gnu), /usr/gcc/14/bin/go comes first in the path, causing the build to fail.

mtelka:
> When using $(PATH.gnu), /usr/gcc/14/bin/go comes first in the path, causing the build to fail.

As a far better workaround you can set `PATH.prepend` _before_ the `shared-macros.mk` include (so on top of the `Makefile`) to `/usr/lib/golang/1.24/bin`, or `/usr/bin`.  Please also add a comment to the `Makefile` explaining why is this needed.  Thank you.

PS: We should do this in a better way in our build framework (and so more transparent for components), but since our support for go in the framework is still very limited let's use the above proposal.

PR #23804, opened 2025-09-13
garage: add garage

jwntree:
Another s3 compatible storage.

garage.toml should be provided as well, but since additional setup steps is required for the configuration, only binary file has been included.

https://garagehq.deuxfleurs.fr/documentation/quick-start/

### DIFF
--- a/components/network/garage/Makefile
+++ b/components/network/garage/Makefile
@@ -1,0 +1,56 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 Jiwon Na
+#
+
+BUILD_STYLE= justmake
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME= garage
+COMPONENT_VERSION= 2.1.0
+COMPONENT_SUMMARY= An open-source distributed object storage service tailored for self-hosting
+COMPONENT_PROJECT_URL= https://garagehq.deuxfleurs.fr
+COMPONENT_SRC= $(COMPONENT_NAME)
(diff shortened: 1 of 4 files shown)


components/network/garage/Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+COMPONENT_LICENSE= AGPL v3
+COMPONENT_LICENSE_FILE= LICENSE
+
+# https://www.illumos.org/issues/15767
+LD_Z_IGNORE=
+
+COMPONENT_BUILD_CMD = cargo build --release
+COMPONENT_BUILD_ENV += TERM=xterm
+COMPONENT_BUILD_ENV += SODIUM_USE_PKG_CONFIG=1
```
AndWac, line 42: If xterm is really needed for the build we need to add it as a build dependency.

And to make sure X11 is really available on a build system additional actions need to be taken
REQUIRED_PACKAGES += x11/server/xvfb
COMPONENT_PRE_BUILD_ACTION += /usr/bin/Xvfb :1 >/dev/null 2>&1 &
COMPONENT_BUILD_ENV += DISPLAY=:1
COMPONENT_POST_BUILD_ACTION += /usr/bin/pkill -x Xvfb ;

components/network/garage/Makefile
```diff
@@ -0,0 +1,57 @@
...
+COMPONENT_BUILD_ENV += CARGO_PROFILE_RELEASE_LTO=false
+COMPONENT_BUILD_ENV += GIT_VERSION=cargo:$(COMPONENT_VERSION)
+
+
+COMPONENT_INSTALL_ACTION = $(INSTALL) -D $(@D)/target/release/garage $(PROTOUSRBINDIR)/garage
+
+# Build dependencies
+REQUIRED_PACKAGES+= developer/lang/rustc
+
```
AndWac, line 51: REQUIRED_PACKAGES += terminal/xterm

jwntree:
xterm is not really required. It was included because, during the build process, an error occurred, if TERM environment variable is empty. That no longer happens.

Also, during, COMPONENT_INSTALL_ACTION, an error occurs in jenkins log, I can't figure out why,

mtelka:
Is this the same as https://crates.io/crates/garage?  If so, then please try to use `BUILD_STYLE = cargo` (and move the component to the `components/rust` directory).

jwntree:
It seems like the build is working now.

PR #23874, opened 2025-09-16
openssl-3: update to 3.5.3

mtelka:


AndWac:
This update seem to have killed sshd and thus the build server is not accessible at the moment.

mtelka:
> This update seem to have killed sshd and thus the build server is not accessible at the moment.

Here is the upstream bugreport: https://github.com/openssl/openssl/issues/28575
It looks the problem is (at least partially) in OpenSSH itself.

To fix the issue it looks like the OpenSSH rebuild is needed. :-(

PR #23939, opened 2025-09-24
Add option to specify separate incorporation repo

Toasterson:


components/Makefile
```diff
@@ -161,7 +161,9 @@ $(COMPONENT_DIRS):	$(WS_LOGS) setup FORCE
 
 incorporation:
 	$(eval COMPONENT_REVISION := $(shell git rev-list HEAD --count))
-	$(WS_TOOLS)/userland-incorporator --repository $(PKG_REPO) \
+	$(eval INCORPORATION_REPO := $(or $(INCORPORATION_REPO),$(PKG_REPO)))
```
mtelka: Why not to do it this way (outside the `incorporation` target)?
```
INCORPORATION_REPO = $(PKG_REPO)
```

AndWac: That should be ?=

mtelka: Not necessarily.

Toasterson: It's only for the incorporation step. We don't need the capability to specify a seperate repo outside of that target and honestly it's just a issue in the official buildsystem since that has that complicated repo setup where i386/repo is cleared every build for performance reasons. @AndWac Both work for our current case but this one also works if we ever define INCORPORATION_REPO outside the target.

mtelka: We do have many cases that are there for a single step (or target) only.  For example, `COMPONENTS_IGNORE` above in this file.

Toasterson: OK, Updated

AndWac (approved):

components/Makefile
```diff
@@ -23,6 +23,9 @@
 
 -include ../make-rules/shared-macros.mk
 
+# If we need to build incorporations on the build system where the final repo is not the same as the our local repo, we need to specify the INCORPORATION_REPO variable.
+$(eval INCORPORATION_REPO := $(or $(INCORPORATION_REPO),$(PKG_REPO)))
```
mtelka: Why not to do it this way instead?
```
INCORPORATION_REPO = $(PKG_REPO)
```
Or, if you really think we need it, then this way:
```
INCORPORATION_REPO ?= $(PKG_REPO)
```
Why do you still prefer the more complex, harder to understand, less versatile and otherwise bringing no additional value way with `eval` and `or`?

components/Makefile
```diff
@@ -23,6 +23,9 @@
 
 -include ../make-rules/shared-macros.mk
 
+# If we need to build incorporations on the build system where the final repo is not the same as the our local repo, we need to specify the INCORPORATION_REPO variable.
```
mtelka: Please make this line shorter (multiline comments are common here).   There is no reason to have a comment with almost 170 characters in one line.  Thank you.

components/Makefile
```diff
@@ -161,7 +164,8 @@ $(COMPONENT_DIRS):	$(WS_LOGS) setup FORCE
 
 incorporation:
 	$(eval COMPONENT_REVISION := $(shell git rev-list HEAD --count))
-	$(WS_TOOLS)/userland-incorporator --repository $(PKG_REPO) \
+
```
mtelka: Empty lines in recipes are (or could be) confusing, so I'd prefer to not see it here.  Especially for such short recipes like `incorporation` is.  Thank you.

mtelka (approved):

PR #24024, opened 2025-10-03
Helix editor package

Toasterson:


components/editor/helix/Makefile
```diff
@@ -0,0 +1,52 @@
...
+# Copyright 2025 Till Wegmueller
+#
+
+BUILD_STYLE=cargo
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME=					helix
+COMPONENT_VERSION=				25.07.1
+IPS_COMPONENT_VERSION=          25.7.1
+COMPONENT_SUMMARY=				A post-modern text editor.
```
AndWac, line 22: Please use
HUMAN_VERSION = 25.07.1
COMPONENT_VERSION= 25.7.1
and drop IPS_COMPONENT_VERSION

(And replace COMPONENT_VERSION with HUMAN_VERSION where appropriate later.)

components/editor/helix/Makefile
```diff
@@ -0,0 +1,52 @@
...
+COMPONENT_BUILD_ENV += HELIX_DEFAULT_RUNTIME=/usr/lib/helix/runtime
+
+COMPONENT_INSTALL_ARGS = --path helix-term
+COMPONENT_INSTALL_ARGS += --root $(PROTOUSRDIR)
+COMPONENT_INSTALL_ARGS += --force
+COMPONENT_INSTALL_ARGS += --no-track
+COMPONENT_INSTALL_ARGS += --offline
+COMPONENT_INSTALL_ARGS += --locked
+
+COMPONENT_POST_INSTALL_ACTION =  ./install.sh $(PROTO_DIR) $(SOURCE_DIR)
```
mtelka, line 47: Please incorporate the install steps from the `install.sh` file into the `Makefile`.  There is no rocket science in the `install.sh` file and so there is no reason to use this kind of dedicated script for the installation.  Thank you.

mtelka (changes requested):

components/editor/helix/Makefile
Toasterson: No. It makes the Makefile very unmaintainable and convoluted. Writing shell inside make is a crutch and breaks things way to easily. Especially having to escape all those escape sequences. And we don't get IDE features like syntax highlighting.

components/editor/helix/install.sh
```diff
@@ -0,0 +1,25 @@
+#/bin/bash
```
mtelka: License header is missing.

components/editor/helix/Makefile
Toasterson: Done

components/editor/helix/install.sh
Toasterson: Done

AndWac (approved):

PR #24027, opened 2025-10-03
firefox: update to 143.0.4

iigs:


iigs:
This has been published and tested locally.

PR #24028, opened 2025-10-03
Forge base commit

Toasterson:
This commit contains the base and safe changes to oi-userland only creating and making forge.mk available. other changes will be put into other PR's


Toasterson:
Will squash this once it's Ok.

make-rules/forge.mk
```diff
@@ -0,0 +1,64 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+#
+# Copyright 2023 Till Wegmueller
+#
+
+# Allow overriding via environment and default to relying on PATH
+PKGDEV ?= pkgdev
```
mtelka, line 16:
```
# which pkgdev
no pkgdev in /usr/sbin /usr/bin
#
```

make-rules/forge.mk
```diff
@@ -0,0 +1,64 @@
...
+
+# Ensure Cargo-installed tools (e.g., pkgdev) are on PATH
+PATH := $(PATH):$(HOME)/.cargo/bin
+export PATH
```
mtelka, line 20: Since you include this file from `common.mk` this will affect virtually all components.  I'm sure we do not want that.  Please do not change `PATH` here and please do not `export` anything here.  Thank you.

make-rules/forge.mk
```diff
@@ -0,0 +1,64 @@
...
+
+.PHONY: generate-package-kdl
+
+# Generate package.kdl using pkgdev with component metadata
+generate-package-kdl:
```
mtelka, line 25: Why do we need an unusable target?  I'm sorry, but this makes no sense.  Please postpone this change once we do have the `pkgdev` tool available.

mtelka (changes requested):

PR #24029, opened 2025-10-03
Draft: Forge global makefile changes

Toasterson:
This adds the global makefile changes to oi-userland and once we merge #24028 this should not display the changes to forge.mk and common.mk anymore. 

These changes should be save but we may need to discuss and find a test strategy.

I will make the changes to the individual components once we coordinated integration of this and #24028


Makefile
```diff
@@ -37,13 +37,37 @@ clobber:	TARGET = clobber
...
 	@+echo "$(TARGET) $@" ; $(GMAKE) -s -C $@ $(TARGET)
 
 FORCE:
+
+# Bootstrap pkgdev before running component setup
+setup: bootstrap-pkgdev $(SUBDIRS)
+
+bootstrap-pkgdev:
+	@echo "Bootstrapping pkgdev via cargo from toasterson/forge"
+	cargo install --git https://github.com/toasterson/forge pkgdev
```
mtelka, line 56: Could you please publish forge at crates.io and then we should package it regularly.  This `cargo install` looks like a nasty hack we should avoid.

Thank you.

mtelka (changes requested):

PR #24032, opened 2025-10-03
add lsd package

Toasterson:


mtelka:
Please add test results.

components/rust/lsd/lsd.p5m
```diff
@@ -0,0 +1,26 @@
...
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 <contributor>
+#
```
AndWac, line 14: Author is missing

components/rust/lsd/test/results-all.master
```diff
@@ -0,0 +1,531 @@
+   Compiling memchr v2.7.2
```
mtelka, line 1: This is not the correct file.

mtelka (changes requested):

components/rust/lsd/test/results-all.master
Toasterson: Then please show me the documentation how to setup tests for rust.

mtelka: This is not about rust.  This is exactly the same problem as [here](https://github.com/OpenIndiana/oi-userland/pull/24033#discussion_r2403911295).

FYI, the documentation for setting up tests for rust is here:

If you use the `cargo` build style then the testing works out of the box and you do not need to do anything special to make it working.  Just run `env - gmake test` and you are done.  If it does not work, then it is a bug.  Either in the crate, or in our build framework.

Toasterson: You forgot the link to the documentation in your response. I would love to know how to setup tests for rust.

mtelka: > You forgot the link to the documentation in your response. I would love to know how to setup tests for rust.

I didn't forget anything.  I wrote the documentation right into the comment.

PR #24033, opened 2025-10-03
add bat package

Toasterson:


components/sysutils/bat/Makefile
```diff
@@ -0,0 +1,39 @@
...
+COMPONENT_ARCHIVE_HASH =	\
+	sha256:2ab792c2ad113a666f08856c88cdec0a62d732559b1f3982eedf0142571e669a
+COMPONENT_FMRI =		utilities/bat
+COMPONENT_CLASSIFICATION =	Applications/System Utilities
+COMPONENT_LICENSE =		Apache-2.0
+COMPONENT_LICENSE_FILE =	LICENSE-APACHE
+
+include $(WS_MAKE_RULES)/common.mk
+
+COMPONENT_BUILD_ENV += RUSTONIG_DYNAMIC_LIBONIG=1
```
mtelka: Could you please add a comment what is this for?  Thanks.

mtelka:
Please add test results.

components/sysutils/bat/Makefile
```diff
@@ -0,0 +1,39 @@
...
+#
+
+BUILD_STYLE = cargo
+
+include ../../../make-rules/shared-macros.mk
+
+COMPONENT_NAME =		bat
+HUMAN_VERSION =			0.25.0
+COMPONENT_SUMMARY =     A cat(1) clone with wings. 
+COMPONENT_ARCHIVE_HASH =	\
```
AndWac, line 23: Does this have a project page?

AndWac, line 23: What does having wings mean here?

mtelka: > Does this have a project page?

The cargo build style points by default to crates.io (which is usually sufficient), so there is no need to add `COMPONENT_PROJECT_URL` to component's `Makefile`.  See `make-rules/cargo-defaults.mk` for details.

mtelka:
@Toasterson please consider to move this to the `components/rust` directory.

components/sysutils/bat/Makefile
Toasterson: It's their official description of the Software https://crates.io/crates/bat The Description is simply a copy of that field from their crates.io page resp. GitHub.

mtelka: > It's their official description of the Software https://crates.io/crates/bat The Description is simply a copy of that field from their crates.io page resp. GitHub.

Yes.  And it looks like a joke.  I like it :-).

components/sysutils/bat/Makefile
Toasterson: Done

Toasterson:
Test results added. About the language component directories. we had this for the interpreted libraries and the current rust packages where rust development specific. the tools I have packaged not use rust to compile but are not language specific so I would say they are in the right place. I am open to other opinions though.

components/sysutils/bat/results-all.master
```diff
@@ -0,0 +1,533 @@
+   Compiling autocfg v1.1.0
```
mtelka: This should go into `test` subdirectory.

Also, it looks like you added wrong file here.  Is this a copy of `build/test/amd64/results-64.snapshot`?

components/sysutils/bat/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+include $(WS_MAKE_RULES)/common.mk
+
+# These three lines are needed to instruct the rust
+# oniguruma wrappers to use the system provided oniguruma
+# library and not the vendored shared library.
+# The Vendored oniguruma library does not compile.
+COMPONENT_BUILD_ENV += RUSTONIG_DYNAMIC_LIBONIG=1
+COMPONENT_INSTALL_ENV += RUSTONIG_DYNAMIC_LIBONIG=1
+COMPONENT_TEST_ENV += RUSTONIG_DYNAMIC_LIBONIG=1
```
mtelka: If you need to add this to all build, install and test steps, then it would be easier to add it to `CARGO_ENV` instead.

mtelka (changes requested):

components/sysutils/bat/Makefile
Toasterson: Seems to work. Can we have a documented template for the two usecases with crates.io and from a releases page as demonstrated with helix?

components/sysutils/bat/results-all.master
Toasterson: That seems to be the file needed so it doesn't show differences.

components/sysutils/bat/Makefile
mtelka: Sure.  Feel free to create it.

components/sysutils/bat/Makefile
```diff
@@ -0,0 +1,46 @@
...
+COMPONENT_LICENSE_FILE =	LICENSE-APACHE
+
+include $(WS_MAKE_RULES)/common.mk
+
+# These three lines are needed to instruct the rust
+# oniguruma wrappers to use the system provided oniguruma
+# library and not the vendored shared library.
+# The Vendored oniguruma library does not compile.
+#COMPONENT_BUILD_ENV += RUSTONIG_DYNAMIC_LIBONIG=1
+#COMPONENT_INSTALL_ENV += RUSTONIG_DYNAMIC_LIBONIG=1
```
mtelka: Please remove these three commented out lines.

mtelka (changes requested):

components/sysutils/bat/bat.p5m
```diff
@@ -0,0 +1,26 @@
...
+# 1.0 of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
+# source.  A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright 2025 <contributor>
+#
```
AndWac, line 14: Author is missing

components/sysutils/bat/Makefile
```diff
@@ -0,0 +1,43 @@
...
+COMPONENT_ARCHIVE_HASH =	\
+	sha256:2ab792c2ad113a666f08856c88cdec0a62d732559b1f3982eedf0142571e669a
+COMPONENT_FMRI =		utilities/bat
+COMPONENT_CLASSIFICATION =	Applications/System Utilities
+COMPONENT_LICENSE =		Apache-2.0
+COMPONENT_LICENSE_FILE =	LICENSE-APACHE
+
+include $(WS_MAKE_RULES)/common.mk
+
+# These three lines are needed to instruct the rust
```
mtelka: _three_?

components/sysutils/bat/Makefile
```diff
@@ -0,0 +1,43 @@
...
+# oniguruma wrappers to use the system provided oniguruma
+# library and not the vendored shared library.
+# The Vendored oniguruma library does not compile.
```
mtelka, line 35: Is it worth reporting upstream?

mtelka (changes requested):

components/sysutils/bat/Makefile
Toasterson: No this intended to be changed by us

components/sysutils/bat/bat.p5m
Toasterson: Done

components/sysutils/bat/Makefile
Toasterson: Fixed

components/sysutils/bat/results-all.master
Toasterson: @mtelka Yes I copied the file you mentioned. Unfortunately the line output does not seem to be stable and there are some compilation results that end in the test output due to the way our framework captures the output. Can you elaborate what you do not like about the results file?

mtelka: @Toasterson, you fixed the problem [here](https://github.com/OpenIndiana/oi-userland/compare/d24414fc7762f34f1ae4f3a4a31b6a1ddaedc954..69e6551449b5fb23ca270b7e55290893320e292d) yesterday, ca 29 minutes after I reported the issue here.  You also marked the issue as resolved.  Could you please elaborate why do you think that I think there is still any problem with the file?

mtelka (approved):

PR #24100, opened 2025-10-08
rsync - fix IPv6

DarkAyron:
The test program, which checks for AF_INET6 fails to build because socket() requires -lsocket. We need to do this test after checking whether we need -lsocket.

I've already sent this patch to the samba project. This commit can be reverted as soon as it is merged upstream.

### DIFF
--- a/components/network/rsync/Makefile
+++ b/components/network/rsync/Makefile
@@ -29,6 +29,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		rsync
 COMPONENT_VERSION=	3.4.1
+COMPONENT_REVISION=	1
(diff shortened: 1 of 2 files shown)


AndWac:
Please add
COMPONENT_REVISION= 1
to the Makefile (after the COMPONENT_VERSION assignment) and add it to the PR.
It is necessary to change the Makefile otherwise jenkins wouldn't rebuild the package.

AndWac:
Fine, and now please squash the commits and get rid of the 2nd message. Increasing the revision is not a meaningful comment.

AndWac:
Welcome to OpenIndiana and thank you for your contribution!

PR #24235, opened 2025-10-19
squeak: add and enable PseudoTTYPlugin

cstes:
plugin for the optional Telnet301  package

code from https://www.squeaksource.com/IanPiumartaGoodies


AndWac:
The PR-checker failed to fetch the sources. So I doubt this will build on the build server. Please provide a more stable download option.

cstes:
It is true that the squeakvm.org subversion server is subject to timeouts.   Sometimes I manage to checkout without connection timeout.

In this particular pull request the subversion revision was kept at 3837 and I did not increase it to 3840, hoping that the sources are already available on the build server (by the DLC_ARCHIVE_MIRROR = https://dlc.openindiana.org/oi-userland/source-archives).

In principe the 3837 sources are available for a build on the build server.

The alternative is to wait and I could submit a new pull request once Squeak is upgraded to 4.21.5 because I requested upstream to include/enable PseudoTTYPlugin.

AndWac:
Your Makefile doesn't provide a download-url but makes use of subversion. There is nothing in our tool chain that creates archives from such fetched files. So you either need to provide an archive manually or you need to enhance the tool chain to create archives.

cstes:
It is true that squeakvm.org often returns "connection timeout" or closes connections.    Sometimes it is possible to succesfully svn export code (prep-svn.mk uses svn export to create a local archive).   I'll come back later on this, I've asked Squeak to enable PseudoTTYPlugin but an update to the (old) squeak repository make take several weeks.   I'll also inform about whether the squeakvm.org server could be improved.

cstes:
There are a few source archives for Squeak at 

https://dlc.openindiana.org/oi-userland/source-archives

-  squeak-platforms-Cross-3834.tar.bz2
-  squeak-platforms-Cross-3835.tar.bz2
-  squeak-platforms-unix-3834.tar.bz2 
-  squeak-platforms-unix-3835.tar.bz2|  
-  squeak-src-3834.tar.bz2 
-  squeak-src-3835.tar.bz2 

Those are created by the openindiana oi-userland prep-svn.mk  from a subversion export and the build server normally would attempt to reuse the .tar.bz2 file before doing a new svn export. 

On my local server there is also a 3837 revision set of three files:

-  squeak-platforms-Cross-3837.tar.bz2
-  squeak-platforms-unix-3837.tar.bz2 
-  squeak-src-3837.tar.bz2 

Because the 4.21.3 release of squeak was already published, the 3837 svn export files should be at 

https://dlc.openindiana.org/oi-userland/source-archives

but it seems they are not.

This could indicate that the 4.21.3 package was published in a different way as the older packages with subversion revision 3835 and 3834.

Anyway the new Pull Request (PR) is still using subversion 3837 so it should in theory not do or not attempt to do a new subversion export.

Also I manage to do succesful new subversion export from time to time, for example for subversion revision 3840 which went fine for me, but this PR is not yet upgrading to 3840.

I understand that the connection timeout problem for the squeakvm.org repository can be a nuissance for publishing packages at OpenIndiana, so I kept the revision at 3837 as an attempt to avoid a new svn export.

AndWac:
We have changed the publishing process a bit in the past as it was necessary when we moved the builds to a new environment. The accesses for squeak by means of subversion were always painful for me as I had to manually retry to successfully fetch and publish the package. But that process is even more complicated as we enhanced it for speed and to reduce the build of multiple unnecessary package versions. Today there is no local repository when jenkins is not running.

cstes:
I'll create a new pull request in the future with the PseudoTTYPlugin for Squeak.   The download source is at the Squeak subversion server which still gets some updates from time to time.   I cannot currently provide an alternative download location for this.   The OpenSmalltalk project is an alternative to Squeak, but it is also in some sense a "fork" of Squeak.

AndWac:
You can create a local archive when you build it. Then send it to me and I will upload it to our dlc.

PR #24475, opened 2025-10-27
qemu: update to 10.1.2

iigs:


iigs:
The files for the libfdt subproject intermittently get added.  Drop those files in qemu.p5m in case they do show up.

PR #24480, opened 2025-10-27
Obsolete python/salt

mtelka:
This project does have several problems that make its maintenance hard, for example:

- salt's testsuite fails early so it is unclear if the package works at all at OpenIndiana
- with almost every new upstream release the project brings several new dependencies that we need to package as well to make the salt installable
- some of these dependencies lacks illumos/Solaris support
- salt relies on some very old python modules that does not work properly with new Pythons

See also: https://openindiana.org/pipermail/openindiana-discuss/2025-October/027782.html


mtelka:
Apparently, nobody cares about Salt anymore, so we should be okay to merge this.

PR #24509, opened 2025-10-29
erlang: don't substitue i386 header files.

klausz65:
Otherwise they don't get delivered on SPARC builds.


components/runtime/erlang/erlang.p5m
```diff
@@ -115,12 +115,12 @@ file path=usr/lib/$(MACH64)/erlang/erts-16.1.1/include/erl_fixed_size_int_types.
 file path=usr/lib/$(MACH64)/erlang/erts-16.1.1/include/erl_int_sizes_config.h
 file path=usr/lib/$(MACH64)/erlang/erts-16.1.1/include/erl_nif.h
 file path=usr/lib/$(MACH64)/erlang/erts-16.1.1/include/erl_nif_api_funcs.h
-file path=usr/lib/$(MACH64)/erlang/erts-16.1.1/include/internal/$(MACH)/atomic.h
```
AndWac, line 118: You should add a comment here why this is needed and hope that the next update will respect that.

klausz65: done

PR #24535, opened 2025-10-31
libreoffice: update to 25.8.3.1

iigs:


iigs:
Removed more lines from test output that contain text unique on every run.

Added memory leak patch for poppler 25.10.0 support that was integrated in the previous libreoffice release on OpenIndiana.

PR #24557, opened 2025-11-03
Don't disable all optional features for emacs-nox

nealsid:
Emacs-nox is configured with --without-all, which disables all optional features, including non-GUI related ones, such as threads, libxml, etc.  Remove this so non-GUI users can still use those features.


nealsid:
This was discussed in #24530

PR #24607, opened 2025-11-12
fish: update to 4.2.0

iigs:
Patches went through a complete overhaul.
* Removed patch where reg got renamed to regr since that change was really only needed for gcc, not rust.
* Test support needed a patch where cc has to be named gcc.
* Added assorted patches from pkgsrc--also used to replace my WIP patches if pkgsrc had a better solution.
* Updated third party crate patches.

gmake test now works.  Updated test results.


iigs:
Updated crate patches now included.

PR #24612, opened 2025-11-12
trousers: update to 0.3.15 depend on openssl-3 now.

klausz65:


AndWac:
Did you verify that system/library works well with the new version?

klausz65:
Will check tomorrow by installing it on a T2000 and afterwards doing a brand new illumos-gate build, I'll keep you updated.

klausz65:
Okay, after that x64 build works well I discovered that I need a header from Solaris 11 just for SPARC, which I don't have and I doubt it would be legal to use: sys/paths.h - trousers-0.3.15 compiles fine with this header on illumos/SPARC, but I will refrain from updating to this version because of these legal reasons, will close this commit and life with the old version for now.

PR #24626, opened 2025-11-14
giflib: update to 5.2.2 - drop 32bit libs/binaries.

klausz65:
I checked all dependent packages - they are all already on 64bit so there should be no harm to them.

### DIFF
--- a/components/library/giflib/patches/01-makefile.patch
+++ b/components/library/giflib/patches/01-makefile.patch
@@ -1,0 +1,107 @@
+--- giflib-5.2.2/Makefile.~1~	2024-02-19 03:01:50.000000000 +0200
++++ giflib-5.2.2/Makefile	2025-11-13 21:38:55.835395949 +0200
+@@ -6,18 +6,17 @@
+ # of code space in the shared library.
+ 
+ #
+-OFLAGS = -O0 -g
+-OFLAGS  = -O2
+-CFLAGS  = -std=gnu99 -fPIC -Wall -Wno-format-truncation $(OFLAGS)
++
++CFLAGS  =  -fPIC -Wall gcc_OPT
+ 
+ SHELL = /bin/sh
+ TAR = tar
+ INSTALL = install
+ 
+-PREFIX = /usr/local
++PREFIX = /usr
+ BINDIR = $(PREFIX)/bin
+ INCDIR = $(PREFIX)/include
+-LIBDIR = $(PREFIX)/lib
++LIBDIR = $(PREFIX)/lib/MACH64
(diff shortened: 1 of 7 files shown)


components/library/giflib/patches/01-makefile.patch
```diff
@@ -0,0 +1,107 @@
...
+ TAR = tar
+ INSTALL = install
+ 
+-PREFIX = /usr/local
++PREFIX = /usr
+ BINDIR = $(PREFIX)/bin
+ INCDIR = $(PREFIX)/include
+-LIBDIR = $(PREFIX)/lib
++LIBDIR = $(PREFIX)/lib/amd64
+ MANDIR = $(PREFIX)/share/man
```
AndWac, line 23: Really amd64 and not $(MACH64)?

AndWac:
Are you sure that nothing uses the old 5.0.6 version that you have removed here?

components/library/giflib/patches/01-makefile.patch
```diff
@@ -0,0 +1,107 @@
...
+ TAR = tar
+ INSTALL = install
+ 
+-PREFIX = /usr/local
++PREFIX = /usr
+ BINDIR = $(PREFIX)/bin
+ INCDIR = $(PREFIX)/include
+-LIBDIR = $(PREFIX)/lib
++LIBDIR = $(PREFIX)/lib/MACH64
+ MANDIR = $(PREFIX)/share/man
```
AndWac, line 23: \$\(MACH64\)

klausz65:
Thanks so much that hardcoded amd64 has been leftover from testing, I checked all apps, which rely on giflib just for the 32 bit usage and there where none. I now also checked /usr/bin /usr/lib and /usr/lib/amd64 on the build zone for libgif.so.6 - the only match I found was gif2raw - which will also be decommed via this commit - well that binary is here since 2016 - hopefully it would not be missed too much - I found it much to complicate to also still support the old library and this binary. if you have a better approach than mine to support also that old stuff - let me know.

components/library/giflib/patches/01-makefile.patch
klausz65: No MACH64 is correct here - because that GSED in the Makefile replaces it with amd64/sparcv9 !

klausz65: And I tested $(MACH64) it's not propagated into the real build - so the need for that sed statement.

AndWac:
We could ask in one of our mailing lists for users. But we shouldn't ask if we don't want to cope with the answers ;-)
If nothing in OI uses the old version I am fine with removing it.

PR #24750, opened 2025-11-19
add R 4.5.2

jwntree:
R is used for statistical and scientific purposes.


components/scientific/R/Makefile
```diff
@@ -0,0 +1,72 @@
...
+COMPONENT_NAME= 	R
+COMPONENT_VERSION= 	4.5.2
+COMPONENT_SUMMARY= 	R is a free software environment for statistical computing and graphics
+COMPONENT_PROJECT_URL=	https://www.r-project.org
+COMPONENT_SRC= 		$(COMPONENT_NAME)-$(COMPONENT_VERSION)
+COMPONENT_ARCHIVE= 	$(COMPONENT_SRC).tar.gz
+COMPONENT_ARCHIVE_URL=	https://cran.r-project.org/src/base/R-4/$(COMPONENT_ARCHIVE)
+COMPONENT_ARCHIVE_HASH= sha256:0d71ff7106ec69cd7c67e1e95ed1a3cee355880931f2eb78c530014a9e379f20
+COMPONENT_FMRI=			runtime/$(COMPONENT_NAME)
+COMPONENT_CLASSIFICATION=	Development/Other Languages
```
AndWac, line 27: Wouldn't scientific be better suited than runtime?

jwntree:
I considered that at first, I changed it to follow the conventions.
Considering R can have subpackages like Python, even though it’s unlikely in OpenIndiana due to packaging overhead.

AndWac:
Where would you search for the package? Its name is too short for a meaningful search. I wouldn't expect to find it in runtime. My first guess would be scientific.

AndWac:
Please also move all files of this PR into the scientific folder.

PR #24751, opened 2025-11-19
icu: use the same x86_64 GCC configuration for SPARC.

klausz65:


components/library/icu/patches/001-genccode.patch
```diff
@@ -1,13 +1,16 @@
...
+
 --- icu/source/configure.orig
 +++ icu/source/configure
 @@ -7013,7 +7013,7 @@
          # We're using gcc, and the simple -a gcc command line works for genccode
          GENCCODE_ASSEMBLY="-a gcc"
      fi ;;
 -i*86-*-solaris*)
-+*86*-solaris*)
++*-solaris*)
```
mtelka: Why we do need this?  The whole block of code is this:
```
*86*-solaris*)
    if test "$GCC" = yes; then
        # When using gcc, look if we're also using GNU as.
        # When using GNU as, the simple -a gcc command line works for genccode.
        asv=`"${CC}" -print-prog-name=as 2>/dev/null`
        asv=`"${asv}" --version 2>/dev/null`
        case "X${asv}" in
        X*GNU*) GENCCODE_ASSEMBLY="-a gcc" ;;
        X*) GENCCODE_ASSEMBLY="-a sun-x86" ;;
        esac
        unset asv
    else
        GENCCODE_ASSEMBLY="-a sun-x86"
    fi ;;
sparc-*-solaris*)
    GENCCODE_ASSEMBLY="-a sun"
    ;;
```
If this is not okay, then the other case (`sparc-*-solaris*`) should be changed, not the `x86` one.  Primarily because you can get `-a sun-x86` which is obviously incorrect for sparc.

mtelka (changes requested):

klausz65:
You're absolutely right - thanks quite much - fixed.

components/library/icu/patches/001-genccode.patch
```diff
@@ -1,5 +1,8 @@
 https://unicode-org.atlassian.net/browse/ICU-23263
 
+OpenIndiana on SPARC follows the same GCC configuration as x86_64.
```
mtelka, line 3: Please add a link to an upstream bugreport.  If you think this is same as ICU-23263 (I do not think it is same bug, but I do not know details about your issue here) then please add a comment there so they are aware about this too.

klausz65:
Hi Marcel, I have registered at this ICU/atlassin side, but for some reason can't modify/create something there. My issue just is that SPARC follows the same GCC/GAS configuration as x86_64 does. In the past SPARC on OpenIndiana used to build GCC compilers using the Sun assembler, this has been changed by me with: https://github.com/OpenIndiana/oi-userland/pull/12724. However I have verified the reproducing steps in ICU-23263 and yes, yor're right it is not the same bug. because genrb works fine on illumos-SPARC, therefore I'll close this pull request - thanks.

PR #24753, opened 2025-11-19
clang-21: add 21.1.5

iigs:


iigs:
CLANG_DEFAULT has not been changed in make-rules/shared-macros.mk

PR #24873, opened 2025-11-30
glib: update to 2.86.2

iigs:
64-bit tests do not run.  Only included 32-bit test results.


iigs:
Added 64-bit tests.  Makefile has new dependencies.

iigs:
This will need to be published before the glibmm 2.86.0 PR

klausz65:
Hi iigs,
please if possible don't use hardcodes names, when ever possible, please use variables - thanks.
For details please see: https://github.com/OpenIndiana/oi-userland/pull/24905

PR #24906, opened 2025-12-04
rustc - update to version 1.91.1

fritzkink:
@Toasterson  it is for reference


iigs:
Will this patch help?
https://github.com/RocketMan/solaris-userland-legacy/blob/master/components/rust/rustc-11.3/patches/temp_v1.90_to_v.1.91.patch

Note: target_os = "solaris" would have to changed to: target_os = "illumos"

fritzkink:
> Will this patch help? https://github.com/RocketMan/solaris-userland-legacy/blob/master/components/rust/rustc-11.3/patches/temp_v1.90_to_v.1.91.patch
> 
> Note: target_os = "solaris" would have to changed to: target_os = "illumos"
Thanks, this solved the issue. No that 1.92.0 is already out it should be possible that the next update build should work smoothly again.

PR #25039, opened 2025-12-16
CGI-Fast: rebuild for perl-542

klausz65:


mtelka:
Why do we need this?  The component is already built for Perl 5.42.

klausz65:
Okay, is this also the case for modules: library/perl-5/crypt-random library/perl-5/email-stuffer library/perl-5/module-install and library/perl-5/sereal-encoder ? If yes then I have probably only missed to build them on SPARC.

mtelka:
Everything is built for Perl 5.42.  You can check yourself.

klausz65:
Okay, thanks.

PR #25117, opened 2025-12-21
openssh: remove unnecessary dependency ssh-askpass for OI server syst…

klausz65:
…ems.
The network/ssh-askpass dependency will be placed into the desktop_common profile in metapackage: install-types in a later commit.

### DIFF
--- a/components/network/openssh/Makefile
+++ b/components/network/openssh/Makefile
@@ -31,6 +31,7 @@ COMPONENT_NAME=		openssh
 #   OpenSSH <x>.<y>.<z>p<n> => IPS <x>.<y>.<z>.<n>
 COMPONENT_VERSION=	10.2.0.1
 HUMAN_VERSION=		10.2p1
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	OpenSSH client and associated utilities
 # OpenSSH made a mistake when releasing openssh 10.0p1 which identifies itself as 10.0p2.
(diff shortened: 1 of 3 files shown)


mtelka:
The `network/ssh-askpass` dependency is optional.  It is installed only in a case there is `x11/server/xserver-common` installed.  If you do not want `network/ssh-askpass` then simply uninstall `x11/server/xserver-common`.  You do not need this change at all.

klausz65:
Sure you need this change if you're doing a brand new install - then it pulls in zenity via ssh-askpass - tested a quite a few times up to now !!!

klausz65:
I know that it's optional - but believe me it will be pulled in on new installs.

mtelka:
I just installed a machine using the `OI-hipster-minimal-20251026.usb` image and this is what I do see right after the first boot:
```
root@openindiana:~# uname -a
SunOS openindiana 5.11 illumos-68259130da i86pc i386 i86pc
root@openindiana:~# pkg list | wc -l
     723
root@openindiana:~# pkg list | grep ssh
library/libssh2                                   1.11.1-2024.0.0.0          i--
network/ssh                                       10.0.0.2-2025.0.0.2        i--
service/network/ssh                               10.0.0.2-2025.0.0.2        i--
root@openindiana:~# pkg list | grep zenity
root@openindiana:~#
```
@klausz65, how can I reproduce your problem?

klausz65:
Okay, It's probably only a problem of my SPARC builds - forget about it  I'll leave it on SPARC as is and cancel this one.

PR #25289, opened 2026-01-06
vte-291: update to 0.82.3

mtelka:


AndWac:
The build dependency on vala is missing, please add it:

bindings/vala/meson.build:19: WARNING: add_languages is missing native:, assuming languages are wanted for both host and build.
Compiler for language vala for the build machine not found.

bindings/vala/meson.build:19:0: ERROR: Could not execute Vala compiler: valac

A full log can be found at /export/home/andreas/oi-userland/components/library/vte-291/build/amd64/meson-logs/meson-log.txt
gmake: *** [/export/home/andreas/oi-userland/make-rules/common-rules.mk:43: /export/home/andreas/oi-userland/components/library/vte-291/build/amd64/.configured] Error 1

mtelka:
> The build dependency on vala is missing, please add it:

Good catch!  Added.  Thanks.

PR #22059, opened 2025-05-16
msgpack: update to 6.1.0

klausz65:

PR #25387, opened 2026-01-16
webkitgtk: re-add SPARC support

klausz65:


klausz65:
Needed to set 20 Jobs otherwise the build system runs out of virtual memory

klausz65:
20 Jobs was also too much - now at 10 jobs

components/library/webkitgtk/Makefile
```diff
@@ -12,16 +12,18 @@
 #
 # Copyright 2017 Alexander Pyhalov
 # Copyright 2020 Michal Nowak
-# Copyright 2021 Klaus Ziegler
 # Copyright 2023 Friedrich Kink
+# Copyright 2026 Klaus Ziegler
 #
 
 USE_PARALLEL_BUILD = yes
+PARALLEL_JOBS = 10
```
mtelka: Please NEVER do this.  Thank you.

mtelka (changes requested):

components/library/webkitgtk/Makefile
mtelka: For more details please read #22059.

klausz65: Please have a look above - if no JOBS are set it always runs out of mem - because it starts too much jobs on the build server - shall I then also disable PARALELL_BUILD ?

mtelka: Where above can I see the failure on the _build_ server?

klausz65: You can't see the error - that's in the log - when it fails - but anyways I have just remove the JOBS line.

mtelka: And you can see it?  If you can see it then you can show it to me too, no?

iigs: > Please have a look above - if no JOBS are set it always runs out of mem - because it starts too much jobs on the build server - shall I then also disable PARALELL_BUILD ?

Please do not remove PARALELL_BUILD since webkitgtk compiles painfully slow without it.  Since I do not like the default setting of PARALLEL_JOBS, I set the environment variable before a build (the default is too low for my use).  This can be set in .bash_login or whatever is relevant for your user shell.

Such as:
PARALLEL_JOBS=4; export PARALLEL_JOBS

klausz65: Thanks for the tip, to make it clear I just added a job count because it ran out of mem on the build server. And yes on my SPARC it also takes quite too long without PARALELL_BUILD I usually set JOBS to 10, but only because I only have 32GB ram available.

PR #25422, opened 2026-01-20
QT: update 6.9.3

cgrzemba:


components/library/qt6/patches/07-python-version.patch
```diff
@@ -25,3 +25,13 @@ diff -Nru qt-everywhere-src-6.4.2.orig/qtbase/mkspecs/features/uikit/devices.py
  # Copyright (C) 2017 The Qt Company Ltd.
  # SPDX-License-Identifier: LicenseRef-Qt-Commercial OR LGPL-3.0-only OR GPL-2.0-only OR GPL-3.0-only
  
+diff --git a/qtbase/libexec/qt-android-runner.py b/qtbase/libexec/qt-android-runner.py
+index bce130c..4a82980 100644
+--- a/qtbase/libexec/qt-android-runner.py
++++ b/qtbase/libexec/qt-android-runner.py
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python3
++#!/usr/bin/python3.9
```
mtelka: Please use `PYTHON_SCRIPTS` instead.

components/library/qt6/Makefile
```diff
@@ -69,6 +76,8 @@ CONFIGURE_OPTIONS += MySQL_LIBRARY=$(MYSQL_LIBDIR)/libmysqlclient.so
 CONFIGURE_OPTIONS += MySQL_INCLUDE_DIR=$(MYSQL_INCDIR)
 CONFIGURE_OPTIONS += PostgreSQL_LIBRARY=$(PG_LIBDIR)/libpq.so
 CONFIGURE_OPTIONS += PostgreSQL_INCLUDE_DIR=$(PG_INCDIR)
+# 3rdparty/open62541
```
mtelka: Could you please explain what is this comment about?

components/library/qt6/Makefile
```diff
@@ -38,15 +39,21 @@ QT_PREFIX= $(QT6_BASEDIR)
 
 QT_ARCHSUBDIR= $(MACH64)
 QT_PLATFORM= solaris-g++-64
+QTV= $(QT6_VERSION)
+PKG_VARS += QTV
 
+CFLAGS += -Wno-unused-variable
 LDFLAGS += -Wl,-rpath,$(QT6_LIBDIR):$(USRLIBDIR)/xorg/$(MACH64)/:$(CLANG_LIBDIR):$(MYSQL_LIBDIR):$(PG_LIBDIR):$(JPEG_LIBDIR)
 
+OPENSSL_VERSION=3
```
mtelka: This is the default, so we should not need this line.

components/library/qt6/patches/16-open62541.patch
```diff
@@ -0,0 +1,52 @@
+diff --git a/qtopcua/src/3rdparty/open62541/open62541.c b/qtopcua/src/3rdparty/open62541/open62541.c
+index 8cafa7b..bb0aca7 100644
+--- a/qtopcua/src/3rdparty/open62541/open62541.c
++++ b/qtopcua/src/3rdparty/open62541/open62541.c
+@@ -123774,6 +123774,10 @@ UA_EventLoopPOSIX_setNoSigPipe(UA_FD sockfd) {
+     return UA_STATUSCODE_GOOD;
+ }
+ 
++#ifndef SO_REUSEPORT
++#define SO_REUSEPORT	15
```
mtelka: Illumos does not have `SO_REUSEPORT` support (see also https://www.illumos.org/issues/12455).  What is this change supposed to do?

mtelka (changes requested):

components/library/qt6/patches/07-python-version.patch
cgrzemba: done

components/library/qt6/Makefile
```diff
@@ -14,15 +14,16 @@
 #
 include ../../../make-rules/shared-macros.mk
 
-QT6_VERSION=				6.6
+QT6_VERSION=				6.9
```
mtelka: Please update `QT6_VERSION` in `make-rules/shared-macros.mk` too.

components/library/qt6/Makefile
```diff
@@ -14,15 +14,16 @@
...
 COMPONENT_NAME=				qt
 COMPONENT_VERSION_MJR=		$(QT6_VERSION)
 COMPONENT_VERSION=			$(COMPONENT_VERSION_MJR).3
 COMPONENT_SUMMARY=			Qt cross-platform application and UI framework
 COMPONENT_PROJECT_URL=		https://www.qt.io/
 COMPONENT_SRC=				$(COMPONENT_NAME)-everywhere-src-$(COMPONENT_VERSION)
 COMPONENT_ARCHIVE=			$(COMPONENT_SRC).tar.xz
-COMPONENT_ARCHIVE_HASH=		sha256:69d0348fef415da98aa890a34651e9cfb232f1bffcee289b7b4e21386bf36104
+COMPONENT_ARCHIVE_HASH=		md5:d8fab224d072df8ffa09f99f6245e446
```
mtelka: Why md5 instead of sha256?

components/library/qt6/Makefile
```diff
@@ -101,6 +109,11 @@ $(BUILD_DIR)/%/.installed:	$(BUILD_DIR)/%/.built
 	$(COMPONENT_POST_INSTALL_ACTION)
 	$(TOUCH) $@
 
+PYTHON_SCRIPTS += /usr/lib/qt/6.9/mkspecs/features/uikit/devices.py
+PYTHON_SCRIPTS += /usr/lib/qt/6.9/bin/amd64/sanitizer-testrunner.py
+PYTHON_SCRIPTS += /usr/lib/qt/6.9/bin/amd64/qt-testrunner.py
+PYTHON_SCRIPTS += /usr/lib/qt/6.9/bin/amd64/qt-android-runner.py
```
mtelka: Please use `$(QT6_BINDIR)` instead of `/usr/lib/qt/6.9/bin/amd64`.

components/library/qt6/Makefile
```diff
@@ -101,6 +109,11 @@ $(BUILD_DIR)/%/.installed:	$(BUILD_DIR)/%/.built
 	$(COMPONENT_POST_INSTALL_ACTION)
 	$(TOUCH) $@
 
+PYTHON_SCRIPTS += /usr/lib/qt/6.9/mkspecs/features/uikit/devices.py
```
mtelka: Please use `$(QT6_BASEDIR)` instead of `/usr/lib/qt/6.9`.

components/library/qt6/manifests/sample-manifest.p5m
mtelka, line 1: Maybe it would be great to replace qt version in paths by something like `$(QT6_VERSION)`.

mtelka (changes requested):

components/library/qt6/Makefile
mtelka: ... and maybe this line could be removed then :-).

cgrzemba: I guess the QT5_ 32bit variables in shared-macros.mk can be removed, we do not have QT5 32bit anymore.

mtelka: Apparently, yes.  Feel free to do so in a separate PR.

components/library/qt6/Makefile
```diff
@@ -14,15 +14,15 @@
 #
 include ../../../make-rules/shared-macros.mk
 
-QT6_VERSION=				6.6
 COMPONENT_NAME=				qt
-COMPONENT_VERSION_MJR=		$(QT6_VERSION)
+COMPONENT_VERSION_MJR=		6.9
```
mtelka, line 18: Why not to keep it `$(QT6_VERSION)`?

components/library/qt6/Makefile
```diff
@@ -38,15 +38,20 @@ QT_PREFIX= $(QT6_BASEDIR)
 
 QT_ARCHSUBDIR= $(MACH64)
 QT_PLATFORM= solaris-g++-64
+QTV= $(COMPONENT_VERSION_MJR)
```
mtelka, line 41: Why not `$(QT6_VERSION)`?

cgrzemba: This recipe do work only for this version, so it makes no sense to define the version outside, I think.

components/library/qt6/Makefile
cgrzemba: MD5 is what upstream provides on download site, but I will change to sha256.

components/library/qt6/Makefile
mtelka: Since the qt version here and `QT6_VERSION` in `shared-macros.mk` should always match (at least for now), then it makes no sense to define it twice.

PR #25446, opened 2026-01-22
bash: update to 5.3p9

iigs:


AndWac:
The downloaded patches bash53-00x need to have a .patch suffix otherwise they won't be applied.

iigs:
Thank you for catching the bash patches were not named correctly.  I did not realize the fetch script only downloaded the patches and not also renamed them for userland's use.

PR #25526, opened 2026-02-01
virtualbox: update 7.2.6

cgrzemba:
fix  CVE's on 7.2.4
draft because I cant test in the moment


iigs:
I tested this yesterday (booted OI Live on a VM).  7.2.6 works fine. The only recommendation would be to remove this first line in Makefile since it is duplicated in the Auto-generated dependencies:
REQUIRED_PACKAGES += library/libvncserver

PR #25531, opened 2026-02-01
heimdal: add 7.8.0

iigs:
This is being added to provide a more modern Kerberos V library to OpenIndiana.  There is no expectation for this to replace the provided Illumos kerberos implementation.  No SMF manifests were developed/provided to start any Kerberos service.

Man pages are not included since gmake publish gives this error message if the man pages are included in the package:
Error: packaged texinfo directory file (%<action.path>)

There is not enough information to troubleshoot this man page problem.

To void conflict with Illumos kerberos, heimdal is being installed in /usr/heimdal and the krb5.conf goes in /etc/heimdal .
Be prepared to add -R /usr/heimdal/lib/amd64 to LDFLAGS if you want to use the library in software

One patch came from openbsd ports and one patch came from debian.  The default krb5.conf comes from openbsd ports.


Toasterson (approved):
I like the general idea. I am not caught up if we still drop the .a files but I saw them in the manifest so i though I mention it but no blocker.

I like that heimdal is in a dedicated location, with this we can at some other time set mediators and provide SMF manifests. Since heimdal is not that Self-Assembly friendly we would need to think about a good admin utility though that assembles heimdal so people can more easily use it.

Should we try to enable AFS support eventually? That would be a good thing to have ontop of our ZFS. Doesn't have to be this PR though.

No blockers from my side. LGTM

Toasterson:
Ah the PR checker seems to have problems with COM interfaces though. My guess would be a missing dependency in REQUIRED_PACKAGES

iigs:
This needs to be published before building freerdp 3.22.0.

heimdal.p5m file drops the .a files so they don't get added to the package.  There are no cmake options in the source to prevent the building of the static libraries.

I did not have the sysutils/e2fsprogs package installed.  Once I added the package, then the system libcom_err.so gets used.  Updated PR.

iigs:
If openafs gets added to OpenIndiana (there is an unmerged closed PR if anybody wants to update it and get it into shape), I would hope it can work with the native Illumos Kerberos.

components/library/heimdal/heimdal.p5m
```diff
@@ -0,0 +1,1624 @@
...
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# work around gmake publish problem: "Error: packaged texinfo directory file (%<action.path>)"
+<transform file path=.+/share/.+$ -> drop>
+<transform hardlink path=.+/share/.+$ -> drop>
+
+# remove static libraries
+<transform file path=.+/lib/.+a$ -> drop>
```
mtelka: Please use `DROP_STATIC_LIBRARIES = yes` in the `Makefile` instead of this.

components/library/heimdal/heimdal.p5m
```diff
@@ -0,0 +1,1624 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+# work around gmake publish problem: "Error: packaged texinfo directory file (%<action.path>)"
+<transform file path=.+/share/.+$ -> drop>
```
mtelka: ~The problem is only with the `/usr/share/info/dir` file, not with all `/usr/share/info/*` files.~

components/library/heimdal/heimdal.p5m
```diff
@@ -0,0 +1,1624 @@
...
+file path=usr/heimdal/libexec/kimpersonate
+file path=usr/heimdal/libexec/kpasswdd
+file path=usr/heimdal/sbin/iprop-log
+file path=usr/heimdal/sbin/kstash
+file path=usr/share/info/dir
+file path=usr/share/info/heimdal.info
+file path=usr/share/info/hx509.info
+file path=usr/share/man/man1/bsearch.1
+file path=usr/share/man/man1/kadmin.1
+file path=usr/share/man/man1/kdestroy.1
```
mtelka: ~This file conflicts with the `service/security/kerberos-5` package making this one hardly installable.~

~Maybe there are more similar conflicts.~

Edit: I realised you drop _all_ `/share/` files, including this one, so there is no conflict.  However, it would be better if these files are not in the manifest to avoid confusion.  Simple postinstall `rm ...` should do the trick.

components/library/heimdal/Makefile
```diff
@@ -0,0 +1,60 @@
...
+# A full copy of the text of the CDDL should have accompanied this
+# source. A copy of the CDDL is also available via the Internet at
+# http://www.illumos.org/license/CDDL.
+#
+
+#
+# Copyright (c) 2026 Geoff Weiss
+#
+
+BUILD_STYLE= configure
```
mtelka: This is default, please remove this line.

mtelka (changes requested):

iigs:
* Fixed license
* Added DROP_STATIC_LIBRARIES = yes to Makefile
* Removed BUILD_STYLE from Makefile
* Added Post Install Action to Makefile to remove and rename files, following debian in most cases
* Removed transforms from p5mfile
* man pages now included

3eka:
Hi all,

I have updated BE and booted into it this morning (11:00 CET)...

shouldn't I be able to have both **kerberos-5** and **heimdal** packages installed at the same time (and use mediator if I need)???
```
:; pfexec pkg install -vn heimdal
Creating Plan (Checking for conflicting actions): -           Packages to install:  1
       Create boot environment: No
          Rebuild boot archive: No

Changed packages:
openindiana.org
  system/library/heimdal
    None -> 7.8.0-2026.0.0.0

pkg install: The following packages all deliver file actions to usr/share/man/man8/kadmind.8:

  pkg://openindiana.org/system/security/kerberos-5@0.5.11,5.11-2026.0.0.22898:20260202T224003Z
  pkg://openindiana.org/system/library/heimdal@7.8.0,5.11-2026.0.0.0:20260203T092444Z

```

And, last but no least: **heimdal** --> **heimdal-7**

Thanks and very best regrads

iigs:
Thank you for this report.  It tuned out I had only had a partial selection of the Kerberos Illumos packages installed during testing.  Since I was missing system/security/kerberos-5 , I did not realize I had missed some more file renames needed for heimdal.  This will be fixed in a future PR for heimdal.  This will avoid the need for mediation.

If there is an expectation for heimadal 7.x and a future 8.x to both exist at the same time, then it makes sense to name the package heimdal-7.  When heimdal 8.x gets released, it will just replace 7.x OI heimdal package so there is no need to add a version number (this is the typical OI package model)  Are you aware if heimdal 8.x is supposed to not provide backwards compatibility and there will be a 7.x and 8.x branch be both maintained for updates and security fixes?  I looked at what other distros did for their package names for heimdal and none have a version number in the package name.





> Hi all,
> 
> I have updated BE and booted into it this morning (11:00 CET)...
> 
> shouldn't I be able to have both **kerberos-5** and **heimdal** packages installed at the same time (and use mediator if I need)???
> 
> ```
> :; pfexec pkg install -vn heimdal
> Creating Plan (Checking for conflicting actions): -           Packages to install:  1
>        Create boot environment: No
> Create backup boot environment: No
>           Rebuild boot archive: No
> 
> Changed packages:
> openindiana.org
>   system/library/heimdal
>     None -> 7.8.0-2026.0.0.0
> 
> pkg install: The following packages all deliver file actions to usr/share/man/man8/kadmind.8:
> 
>   pkg://openindiana.org/system/security/kerberos-5@0.5.11,5.11-2026.0.0.22898:20260202T224003Z
>   pkg://openindiana.org/system/library/heimdal@7.8.0,5.11-2026.0.0.0:20260203T092444Z
> ```
> 
> And, last but no least: **heimdal** --> **heimdal-7**
> 
> Thanks and very best regrads

3eka:
Well,

I am fine with any name, but why not?
Thanks

P.S: I have never used **heimdal**, so no idea if versioning is important - just know that OI delivers some packages in 2+ versions

iigs:
Multiple versions of packages are used for:
* provide an older library for packages that can't or won't be updated (typically ABI incompatibility), i.e icu, ffmpeg, etc.
* developer maintains multiple branches which OI wants to provide, i.e php, perl, python, etc.

At this time, there does not appear for any reason to expect that multiple versions of heimdal will be needed in the future.

PR #25533, opened 2026-02-01
freerdp: update to 3.22.0

iigs:
Needs heimdal 7.8.0 to be published/installed before this can be built.  This relies on heimdal kerberos since Illumos kerberos is missing functionality.


iigs:
Disabled SDL3 in Makefile.  SDL3 support was tested--the Remote Desktop window does not open with SDL3--it does open with SDL2.

Tested RDP connection to a Windows 11 system (local workgroup, not a member of a domain).  Login works.  Windows 11 desktop access works.

PR #25549, opened 2026-02-03
heimdal: move man pages to /usr/heimdal

iigs:


AndWac:
What about moving all man pages also into usr/heimdal. This way a user can decide to see heimdal's man pages by explicitly setting MANPATH accordingly.

iigs:
Upon request, move man pages to /usr/heimdal.  PR name does not update.

components/library/heimdal/Makefile
```diff
@@ -35,8 +36,10 @@ CFLAGS += -Wno-error=implicit-function-declaration
 
 CONFIGURE_OPTIONS += --bindir=/usr/heimdal/bin
 CONFIGURE_OPTIONS += --includedir=/usr/heimdal/include
+CONFIGURE_OPTIONS += --infodir=/usr/heimdal/share/info
 CONFIGURE_OPTIONS += --libdir=/usr/heimdal/lib/$(MACH64)
 CONFIGURE_OPTIONS += --libexecdir=/usr/heimdal/libexec
+CONFIGURE_OPTIONS += --mandir=/usr/heimdal/share/man
 CONFIGURE_OPTIONS += --sbindir=/usr/heimdal/sbin
```
mtelka, line 43: Isn't it possible to replace all of these options by single `CONFIGURE_PREFIX = $(USRDIR)/heimdal` or so?

components/library/heimdal/Makefile
mtelka, line 29: Where the `heimdal.license` file comes from?  It significantly differs from the `heimdal-7.8.0/LICENSE` file.

mtelka (changes requested):

mtelka:
> Upon request, move man pages to /usr/heimdal. PR name does not update.

Could you please rename the PR (by clicking on the `Edit` button next to the PR name) to avoid confusion in future.  Thank you.

components/library/heimdal/Makefile
mtelka: ... and since this is the default you can safely remove this line.  Please do so.  Thank you.

mtelka: Since you use SPDX identifiers then please also use [SPDX license expressions](https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/) to combine them.  It looks like `AND` should be used here.

iigs:
That was not obvious--Edit button was visible under the Checks tab, but not visible under the default Conversation tab.

I loosely followed https://github.com/openbsd/ports/blob/master/security/heimdal/Makefile when setting prefixes.  I have zero faith that PREFIX is set to /usr in some random file in userland so I hard coded /usr in the paths (sorry, I'm not grepping the entire userland tree to see where "usr" is used--I would not know how else to find variables that have the contents of usr).

Does anything use COMPONENT_LICENSE= to where needs to be in SPDX forma?t  Isn't the field used only by humans to read?

mtelka:
> Does anything use COMPONENT_LICENSE= to where needs to be in SPDX forma?t Isn't the field used only by humans to read?

Currently no, but that could change in future so why not to do it properly right now?

iigs:
Added contents of LICENSES to heimdal.license file.  Since this text has no title if it based on an existing license, I have called it Heimdal.  heimdal.license is not in SPDX format--it has been edited to remove extraneous lines.

mtelka:
> Added contents of LICENSES to heimdal.license file. Since this text has no title if it based on an existing license, I have called it Heimdal. heimdal.license is not in SPDX format--it has been edited to remove extraneous lines.

The `heimdal-7.8.0/LICENSE` file is `BSD-3-Clause`.

mtelka:
> I loosely followed https://github.com/openbsd/ports/blob/master/security/heimdal/Makefile when setting prefixes. I have zero faith that PREFIX is set to /usr in some random file in userland so I hard coded /usr in the paths (sorry, I'm not grepping the entire userland tree to see where "usr" is used--I would not know how else to find variables that have the contents of usr).

My question was not about hardcoded `/usr`, but about seven lines in the `Makefile` that could be possibly replaced by single `CONFIGURE_PREFIX =  ...` line.

iigs:
Removed Heimdal from COMPONENT_LICENSE.  

I think it is more intuitive to list out all of the install directories that need to be customized.  I do not want spend a couple of hours testing out your suggestion to find out it may not work.

PR #25561, opened 2026-02-05
heimdal: rebuild

iigs:
I never know if a package needs to be rebuilt or not if there is a jenkins build failure.  I am providing a changed Makefile to heimdal if it needs to be rebuilt.  Please close PR if it is not needed.


mtelka:
This is not needed.

PR #25592, opened 2026-02-10
gcc-14: Add patches to allow nontrivial use of the C++ <locale> API

Bill-Sommerfeld:
The generic locale implementation chosen naturally by gcc's configure throws an exception if a locale other than "C" is selected.  We can do better.

Use the DragonflyBSD locale support with a few minor changes.

From my illumos-gcc pull request, https://github.com/illumos/gcc/pull/56


AndWac:
Please set COMPONENT_REVISION=1 in the Makefile

PR #25666, opened 2026-02-17
gdk-pixbuf: update 2.44.5

cgrzemba:
minor update


components/library/gdk-pixbuf/Makefile
```diff
@@ -44,9 +44,12 @@ CONFIGURE_OPTIONS += -Dothers=enabled
 CONFIGURE_OPTIONS += -Dbuiltin_loaders=none
 # Do not install tests
 CONFIGURE_OPTIONS += -Dinstalled_tests=false
+CONFIGURE_OPTIONS += -Ddocumentation=true
+CONFIGURE_OPTIONS += -Dgtk_doc=true
+CONFIGURE_OPTIONS += -Dman=true
```
mtelka, line 49: Why do we need these?

cgrzemba: Its now necessary to ship the same files like version before

PR #25971, opened 2026-03-21
dtlsqueak: squeak 4.6 from dtlewis290 github

cstes:
Same package as squeak:  provides a VM for running Squeak 4.6 and older images (perhaps Squeak 1.x or Squeak 2.x)

tested "SUnit Tests" and same results as with the "squeak" package

the IPS mediator can be set :

pkg set-mediator -V 4 -I dtl squeak

### DIFF
--- a/components/runtime/smalltalk/dtlsqueak/Makefile
+++ b/components/runtime/smalltalk/dtlsqueak/Makefile
@@ -1,0 +1,235 @@
+#
+# This file and its contents are supplied under the terms of the
+# Common Development and Distribution License ("CDDL"). You may
+# only use this file in accordance with the terms of the CDDL.
+#
+# A full copy of the text of the CDDL should have accompanied this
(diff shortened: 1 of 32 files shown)


cstes:
advantage can be that the github checks pass (or seem to pass) for this package

this avoids the timeout issue with the svn-prep.mk script

however essentially there is no reason to change the subversion package

cstes:
fixed the mediator link for ckformat
(the link was set in the opensmalltalk cog-spur and stack-spur packages but not in the classical squeak package)

ckformat is a utility from dtlewis290 and is in the source code distribution of both classical squeak and opensmalltalk 

in fact the utility is provided here as a C program but is maintained by dtlewis290 in the ImageFormat Smalltalk package (the C source code is generated from Smalltalk)

more info at http://wiki.squeak.org/squeak/676 and http://wiki.squeak.org/squeak/6290

PR #25988, opened 2026-03-24
install-types: add new storage drivers from illumos-2023 to OI distri…

klausz65:
…bution


mtelka:
@klausz65 this change is okay, but I'm curious: What exactly is illumos-2023?

klausz65:
Just that they where already introduced in 2023 and are stale since then for OI

PR #26044, opened 2026-03-31
webkitgtk: update to 2.52.1

iigs:


iigs:
Added missing patches.  Refactored some patches. yelp still shows a blank page.

PR #26086, opened 2026-04-04
squeak: fix ckformat mediator link

cstes:
fix the mediator link for ckformat
(the link was set in the opensmalltalk cog-spur and stack-spur packages but not in the classical squeak package)

ckformat is a utility from dtlewis290 and is in the source code distribution of both classical squeak and opensmalltalk 

in fact the utility is provided here as a C program but is maintained by dtlewis290 in the ImageFormat Smalltalk package (the C source code is generated from Smalltalk)

more info at http://wiki.squeak.org/squeak/676 and http://wiki.squeak.org/squeak/6290

### DIFF
--- a/components/runtime/smalltalk/squeak/Makefile
+++ b/components/runtime/smalltalk/squeak/Makefile
@@ -71,7 +71,7 @@ include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		squeak
 COMPONENT_VERSION=	4.21.3
-COMPONENT_REVISION=	1
+COMPONENT_REVISION=	2
 COMPONENT_SUMMARY=	The Squeak Virtual Machine
 COMPONENT_PROJECT_URL=	http://www.squeak.org
 COMPONENT_ARCHIVE_URL=  http://squeakvm.org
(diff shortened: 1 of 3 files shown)


cstes:
There seems to be a check that fails as the pull request for squeak is flagged as "one check failed", related to download or svn export of the subversion source code ...

However for the companion dtlsqueak package which provides the same binaries but built from github , the pull request is flagged as OK

it would be nice if dtlsqueak could be added to the oi-userland so that the subversion and github sources IPS package can both be provided.

AndWac:
I am still not convinced that we need two variants of the old squeak when they only differ by the source of the download. The subversion download is unreliable for us so please simply replace it with the github based one.

cstes:
With the oi-userland component Makefile of squeak, I can relatively easily build a VM by subversion revision number.  For me locally the svn export (subversion export) works and saves a .tar.gz file in my local archives directory.  So for example for this update (which did not change the svn subversion revision) no actual download or export was needed, and this builds easily using the archives in my local archives directory.   So I'd prefer to keep this squeak package using the subversion support, because it allows me to build locally (for my local package repo) older VMs given a certain subversion number (where the history of the subversion repository is going back many years).    The dtlsqueak package which downloads from github, was meant as a service to oi-userland builds and the jenkins framework because dtlsqueak builds more easily for oi-userland with jenkins.    Also the contents of those packages is relatively small, I think it is about 2 megabytes of binaries.  OI-Userland and IPS should have no problems supporting many packages, those squeak packages are just 5-10 packages.

PR #26120, opened 2026-04-07
thunderbird: update to 149.0.2

iigs:


iigs:
Files in thunderbird release directory are now accessible.

PR #26185, opened 2026-04-13
liburcu - new package (prerequisite for latest bind version)

fritzkink:


components/library/liburcu/Makefile
```diff
@@ -0,0 +1,62 @@
...
+COMPONENT_ARCHIVE_URL=	https://lttng.org/files/urcu/$(COMPONENT_ARCHIVE)
+COMPONENT_FMRI=         library/liburcu
+COMPONENT_CLASSIFICATION= System/Libraries
+COMPONENT_LICENSE=      LGPLv2.1
+COMPONENT_LICENSE_FILE= LGPL-2.1-or-later.txt
+
+include $(WS_MAKE_RULES)/common.mk
+
+COMPONENT_POST_CONFIGURE_ACTION = ( \
+	$(CP) $(SOURCE_DIR)/LICENSES/LGPL-2.1-or-later.txt $(SOURCE_DIR)/LGPL-2.1-or-later.txt; \
```
AndWac: You don't need to copy the license file to the top level source folder. Instead you can use the path below $SOURCE_DIR directly eg. with
LICENCE_FILE= LICENSES/LGPL-2.1-or-later.txt

PR #26236, opened 2026-04-18
rustc - update to version 1.95.0

fritzkink:


klausz65:
Hi Fritz, could you leave this one as is for now, because I do want to intergrate it for SPARC, I have rustc for SPARC cross compiled from 11.4 to OI since a few month now, and was just waiting that stable gets up to 1.95.0. I would like to have a native one for OI with this commit.

PR #26262, opened 2026-04-21
thunderbird: update to 150.0

iigs:


iigs:
If built without using mach, then it won't use ~/.mozbuild for storing files (non-mach builds do not work).  ~/.mozbuild still being used on mach builds.

Add Bug2031958.patch to fix compile of in-tree media/ffvpx.
Add 12-mach-build-not-use-python-modules.patch since freebsd added it (no reason given).  gmake publish did not break.

PR #26758, opened 2026-06-11
openssl-3: update to 3.5.7

iigs:


mtelka:
@iigs I do see this testing failure (32-bit only): https://github.com/openssl/openssl/issues/31463
Do you see the same, or all tests pass for you?  Thank you.

iigs:
All of the 70-test_quic tests pass for me in the build/test/i86/test-32-results file.  The results-all.master is a copy of the 32-bit test snapshot.  The 64-bit test snapshot matches the same file.

mtelka:
@iigs could you please send me the list of installed packages (`pkg list`) and also the output of `pkg mediator` and `pkg facet` commands from your machine?

iigs:
Between  the time I built openssl 3.5.7 and the time you asked for the pkg list and other stuff, I updated my BE.  Now I can confirm that there is test failure in 32-bit for 70-test_quic_tserver.t .

PR #26873, opened 2026-06-24
fish: update to 4.8.0

iigs:
themes and completions have not been used since 4.2. https://fishshell.com/docs/current/relnotes.html#fish-4-8-0-released-june-24-2026 They have been removed.


iigs:
Removed REVISION

PR #26889, opened 2026-06-27
Add gcc-15.3.0

Bill-Sommerfeld:
Add gcc-15.3.0

Mostly from omnios pull requst to illumos/gcc


AndWac:
Any progress here?

Bill-Sommerfeld:
It's been a long, slow, slog.  regression tests take about two days to run (maybe I'm missing something to parallelize them..)

I think this is worth publishing as a first cut to make it more available; there's an additional revision to the `-msave-args` patch in solaris that I'd like to take a closer look at but I don't think it's is worth holding the initial integration for that patch.

PR #26905, opened 2026-06-29
gcc-13: enhance SPARC support, fix install-strip for libs, add CVE fi…

klausz65:
…xes for gccgo.


components/developer/gcc-13/Makefile
```diff
@@ -78,7 +80,6 @@ CONFIGURE_OPTIONS += --enable-plugins
 CONFIGURE_OPTIONS += --enable-__cxa_atexit
 CONFIGURE_OPTIONS += --enable-initfini-array
 CONFIGURE_OPTIONS += --with-diagnostics-urls=auto-if-env
-CONFIGURE_OPTIONS += enable_frame_pointer=yes
```
AndWac, line 83: Why is this change necessary? All other gcc versions have frame pointers enabled!

klausz65: If you have a look in gcc/configure script you  will see the following:
case $target_os in
linux* | gnu* | darwin[8912]* | cygwin* | mingw*)
  # Enable -fomit-frame-pointer by default for these systems with DWARF2.
  enable_frame_pointer=no
  ;;
*)
  enable_frame_pointer=yes
  ;;
esac

I understand it - that it's always enabled but on named systems omitted, so the setting in Makefile is doubled.

klausz65: So the removal is just a cleanup in our Makefile.

klausz65: klausz@oibuild % cd oi-userland/components/developer/gcc-13
klausz@oibuild % grep enable_frame_pointer Makefile 
klausz@oibuild % grep enable_frame_pointer build/amd64/config.log 
....
--with-diagnostics-urls=auto-if-env enable_frame_pointer=yes

klausz65: But heee, if you feel better I put it back in again - give me 5 minutes..

klausz65: changed - also removed the LD_OPTIONS line for SPARC as it didn't effected my problem with RUST

PR #26992, opened 2026-07-09
tomcat-11: make used Java runtime variable

klausz65:


components/web/tomcat-11/Makefile
```diff
@@ -89,5 +90,5 @@ COMPONENT_TEST_TARGETS = test
 
 # Auto-generated dependencies
 REQUIRED_PACKAGES += SUNWcs
-REQUIRED_PACKAGES += runtime/java/openjdk21
+REQUIRED_PACKAGES += runtime/java/openjdk$(JAVA_VERSION)
```
mtelka: Was this really auto-generated?  I'm sure it was not.

mtelka (changes requested):

components/web/tomcat-11/Makefile
klausz65: changed

components/web/tomcat-11/Makefile
```diff
@@ -87,7 +88,9 @@ COMPONENT_POST_INSTALL_ACTION += \
 # Override the default test target
 COMPONENT_TEST_TARGETS = test
 
+# set Java_Runtime version variable, SPARC only has 17 for now.
+REQUIRED_PACKAGES += runtime/java/openjdk$(JAVA_VERSION)
+
 # Auto-generated dependencies
 REQUIRED_PACKAGES += SUNWcs
-REQUIRED_PACKAGES += runtime/java/openjdk21
```
mtelka, line 92: Was this line removed by the `gmake REQUIRED_PACKAGES` command run (on intel)?  I'm sure it was not.

mtelka (changes requested):

components/web/tomcat-11/Makefile
```diff
@@ -87,7 +88,13 @@ COMPONENT_POST_INSTALL_ACTION += \
 # Override the default test target
 COMPONENT_TEST_TARGETS = test
 
+# set Java_Runtime arch dependent.
+ifeq ($(MACH), i386)
+REQUIRED_PACKAGES += runtime/java/openjdk21
```
mtelka, line 93: This is not needed here because `gmake REQUIRED_PACKAGES` will add this line automatically below to the `Auto-generated dependencies` section.

mtelka (changes requested):

components/web/tomcat-11/Makefile
klausz65: You obviously see what my problem/intention is, I know that it's not needed ! - but every time a new tommcat get's released I have to edit this Makefile. So why do you give me such a hard time?

klausz65: Do you want me to change shared-makros and put arch depending if for JDK Runtime in there, which I then reference here as $(JDK) or $(JDK_RUNTIME) - better compromise?

mtelka: The `# Auto-generated dependencies` section is automatically generated.  You should never edit it manually.  If you do so, then you do something wrong.  It really makes no sense to edit the section manually (remove a line) because on the next component touch the line will be added back.

I'm sorry, but you need to come with a solution that is NOT based on a manual edit of the `# Auto-generated dependencies` section.

klausz65:
closed

PR #26997, opened 2026-07-10
gcc-13, gcc-14: fix COMPONENT_TEST_TRANSFORMS typo

Bill-Sommerfeld:
Variable was misspelled which broke test result reduction.  Does not affect package contents; patching it so the typo doesn't get copied again.


AndWac:
Please set COMPONENT_REVISION here, too.

Bill-Sommerfeld:
> Please set COMPONENT_REVISION here, too.

Done.

PR #26998, opened 2026-07-10
library/poppler: fix misspelling of COMPONENT_TEST_ENV

Bill-Sommerfeld:


AndWac:
Any change in a Makefile will trigger a new release of the package. So you should also increment COMPONENT_REVISION (or set it to 1 if not yet defined).

Bill-Sommerfeld:
> Any change in a Makefile will trigger a new release of the package. So you should also increment COMPONENT_REVISION (or set it to 1 if not yet defined).

Done.

PR #27017, opened 2026-07-11
install_types: add I2C Framework and Utilities / make prtconf fully f…

klausz65:
…unctional for installer.


mtelka:
It is not possible to update to version 27:
```
# pkg update pkg://openindiana.org/mate_install@27-2026.0.0.0
Creating Plan (Solver setup): \
pkg update: No matching version of mate_install can be installed:
  Reject:  pkg://openindiana.org/mate_install@27-2026.0.0.0
  Reason:  No version for 'require' dependency on system/data/hardware/registry can be found
#
```
@klausz65 please fix.

klausz65:
done see: https://github.com/OpenIndiana/oi-userland/pull/27028

PR #27039, opened 2026-07-14
openssh: update to 10.4p1

iigs:
Used omnios patches.  Renamed patch 0032 to 00006a to make sure it would be used in the correct order in oi-userland

### DIFF
--- a/components/network/openssh/Makefile
+++ b/components/network/openssh/Makefile
@@ -29,15 +29,14 @@ include ../../../make-rules/shared-macros.mk
 COMPONENT_NAME=		openssh
 #   OpenSSH <x>.<y>p<n>     => IPS <x>.<y>.0.<n>
 #   OpenSSH <x>.<y>.<z>p<n> => IPS <x>.<y>.<z>.<n>
-COMPONENT_VERSION=	10.3.0.1
-COMPONENT_REVISION=	1
-HUMAN_VERSION=		10.3p1
+COMPONENT_VERSION=	10.4.0.1
+HUMAN_VERSION=		10.4p1
 COMPONENT_SUMMARY=	OpenSSH client and associated utilities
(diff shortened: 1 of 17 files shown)


AndWac:
This fails to build here:checking for /proc/pid/fd directory... yes
checking for libwrap... configure: error: *** libwrap missing
gmake: *** [/export2/home/andreas/oi-userland/make-rules/common-rules.mk:43: /export2/home/andreas/oi-userland/components/network/openssh/build/amd64/.configured] Error 1

AndWac:
And the reason for the failure is
configure:16433: checking for libwrap
configure:16454: /usr/gcc/14/bin/gcc -std=gnu23 -o conftest -m64 -O3  -DSET_USE_PAM -DPAM_ENHANCEMENT -DPAM_BUGFIX -DDTRACE_SFTP -I/
conftest.c: In function 'main':
conftest.c:140:9: error: too many arguments to function 'hosts_access'
  140 |         hosts_access(0);
      |         ^~~~~~~~~~~~
In file included from conftest.c:133:
/usr/include/tcpd.h:156:12: note: declared here
  156 | extern int hosts_access();              /* access control */
      |            ^~~~~~~~~~~~
configure:16454: $? = 1
configure: failed program was:
| /* confdefs.h */

iigs:
Removed patch 31 from omnios.

Bill-Sommerfeld:
instead of dropping the patch, we should build this with something pre-C23 until tcpd.h is fixed to have a proper prototype

Bill-Sommerfeld:
I filed https://www.illumos.org/issues/18246 against libwrap in illumos.

iigs:
I did not drop the patch.  I just removed the patch from the PR.  The patch that existed with 10.3p1 is what will be used with 10.4p1.

PR #27069, opened 2026-07-16
openjdk-21: update to 21.0.11

iigs:
Replaced all of the patches with the native patches from tribblix.  p0 patches put in patches-p0 and p1 patches put in patches-p1.  The reverse patch put  in patches-reverse. The standard userland patches directory and relevant make stuff is not used.  openindiana specific patches put in patches-oi.  

Added missing build dependencies.


iigs:
Fixed patches-reverse file contents

PR #27116, opened 2026-07-22
add nvidia-535 and nvidia-580 to xorg-video

timmooney:
Add the missing driver/nvidia-535 and driver/nvidia-580 to the xorg-video meta-package.

I understand that this won't have any impact on standard hipster Live media, but it will make it easier to build custom Live images using distribution-constructor.


AndWac:
This already has been fixed with #27099

PR #27193, opened 2026-07-31
history: sort file, add component history files.

klausz65:
…o it.

This is the first PR for the upcomming cleanup/Change in history handling


mtelka (changes requested):
Please do not do this.  One big `history` file is far harder to manage than individual small ones.

PR #27194, opened 2026-07-31
history: new Makefile to only build the last commit

klausz65:


mtelka (changes requested):
In general this is interesting (and good) idea, but the implementation should go somewhere else, IMHO.

To consider this we should start with understanding what exactly is the problem you are trying to solve by this.  Is the problem that there is a lot of unnecessary obsoleted packages published to the IPS repo?  Or, is the problem that the build time of the `history` component is too long?  Or something else?

Thank you.

klausz65:
Well, both of the reasons you name, atm. the metadata is more than ~700mb which causes every pkg refresh --full to run quite some time, and downloads all that old metadata for all these old obsolete packages. The new installer I'm trying to bring onto the road, will give the installing user the possibility to choose between several software profiles. To do a successfull up-to-date install of these I need a refresh of these metadata right after the basic os installation. This particular step (the refresh) takes quit some minutes on old systems (longer if the connection of that user is slow). Till is evaluating these #PRs on a copy ot the main repo today. I did all this while having him involved, let's wait what today brings up regarding this.

mtelka:
> Well, both of the reasons you name, atm. the metadata is more than ~700mb which causes every pkg refresh --full to run quite some time, and downloads all that old metadata ...

Both?  You are elaborating on the size of the IPS repo only...

Anyway, the IPS repo size should be handled elsewhere, and generally.  For example, the final publish step (the one that copies the built packages to the ips server) should compare the current list of published packages (at the ips server) with the list of currently built packages and publish only packages with different fmri.  This would solve the problem generally for all components, not only for the `history`.

Re the `history` build time: Splitting out the single big beast to several small `history` files (the exact opposite of #27193) is the way to shrink down the `history` build time without build hacks like proposed here.

BTW, with this change if there is a `history` build failure and there is some newer `history` commit added we will never re-build the failed one and hacks with modifying perfectly okay `history` lines will be needed to be done.  We do something similar by touching `Makefile`s to see a component rebuild, but that's harmless since an empty line addition to the `Makefile` both does the job and is harmless.  With the `history` lines, what will you change in the line like `database/mongodb-34@3.4.19,5.11-2022.0.0.3` to be harmless?   Add an empty space at the end of the line (please no!)?

PR #27196, opened 2026-07-31
history: add consolidation-xvm-xvm-incorporation to history file.

klausz65:


mtelka (commented):
So far, the `xvm-incorporation` package was a regular package.  I'm not sure why exactly it is (or was) needed (do you or anyone else happen to know?).  With this change you simply obsolete it.  Is it okay to obsolete the `xvm-incorporation`?

PR #27207, opened 2026-08-01
php-8_5: update to 8.5.9

iigs:
Use only openssl 3.

### DIFF
--- a/components/web/php/php-8_5/Makefile
+++ b/components/web/php/php-8_5/Makefile
@@ -12,18 +12,19 @@
 # Copyright 2025-2026 Friedrich Kink. All rights reserved.
 #
 
+USE_PARALLEL_BUILD = yes
 include ../../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=         php
 COMPONENT_MJR_VERSION=	8.5
-COMPONENT_MNR_VERSION=	7
+COMPONENT_MNR_VERSION=	9
 COMPONENT_SUMMARY=      PHP 8.5
 COMPONENT_DESCRIPTION=	PHP is a popular general-purpose scripting language that is especially suited to web development.
 COMPONENT_VERSION=      $(COMPONENT_MJR_VERSION).$(COMPONENT_MNR_VERSION)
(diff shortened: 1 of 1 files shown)


AndWac:
The changes related to openssl look strange. If gmake REQUIRED_PACKAGES doesn't detect the openssl libraries then no binary uses the libs. If they were used before then this would be strange.

iigs:
Oh... I missed that REQUIRED_PACKAGES += $(OPENSSL_PKG) was in manually added.

iigs:
The openssl library is used by the extensions in usr/php/8.5/extensions/ .  I do not think gmake REQUIRED_PACKAGES knows to look in  that directory for libraries. Please advise what is needed to have that library scanned for libraries.

iigs:
Er. have that directory scanned for libraries.

klausz65:
Hi, I just have built this commit on SPARC and can confirm that it does not use old SSL libs - just the ones from openssl-3 - however the REQUIRED_PACKAGES target does not pull in the variable $(OPENSSL_PKG). But verifying with ldd on all the extensions - I only find the said openssl-3 libs referenced. Therefore from my point of view this good to go. Thanks quite much!

AndWac:
Ok, then please add a comment with a short explanation in front of the manually added openssl line.
Still, this looks strange if the target doesn't catch the dependency at all.

klausz65:
Hmm, there is no dependency line for openssl in REQUIRED_PACKAGES section :-) that' why I used ldd in the prototype area, and yes I see references to openssl/cyrpto libs, which are all pointing to version 3 ! iigs has already removed the REQUIRED_PACKAGES += $(OPENSSL_PKG) line from the Makefile. The reason why REQUIRED_PACKAGES target does not reveal the dependency is the fact that these libs are probably only loaded when the shared-modules used in php will only be pulled in at runtime - the only explanation which does make sense to me.

iigs:
Removed duplicate dependencies from manual dependency section. Added comment regarding openssl in manual dependency section. Enable parallel build.

PR #27372, opened 2026-08-19
bash-completion: update 2.18.0, fix ps command

cgrzemba:


cgrzemba:
https://github.com/scop/bash-completion/commit/7cccbd2cb9795246ad56edfa8f7dbb13f2c287fe
renames the completion file by adding suffix .bash
I would recommend also renaming our files in https://github.com/OpenIndiana/openindiana-completions

PR #27529, opened 2026-09-07
fix(illumos-gate): resolve gettext locale fallbacks

Toasterson:
Resolve gettext locale and per-message fallbacks without locale symlinks or generated LANGUAGE lists.

Native libc builds and 44 public API checks passed. Live desktop testing is ongoing.

Related: OpenIndiana/sysding#9


Toasterson:
@Bill-Sommerfeld This might be related to your locale work. The idea is that it is closer to glibcs handling of locales so we don't run into problems with desktops.

PR #27589, opened 2026-09-14
webkitgtk-60: add 2.53.92

iigs:
This is to provide the 6.0 api for webkitgtk.  Only the library/headers are provided.  This is the version that uses GTK4.

This is Part 1.

This needs to be published before building the yelp 49 packages.


components/library/webkitgtk-60/manifests/sample-manifest.p5m
```diff
@@ -0,0 +1,1782 @@
...
+set name=pkg.summary value="$(COMPONENT_SUMMARY)"
+set name=info.classification value="$(COMPONENT_CLASSIFICATION)"
+set name=info.upstream-url value=$(COMPONENT_PROJECT_URL)
+set name=info.source-url value=$(COMPONENT_ARCHIVE_URL)
+set name=org.opensolaris.consolidation value=$(CONSOLIDATION)
+
+license $(COMPONENT_LICENSE_FILE) license='$(COMPONENT_LICENSE)'
+
+file path=usr/bin/WebKitWebDriver
+file path=usr/include/webkitgtk-6.0/jsc/JSCClass.h
```
AndWac, line 27: This file collides with the one from the webkitgtk2 package. You have to rename it and probably should introduce a mediator (thus, also rename the one from webkitgtk2).

iigs: This is why I have these lines in the .p5m file:

<transform file path=usr/bin/WebKitWebDriver -> drop>
<transform file path=usr/libexec/MiniBrowser -> drop>

I was able to install webkitgtk-60 on a system with webkitgtk already installed.

AndWac: Ok, I didn't recognise that. I still prefer renaming both files.

PR #27608, opened 2026-09-15
webkitgtk-60: fix missing mpegts

iigs:
enabled webkitgtk-60 package


components/library/webkitgtk-60/Makefile
```diff
@@ -22,6 +22,7 @@ include ../../../make-rules/shared-macros.mk
 
 COMPONENT_NAME=		webkitgtk-60
 COMPONENT_VERSION=	2.53.92
+COMPONENT_REVISION=	1
 COMPONENT_SUMMARY=	Web content engine library for GTK+
```
AndWac, line 25: You don't need to set COMPONENT_REVISION because the original build failed an no packag were created yet. So the repository doesn't know of any webkitgtk-60 package at the moment.

iigs:
Removed COMPONENT_REVISION line from Makefile

PR #27622, opened 2026-09-17
webkitgtk: update to 2.54.0

iigs:
Add mediator support for binaries that would conflict with those that would be in webkitgtk-60.  The mediator version is api version, not webkitgtk version

This needs to be published before building webkitgtk-60 version 2.54


AndWac:
You have deleted the Makefile

iigs:
Added missing Makefile.

PR #27623, opened 2026-09-17
webkitgtk-60: update to 2.54.0

iigs:
Include missing binaries and mediate by api version.


iigs:
This has to be built after webkitgtk 2.54 is published.

PR #27639, opened 2026-09-20
emacs: update to 31.1; provide workaround for missing load-path entries;

AndWac:
add test results


AndWac:
I am not sure why setting the default load-path is not working correctly during bootstrap. Setting it in site-lisp.el is just a workaround.

